PR #35, opened 2019-08-05
Describe folder structure

cgreene:
This has an initial draft of a folder structure for analyses. It would be nice to do a couple analyses with this from closely-associated folks. We can see how well these work out and make modifications to improve our description.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ OpenPBTA-analysis
 │       └── results
 │           └── tabular_summary.csv
 ├── data
-├── existing-analysis-2
 └── scratch
 ```
 
--- a/README.md
+++ b/README.md
@@ -59,6 +59,11 @@ Choose a name that is unique from other analyses and somewhat detailed.
 For example, instead of `gene-expression`, choose `gene-expression-clustering` if you are clustering samples by their gene expression values.
 You should assume that any data files that you need to read live in the `../../data` directory and that their file names match what the `data_downloader.sh` script produces.
 These files should be read in at their relative path, so that we can re-run analyses if the underlying data change.
+Files that are primarily graphic should be placed in a `plots` subdirectory.
+Files that are primarily tabular results files should be placed in a `results` subdirectory.
+Intermediate files that are useful within the processing steps but that do not represent final results should be placed in `../../scratch/`.
+It is safe to assume that files placed in `../../scratch` will be available to all analyses within the same folder.
+It is not safe to assume that files placed in `../../scratch` will be available from analyses in a different folder.
 
 ```
 OpenPBTA-analysis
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -53,10 +53,44 @@ We are using continuous integration software applied to the supplied test datase
...
+│       └── figure2.pdf
+└── scratch
+```
+
 ### Analysis Script Numbering
 
+As shown above, analysis scripts within a folder should be numbered from `01` and are intended be run in order.
+If the script produces any intermediate files, these files should be placed in `../../scratch`.
+It is safe to assume that files placed in `../../scratch` will be available to all analyses within the same folder.
+It is not safe to assume that files placed in `../../scratch` will be available from analyses in a different folder.
```
jaclyn-taroni: Hm, is this preferable to an intermediate files folder _within_ a `new-analysis` directory? Alternatively:
```
└── scratch
  └── new-analysis
         ├── intermediate-file1.tsv
         └── intermediate-file2.tsv
```
which is more consistent with what else you have

README.md
```diff
@@ -53,10 +53,44 @@ We are using continuous integration software applied to the supplied test datase
...
+├── analyses
+│   ├── existing-analysis-1
+│   ├── existing-analysis-2
+│   └── new-analysis
+│       ├── 00.run_all.sh
+│       ├── 01.preprocess-data.Rmd
+│       ├── 02.run-analyses.Rmd
+│       └── 03.make-figures.Rmd
+└── data
+└── results
```
jaclyn-taroni: I typically keep a `results` and `plots` directory for my projects, where `results` would include items like stats for differential gene expression analysis in tabular format. Would such a table be in the `results` folder here? If so, I would include an example filename here that is something _other than a figure_.

README.md
```diff
@@ -53,10 +53,44 @@ We are using continuous integration software applied to the supplied test datase
...
+│   └── new-analysis
+│       ├── figure1.pdf
```
cansav09: Does this mean we generally want things to be in PDF form when possible or is this not actually a preference you're meaning to imply by this?

README.md
```diff
@@ -53,10 +53,44 @@ We are using continuous integration software applied to the supplied test datase
...
+```
+OpenPBTA-analysis
+├── CONTRIBUTING.md
+├── README.md
+├── analyses
+│   ├── existing-analysis-1
+│   ├── existing-analysis-2
+│   └── new-analysis
+│       ├── 00.run_all.sh
+│       ├── 01.preprocess-data.Rmd
```
cansav09: Should it be emphasized somewhere when a `.Rmd` vs `.R` should be used or do we just want people to always heavily default to `.Rmd`?

README.md
```diff
@@ -53,10 +53,44 @@ We are using continuous integration software applied to the supplied test datase
...
+You should assume that any data files that you need to read live in the `../../data` directory and that their file names match what the `data_downloader.sh` script produces.
+These files should be read in at their relative path, so that we can re-run analyses if the underlying data change.
+
+```
+OpenPBTA-analysis
+├── CONTRIBUTING.md
+├── README.md
+├── analyses
+│   ├── existing-analysis-1
+│   ├── existing-analysis-2
```
cansav09: What's the significance of the terms "existing" vs "new" analysis here? Or is there none?

README.md
cgreene: 👍

README.md
cgreene: Tried to add more specificity.

README.md
cgreene: Added additional info that says this is an example, but other formats are fine 👍

README.md
cgreene: I added:
> An example highlighting a `new-analysis` directory is shown below.
> The directory is placed alongside existing analyses within the `analyses` directory.

In an upcoming commit.

README.md
cgreene: Updated to say:
> The author has produced their output figures as `.pdf` files.
> We have a preference for vector graphics as PDF files, though other forms of vector graphics are also appropriate.

jharenza:
I like this!

Issue #30, opened 2019-07-29
Evaluate concordance between Mutect2 and Strelka2 and decide on next steps


jharenza:
Evaluate concordance and determine overall set of high-confidence somatic variants for PBTA.

cansav09:
I've gotten the data fairly well wrangled and loaded. What kinds of stats should I focus on evaluating the similarities/differences between these two? What kind of comparison plots would you like to see?

cansav09:
Here's what I'm thinking for *first line* analyses: 

- Venn diagrams of individual mutations detected by Mutect2 vs Strelka2
- Correlation of total # mutations called for each sample
- Correlation of total # mutations called for each gene

We can then do these analyses and subset these by the type of mutation. Does this sound reasonable as a first pass?

cgreene:
I like those ideas!

cansav09:
Here's the equivalents of the [figure examples noted here](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/38#issuecomment-518838416). I need to neaten up the code some more, will get that in a better spot tomorrow. 

Figure A:
<img width="783" alt="Screen Shot 2019-08-07 at 3 51 25 PM" src="https://user-images.githubusercontent.com/23458084/62653485-abe2a380-b92b-11e9-96a4-f9b5b58f10b3.png">

Figure B:
<img width="871" alt="Screen Shot 2019-08-07 at 3 01 10 PM" src="https://user-images.githubusercontent.com/23458084/62653508-b7ce6580-b92b-11e9-9a3e-8d38e007d238.png">

Figure C:
<img width="871" alt="Screen Shot 2019-08-07 at 3 50 34 PM" src="https://user-images.githubusercontent.com/23458084/62653522-be5cdd00-b92b-11e9-82e9-9e7493409dae.png">

jharenza:
This looks really great! Interesting - most of the mutations called by only one algorithm are subclonal (VAF <0.2). I think the next question is, are any of these artifacts/low confidence, and do we keep all/some/none?

cansav09:
> I think the next question is, are any of these artifacts/low confidence, and do we keep all/some/none?

Do you have suggestions for analyses to approach this question?

jharenza:
[...]

Issue #33, opened 2019-07-30
Script to generate a subset of the data


cgreene:
It would be really handy if there was a script that could be run over the input data files to produce files containing data for only a small, specified number of matched participants. I'm imagining that we would use this to produce data files with information for 10 or so participants. We would then use this code via CircleCI to test the analytical code that people contribute without having to run over the full set of input data.

jharenza:
@yuankunzhu ^

jaclyn-taroni:
@cgreene and I discussed this in person as a follow-up to #44. We'll aim for 5 matched participants across all files. They will contain all the data for these 5 participants.

I can generated these subset files and stick them on S3 to facilitate getting CI up and running.

jaclyn-taroni:
Closed via #56.

PR #38, opened 2019-08-06
In Progress: Mutect2 vs Strelka2 First Pass

cansav09:
## Purpose:
Ultimately this PR pertains to issue #30, but this is still still very much an "in-progress" PR. 
I'm only uploading this so that I can ask about directions for this analysis, so do not bother giving a detailed review just yet. 

A lot of these chunks will ultimately change a lot after #2 is established and as other contributing guidelines for this repository are further defined. 

## Results/Concerns

Overall these *very* preliminary results seem to suggest MuTect2 and Strelka2 agree most of the time. 

First my over-arching question is are these analyses appropriate handlings of these data? 
But then I have questions related to what I'm seeing in the results that it would be great to get more thoughts on. 

These are my main questions that you will see sprinkled in the notebook: 
**Question 1)** Is this a reasonable amount of concordance for these methods 
across genes? 
**Question 2)** Is this a reasonable amount of concordance for these methods 
across samples? 
**Question 3)** Is this an acceptable way of looking for overlap between these methods? 
If not, what variables or data should I be basing this Venn Diagram on? 
**Question 4)** I made a gene concordance table for each, any idea why everything appears to be so highly correlated for Strelka2 results but not MuTect2 results look more reasonable? 

## Pull Request Check List:
* [ ] Run a linter - have not run one yet. 
* [x] Set the seed (if applicable).
* [x] Comments and/or documentation up to date.
* [x] Double check your paths.
* [x] Check headline numbering.
* [x] Spell check any Rmd file or md file.
* [x] Restart R and run all notebooks fresh and save.


cgreene:
I have just a quick visualization question. It looks like there's a lot of overplotting in
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/73df39768a17bbdc4b930294705976d8e53146a2/analyses/mutect2-vs-strelka2/plots/gene_cor_mutect2_vs_strelka2.pdf
and
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/73df39768a17bbdc4b930294705976d8e53146a2/analyses/mutect2-vs-strelka2/plots/sample_cor_mutect2_vs_strelka2.pdf

It might be good to use something that helps to deal with this when making these plots (one option is something like https://ggplot2.tidyverse.org/reference/geom_hex.html )

It's nice to see this coming along! I look forward to @jharenza's perspective on the level of agreement.

cansav09:
> Would it be good to use something that helps to deal with this when making these plots (one option is something like https://ggplot2.tidyverse.org/reference/geom_hex.html )?

I'll try this out instead.

cansav09:
Here's a sample of using that: 
Screen Shot 2019-08-06 at 2 51 01 PM

It certainly makes it more clear to the reader how skewed these data are.

jharenza:
Hi @cansav09! This looks like a great start. I think we do expect more calls by Strelka2 - it detects mutations at lower VAF than Mutect2. The major thing we want to gain from this is a final MAF of all disease-relevant somatic mutations. If a mutation is not concordant between the two algorithms, can you rule it out as a false positive/artifact or should we keep it as a a true positive? So, to look at the concordance, I think it is important to look at the actual variants per patient across the two algorithms. Take a look at A,B,C below (figures created by @gonzolgarcia).
concordance-plots

cansav09:
> Take a look at A,B,C below (figures created by @gonzolgarcia).

Do you have the code you used to make these figures available on GitHub somewhere, @gonzolgarcia?

cansav09:
> A) Variant overlap for a set of 59 samples with both WGS and WES somatic variant calls.

This seems like it should be a simple question, but it is difficult to find an exact answer anywhere, which is why I was wondering if I could see the code for A. What criteria do you use to determine if the somatic variant calls for the two sets here are the same? To make this question even more specific, what [`maf` format categories](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) would one use to compare? E.g. if a somatic variant call has the same sample, same gene, and same chromsomal start site in both algorithms, would you consider it to be the same call and hence be in the center part of that Venn Diagram (A)?

jharenza:
> > A) Variant overlap for a set of 59 samples with both WGS and WES somatic variant calls.
> 
> This seems like it should be a simple question, but it is difficult to find an exact answer anywhere, which is why I was wondering if I could see the code for A. What criteria do you use to determine if the somatic variant calls for the two sets here are the same? To make this question even more specific, what [`maf` format categories](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) would one use to compare? E.g. if a somatic variant call has the same sample, same gene, and same chromsomal start site in both algorithms, would you consider it to be the same call and hence be in the center part of that Venn Diagram (A)? @jharenza @gonzolgarcia

Almost, @cansav09!  - you would use same sample, gene, chr, and cDNA change (I would not use start site because in rare cases, there may be two changes at the one site or the change in one algorithm may not match the change in the other, so to confirm it is identical, you can concatenate all columns mentioned and match them in both algorithms).

cansav09:
@jharenza! Thanks for this clarification!

gonzolgarcia:
> > A) Variant overlap for a set of 59 samples with both WGS and WES somatic variant calls.
> 
> This seems like it should be a simple question, but it is difficult to find an exact answer anywhere, which is why I was wondering if I could see the code for A. What criteria do you use to determine if the somatic variant calls for the two sets here are the same? To make this question even more specific, what [`maf` format categories](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) would one use to compare? E.g. if a somatic variant call has the same sample, same gene, and same chromsomal start site in both algorithms, would you consider it to be the same call and hence be in the center part of that Venn Diagram (A)? @jharenza @gonzolgarcia

Hi @cansav09 , I am refining the code for uploading, but in the meantime; variants were consider the same only when the SampleID/Chromosome/Start/Stop were the same. (Indeed I did concatenate the fields and intersect...). In this figure I did not considered whether the exact nucleotide change was the same as well;  there is one case in which the the same position suffers two different nucleotide changes; presumably different alleles affected or different clones.

analyses/mutect2-vs-strelka2/01-set-up.Rmd
```diff
@@ -0,0 +1,256 @@
...
+
+## Set Up
+
+\* Some of these set up steps will be removed once we have a Dockerfile that installs
+maftools automatically. 
+
+```{r}
+# We need maftools - this will be added to the running Docker issue whenever it is up
+if (!('maftools' %in% installed.packages())) {
+  devtools::install_github("PoisonAlien/maftools")
```
jaclyn-taroni: @cansav09 I am working on #33 and was planning on using `maftools` to do so. Is there a reason you're installing what I am assuming is a development version of the package from GitHub here rather than installing from bioconductor? If you're using functionality that is not available from the bioconductor version, you should tie this to a specific commit via the `ref` argument to `devtools::install_github`.

cansav09: I switched versions of `maftools` when I was having problems running `maftools` in Docker and hadn't yet identified that it was definitely an OOM problem. I will return to `maftools_2.0.15` which is the version that will be default installed with our tidyverse 3.6.0 Docker image and make sure everything run smoothly with this version. When this PR is ready for prime time, I'll make sure `maftools` is added installation to the Dockerfile.

cansav09:
Going to close this draft PR in preparation for the 'real thing' PR.

PR #45, opened 2019-08-12
Improve Downloads / Reduce Maintainence Burden

cgreene:
This PR improves the download code by first downloading the md5sum file and then downloading each of the files listed within that plus the release-notes.md file. It also substantially reduces the number of lines of the README that have specific information that we have to maintain. Instead of directly maintaining the text descriptions of processing, the README links to the OpenPBTA-manuscript headers.


cgreene:
@jaclyn-taroni - done 👍 . Also put a stub in the analysis section noting that the `data/` symlinks and not anything in release should be used. I could see that as a frequent issue, so we'll want to make sure we reiterate it somewhere there as that gets filled in.

PR #48, opened 2019-08-12
BUGFIX: Data download script name in README

jaclyn-taroni:
The data download shell script (https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/1ace53d0b8fb1d2ec321721bd80e981969e408d8/download-data.sh) has a hyphen in the name. The README used an underscore.


jaclyn-taroni:
If I run:
```
sh download-data.sh
```
on Linux, I get the following error:
```
download-data.sh: 3: set: Illegal option -o pipefail
```
Some light googling tells me that this is because we're trying to use a `bash` feature. I'm going to change the instructions to be `bash download-data.sh`.

cgreene:
Oh wow - didn't know pipefail was bash specific. This looks terrible:
https://stackoverflow.com/questions/13084352/how-to-implement-set-o-pipefail-in-a-posix-way-almost-done-expert-help-nee

I agree with the switch to just recommending bash.

Issue #50, opened 2019-08-12
Add histologies file to md5sum.txt


cgreene:
This file will also need to be added since #45 has landed.

jaclyn-taroni:
This is blocking #33

jharenza:
@yuankunzhu - can you add md5sum for the histologies file?

jharenza:
This has been completed.

cgreene:
Can confirm. Download via script now includes this file 👍

PR #54, opened 2019-08-12
Addition of `01-filter-across-types.R` script

cbethell:
This PR is the first step of the analysis addressing issue #5  

- `01-filter-across-types.R` performs the initial filtering of the dataset. 
- The script produces a PDF of a barplot representing the sample distribution across cancer types, as well as a TSV with the raw values used to do so.  
- The script also produces TSV files of the cancer type distribution across primary sites of the brain (where `primary_sites.tsv` displays a table of all the cancer types expressed within each site and `primary_sites_counts.tsv` displays the total number of types, along with the highest and second highest expressed types for each unique site). 

### Questions

1. Are there any additional data filtering steps that may need to be included for this particular dataset? 

2. Are the results/plots suffice for this particular issue? If not, what other forms of tables or plots would you like to see included? *Please keep in mind that the interactive treemap and pie chart are produced in the upcoming script.*

### Pull Request Check List:
* [x] Run a linter
* [x] Set the seed (if applicable).
* [x] Comments and/or documentation up to date.
* [x] Double check your paths.
* [x] Check headline numbering.
* [x] Spell check any Rmd file or md file.
* [x] Restart R and run all notebooks fresh and save.


cbethell:
@jharenza 
> * Can you update the code to include the data release V2, which will change the name of the clinical file?
> * One potentially relevant thing to do for primary site may be to create a new variable for region, where we separate into midbrain and hemispheric. I would use `Thalamus|Pons|Spinal Cord|Brain Stem|Midbrain` to denote midbrain and the `various lobes|cerebellum|hippocampus` to be hemispheric, and we can call the remaining as other.

Great, thank you for your suggestions! I can work on both of those today and notify you when I've committed the changes.

cgreene:
@jharenza can you see about addressing #50? Without that file in the md5sum, the download script won't acquire it.

cgreene:
Also, @jharenza : can you add that additional variable to the clinical metadata file? It seems better to have it in one place as opposed to defined separately in many scripts, since that grouping may be used throughout the project.

jaclyn-taroni:
I agree with updating the code to include the data release v2 file which will be called `pbta-histologies.tsv` once #50 is addressed.

As far as the region point, I agree with @cgreene that region information should be in the histologies file to serve as a single source of truth across the project. Anything that will be used across the project in multiple places should be defined in as few places as possible and as part of the files obtained with `bash download-data.sh` where possible.

jharenza:
> Also, @jharenza : can you add that additional variable to the clinical metadata file? It seems better to have it in one place as opposed to defined separately in many scripts, since that grouping may be used throughout the project.

Yes - I can do this - we can get this updated after our next meeting.

jharenza:
Just a note - I will add the regions to the clinical file in the next version, since I want to double check my classifications with an MD. The md5sum is now included.

jharenza:
Referring to #57 for the brain region classification. Angela can get to this by Friday.

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,225 @@
+# This script filters the given dataset to produce a summarized visualization
+# of key variables within the dataset.
+#
+# Chante Bethell for CCDL 2019
+
+
+# Load/install packages. This will be removed once a Dockerfile is established.
+source("00-install-packages.R")
```
jaclyn-taroni: I am looking forward to having this removed because if you are installing a new package, it is also being loaded into the namespace which may or may not be what you want to do.

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,225 @@
...
+      meninges = meninges,
+      basal_ganglia = basal_ganglia,
+      optic_pathway = optic_pathway,
+      hippocampus = hippocampus,
+      cranial_nerves = cranial_nerves
+    )
+  )
+
+# Write to tsv file
+readr::write_tsv(primary_sites, file.path("results", "primary_sites.tsv"))
```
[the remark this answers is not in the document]
cbethell: The idea behind this data.frame was to provide an easy visualization of this particular subset for the brain graphic. However, it isn't completely necessary given that `primary_sites_counts` specifies the highest and second highest expressed types, along with the counts (all the information that seems to be needed for the brain graphic).

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,225 @@
...
+
+# data.frame with the location where each cancer type in the dataset is expressed 
+# sorted to show highest expression
+brain_location <- df2 %>%
+  dplyr::select(disease_type_new, primary_site) %>%
+  dplyr::group_by(primary_site) %>%
+  dplyr::arrange(primary_site) %>%
+  dplyr::group_by(disease_type_new, primary_site) %>%
+  dplyr::tally() %>%
+  dplyr::arrange(desc(n)) %>%
```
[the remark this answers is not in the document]
cbethell: Yes, good catch

analyses/sample-distribution-analysis/01-filter-across-types.R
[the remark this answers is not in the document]
jaclyn-taroni: 👍 I would take it out in that case. Doing so will facilitate the suggested changes.

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,225 @@
...
+  dplyr::count() %>%
+  dplyr::arrange(desc(n)) %>%
+  dplyr::rename(count = n)
+
+# Calculate the total count of the dataset
+sum_count <- sum(disease_expression$count)
+
+# Create a percent variable
+disease_expression <- disease_expression %>%
+  dplyr::mutate(percent = (count / sum_count))
```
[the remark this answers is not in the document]
cbethell: Out of the above suggestions, this is what worked: `dplyr::mutate(percent = paste0(((count / sum_count) * 100)`.
Would this still be a better solution compared to the inclusion of lines 65-66?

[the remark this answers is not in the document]
jaclyn-taroni: Ah yes good call. Wouldn't you still need to add the `"%"` or am I misunderstanding the other part of what `formattable::percent` does?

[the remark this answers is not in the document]
cbethell: Ah yes, good catch. I have the `"%"` in the script but left it out when I copied the line here.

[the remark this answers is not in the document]
jaclyn-taroni: I would say since we will have a monolithic Docker container for this project, any place where we can get rid of dependencies we should try and do that. This seems like something where we don't need a well-used function from a package to do it.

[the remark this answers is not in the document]
cbethell: Understood. I'll remove the use of that package and keep this in mind before implementing any other new packages in other scripts.

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,146 @@
...
+  dplyr::group_by(disease_type_new) %>%
+  dplyr::count(name = "count") %>%
+  dplyr::arrange(dplyr::desc(count))
+
+# Calculate the total count of the dataset
+sum_count <- sum(disease_expression$count)
+
+# Create a percent variable
+disease_expression <- disease_expression %>%
+  dplyr::mutate(percent = paste0(((count / sum_count) * 100), "%"))
```
jaclyn-taroni: Now that you've pushed the changes. I wanted to note that the numbers look ever so slightly different to make sure you are aware of that and I think the last thing that function was doing was performing a rounding step, which can likely be accomplished with something like:
```suggestion
  dplyr::mutate(percent = paste0((round(count / sum_count, 2) * 100), "%"))
```

cbethell: Ah, very true. I will make that change now and check the numbers before committing the change.

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,143 @@
...
+# of key variables within the dataset.
+#
+# Chante Bethell for CCDL 2019
+
+# magrittr pipe
+`%>%` <- dplyr::`%>%`
+
+# Function to filter based on primary_site 
+location_fn <- function(location) {
+  disease_type_vector <- brain_location %>%
```
jaclyn-taroni: Can you add more documentation to this function? I tend to favor the [old R Google Style Guide recommendations for function documentation](https://web.archive.org/web/20190620164620/https://google.github.io/styleguide/Rguide.xml). Here's an example in the `create-subset-files` analysis: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/create-subset-files/01-create_subset_files.Rmd#L47

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,143 @@
+# This script filters the given dataset to produce a summarized visualization
+# of key variables within the dataset.
+#
+# Chante Bethell for CCDL 2019
+
```
jaclyn-taroni: Can you add a statement about how this script is intended to be run? An example would be:

```
#### USAGE
This script is intended to be run via the command line from the top directory of the repository:
Rscript analyses/sample-distribution-analysis/01-filter-across-types.R
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,143 @@
...
+# Create a bar plot of sample distribution across cancer types
+gg_types <- disease_expression %>%
+  ggplot2::ggplot(ggplot2::aes(x = disease_type_new, y = count, fill = count)) +
+  ggplot2::geom_col() +
+  ggplot2::theme_bw() +
+  ggplot2::labs(x = "Cancer Types", y = "Count",
+                title = "Sample Distribution Across Cancer Types") +
+  ggplot2::scale_y_continuous(breaks = seq(0, 500, by = 100)) +
+  ggplot2::theme(axis.text.x = element_text(
```
jaclyn-taroni:
```suggestion
  ggplot2::theme(axis.text.x = ggplot2::element_text(
```
Based on the test failure

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,143 @@
...
+    angle = 75,
+    hjust = 1,
+    size = 8
+  ),
+  panel.grid = element_blank()) +
```
jaclyn-taroni:
```suggestion
  panel.grid = ggplot2::element_blank()) +
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+  disease_type_vector <- brain_location %>%
+    dplyr::arrange(dplyr::desc(n)) %>%
+    dplyr::filter(stringr::str_detect(primary_site, location)) %>%
+    dplyr::pull(disease_type_new)
+  unique(disease_type_vector)
+}
+
+# Define the file.path to output directories
+outputDir <- file.path("analyses", "sample-distribution-analysis")
+
```
jaclyn-taroni: If you add this here:
```R
results_dir <- file.path(outputDir, "results")
plots_dir <- file.path(outputDir, "plots")
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+# Create directories to hold the output.
+if (!dir.exists(file.path(outputDir, "results"))) {
```
jaclyn-taroni:
```suggestion
if (!dir.exists(results_dir)) {
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+  dir.create(file.path(outputDir, "results"))
```
jaclyn-taroni:
```suggestion
  dir.create(results_dir)
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+}
+if (!dir.exists(file.path(outputDir, "plots"))) {
```
jaclyn-taroni: Would change these to `plots_dir` -- same pattern as `results_dir`

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+disease_expression <- disease_expression %>%
+  dplyr::mutate(percent = paste0((round(count / sum_count, 4) * 100), "%"))
+
+# Reorder the columns to be displayed in descending order by count on the plot
+disease_expression$disease_type_new <- with(disease_expression,
+                                            reorder(disease_type_new, -count))
+
+# Write to tsv file
+readr::write_tsv(disease_expression,
+                 file.path(outputDir, "results",
```
jaclyn-taroni:
```suggestion
                 file.path(results_dir,
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+    size = 8
+  ),
+  panel.grid = ggplot2::element_blank()) +
+  ggplot2::geom_text(nudge_y = 6.5, size = 2,
+                     ggplot2::aes(label = paste0(disease_expression$percent)))
+
+# Save plot
+ggplot2::ggsave(
+  gg_types,
+  file = file.path(outputDir, "plots", "distribution_across_cancer_types.pdf"),
```
jaclyn-taroni: Could then change this to
```suggestion
  file = file.path(plots_dir, "distribution_across_cancer_types.pdf"),
```

analyses/sample-distribution-analysis/01-filter-across-types.R
```diff
@@ -0,0 +1,162 @@
...
+  dplyr::mutate(max_type = unlist(lapply(cancer_types_list,
+                                         function(x) x[1])),
+                second_max_type = unlist(lapply(cancer_types_list,
+                                                function(x) x[2]))) %>%
+  # Reorder the rows
+  dplyr::arrange(dplyr::desc(number_of_types))
+
+# Write to tsv file
+readr::write_tsv(primary_sites_counts,
+                 file.path(outputDir, "results", "primary_sites_counts.tsv"))
```
jaclyn-taroni:
```suggestion
                 file.path(results_dir, "primary_sites_counts.tsv"))
```

jaclyn-taroni:
@jharenza the change to the histologies file you requested has been made and the second point you raised is now tracked at #57. If this looks good to you, please feel free to merge.

jharenza:
> Just a note - I will add the regions to the clinical file in the next version, since I want to double check my classifications with an MD. The md5sum is now included.

I talked with Angela Waanders and she recommended we only do this for gliomas (and only if regions are not ambiguous - ie tumor is not in both midline and hemispheric regions) and not all samples, so will add that later.

cgreene:
@jharenza: Ok! It also looks like you still have this PR marked as changes requested. Is there anything else that needs to be changed in the code before it can be merged?

cbethell:
@jharenza Thank you for that confirmation!

cgreene:
Going to merge this 👍

PR #56, opened 2019-08-13
Generate subset files for continuous integration

jaclyn-taroni:
#### Issue

This is a draft PR that addresses #33. There may be a better way to tackle the SNV data, specifically. (More on that below.)

#### Purpose/Implementation

Here, I'm adding an R Notebook where I subset each of the files obtained via the download script. I first read in all of these files and extract the biospecimen identifiers. Then I map the _biospecimen_ identifiers to _participant_ identifiers and find the participant IDs that are present in all files. I randomly select 5 participant identifiers and use these to subset the files. As a result, we have the _full features_ for only 5 participants and the 5 participants are matched across all files as outlined in #33.

#### Soliciting feedback/discussion on the following:

* How I'm subsetting and writing the subset MAF files is probably not ideal. Specifically, I'm reading the file into R with [`maftools::read.maf`](https://rdrr.io/bioc/maftools/man/read.maf.html). This function appears to be opinionated about removing duplicates, etc. by default. I'm subsetting with [`maftools::subsetMaf`](https://rdrr.io/bioc/maftools/man/subsetMaf.html). I then write the subset object to file with an admittedly hack-y custom function `write_maf_file`. It appears to be a valid MAF file, as I can read it back in with `maftools::read.maf` but the [`maf.silent` slot](https://rdrr.io/bioc/maftools/man/MAF-class.html) is missing. 

  It might be better to perform the subsetting for these files upstream by merging only the information for the 5 randomly selected participants. I don't have enough knowledge of what is upstream to say for sure. If we continue to take the `maftools` approach, we may want to deviate from the default behavior of `read.maf`.

* I'm hardcoding the file names here (starting at line 98). I'm wondering if I should figure out a way to use `md5sum.txt` and detect the data type from the file name. OTOH, each data 'type' (e.g., CNV vs. gene expression) has a different structure and there are 2 or less of each type, so I'm not sure digging into this will be that helpful.

#### Notebook HTML for easily viewing: [01-create_subset_files.nb.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3498464/01-create_subset_files.nb.zip)


jharenza:
Thanks for working on this, @jaclyn-taroni! As you alluded to above, I would definitely suggest reading the maf into R as a dataframe and then subsetting the samples from the df and then just writing that output maf as your data subset, rather than subsetting via `maftools::subsetMaf`, as this may remove some of the initial information from the files (eg duplicate variants, synonymous, depending on the version of maftools you use).

jaclyn-taroni:
> As you alluded to above, I would definitely suggest reading the maf into R as a dataframe and then subsetting the samples from the df and then just writing that output maf as your data subset

I initially gave this a try using `#` as the string to identify comments to skip the version information and I got some parsing errors. I can try reading it in with something other than `readr::read_tsv`.

jharenza:
try this `maf <- read.delim("~/OpenPBTA-analysis/data/release-v2-20190809/test.maf", comment.char = "#", sep = "\t", header = T)`

jaclyn-taroni:
```R
> mutect2_df <- readr::read_tsv(file.path(data_dir, mutect2_file), comment = "#")
Parsed with column specification:
cols(
  .default = col_character(),
  Entrez_Gene_Id = col_double(),
  Start_Position = col_double(),
  End_Position = col_double(),
  dbSNP_Val_Status = col_logical(),
  Tumor_Validation_Allele1 = col_logical(),
  Tumor_Validation_Allele2 = col_logical(),
  Match_Norm_Validation_Allele1 = col_logical(),
  Match_Norm_Validation_Allele2 = col_logical(),
  Verification_Status = col_logical(),
  Validation_Status = col_logical(),
  Mutation_Status = col_logical(),
  Sequencing_Phase = col_logical(),
  Sequence_Source = col_logical(),
  Validation_Method = col_logical(),
  Score = col_logical(),
  BAM_File = col_logical(),
  Sequencer = col_logical(),
  Tumor_Sample_UUID = col_logical(),
  Matched_Norm_Sample_UUID = col_logical(),
  t_depth = col_double()
  # ... with 53 more columns
)
See spec(...) for full column specifications.
3559 parsing failures.
 row    col               expected   actual                                     file
1685 PUBMED 1/0/T/F/TRUE/FALSE     29228579 '../../data/pbta-snv-mutect2.vep.maf.gz'
1710 PUBMED 1/0/T/F/TRUE/FALSE     21194676 '../../data/pbta-snv-mutect2.vep.maf.gz'
7495 PUBMED 1/0/T/F/TRUE/FALSE     25643325 '../../data/pbta-snv-mutect2.vep.maf.gz'
7707 AF     no trailing characters ,0.1556  '../../data/pbta-snv-mutect2.vep.maf.gz'
9631 PUBMED 1/0/T/F/TRUE/FALSE     24549054 '../../data/pbta-snv-mutect2.vep.maf.gz'
.... ...... ...................... ........ ........................................
See problems(...) for more details.
```

I'm not familiar enough with this file type to know how big a deal this is. Going this route would still require me to include the string with the version info at the top of the file, correct?

jaclyn-taroni:
Okay let me try your `read.delim` solution @jharenza, standby.

jharenza:
Outputting the string version in the file is not necessary but maybe should remain to keep consistency with the master file in case that comment line interferes with code tests.

jaclyn-taroni:
> Outputting the string version in the file is not necessary but maybe should remain to keep consistency with the master file in case that comment line interferes with code tests.

My thoughts exactly! If we want to remove the version information, we should do that in the original file, not at this step.

jaclyn-taroni:
Okay - I was impatient and tried a `data.table` solution that allows me to read the compressed file relatively quickly:
```R
> mutect2_fread_df <- data.table::fread(file.path(data_dir, mutect2_file),
+                                       skip = 1,  # version info
+                                       data.table = FALSE)
Registered S3 method overwritten by 'R.oo':
  method        from       
  throw.default R.methodsS3
|--------------------------------------------------|
|==================================================|
|--------------------------------------------------|
|==================================================|
```

I'm going to iterate on this route.

jaclyn-taroni:
The `data.table::fread` solution opened a can of worms re: adding the version information. I'm going to try setting `removeDuplicatedVariants = FALSE` and `useAll = TRUE` in `read.maf`. I'm using `maftools` v2.0.15.

jaclyn-taroni:
If the version information comes out of the master file, I am happy to revisit filtering a `data.frame`.

jharenza:
Ok, did the `read.delim` solution not work? It's my go-to! In that case, I would just be sure your `nrow` of the MAF for those patients match before and after!

jaclyn-taroni:
To clarify, the issue was not reading the files into R and skipping the version information. The problem was with my original strategy for building the subset file to be written that included the version information. 

(Side note: I like the `data.table::fread` strategy because it's usually faster -- that may not be the case with compressed files -- and when I run things interactively, it gives me a progress bar so I know it's not hanging.)

I've taken a new approach to writing the subset MAF files in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/56/commits/30220130916400452831ac9e9bd51ee5afed3ad6.

I've also now included a proof of concept notebook to demonstrate this strategy works. It is available here: https://jaclyn-taroni.github.io/openpbta-notebook-concept/analyses/create-subset-files/02-subset_maf_concept.nb.html

jaclyn-taroni:
Okay I think this is ready for review. Note there are a few `TODO` in here for the additional SV file we're expecting. 

@cansav09 can you take a look at the code for clarity and correctness?
@jharenza can you focus on the new strategy for subsetting the MAF files? Please feel free to comment on other things as well of course.

jaclyn-taroni:
> It seems that the all.equal(mutect2_subset, mutect2_post_subset) test is only testing whether two different ways of subsetting mafs using maftools & writing mafs are the same, but I am still not clear as to whether maftools subset is losing information from these files by using maftools::subsetMaf rather than subsetting these from the dataframe prior to reading into maftools as a maf. Can you perform this check?

I'm not sure I followed this correctly.

In https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/56/commits/386af3d789756b5f575397cac29f5fa184950398, I've added a `data.frame`-based check. I believe this is what you are asking for @jharenza.

As of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/56/commits/30220130916400452831ac9e9bd51ee5afed3ad6, I subset the MAF files using a `data.frame` based strategy (see [this line](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/56/files#diff-708e4a03314fa9bd697d8e803b8feee5R327) for the MuTecT2 filtering) rather than using `maftools::subsetMaf` as you recommended in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/56#issuecomment-520972863:

>  I would definitely suggest reading the maf into R as a dataframe and then subsetting the samples from the df and then just writing that output maf as your data subset, rather than subsetting via maftools::subsetMaf

That `data.frame` is what is written to the testing data directory. Changing how I wrote to file allowed me to make your suggested change.

analyses/create-subset-files/01-create_subset_files.Rmd
```diff
@@ -0,0 +1,370 @@
...
+  }
+)
+```
+
+Now we can take the intersection of elements of `participant_id_list` to
+identify participant identifiers represented in each of the files we're trying
+to subset.
+
+```{r}
+participants_in_all <- Reduce(intersect, participant_id_list)
```
cansav09: To prevent people from shooting themselves in the foot, and since there are a lot of functions named "Reduce", you could specify that this is the Reduce from `base::Reduce`. I thought maybe you were using the `BiocGenerics::Reduce` for a sec. People *should* be refreshing their R sessions, but we know this often won't happen.

analyses/create-subset-files/01-create_subset_files.Rmd
```diff
@@ -0,0 +1,370 @@
...
+```
+
+To keep things less verbose below, we'll keep the biospecimen IDs as a vector.
+
+```{r}
+selected_biospecimen_ids <- selected_biospecimen_df %>%
+  dplyr::pull(Kids_First_Biospecimen_ID)
+```
+
+### Subset files
```
cansav09: Are there samples that are not in all the files? Should we have some sort of check or error handling to let the person know that a particular sample they are trying to obtain in the subset doesn't exist in a particular dataset?

analyses/create-subset-files/02-subset_maf_concept.Rmd
```diff
@@ -0,0 +1,108 @@
...
+```
+
+```{r}
+mutect2_post_df <- mutect2_full_df %>%
+  dplyr::filter(Tumor_Sample_Barcode %in% subset_tsb)
+```
+
+```{r}
+all.equal(mutect2_subset_df, mutect2_post_df)
+```
```
cansav09:
```suggestion
sessionInfo()
```

analyses/create-subset-files/01-create_subset_files.Rmd
```diff
@@ -0,0 +1,370 @@
...
+
+_This assumes you are in the top directory of the repository._
+
+## Setup
+
+#### Packages and functions
+
+```{r}
+# R.utils is required to gzip files here
+if (!("R.utils" %in% installed.packages())) {
```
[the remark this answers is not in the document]
jaclyn-taroni: > Do we wanna keep both installation with the Dockerfile as well as installation in this method just in case people are using these notebooks outside of a Docker image?

This was my plan. I don't currently see a downside to leaving it in, particularly when we will often be using a versioned `rocker` container during development.

analyses/create-subset-files/01-create_subset_files.Rmd
jaclyn-taroni: This is beyond the scope of this pull request (and certainly beyond the scope of my knowledge of these data). My intention with this PR is to generate files that can be used as test data for continuous integration, so in this specific case, it does not matter.

More generally, I would expect folks to find the 'missing' samples organically through their own analyses and as we get the first couple ones merged in.

analyses/create-subset-files/02-subset_maf_concept.Rmd
jaclyn-taroni: Good call.

analyses/create-subset-files/01-create_subset_files.Rmd
jharenza: As an FYI, I had added the `experimental_strategy` column in the `pbta-histologies.tsv` file, so all samples with RNA-Seq will have fusions, expression, etc, those with WGS/Panel/WES should have SNVs, those with WGS should have CN and SVs.

jaclyn-taroni:
I'm going to merge this because it is required to get CI fully set up. I would not recommend merging any (true) analyses until we get that set up.

PR #65, opened 2019-08-15
Add pull request template

jaclyn-taroni:
As analyses come in, it will be helpful to have a pull request template for folks to fill in. I've based this largely off of our experience at the CCDL. Specifically, the directions for reviewers and summary of results can help kick-start discussions when necessary.


jaclyn-taroni:
> Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

Can you tell me more about what you want to know when posing this question @jharenza? If the answer is no, it may still be appropriate for reviewers to make suggestions. I'm also concerned that this might unintentionally convey that we only want to see things that are publication ready.

jaclyn-taroni:
> Question - at which point do you foresee people creating figure/table captions? Should this be a part of the analysis text or part of the manuscript text?

I think this should be up to the contributor. Personally, I like to have some text around my figures in my notebooks. The other rationale for including text around results in a pull request is that I've found that it's helpful to have that when writing manuscript text or captions. (Boy am I glad I did this: https://github.com/greenelab/multi-plier/pull/54#issue-239508406)

jharenza:
> > Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
> 
> Can you tell me more about what you want to know when posing this question @jharenza? If the answer is no, it may still be appropriate for reviewers to make suggestions. I'm concerned that this might unintentionally convey that we only want to see things that are publication ready.

I was suggesting this as a 3rd option to the ones you have listed, which are tailored more to works in progress - eg: if the analyst thinks the analysis is publication-ready at the time of the PR. But as an alternative we can add language that we can/will review PRs at any stage - from initial questions/analyses to publication ready figures.

jaclyn-taroni:
Ah I see, thank you. Added in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/65/commits/f42fbe7b3292d74c1295c0a8647242c0dbd23c5e.

jaclyn-taroni:
I picked the language in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/65/commits/57872b0d7109a7443d35b022374c5c2f14c99998 because I expect some contributions may be important for driving how the project moves forward (e.g., a comparison of two methods), but may not end up in the manuscript as a display item.

jharenza:
> I picked the language in [57872b0](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/57872b0d7109a7443d35b022374c5c2f14c99998) because I expect some contributions may be important for driving how the project moves forward (e.g., a comparison of two methods), but may not end up in the manuscript as a display item.

Agree - we may have a spectrum.

PR #70, opened 2019-08-19
Continue CircleCI setup

jaclyn-taroni:
#### Purpose/implementation

Here, I'm setting up our Docker image and Circle CI which results in changes to the Dockerfile and the Circle config file. Also adds `scripts/run_in_docker.sh` and adds the `data` folder to `.dockerignore`.

*In CI:* 
We're downloading the subset data files from S3. Folks can use `./scripts/run_in_docker.sh '<RUN ANALYSIS>'` as a command to add analyses to CI. When something gets successfully merged into master, this should push the Docker image to `ccdlopenpbta/open-pbta:latest`.

#### Issue

N/A


.circleci/config.yml
```diff
@@ -1,16 +1,41 @@
...
       - run:
-          name: Greeting
-          command: echo Hello, world.
+          name: Data Download
+          command: URL=https://open-pbta.s3.amazonaws.com/data RELEASE=testing ./download-data.sh
 
       - run:
-          name: Print the Current Time
-          command: date
+          name: Analysis 1
```
cgreene: What do you think about naming this something more specific? Just thinking out loud here that we might want folks who add their own analyses to have a non-numeric name.
> List Data Directory
or something

.circleci/config.yml
```diff
@@ -1,16 +1,41 @@
 version: 2
 jobs:
-  build:
-    docker:
-      - image: rocker/tidyverse:3.6.0
-
+  run_tests:
```
cgreene: Can you make this `run_analyses` so that it's easier to for external contributors to get into what's happening here?

PR #74, opened 2019-08-20
Fix: volume in run_in_ci.sh

jaclyn-taroni:
This workflow failed because it couldn't find the files: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/127?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

I think these changes may remedy that because of the following functional tests:

* `./scripts/run_in_ci.sh ls` returns the contents of my current working directory (the top directory of the repository).
* If I start running: `./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/create-subset-files/01-create_subset_files.Rmd',  clean = TRUE)"`, it gives me output consistent with running that file from the command line (e.g., starts installing packages).

I can confirm that image built from this changed Dockerfile still allows you to access the RStudio server via method in the README.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,6 @@
 FROM rocker/tidyverse:3.6.0
 MAINTAINER ccdl@alexslemonade.org
+WORKDIR /rocker-build/
 
 RUN apt-get update && apt-get install -y --no-install-recommends apt-utils
 RUN apt-get install dialog apt-utils -y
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
Having taken a closer look at that workflow, this may not be necessary for CI and could be a case of confirmation bias! But it would probably make this script friendlier for running as originally intended when it was named `run_in_docker.sh`.

jaclyn-taroni:
No, fairly sure it's still necessary: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/129?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #75, opened 2019-08-21
Tweak to download in CI

jaclyn-taroni:
I don't think we were getting the subset files as intended because `ls data/testing` included `release-notes.md` which are not in the subset file bucket. Trying this out.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ jobs:
 
       - run:
           name: Data Download
-          command: URL=https://open-pbta.s3.amazonaws.com/data RELEASE=testing ./download-data.sh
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
Ah okay, I think the issue is: 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/af30c37d50db2cc8a0e87f2862879ef0bbe11685/download-data.sh#L13

Because when I run the following locally:
```sh
 URL=https://open-pbta.s3.amazonaws.com/data RELEASE=testing ./download-data.sh
```
and then `ls data/testing` I also see `release-notes.md` but the file is filled with XML that says Access Denied.

Nevermind, closing this.

PR #76, opened 2019-08-22
(PR #2 of 2) Determine concordance between MuTect2 and Strelka2 SNV calls

cansav09:
#### Purpose/implementation

This notebook analyzes the overlap and distinctions between between MuTect2 and 
Strelka2 results after having set the data up in #69 

#### Issue

It closes [issue \# 30 in OpenPBTA](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/30).

#### Directions for reviewers

Due to some of the findings here with Mutect2 data, I suggest we should move 
forward with only Strelka2 data *OR* move forward with the variants that are 
detected by both algorithms. 

1) Do these conclusions make sense based on the analyses shown?
2) Are there any additional analyses we need? 

#### Results

- MuTect2 and Strelka2 detect 55,808 of the same variants. This as defined by having the same Hugo gene symbol, base change, chromosomal start site, and sample of origin (see notebook 01-set-up.Rmd for more details). If moving forward we want only the most reliably called variants, this set of 55,808 variants would give us plenty to work with. 

- MuTect2 and Strelka2 highly agree in their Variant allele frequency (VAF) calculations. 
This is good regardless of our choices moving forward. 

- Variants detected only by MuTect2 have a particularly low VAF compared to
variants detected only by Strelka2. (See VAF Violin plots). 
These density plots suggest some of these MuTect2 calls may be noise. 
Although these low-VAF of MuTect2 could be identifying true variants, our further
analyses would probably benefit from a more robust, higher confidence set of 
variants. 

- MuTect2 also registers dinucleotide and larger variants where Strelka2 seems 
to break these variants into their single nucleotide changes.
In these analyses, these base changes have been grouped together and collectively
called `long_changes`. The higher base resolution of Strelka2, and its ability to parse apart the SNVs from each other, is more useful to us for this particular analyses, as the larger structural variants are better detected in the Manta or LUMPY analyses.

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile. (Requirements are the same as was added for #69). 

## Pull Request Check List:

* [x] Run a linter
* [x] Double checked paths
* [x] Connect the pertinent issues on ZenHub
* [x] Updated Dockerfile and built Docker image successfully


.circleci/config.yml
```diff
@@ -15,8 +15,16 @@ jobs:
...
 
-    #### Add your analysis here ####
-    
+      - run:
+          name: Mutect2 vs Strelka2
+          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/mutect2-vs-strelka2/01-set-up.Rmd', clean = TRUE)"
+
+      - run:
+          name: Mutect2 vs Strelka2 part 2
+          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/mutect2-vs-strelka2/02-analyze-concordance.Rmd', clean = TRUE)"
```
cgreene: Can you add this above?

.circleci/config.yml
```diff
@@ -15,8 +15,16 @@ jobs:
           name: List Data Directory Contents
           command: ./scripts/run_in_ci.sh ls data/testing
 
-    #### Add your analysis here ####
-    
+      - run:
+          name: Mutect2 vs Strelka2
+          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/mutect2-vs-strelka2/01-set-up.Rmd', clean = TRUE)"
```
cgreene:
```suggestion
          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/mutect2-vs-strelka2/01-set-up.Rmd', clean = TRUE)"
          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/mutect2-vs-strelka2/02-analyze-concordance.Rmd', clean = TRUE)"
```

.circleci/config.yml
cgreene: Then you could delete this extra analysis. It'd be nice to have all of the runs for one analysis under the same header.

cansav09: Will do.

.circleci/config.yml
cansav09: This doesn't seem to work. It doesn't like having two "commands".

cansav09: I put it in the same `command` and that works.

cgreene: :today_i_learned:

jaclyn-taroni: I think we want to do this:
* As individual notebooks or scripts are coming in for an analysis module, they should be added to CI as multiple runs. This validates that they can be executed at the time of review.
* Once an analysis has been reviewed and validated, the last PR for an analysis module should be a shell script used to run everything in the module. That PR also replaces the multiple runs in the config file with a single run that runs the shell script.

cgreene: I like this idea a lot 👍

cansav09: Sounds good!

cgreene:
@jharenza : would you be worried in that scenario that you'd be including variants that are labeled pathogenic but that aren't really there? That would be my biggest concern, since Mutect2 looks very aggressive at calling even low VAF things. These seem more likely to be enriched with errors (whether they are pathogenic or not). If we're saying that we should retain anything from Mutect2 labeled pathogenic, then we're basically just saying we should do the union of Strelka2 and Mutect2.

jharenza:
> @jharenza : would you be worried in that scenario that you'd be including variants that are labeled pathogenic but that aren't really there? That would be my biggest concern, since Mutect2 looks very aggressive at calling even low VAF things. These seem more likely to be enriched with errors (whether they are pathogenic or not). If we're saying that we should retain anything from Mutect2 labeled pathogenic, then we're basically just saying we should do the union of Strelka2 and Mutect2.

I think it depends on whether these are likely true subclonal mutations or artifacts. I'm not clear how many of the potentially damaging mutations are in hotspot/are recurrent (COSMIC) vs one-off, potentially private mutations. Eg: if they are mostly the latter, then it doesn't make sense to keep them. We can look at some of these low level nonsynonymous damaging mutations in IGV using the BAM files here: https://cavatica.sbgenomics.com/u/cavatica/pbta-cbttc/ or https://cavatica.sbgenomics.com/u/cavatica/pbta-pnoc003/?

cgreene:
If someone is willing/able to look at the set in the BAM files then I think we should ask for @cansav09 to produce a list of pathogenic ones that are Mutect2 only, Strelka2 only, and both callers.

How many would folks be willing to look at? We could ask @cansav09 to create files that dump up to that number, spread across the three categories? Perhaps even better if we randomly shuffle the three lists together before handing them to whoever is looking by hand, and we keep that person blinded.

jharenza:
> If someone is willing/able to look at the set in the BAM files then I think we should ask for @cansav09 to produce a list of pathogenic ones that are Mutect2 only, Strelka2 only, and both callers.
> 
> How many would folks be willing to look at? We could ask @cansav09 to create files that dump up to that number, spread across the three categories? Perhaps even better if we randomly shuffle the three lists together before handing them to whoever is looking by hand, and we keep that person blinded.

Let me discuss with @yuankunzhu today during our sprint planning. @cansav09 can you give us an N of variants that would need to be inspected?

cansav09:
> @cansav09 can you give us an N of variants that would need to be inspected? 

@jharenza Couple clarification questions I have after this discussion:

1) Did you want these numbers for the non-synonymous variants only? Or all variants? 

2) Also did you want “possibly_damaging” category variants to be included or only “probably_damaging” variants? 

3) Or did you prefer I use something besides PolyPhen, like SIFT or ClinVar? 

4) For linking back to BAM files what information would you like included in these lists to make this easiest to trace back? Or should I just include all the information and you guys will figure it out?

jharenza:
@cansav09 
1) Was thinking non-synonymous only, because I am not sure if any synonymous would be damaging - I guess we can do a quick check on that.
2&3) Possibly and probably damaging/deleterious from at least 1 of the three tools you mention
4) If you gave the variant coordinates, nucleotide change, and biospecimen ID (which is probably the TSB), then that should suffice.

Issue #77, opened 2019-08-22
How we add completed analysis modules to CI


jaclyn-taroni:
Context is https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/76#discussion_r316751701, to quote that here

> * As individual notebooks or scripts are coming in for an analysis module, they should be added to CI as multiple runs. This validates that they can be executed at the time of review.
> * Once an analysis has been reviewed and validated, the last PR for an analysis module should be a shell script used to run everything in the module. That PR also replaces the multiple runs in the config file with a single run that runs the shell script.

I propose that we try this model out with @cbethell's addition of #5 (relevant PRs: #54, #55). If successful, we can update the instructions in the README.

PR #55, opened 2019-08-13
Addition of `02-multilayer-pie.R` and plots

cbethell:
This PR continues to address issue #5 

### Additions: 
- `02-multilayer-pie.R` script produces the interactive treemap and pie chart HTML plots, which display the `broad_histology`, `short_histology`, and `disease_type_new` across the dataset. `disease_type_new` will be replaced with `molecular_subtype` once we have this data. 
- The plots `histology-treemap.html` and `histology-pie.html` can be found in `sample-distribution-analysis/plots`
- The data used to produce the plots can be found at `sample-distribution-analysis/results/sunburst_plot_df.tsv`

I have attached the links to the interactive plots below for convenient access. The links are currently attached to my personal Github Pages and will be transferred to `alexslemonade.github.io`. 

[histology-treemap](https://cbethell.github.io/histology-pie-data/histology-treemap.html)
[histology-pie](https://cbethell.github.io/histology-pie-data/histology-pie.html)

### Questions: 
- Do the plots appear to represent the data well?
- Does the interactive treemap plot appear to add significant value? In other words, would we want to keep this plot going forward? 
- Are the color schemes of the plots satisfactory?  
- Should there be still PDF versions of each plot living in the `sample-distribution-analysis/plots` directory? 
- Should `sample-distribution-analysis/results/sunburst_plot_df.tsv` be moved to `scratch`?

### Pull Request Check List:
* [x] Run a linter
* [x] Set the seed (if applicable).
* [x] Comments and/or documentation up to date.
* [x] Double check your paths.
* [x] Check headline numbering.
* [x] Spell check any Rmd file or md file.
* [x] Restart R and run all notebooks fresh and save.

PR #78, opened 2019-08-23
Addition of `run-sample-distribution.sh` 

cbethell:
#### Purpose/implementation

This shell script is intended to run everything in the `sample-distribution-analysis` module.
This PR replaces multiple runs in `circleci/config.yml` with a single run that runs the shell script. 

#### Issue

This PR addresses issue #77.  

#### Directions for reviewers

- `Rplots.pdf` was also added to .gitignore (it is produced when we run the shell script).
-  Should this shell script be noted in `analyses/sample-distribution-analysis/README/md`?

#### Results

The results here would mimic the combined results and plots produced in PR #54 and PR #55. 
They include:
- `analyses/sample-distribution-analysis/plots/distribution_across_cancer_types.pdf`
- `analyses/sample-distribution-analysis/plots/histology-pie.html`
- `analyses/sample-distribution-analysis/plots/histology-treemap.html`
- `analyses/sample-distribution-analysis/results/disease_expression.tsv`
- `analyses/sample-distribution-analysis/results/plots_df.tsv`
- `analyses/sample-distribution-analysis/results/primary_sites_counts.tsv`


jaclyn-taroni:
Re: the two questions for reviewers

> `Rplots.pdf` was also added to .gitignore (it is produced when we run the shell script).

This ends up in the root directory of the repository because of how this is run and that's why it is ignored in the `.gitignore` in the root of the repository, is that correct? If so, can you put a comment in `.gitignore`.

> Should this shell script be noted in `analyses/sample-distribution-analysis/README`

Yes I would put something similar to the Usage information in my and @cansav09's notebooks in that README.

cgreene:
What about putting the shell script in the analysis folder? I think I prefer that unless it creates problems.

jaclyn-taroni:
Agree with that @cgreene, I was headed there next. Even if it was placed there, if it's run from the top directory the `Rplots.pdf` may still end up in the root directory (completely untested).

run-sample-distribution.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
jaclyn-taroni: We should move this to within `analyses/sample-distribution-analyses` per @cgreene's point.

.gitignore
```diff
@@ -1,3 +1,4 @@
 .DS_Store
 .Rhistory
 .spelling  # The dictionary of words accrued by `mdspell` for spellcheck
+Rplots.pdf
```
jaclyn-taroni: Depending on the outcome of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/78#issuecomment-524402332...
* `Rplots.pdf` ends up in the analysis module itself = put this in the module specific `.gitignore` 
* `Rplots.pdf` is in the root directory = change this to `/Rplots.pdf` (see [this StackOverflow post](https://stackoverflow.com/questions/3637660/how-to-exclude-file-only-from-root-folder-in-git)) to avoid affecting other modules

cbethell:
> This ends up in the root directory of the repository because of how this is run and that's why it is ignored in the `.gitignore` in the root of the repository, is that correct? If so, can you put a comment in `.gitignore`.

Yes, that is correct. I will add the relevant comment. 

> Yes I would put something similar to the Usage information in my and @cansav09's notebooks in that README.

Great, I will take a look at these notebooks and follow suit. 

> Agree with that @cgreene, I was headed there next. Even if it was placed there, if it's run from the top directory the Rplots.pdf may still end up in the top directory (completely untested).

I will test this out now.

.gitignore
cbethell: > Depending on the outcome of [#78 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/78#issuecomment-524402332)...
> 
> * `Rplots.pdf` ends up in the analysis module itself = put this in the module specific `.gitignore`
> * `Rplots.pdf` is in the root directory = change this to `/Rplots.pdf` (see [this StackOverflow post](https://stackoverflow.com/questions/3637660/how-to-exclude-file-only-from-root-folder-in-git)) to avoid affecting other modules

`Rplots.pdf` was produced in the root directory so I made the latter change suggested above and added a comment.

PR #79, opened 2019-08-23
Make it such that sample distribution scripts can be run from anywhere in project

jaclyn-taroni:
#### Purpose/implementation

This is the possibly overly complicated thing mentioned https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/78#pullrequestreview-279222422. It occurred to me that someone might navigate to the sample distribution directory and try to run the Rscripts, i.e.:

```sh
cd analyses/sample-distribution-analyses
Rscript 01-filter-across-types.R
```

This would raise an error because this is supposed to be run from the root of the project and the data, results, etc. directories are hardcoded as such. 

This prompted me to try a strategy like what we have in [lines 3-8 of `./scripts/run_in_ci.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e8e23b53d9284a859073d30f9fe1d96d6f4d405f/scripts/run_in_ci.sh#L3)

@cbethell you were onto something when you mentioned `setwd()` in person on Wednesday. Using it is [generally considered](http://plantarum.ca/code/setwd/) a [bad idea](https://www.tidyverse.org/articles/2017/12/workflow-vs-script/).

By adding the changes I have to the shell and R scripts in this module it ensures:
1. The shell script will always run as it is being called in the directory it lives in = will always be run as if it is in the same directory that the Rscripts it runs are in. The [`--vanilla` addition](https://stat.ethz.ch/R-manual/R-devel/library/base/html/Startup.html) is probably more important if this is run locally.
2. The working directory for the Rscripts will always be the root of the project, as is intended. We can use `rprojroot::find_root(rprojroot::has_dir(".git"))` to find the root of the project. The criterion is where the `.git` folder lives. (The irony of using this in conjunction with `setwd()` is not lost on me.)

#### Issue

N/A -  I don't think this should be a requirement or included as part of the instructions for #77. I was curious wanted to see if it worked given how this project structure is a bit different from how I usually set up my own projects.

#### Directions for reviewers

If you are willing to check out my branch, I would try running these scripts using different places as your working directory.


jaclyn-taroni:
With https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/79/commits/0a26f6a573e04d28203adc19e65f6ecf18ca194e, I've taken out `setwd()` and used a strategy that is more consistent with the intended use of `rprojroot`.

analyses/sample-distribution-analysis/.gitignore
```diff
@@ -1,2 +1,4 @@
 histology-pie_files
 histology-treemap_files
+# The nonessential output produced in the root directory of this repository when we run `analyses/sample-distribution-analysis/run-sample-distribution.sh`
```
cgreene:
```suggestion
# The nonessential output produced in the root directory of this repository when we run `run-sample-distribution.sh`
```

cgreene: I think that since it's in the same directory, the path should only be the script name.

jaclyn-taroni: Good idea -- added in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/79/commits/148f37617b15ab575980249f65ba18c020eb3b17

jaclyn-taroni:
This is ready for review.

jaclyn-taroni:
From CircleCI output:
```
[WARNING]: Empty continuation line found in:
    RUN apt-get update -qq && apt-get -y --no-install-recommends install     libgdal-dev     libudunits2-dev     libmagick++-dev
[WARNING]: Empty continuation line found in:
    RUN apt-get update -qq && apt-get -y --no-install-recommends install     && install2.r --error     --deps TRUE     gdalUtils     leafem     lwgeom     stars     leafpop     plainview     sf     mapview
[WARNING]: Empty continuation line found in:
    RUN apt-get update -qq && apt-get -y --no-install-recommends install     && install2.r --error     --deps TRUE     R.utils     treemap     d3r     hexbin     rprojroot
[WARNING]: Empty continuation lines will become errors in a future release.
```

What I've read suggested that is due to the presence of comments within the RUN statements, so I moved them in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/79/commits/dcd9b23ac52f19414106d07134af0fddcdd3059f while I was at it.

PR #91, opened 2019-08-30
Fix gene expression RDS testing files

jaclyn-taroni:
#### Purpose/implementation

I used the wrong function to write the gene expression RDS files to file, causing this failure: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/244?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This fixes that. This has also been rerun on the V3 release files and I'm about to put them on S3.


jaclyn-taroni:
It fixed the original problem, but we may want to consider using a larger number of samples: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/246

cgreene:
I agree that it may be better to include more samples. The number is currently very small to make CI as fast as possible. It's good that the perplexity issue is caught in the sense that it's a parameter that doesn't generalize to small datasets, but probably if one can't t-SNE we've got it too small :)

jaclyn-taroni:
@cgreene I added the last two commits to make this easier to use with a higher number or different subset of matched participants while I'm away next week. I can confirm that https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/91/commits/a697cbdde2a9d54e6b0bb3df9d452ba22de11706 gives me the same checksums.

cgreene:
I'm going to propose 50 samples. Should be fast, and also hopefully enough for things like building low-dimensional representations. We can continue to evaluate it as new analyses come in.

jaclyn-taroni:
Ok I can do that right now and put them up on S3.

jaclyn-taroni:
A notebook is probably not best for this now that I have a better idea about the file formats, but will have to do for now.

jaclyn-taroni:
I think we're gonna need a bigger boat: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/262

cgreene:
We can get a bigger boat!
https://circleci.com/docs/2.0/configuration-reference/#resource_class

cgreene:
I filed https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/98 which asks for a bigger boat.

PR #99, opened 2019-09-01
Set resource class to large

jaclyn-taroni:
#### Purpose/implementation

Related: #91 

If we increase the size of the subset files, we'll need more RAM for CI.


jaclyn-taroni:
This works when I set the number of matched participants to 25 (upcoming commit over on #91).

cgreene:
To close the loop - unless we're pulling a docker container and working off of that, the highest one can go is `large`. 😭

PR #83, opened 2019-08-27
Addition of `01-transcriptomic-analysis-prep.Rmd`

cbethell:
#### Purpose/implementation

The purpose of this draft PR is to conduct an unsupervised transcriptomic analysis using three dimensionality reduction techniques. The techniques used include PCA, t-SNE, and UMAP. 

You can find the html output of this notebook [here](https://cbethell.github.io/open-pbta-output/01-transcriptomic-analysis-prep.nb.html).

#### Issue

This draft PR addresses issue #9.

#### Directions for reviewers

Given the first set of visualizations produced, it was suggested that I backtrack using the `Kids_First_Participant_ID` and `experimental_strategy` information to find the sequencing strategies used for each observation. The unique strategies were found to be `WGS`, `WXS`, `Panel`, and `RNA-Seq` (which is renamed and treated as `Unknown`). 

1. My attempt to correct for batch effects did not produce significant results according to the visualizations. Should I remove this attempt and open a separate issue for this purpose? 
2. I have made this a draft PR because I believe the code can be further simplified. Please feel free to provide suggestions you believe may be useful on this matter. 
3. Are there any additional steps I should add to this analysis? 

#### Results

- Visualizations of the reduction scores for each of the dimension reduction techniques executed in this notebook are available on the html output and in the directory `analyses/unsupervised-trancriptomic-analysis/plots`

- The .PDF files in the `plots` directory include visualizations before and after the attempt to correct for batch effects. Those beginning with `meta_grid` represent the data before batch correction attempt, while those beginning with `no_batch_grid` represent the data after the batch correction attempt. 

#### Docker and continuous integration

_Check all those that apply or remove this section if it is not applicable._

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Pull Request Check List:

* [ ] Run a linter (not done as yet) 
* [x] Set the seed (if applicable)
* [x] Comments and/or documentation up to date
* [x] Double checked paths
* [x] Spell check any Rmd file or md file
* [x] Restart R and run all notebooks fresh and save
* [x] Connect the pertinent issues on ZenHub
* [x] Updated Dockerfile and built Docker image successfully

PR #100, opened 2019-09-09
Addition of `02-transcriptomic-analysis-plotting.Rmd`

cbethell:
#### Purpose/implementation

The purpose of this draft PR is to plot the dimension reduction score data obtained in the previous notebook found in draft PR #83. 

You can find the html output of this notebook [here](https://cbethell.github.io/open-pbta-output/02-transcriptomic-analysis-plotting.nb.html).

#### Issue

This draft PR addresses issue #9.

#### Directions for reviewers

- Do the plots sufficiently represent the data? Would you recommend the use of a different R package to produce the plots? 
- Please note that I attempted to increase the size of the text on the legend of the plots, but this resulted in part of the legend being trimmed. I am still looking into the best solution for this. 

#### Results

- The `PDF` files in the `plots` directory include plots for each expression data set (RSEM and Kallisto) across the three dimension reduction techniques score datasets produced in `01-transcriptomic-analysis-prep.Rmd`.

#### Pull Request Check List:

* [x] Run a linter (not done as yet) 
* [x] Double checked paths
* [x] Connect the pertinent issues on ZenHub
* [x] Updated Dockerfile and built Docker image successfully


jaclyn-taroni:
The selection strategy issue has come up before on #83: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/83#issuecomment-525357478 and https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/83#issuecomment-525386784. We should file an issue for exploring how to deal with this. Please feel free to file that @cbethell or @jashapiro. I can also circle back and file it.

cbethell:
> The selection method definitely matters! That is probably worth highlighting in the PR notes. It also suggests that we might want to repeat these analyses with the two methods separated? And we definitely want to think about how it is affecting all other analyses.

@jashapiro The last 2 commits I pushed address this comment as well as the following comments in your review. Please let me know what you think now that I have separately plotted the `poly-A` selected data from the `stranded` ribosomal deplete data. 

Also, I attempted to create an optimal color palette for the plots but I do not believe I have done that just yet. So please feel free to provide suggestions here.

analyses/transcriptomic-dimension-reduction/01-transcriptomic-analysis-prep.R
```diff
@@ -155,8 +163,10 @@ dimension_reduction_wrapper <- function(transposed_expression_matrix,
                                         method,
                                         seed,
                                         metadata_df,
-                                        filename,
+                                        model_filename,
+                                        scores_filename,
                                         output_directory,
+                                        strategy,
```
jashapiro: Add strategy to the args list below.

analyses/transcriptomic-dimension-reduction/01-transcriptomic-analysis-prep.R
```diff
@@ -304,75 +316,241 @@ transposed_kallisto_data <- t(exp_kallisto)
 #### RSEM ----------------------------------------------------------------------
 
 # Run `dimension_reduction_wrapper` function using PCA
+rsem_pca_aligned_scores_poly <-
```
jashapiro: Probably call this `rsem_pca_aligned_scores_polyA` as there are many kinds of poly, and I was confused for a sec. (see also other places this appears)

analyses/transcriptomic-dimension-reduction/02-transcriptomic-analysis-plotting.Rmd
```diff
@@ -0,0 +1,650 @@
...
+  dir.create(plots_dir)
+}
+```
+
+Read in the tsv files containing the dimension reduction scores aligned with 
+the metadata, which were produced in `01-transcriptomic-analysis-prep.R`.
+```{r, message = FALSE}
+### RSEM -----------------------------------------------------------------------
+poly_rsem_pca_aligned_df <-
+  (readr::read_tsv(file.path(
```
jashapiro: Extra parens here (and in all the read_tsv lines below)

analyses/transcriptomic-dimension-reduction/02-transcriptomic-analysis-plotting.Rmd
```diff
@@ -0,0 +1,650 @@
...
+(poly-A selection versus stranded(ribosomal depletion)).
+This can be further determined using the `RNA_library` column of the
+metadata to attempt to correct batch effects as suggested by the selection
+strategy grid of plots.
+
+```{r, fig.width = 10, fig.height = 18}
+# Run the `plot_dimension_reduction` function to plot each set of dimension
+# reduction scores using `broad_histology` to color points
+rsem_pca_plot_histology <-
+  plot_dimension_reduction(rsem_pca_aligned_df, "broad_histology", 1, 5)
```
jashapiro: I know I said I was curious about other PCs, but lets just stick to 1 & 2 for the main plots here (now that I've seen more).

cbethell:
@jaclyn-taroni Do you have any additional suggestions to make on this PR?

jaclyn-taroni:
It looks like you are filtering by strategy after creating a model using all the data. I would expect that you might want to filter by strategy before performing dimension reduction to get a better look downstream. Otherwise you could perform the filtering on the aligned `data.frame` in your plotting notebook. Thoughts @cbethell @jashapiro ?

jashapiro:
> It looks like you are filtering by strategy after creating a model using all the data. I would expect that you might want to filter by strategy before performing dimension reduction to get a better look downstream. Otherwise you could perform the filtering on the aligned `data.frame` in your plotting notebook. Thoughts @cbethell @jashapiro ?

I agree, filtering first, then running the dimension reduction seems the way to go. I don't know enough about the behaviors of tSNE and UMAP, but for PCA it can dramatically change the results to eliminate a set of outliers like that.

jaclyn-taroni:
My guess is that there are probably too few polyA samples in the subset data run in CI. Does this run successfully locally on the polyA samples and if so, what perplexity do you use?

cbethell:
> My guess is that there are probably too few polyA samples in the subset data run in CI. Does this run successfully locally on the polyA samples and if so, what perplexity do you use?

This is also my guess. It does run successfully locally, using perplexity = 10.

jaclyn-taroni:
What happens if you set it to 1 in CI? My next step after that would be to snag the subset histologies file from S3 (you can find the bucket info in the CI config file in the data download step).

cbethell:
> What happens if you set it to 1 in CI? My next step after that would be to snag the subset histologies file from S3 (you can find the bucket info in the CI config file in the data download step).

Looks like I will be using the subset histologies file from S3.

jaclyn-taroni:
I think the move for now, which we should use judiciously/try to avoid in the future, is to comment out the polyA stuff so the build isn’t broken. I say this because the first script is now in the 500 lines long range and that is a hint to me that it’s probably time for a bit of a refactor, which is beyond the scope of this PR but should be addressed before the shell script + README get put together. I can draft out what I think the refactor would look like.

jaclyn-taroni:
@jashapiro are you okay with this going forward and getting merged? The caveat is that I am going to take a pass at a refactor once this PR lands.

Issue #103, opened 2019-09-10
vardict and lancet variant calls on deck


jharenza:
#### File(s)

_What data file(s) does this issue pertain to?_

#### Release

_What release are you using?_

#### Link to [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/)

_Put a link to the relevant section of the OpenPBTA manuscript here._

#### Question/issue

_Put your question or report your issue here._
We are currently generating lancet and vardict calls and will release in V5

jharenza:
update on this - we have vardict completed and annotated and 100 more samples to run through lancet + annotation - will have these for v5 release on 09/24/19

jaclyn-taroni:
@jharenza was this closed via #127?

jharenza:
> @jharenza was this closed via #127?

yup, closed!

Issue #11, opened 2019-07-12
Planned Analysis: Cosmic Mutational Signatures


PichaiRaman:
Characterize PBTA with COSMIC Mutational Signatures and correlate back to cancer type, genetic lesions, and pathway activity. Visualization of this would be important as well.

jharenza:
Though in pediatric cancers, these should not be as prevalent, could also consider environmental signatures as reported [here](https://www.cell.com/cell/pdf/S0092-8674(19)30263-6.pdf).

cansav09:
Since I've been getting more comfortable with the variant data, I can also work on setting this analysis up.

jharenza:
@cansav09 great! I have used the [deconstructSigs](https://github.com/raerose01/deconstructSigs) package - super easy and will calculate signatures for each sample individually (some algorithms only look across a group and those are less informative). There are some newer COSMIC signatures with the latest COSMIC release, but I have not had a chance to look into those at all. Can definitely help with feedback along the way!

cansav09:
@jharenza, thanks for that recommendation I will look into it. Meanwhile, if you wanna see how I’m comparing our data to the COSMIC mutations for the first and admittedly rough pass, you can take a look at the documentation I’ve prepared on PR 114.

cansavvy:
So I have mutational signature information for all the samples and made a graph for each sample,  but how would we like this information presented overall?  (Since looking over ~980 graphs is not practical.) How do we feel about something like the graph in [Alexandrov et al, 2018](https://www.biorxiv.org/content/10.1101/322859v2)? In our case we would have the signatures as rows and the histology as columns, then report the median weight for each histology's  signatures as reported by `deconstructSigs::whichSignatures`? 
![Screen Shot 2019-11-04 at 9 30 36 AM](https://user-images.githubusercontent.com/23458084/68128949-af63d780-fee6-11e9-8a10-c68a907dc27a.png)

cansavvy:
Then when we have molecular subtype information we could make a version with that information instead of the histology?

cansavvy:
We can also present the data like this: ![Screen Shot 2019-11-04 at 9 45 17 AM](https://user-images.githubusercontent.com/23458084/68129597-d4a51580-fee7-11e9-8b71-9e0a25f23cb6.png)

jaclyn-taroni:
[...]

PR #118, opened 2019-09-20
PR 2 of n for SNV Caller Analysis: Wrangling functions

cansav09:
#### Purpose/implementation

This is the second PR in a series that are preparing a pipeline to perform an initial analysis and calculation of each MAF file for each SNV caller. At this point, the SNV callers that will be evaluated are MuTect2, Strelka2, VarDict, and Lancet. 

This includes all the functions needed to calculate and wrangle the MAF files. 

#### Issue

#### Directions for reviewers

* Do the functions make sense? Their naming? Their arguments? 
* Should I break any out into separate smaller functions? 
* Is the documentation thorough enough? Are their gaps anywhere that should be filled in? 

#### Results

These things will be in a future PR. 


analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+  #              to GenomicRanges::findOverlaps's `maxgap` argument.
+  #
+  # Read in the genomic regions annotation object
+  annotations <- readr::read_rds(annot_rds)
+
+  # Use custom function to turn our MAF data into a GRanges
+  maf_ranges <- maf_to_granges(maf_df)
+
+  # Intersect the regions we read in with the annotations
+  annot_matches <- GenomicRanges::findOverlaps(annotation_obj, maf_ranges,
```
jaclyn-taroni:
```suggestion
  annot_matches <- GenomicRanges::findOverlaps(
    annotation_obj, 
    maf_ranges,
```

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+  # Args:
+  #   maf_df: MAF formatted data that has been turned into a data.frame and has
+  #           been run through `set_up_variables`
+  #   bp_window: how many base pairs away can it be from the BED region to still
+  #              be included? Default is 0 bp. This argument gets forwarded
+  #              to GenomicRanges::findOverlaps's `maxgap` argument.
+  #
+  # Turn both these datasets into GRanges
+  maf_granges <- maf_to_granges(maf_df)
+  cosmic_granges <- maf_to_granges(suppressMessages(readr::read_tsv(cosmic_clean_file)))
```
jaclyn-taroni: Is there no argument to `readr::read_tsv` that will accomplish the same thing as `suppressMessages`?

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+
+  # Find the overlap of the MAF and COSMIC mutations.
+  overlap <- GenomicRanges::findOverlaps(maf_granges, cosmic_granges,
```
jaclyn-taroni:
```suggestion
  overlap <- GenomicRanges::findOverlaps(
    maf_granges, 
    cosmic_granges,
```

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+  #
+  # Args:
+  #   maf_df: maf data that has been turned into a data.frame.
+  #   annotation_obj: an annotation object with the desired annotations.
+  #   bp_window: how many base pairs away can it be from the BED region to still
+  #              be included? Default is 0 bp. This argument gets forwarded
+  #              to GenomicRanges::findOverlaps's `maxgap` argument.
+  #
+  # Read in the genomic regions annotation object
+  annotations <- readr::read_rds(annot_rds)
```
jaclyn-taroni:
```suggestion
  annotations <- readr::read_rds(annotation_obj)
```

analyses/snv-callers/util/wrangle_functions.R
jaclyn-taroni: Should this be `annotations` instead of `annotation_obj`?

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+  maf_ranges <- maf_to_granges(maf_df)
+
+  # Intersect the regions we read in with the annotations
+  annot_matches <- GenomicRanges::findOverlaps(annotation_obj, maf_ranges,
+    maxgap = bp_window
+  )
+
+  # Extract the annotation from the regions that overlap the mut
+  annot <- data.frame(
+    annotation_obj[annot_matches@from]@elementMetadata@listData,
```
jaclyn-taroni: Same comment about `annotation_obj` vs. `annotations`

analyses/snv-callers/util/wrangle_functions.R
jaclyn-taroni: Where is `cosmic_clean_file` coming from? Is your plan to have it be in your global environment when you use this function? This can be sourced _anywhere_ so `cosmic_clean_file` should be passed as an argument to this function.

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+    dplyr::filter(experimental_strategy == "WXS")
+
+  # Error catcher
+  if (nrow(maf_wxs) == 0) {
+    warning("No WXS samples found underneath column 'experimental_strategy'
+            double check filtering steps and data.")
+  }
+
+  # Find the overlap of the BED regions and the mutations.
+  wxs_maf_granges <- GenomicRanges::GRanges(
```
jaclyn-taroni: Can you explain why you're not using `maf_to_granges` here?

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+      start = maf_df$Start_Position,
+      end = maf_df$End_Position
+    ),
+    strand = maf_df$Strand,
+    mcols = maf_df
+  )
+}
+
+wxs_bed_filter <- function(maf_df, wxs_bed = NULL, bp_window = 0) {
+  # Given a MAF formatted data.frame and a BED regions data.frame; filter out
```
jaclyn-taroni: Can you add documentation about what this returns please? I didn't find what is being returned to be intuitive.

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+      end = maf_wxs$End_Position
+    )
+  )
+
+  # Find the overlap of the BED regions and the mutations.
+  overlap <- GenomicRanges::findOverlaps(wxs_maf_granges, wxs_bed_granges,
+    maxgap = bp_window
+  )
+
+  # Calculate of ratio of variants in this BED
```
jaclyn-taroni: I am not the most familiar with `GenomicRanges` but I would not intuit that the `from` slot to contain "variants in this BED" information. Can you add a comment?

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+  # TMB = # variants / size of the genome or exome surveyed
+  #
+  # Args:
+  #   maf_df: maf data.frame that has been turned into a data.frame and has
+  #           WXS mutations filtered using `wxs_bed_filter` function.
+  #   wgs_size: genome size in bp to be used for WGS samples
+  #   wxs_size: genome size in bp to be used for WGS samples
+  #
+  # Make a genome size variable
+  tmb <- maf_df %>%
```
jaclyn-taroni: Remind me - does the `experimental_strategy` and `short_histology` information come with the MAF files or is that part of your set up?

cansav09:
> Because these wrangle functions can be sourced from anywhere, you should be careful about scope, i.e., you don't want to assume that something will be in the global environment when you run them. I think I caught one issue along these lines but there may be more. Can you go back through and check for any addition cases like this?

Yes, this definitely is an existing problem due to how these functions originally came to be. I started to try to address those, but wasn't sure how much that would be a problem for this use. Since you've mentioned it though now too, I'll go back and make sure that these functions separately from things in the global environment. Right now these scripts are still a little too interdependent.

analyses/snv-callers/util/wrangle_functions.R
cansav09: Yes. This whole argument handling is messed up. I'm changing it so `annotation_file` is for specifying a file.path to the annotation RDS.

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,260 @@
...
+}
+
+wxs_bed_filter <- function(maf_df, wxs_bed = NULL, bp_window = 0) {
+  # Given a MAF formatted data.frame and a BED regions data.frame; filter out
+  # any variants of the MAF df that are not within the BED regions.
+  #
+  # Args:
+  #   maf_df: maf data that has been turned into a data.frame. Can be a maf object
+  #           that is subsetted using `@data`.
+  #   wxs_bed: a data.frame that has the windows to be used for TMB calculation.
```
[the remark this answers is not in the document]
cansav09: I don't have the exact BED regions files yet, but the plan is to have a script that sets them up and now I'm changing this argument so that it is a file.path to that BED file.

analyses/snv-callers/util/wrangle_functions.R
cansav09: > Is there no argument to `readr::read_tsv` that will accomplish the same thing as `suppressMessages`?

I had looked in the docs to see if there was an argument for this and didn't see anything but Dr. Google suggests that `message = FALSE` should do the trick.

analyses/snv-callers/util/wrangle_functions.R
cansav09: No reason besides I forgot about my own function I made for this.

analyses/snv-callers/util/wrangle_functions.R
cansav09: I'll add more documentation surrounding all the GRanges steps.

analyses/snv-callers/util/wrangle_functions.R
cansav09: Part of the set up, but because these function requires it I've added it as an option in the function that sets up these data so that is more apparent.

analyses/snv-callers/util/wrangle_functions.R
jaclyn-taroni: Thank you, much appreciated!

analyses/snv-callers/util/wrangle_functions.R
cansav09: Dr. Google lied. This does not work.

jaclyn-taroni: Dr. Google! :shakes fist:

cansav09: This as an argument works: `col_type = readr::cols()`.

cansav09:
Okay. I think this is ready for another look, @jaclyn-taroni although I will not be surprised if some of this still needs to be arranged later as more pieces of this analysis are added and likely changed. I re-ran it through a linter as well as re-ran my dummy datasets through the whole thing after implementing these changes and things are still working.

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,334 @@
...
+  #              to GenomicRanges::findOverlaps's `maxgap` argument.
+  #
+  # Returns:
+  # The same MAF formatted data.frame given but having filtered out the WXS
+  # mutations that lie outside the supplied WXS BED regions.
+
+  # Read in the BED regions file and make sure the column names are the same
+  # as the MAF column format names
+  wxs_bed_ranges <- readr::read_tsv(wxs_bed_file, col_names = FALSE) %>%
+    dplyr::rename(Chromosome = X1, Start_Position = X2, End_Position = X3)
```
jaclyn-taroni: 👍

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,334 @@
...
+  #   bp_window: how many base pairs away can it be from the BED region to still
+  #              be included? Default is 0 bp. This argument gets forwarded
+  #              to GenomicRanges::findOverlaps's `maxgap` argument.
+  #
+  # Returns:
+  # A large data.frame that contains every mutation and every region type it
+  # overlaps. Genomic region types are noted in the `type` column.
+  #
+  # Read in the genomic regions annotation object
+  annotation_ranges <- readr::read_rds(annot_rds)
```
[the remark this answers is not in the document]
cansav09: Ah! You're right. I messed it up again. Don't know why this argument is confusing me so much.

analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -0,0 +1,334 @@
...
+  wxs_bed_granges <- maf_to_granges(wxs_bed_ranges, strand = FALSE)
+
+  # Obtain a MAF data.frame of only the WXS samples since this filter will only
+  # be applied to those samples.
+  maf_wxs <- maf_df %>%
+    dplyr::filter(experimental_strategy == "WXS")
+
+  # Error catcher in case there are no `WXS` samples
+  if (nrow(maf_wxs) == 0) {
+    warning("No WXS samples found underneath column 'experimental_strategy'
```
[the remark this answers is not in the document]
cansav09: Ah. Probably should just stop it.

cansav09:
> I had a few more comments @cansav09. After that, I think it's better to see what shakes out as we review more parts of the analysis as you pointed out.

Does this mean you don't want to merge it just yet? Or can we merge it and then I will make those changes concurrent with whatever future changes are also needed?

PR #122, opened 2019-09-22
Make it such that poly-A and WXS samples are over-represented in testing

jaclyn-taroni:
#### Purpose/implementation

Samples that were assayed with poly-A selection or via WXS are currently a small subset of samples. Given that we were randomly selecting participant IDs without that in mind, we only had 1 poly-A sample in the subset files for CI. This caused failures. Here, I'm updating the step where we generate subset files to make sure these two assays are over-represented in the subset files we use for CI.


jaclyn-taroni:
As long as this is not incorrect, I think we should get this merged. It will get changed shortly with #124.

jaclyn-taroni:
> It feels like these 4 dplyr pipe step chunks could be shortened into one, but because it's pretty easy to follow as is, I don't think it should be changed. Just going to voice this out-loud.

In my opinion, this probably should not be a notebook and will likely cease to be a notebook when #124 is addressed.

> Second comment/question is I assume there's no dataset where we wouldn't have one of these WXS/PolyA samples represented?

When the poly-A and stranded RNA-seq samples get split into different files, this will have to be revisited.

cansav09:
> In my opinion, this probably should not be a notebook and will likely cease to be a notebook when #124 is addressed.
> 
> When the poly-A and stranded RNA-seq samples get split into different files, this will have to be revisited.

Cool. This looks good for now then.

Issue #3, opened 2019-07-11
Planned Analysis: Tumor Mutation Burden


cgreene:
We have planned to include an analysis of tumor mutation burden as well as a comparison to the tumor mutation burden of adult tumors in TCGA. This issue is for fleshing out the analysis. The manuscript issue for the methods description is: https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/7

cansav09:
This may be a simplistic question, but in our case `TMB = Median(# coding indels + # coding base substitutions) / Megabase`?  Second question is, I am presuming we need to do some kind of analysis that looks at differences between synonymous and nonsynonymous mutations? This perhaps may needs to be related to age of diagnosis?

cgreene:
@PichaiRaman : I think this came from you - how were you planning to do this analysis?

jharenza:
@cansav09 I think we are looking for something like this - Figure 1 from the [Grobner, 2018](https://www.ncbi.nlm.nih.gov/pubmed/29489754) landscape, but for brain tumors:
<img width="928" alt="grobner-figure1" src="https://user-images.githubusercontent.com/207034/62128604-c28d5880-b2a2-11e9-97f3-a59e0580129d.png">

For reference, here is the second pan-pediatric cancer landscape from 2018:
[Ma, 2018](https://www.nature.com/articles/nature25795)

jharenza:
> This may be a simplistic question, but in our case `TMB = Median(# coding indels + # coding base substitutions) / Megabase`? Second question is, I am presuming we need to do some kind of analysis that looks at differences between synonymous and nonsynonymous mutations? This perhaps may needs to be related to age of diagnosis?

cansav09:
Now that I've gotten somewhat familiar with the SNV data, I can do this analysis as well. It looks like Grobner says they used `Number of SNV of coding mutations per Mb`? Just to try to clarify, does this mean we include synonymous and non-synonymous mutations alike in this number? They don't mention distinguishing between this in their Methods, but I've seen other papers that look into it. 

Here's my tentative plan, let me know what you think: 

Step 1) Start with Mutect2 results only. (Once I make the script for this, and then later after #30 is determined, I can easily go back and change it). 
Step 2) Calculate total number of mutations per Mb
Step 3) Graph this in a jitterplot like above with the disease type as the x axis. 
Step 4) Split this out into a jitterplot that examines nonsynonymous vs synonymous.

[...]

PR #125, opened 2019-09-23
PR 3 of n for SNV Caller Analysis: Plotting functions

cansav09:
#### Purpose/implementation

This is the third PR in a series that are preparing a pipeline to perform an initial analysis and calculation of each MAF file for each SNV caller. At this point, the SNV callers that will be evaluated are MuTect2, Strelka2, VarDict, and Lancet. 

This includes all the functions needed to plot the MAF data that has been wrangled by the wrangle functions #118 

#### Issue

#### Directions for reviewers

* Do the functions make sense? Their naming? Their arguments? 
* Should I reorganize them in some way? 
* How do you feel about how these functions relate to the [wrangling functions](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/118)?
* Is the documentation thorough enough? Are their gaps anywhere that should be filled in? 

#### Results

These things will be in the next PR. 


analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  #   vaf_df: MAF formatted data that has been turned into a data.frame and has
+  #           been run through `set_up_maf`.
+  #   exp_strategy: argument to specify whether to plot `wgs`, `wxs` samples or
+  #                 `both` Case insensitive.
+  #   filter_cutoff: A numeric number to only keep groups larger than it.
+  #
+  # Returns:
+  # A barplot with the number of mutations with each type of base change noted
+  # in the `base_change` column made in `set_up_maf`
+  # Make this argument case insensitive
```
jashapiro: Add whitespace
```suggestion

  # Make this argument case insensitive
```

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  exp_strategy <- toupper(exp_strategy)
+
+  # Filter out based on experimental stategy if specified
+  if (exp_strategy != "BOTH") {
+    vaf_df <- vaf_df %>%
+      dplyr::filter(experimental_strategy == exp_strategy)
+  }
+
+  # Count the number of each type of base change
+  base_count <- vaf_df %>%
```
jashapiro: I don't like data frames having the same names as columns within, as it gets quite confusing (especially as in the call to ggplot below).

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+}
+
+depth_vs_vaf_plot <- function(vaf_df, exp_strategy = "BOTH") {
+  # Plot read depth versus variant allele fraction. If both experimental strategies
+  # are included, the plot is made color-coded by strategy.
+  #
+  # Args:
+  #   vaf_df: MAF formatted data that has been turned into a data.frame and has
+  #           been run through `set_up_maf` with metadata (specifically the
+  #           `experimental_strategy column added.
```
jashapiro:
```suggestion
  #           `experimental_strategy` column added.
```

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  if (exp_strategy != "BOTH") {
+    barplot <- barplot +
+      ggplot2::geom_point(color = "navyblue", alpha = 0.15) +
+      ggplot2::theme(legend.position = "none")
+  } else {
+    barplot <- barplot +
+      ggplot2::geom_point(ggplot2::aes(color = experimental_strategy), alpha = 0.15) +
+      colorblindr::scale_color_OkabeIto()
+  }
+  return(barplot)
```
jashapiro: Do you want to add an xlab here? n_depth isn't the clearest variable name.

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+      ggplot2::geom_violin(fill = "navyblue") +
+      ggplot2::theme(legend.position = "none")
+  } else {
+    barplot <- barplot +
+      ggplot2::geom_violin(ggplot2::aes(fill = experimental_strategy))
+  }
+  # Add some aesthetics
+  barplot <- barplot +
+    ggplot2::theme_classic() +
+    ggplot2::xlab("Same mutation and base change found in COSMIC") +
```
jashapiro: Looking at the docs for `find_cosmic_overlap` it appears you are plotting the overlap here, not the same exact base change (that would be `same_cosmic`, right? 

Do you want to consider making this plot with both the `same_cosmic` and `overlap_cosmic` on the same plot?

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  }
+
+  # Plot this as a jitter plot
+  barplot <- tmb_df %>%
+    ggplot2::ggplot(ggplot2::aes(x = eval(parse(text = x_axis)), y = tmb))
+
+  # Get rid of legend if both data aren't being plotted
+  if (exp_strategy != "BOTH") {
+    barplot <- barplot +
+      ggplot2::geom_jitter(color = "navyblue", alpha = 0.2) +
```
jashapiro: I don't like jitter plots (ask me why!) maybe a ggforce::geom_sina?

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+      ggplot2::theme(legend.position = "none")
+  } else {
+    barplot <- barplot +
+      ggplot2::geom_violin(ggplot2::aes(fill = experimental_strategy))
```
jashapiro: Why a violin here, should this be the same?

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+      dplyr::filter(experimental_strategy == exp_strategy)
+  }
+
+  # Plot this as a jitter plot
+  barplot <- tmb_df %>%
+    ggplot2::ggplot(ggplot2::aes(x = eval(parse(text = x_axis)), y = tmb))
+
+  # Get rid of legend if both data aren't being plotted
+  if (exp_strategy != "BOTH") {
+    barplot <- barplot +
```
jashapiro: Also not a barplot :smile:

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  # Count the number of each type of base change
+  type_count_df <- maf_annot %>%
+    dplyr::group_by(type, experimental_strategy) %>%
+    dplyr::summarise(type_count = dplyr::n()) %>%
+    dplyr::filter(type_count > filter_cutoff)
+
+  # Plot this as a barplot
+  barplot <- ggplot2::ggplot(
+    type_count_df,
+    ggplot2::aes(x = reorder(type, -type_count), y = type_count)
```
jashapiro: Do we want to reorder by count here, or have a consistent order? Consistency will make it easier to compare between plots with different data sets. Note that in the consistent order case, if for some amazing reason there were none of a particular mutation, we _would_ still want to include that zero bar to preserve order.

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  if (exp_strategy != "BOTH") {
+    vaf_df <- vaf_df %>%
+      dplyr::filter(experimental_strategy == exp_strategy)
+  }
+
+  # Get the overlap with COSMIC mutations using special function
+  maf_cosmic <- find_cosmic_overlap(vaf_df, cosmic_clean_file)
+
+  # Plot this as a violin plot
+  barplot <- maf_cosmic %>%
```
jashapiro: not a barplot

analyses/snv-callers/util/plot_functions.R
cansav09: Ah. Good point.

analyses/snv-callers/util/plot_functions.R
cansav09: Will do!

analyses/snv-callers/util/plot_functions.R
cansav09: I could make both. IDK. This may be a thing that will be easier to make a decision on when we have the real data and I see what those plots look like.

jashapiro: Agreed. But check the axis label...

analyses/snv-callers/util/plot_functions.R
cansav09: Oh my. My brain has been reading over me calling this `barplot` for probably a week and a half by now and I didn't think anything of it. Will go back and make all these object names for my plots more sensible.

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+      ggplot2::geom_bar(
+        position = "dodge", stat = "identity",
+        ggplot2::aes(fill = experimental_strategy)
+      )
+  }
+  # Add some other aesthetics
+  barplot <- barplot +
+    ggplot2::theme_classic() +
+    ggplot2::theme(axis.text.x = ggplot2::element_text(angle = 45, hjust = 1)) +
+    ggplot2::xlab("") +
```
cbethell: Perhaps add an xlab label here.

analyses/snv-callers/util/plot_functions.R
cansav09: The main reason this is a jitterplot is because of the [plot that it was requested to be based on in the issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/3#issuecomment-516395431) I can play around with a sina plot though and see what that looks like.

analyses/snv-callers/util/plot_functions.R
cansav09: I'll make it a consistent order.

analyses/snv-callers/util/plot_functions.R
cansav09: Probably.

analyses/snv-callers/util/plot_functions.R
```diff
@@ -0,0 +1,274 @@
...
+  return(barplot)
+}
+
+snv_region_plot <- function(maf_annot, exp_strategy = "BOTH", filter_cutoff = 0) {
+  # Plot the genomic region analysis from the annotation added from `annotatr_maf`
+  # as a barplot with the number of mutations found in each type of genomic region.
+  #
+  # Args:
+  #   maf_annot: MAF formatted data that has been turned into a data.frame and has
+  #           been run through `annotr_maf`
```
cbethell: The spelling of `annotr_maf` here is not the same as the `annotatr_maf` mentioned above. I am not sure which is the correct spelling here

PR #126, opened 2019-09-23
Part 4 of n: SNV Caller Analysis: Set Up Script

cansav09:
#### Purpose/implementation

This is the fourth PR in a series that are preparing a pipeline to perform an initial analysis and calculation of each MAF file for each SNV caller. At this point, the SNV callers that will be evaluated are MuTect2, Strelka2, VarDict, and Lancet. 

This includes all the set up for the initial analyses. It can be run at the beginning of the next two scripts and if files were already set up, they will be skipped. By running this at the beginning of the later scripts, it makes it so if I change a directory I only have to change it once.  However, it's very likely aspects of this script need to be rearranged and some of these things made into `optparse` options so this script and subsequent scripts are more flexible. 

**What this script does:** 

1) Import functions from the `util` folder
2) Installs packages needed and imports the `%>%` if any of the packages have not been installed. 
3) Declares already existing files' paths and sets up directories used by all 
caller's analyses. 
4) Build `AnnotatR` annotation object for use in genomic region analysis. 
5) Has the set up for COSMIC mutations, but because it is not assumed whomever downloads the repository necessarily has registered and downloaded this data, the genomic coordinates of mutations from the brain-related samples COSMIC is already there and this code chunk will not be run since the needed file already exists. 

#### Issue

#### Directions for reviewers

*Things you should know will change* 
* The COSMIC file set up will probably change after the real data is analyzed and we see certain findings and/or other questions that arise. 
* Unknown unknowns

* Should some of these directories and other items that are specified here be specified instead in the bash script that will call these scripts, using `optparse` or the like? I'm concerned too many important items I've squirreled away into this set-up script. Its nice that I only have to change it in one spot but if this is not a good way to get these things set up, let me know alternative suggestions. 
* Should some of these things be divied into the subsequent scripts? (I know this is hard to say without seeing those subsequent scripts yet.
* Is there a better way I can have the COSMIC mutation stuff set up without breaking their data agreement or having to set a password that people who download this repository would then need? 
* Should I make an option for setting the `AnnotatR` regions? Perhaps this is a thing to do at a later point in time. 

#### Results

- `data.table` and `annotatr` are added to the Dockerfile
- I noticed that previously `optparse` was being installed twice so I deleted one of those. 
- When other parts of this analysis are added later, I will be adding subsequent tests to this same command. 

# PR Checklist
* [x] Run a linter
* [x] Double check your paths


jaclyn-taroni:
Something has gone wrong in the CircleCI config file: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/126/commits/a25be13aca86652ee8f5242edc1045b52c83609d

This was probably out of date with AlexsLemonade master when you made your changes.

cansav09:
> Something has gone wrong in the CircleCI config file: [a25be13](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/a25be13aca86652ee8f5242edc1045b52c83609d)
> 
> This was probably out of date with AlexsLemonade master when you made your changes.

I was working on getting the Circle CI worked out, which is why no one has been requested to review yet.

Dockerfile
```diff
@@ -41,11 +41,15 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
...
     rprojroot \
-    optparse
+    optparse \
+    data.table
 
-# Use maftools for reading MAF files
+# maftools for proof of concept in create-subset-files
 RUN R -e "BiocManager::install(c('maftools'), update = FALSE)"
 
+# These two packages are for the genomic region analysis for snv-callers
```
cansav09:
```suggestion
# These packages are for the genomic region analysis for snv-callers
```

analyses/snv-callers/scripts/00-set_up.R
```diff
@@ -0,0 +1,144 @@
...
+# Import special functions
+source(file.path(root_dir, "analyses", "snv-callers", "util", "wrangle_functions.R"))
+source(file.path(root_dir, "analyses", "snv-callers", "util", "plot_functions.R"))
+
+#################################### Set Up ####################################
+
+# Magrittr pipe
+`%>%` <- dplyr::`%>%`
+
+# Will need R.utils for zipping up the results file
```
cbethell: This comment does not look to be relevant to `optparse`

analyses/snv-callers/scripts/00-set_up.R
```diff
@@ -0,0 +1,144 @@
...
+}
+
+# Declare reference file names
+original_metadata <- file.path(data_dir, "pbta-histologies.tsv")
+
+# These data were obtained from https://cancer.sanger.ac.uk/cosmic/download
+# These data are available if you register.
+# The full, unfiltered somatic mutations file CosmicMutantExport.tsv.gz for grch38
+# is used here.
+cosmic_file <- file.path(snv_dir, "CosmicMutantExport.tsv.gz")
```
jashapiro: We might consider adding a check for this file to ensure the version is as expected. We could include an md5 hash of the file, checked with md5sum(). This might want to be separate from this script, perhaps in the readme that describes how to get the file.

analyses/snv-callers/scripts/00-set_up.R
```diff
@@ -0,0 +1,144 @@
...
+    # Separate the genome coordinates into their own BED like variables
+    dplyr::mutate(
+      Chromosome = paste0(
+        "chr",
+        stringr::word(Mutation_genome_position, sep = ":", 1)
+      ),
+      Start_Position = stringr::word(Mutation_genome_position, sep = ":|-", 1),
+      End_Position = stringr::word(Mutation_genome_position, sep = "-", 2),
+      # Make a base_change variable so we can compare to our set up for PBTA data
+      base_change = substr(Mutation_CDS, nchar(Mutation_CDS) - 2, 10)
```
jashapiro: I'm a bit confused here. Why the chosen start and stop positions? It looks like you want the last 3 characters, but if the string is longer than 10, you will not get that?

cansav09: The strings are showing the base changes which are all 3 characters long. They don't have the insertions or deletions spelled out so all of them are `ins` `del` or `G>T`-like. I could change it to use `nchar(Mutation_CDS)` if that is less confusing.

analyses/snv-callers/scripts/00-set_up.R
cansav09: I think this should be done eventually, but because this whole analysis and its associated files is in a rather fluid state this might be a future to-do list item. What do you think?

analyses/snv-callers/scripts/00-set_up.R
cansav09: I went ahead and changed it to `nchar(Mutation_CDS)` because I don't think you are the first person to be confused by that.

cansav09:
Regarding both of your comments on usage documentation, the script that will a person would actually call for this to run, is a bash script that is added in #134 so I will add the usage documentation to that PR. If I add it in this PR, the usage won't make sense because the scripts don't yet exist in this PR. Does that make sense? 

That being said, is this ready to be merged? Did I cover your suggestions (except the md5sum suggestion thing which we will add later when files are better defined.)

analyses/snv-callers/scripts/00-set_up.R
cansav09: This item is [tracked here](https://github.com/cansav09/Notes-and-goals/issues/1)

PR #134, opened 2019-09-27
Part 5 of n: SNV caller initial calculations 

cansav09:
#### Purpose/implementation

This is the fifth PR in a series that are preparing a pipeline to perform an initial analysis and calculation of each MAF file for each SNV caller. At this point, the SNV callers that will be evaluated are MuTect2, Strelka2, VarDict, and Lancet. 

**What these scripts do:** 
`01-calculate_vaf_tmb.R` is called from the `run_caller_evals.sh` bash script and sets up and creates the VAF, TMB, and other files for each dataset based on the options provided. It uses functions from `util/wrangle_functions.R` and will run the `00-set_up.R` script at the very beginning so it makes sure the files needed exist. 

**These are the files it creates in a results folder for each dataset:**
1) Filtered metadata file
2) Set up MAF version of the file with VAF 
3) Create an genomic regions annotation file
4) A TMB file with TMB stats for each sample. 

#### Issue

#### Directions for reviewers

* How do you feel about the relationship of these scripts in `00-set_up.R`? What should be changed/optimized? 
* Where might be some "shoot in the foot" items that should be fixed? 
* How do you feel about the `optparse` options? 
* As we have accumulated more scripts now, how do you feel about the overall structure and interactions of this analysis? How can it be made less buggy? 
* What do you think about the `rprojroot::find_root` for getting the directories method? 

#### Results

(From #126, no changes have been made to the Dockerfile.)
- It is called in the same line as the other SNV set up

# PR Checklist
* [x] Run a linter
* [x] Double check your paths


analyses/snv-callers/run_caller_evals.sh
```diff
@@ -7,7 +7,8 @@
 
 # Change directory
 cd kitematic
- 
+
+# The files named in these arrays will be ran in the ana,ysis.
```
jashapiro: fix typos... 
"will be run in the analysis"

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+
+# Check that the files we need are in the paths specified
+needed_files <- c(original_metadata, wxs_bed_file, opt$maf, opt$bed_wgs)
+
+# Get list of which files were found
+files_found <- file.exists(needed_files)
+
+# Report error if any of them aren't found
+if (!all(files_found)) {
+  stop(paste("\n This needed file wasn't found:",
```
jashapiro:
```suggestion
  stop(paste("\n Could not find needed file(s):", 
```

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+files_found <- file.exists(needed_files)
+
+# Report error if any of them aren't found
+if (!all(files_found)) {
+  stop(paste("\n This needed file wasn't found:", 
+              needed_files[which(!files_found)], 
+              "Check your options and set up.", sep = "\n"))
+}
+################## Create output directories for this caller ###################
+caller_results_dir <- file.path(base_results_dir, opt$label)
```
jashapiro: `base_results_dir` is not defined?

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+option_list <- list(
+  make_option(opt_str = c("-m", "--maf"), type = "character", default = "none",
+              help = "Relative file path (assuming from top directory of 
+              'OpenPBTA-analysis') to MAF file to be analyzed. Can be .gz compressed.",
+              metavar = "character"),
+  make_option(opt_str = c("-b", "--bed_wgs"), type = "character",
+              default = "none", help = "File path that specifies the caller specific
+                BED regions file. Assumes from top directory, 'OpenPBTA-analysis'",
+              metavar = "character"),
+  make_option(opt_str = c("-l", "--label"), type = "character",
```
jashapiro: Do you trust people to set this correctly? I worry that it will be too easy for people to accidentally overwrite directories with the wrong data.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+# Rscript 01-calculate_vaf_tmb.R \
+# -m data/pbta-snv-strelka2.vep.maf.gz \
+# -b data/WGS.hg38.strelka2.unpadded.bed \
+# -l strelka2
+
+# Establish base dir
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+
+# Run set up script
+source(file.path(root_dir, "analyses", "snv-callers", "scripts", "00-set_up.R"))
```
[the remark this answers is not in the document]
jashapiro: Oh, I think I see... this is also defining variables for later. I don't like that. It is makes it hard to figure out where variables are coming from, and makes me worry that there will be errors introduced if variable names or file locations change.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+
+######################### Annotate genomic regions #############################
+message(paste("Annotating genomic regions for", opt$label, "MAF data..."))
+
+# Annotation genomic regions
+maf_annot <- annotr_maf(vaf_df, annotation_file = annot_rds) %>%
+  readr::write_tsv(region_annot_file)
+
+# Print out completion message
+message(paste0("Genomic region annotations saved to:", region_annot_file))
```
jashapiro:
```suggestion
message(paste("Genomic region annotations saved to:", region_annot_file))
```

.circleci/config.yml
```diff
@@ -28,10 +28,10 @@ jobs:
       - run:
           name: fusion filtering
           command: ./scripts/run_in_ci.sh Rscript analyses/fusion_filtering/01-fusion-standardization.R -f "data/pbta-fusion-starfusion.tsv.gz" -c "STARfusion" -o "scratch/star.RDS" 
-
+                                                
       - run:
-          name: SNV Caller Set Up    
-          command: ./scripts/run_in_ci.sh Rscript analyses/snv-callers/scripts/00-set_up.R      
+          name: SNV Caller Analysis 
+          command: ./scripts/run_in_ci.sh Rscript analyses/snv-callers/scripts/00-set_up.R \ bash analyses/snv-callers/run_caller_evals.sh
```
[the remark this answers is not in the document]
cansav09: In this specific case, as the scripts stand, I think you are correct that it is redundant to call 00-set_up.R here. However, for your larger discussion question, I do not know yet.

cansav09:
>The other part is the definition of file locations and defining variables that will be used later, that I am less fond of, as I feel like those should be more explicitly in the file that uses them. 

I suspected this may be confusing. The reason I originally did it was so that those variables didn't need to specified in both `01-calculate_vaf_tmb.R` and `02-run_eval.R` scripts (02 you haven't seen yet). So I need a more apparent way to specify those variables but preferably not in a way that I have to specify them in both scripts. 

I can just use a combo of opt-parse and the bash script to do this. However, that also brings us back to how are we going to run this pipeline (and ones like it) in the long run.

jashapiro:
> > The other part is the definition of file locations and defining variables that will be used later, that I am less fond of, as I feel like those should be more explicitly in the file that uses them.
> 
> I suspected this may be confusing. The reason I originally did it was so that those variables didn't need to specified in both `01-calculate_vaf_tmb.R` and `02-run_eval.R` scripts (02 you haven't seen yet). So I need a more apparent way to specify those variables but preferably not in a way that I have to specify them in both scripts.
> 
> I can just use a combo of opt-parse and the bash script to do this. However, that also brings us back to how are we going to run this pipeline (and ones like it) in the long run.
> 
> Do you have suggestions about fixing this script organization, @jashapiro? I've kinda known/suspected this was a problem, but haven't come to a good solution for it yet.

I think I would do exactly what you suggest in the paragraph above. Use optparse and define the files needed in the bash script (or whatever other system we use later).

analyses/snv-callers/run_caller_evals.sh
```diff
@@ -0,0 +1,25 @@
...
+# CCDL for ALSF 2019
+
+# Purpose: 1) Run an intial evaluation of each variant caller's MAF file
+#          2) Compare the variant callers to each other. (Coming soon)
+
+# Change directory
+cd kitematic
+
+# The files named in these arrays will be ran in the analysis. 
+declare -a datasets=("strelka2" "mutect2" "lancet" "vardict")
```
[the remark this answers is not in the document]
cansav09: I just tried it without and it still works. My interpretation of [this guide](https://www.tldp.org/LDP/Bash-Beginners-Guide/html/sect_10_02.html) tells me that for this example, either works fine.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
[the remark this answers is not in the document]
cansav09: I am working on re-configuring all of this as we discussed below.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansav09: I have an `--overwrite option` to avoid this problem on the 02-run_eval.R script, unsure why I didn't add the same option here. Will do.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,177 @@
...
+#
+# C. Savonen for ALSF - CCDL
+#
+# 2019
+#
+# Option descriptions
+#
+# -m : Path to MAF file to be analyzed. Can be .gz compressed. Give path relative
+#      to top directory, 'OpenPBTA-analysis'.
+# -b : File path that specifies the caller specific WGS BED regions file that is
```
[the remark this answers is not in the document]
cansav09: It just is easier/more straightforward to have all file path related options using the same directory as their assumption. So all my file path options want the file paths given in relation to `OpenPBTA-analysis`.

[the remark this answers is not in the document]
jashapiro: I worry about this, as convention for file paths at a command line is to specify either a relative or absolute path from the calling directory. This is neither and seems likely to cause confusion. (Mine isn't really convention either, so take that with a grain of salt). But moving the logic of where to look to the shell script and specifying absolute path may be safer.

analyses/snv-callers/run_caller_evals.sh
```diff
@@ -0,0 +1,25 @@
...
+declare -a wgs_files=("WGS.hg38.strelka2.unpadded.bed" "WGS.hg38.mutect2.unpadded.bed" "WGS.hg38.lancet.unpadded.bed" "WGS.hg38.vardict.100bp_padded.bed")
+
+########################## Calculate and Set Up Data ###########################
+i=-1
+for dataset in ${datasets[@]}
+  do
+  let i=${i}+1
```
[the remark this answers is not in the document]
cansav09: This works if I run it in Terminal on my computer, but doesn't seem to work on the container. it prints out blanks for `${datasets[$i]}` according to Circle CI's error report. Does Debian not use part of the syntax you suggested here? It works locally but not on the container; which means my Circle CI tests fail.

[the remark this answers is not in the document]
cansav09: It also might be something else more dumb. Still investigating.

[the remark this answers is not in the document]
cansav09: It was something more dumb. I was referencing the wrong array.

cansav09:
The first three datasets passed through the Circle CI check fine, but it has failed because of VarDict. The problems with the [VarDict file have been noted](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/135). But unfortunately my test won't work until that's resolved.

analyses/snv-callers/run_caller_evals.sh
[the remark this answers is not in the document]
jashapiro: one change in the syntax I gave: 
```for ((i=0;i<${#datasets[*]};i++)); ```
 may be safer

File: analyses/snv-callers/run_caller_evals.sh [hunk 11]
```diff
@@ -0,0 +1,41 @@
...
+  echo "Processing dataset: ${datasets[$i]}"
+  Rscript analyses/snv-callers/scripts/01-calculate_vaf_tmb.R \
+  --label ${datasets[$i]} \
+  --output analyses/snv-callers/results/${datasets[$i]} \
+  --maf data/pbta-snv-${datasets[$i]}.vep.maf.gz \
+  --metadata data/pbta-histologies.tsv \
+  --bed_wgs data/${wgs_files[$i]} \
+  --bed_wxs data/WXS.hg38.100bp_padded.bed \
+  --annot_rds $annot_rds \
+  --overwrite
```
jashapiro: entabben

analyses/snv-callers/scripts/00-set_up.R
```diff
@@ -34,50 +42,46 @@ if (!("data.table" %in% installed.packages())) {
...
-cosmic_file <- file.path(snv_dir, "CosmicMutantExport.tsv.gz")
-
-# This is the cleaned version that only contains the genomic coordinates and
-# base changes from the original file for brain sample mutations only.
-cosmic_clean_file <- file.path(cosmic_dir, "brain_cosmic_variants_coordinates.tsv")
+################################ Set up options ################################
+# Set up optparse options
+option_list <- list(
+  make_option(
+    opt_str = "--cosmic_og", type = "character", default = "none",
```
jashapiro: Ha!

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+# Command line example:
+#
+# Rscript analyses/snv-callers/scripts/01-calculate_vaf_tmb.R \
+# --label strelka2 \
+# --output analyses/snv-callers/results \
+# --maf scratch/snv_dummy_data/strelka2 \
+# --metadata data/pbta-histologies.tsv \
+# --bed_wgs data/WGS.hg38.mutect2.unpadded.bed \
+# --bed_wxs data/WXS.hg38.100bp_padded.bed \
+# --annot_rds scratch/hg38_genomic_region_annotation.rds
```
jashapiro: Entabben

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+  # If all files exist; stop
+  if (all(existing_files)) {
+    stop(cat(
+      "Stopping; --overwrite is not being used and all output files already exist: \n",
+      vaf_file, "\n",
+      region_annot_file, "\n",
+      tmb_file
+    ))
+  }
+  # If some files exist, print a warning:
```
jashapiro: Good idea. Of course, overwriting should be a warning sometimes too... but we can't save everyone!

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+if (!all(unique(maf_df$Tumor_Sample_Barcode) %in% metadata$Tumor_Sample_Barcode)) {
+  stop("There are samples in this MAF file that are not in the metadata.")
+}
+
+################## Calculate VAF and set up other variables ####################
+# If the file doesn't exist or the overwrite option is being used, run this.
+if (any(c(!file.exists(vaf_file), opt$overwrite))) {
+
+  # Print out warning if this file is going to be overwritten
+  if (all(c(opt$overwrite, file.exists(vaf_file)))) {
```
jashapiro: If the file exists, you must be overwriting, or you wouldn't be here:
```suggestion
  if (file.exists(vaf_file){
```

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+  # Print out progress message
+  message(paste("Calculating VAF for", opt$label, "MAF data..."))
+
+  # Use the premade function to calculate VAF this will also merge the metadata
+  vaf_df <- set_up_maf(maf_df, metadata) %>%
+    readr::write_tsv(vaf_file)
+
+  # Print out completion message
+  message(paste("VAF calculations saved to: \n", vaf_file))
+} else {
```
jashapiro: Personally, I like to put the simpler case first, which in this case would be 

```r
if (file.exists && !overwrite){
  # skip and warn
}else{
  # do the thing which takes a lot of lines
}
```
but that is a style thing. I just like to keep the if and else as close together as possible.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+message(paste("Filtered metadata file saved to: \n", metadata_file))
+
+# Make sure that we have metadata for all these samples.
+if (!all(unique(maf_df$Tumor_Sample_Barcode) %in% metadata$Tumor_Sample_Barcode)) {
+  stop("There are samples in this MAF file that are not in the metadata.")
+}
+
+################## Calculate VAF and set up other variables ####################
+# If the file doesn't exist or the overwrite option is being used, run this.
+if (any(c(!file.exists(vaf_file), opt$overwrite))) {
```
jashapiro: With only two things being tested, I would usually go with a simple or:
```suggestion
if (!file.exists(vaf_file) || opt$overwrite) {
```

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+  # Stop if this file exists and overwrite is set to FALSE
+  warning(cat(
+    "The VAF file already exists: \n",
+    vaf_file, "\n",
+    "Use --overwrite if you want to overwrite it."
+  ))
+}
+######################### Annotate genomic regions #############################
+# If the file doesn't exist or the overwrite option is being used, run this.
+if (any(c(!file.exists(region_annot_file), opt$overwrite))) {
```
jashapiro: See above about `||` and simplifying the overwrite warning case.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,294 @@
...
+  # Stop if this file exists and overwrite is set to FALSE
+  warning(cat(
+    "The regional annotation file already exists: \n",
+    region_annot_file, "\n",
+    "Use --overwrite if you want to overwrite it."
+  ))
+}
+############################# Calculate TMB ####################################
+# If the file doesn't exist or the overwrite option is being used, run this.
+if (any(c(!file.exists(region_annot_file), opt$overwrite))) {
```
jashapiro: One more time... see above.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansav09: Well, I have it written so if some files exist but others don't and overwrite is off it will only write the files that don't exist. Does that make sense? And also is there a better way to reflect that here?

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansav09: I like that idea. Will do.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansav09: AH. I see what you are saying now. Yes. This is redundant. Will delete. Took a second for that to sink in.

jashapiro: Yes, but If you play with the logic a bit, you will see that my version will still work.
There are 4 possible cases for each file

1.  file doesn't exist, overwrite off
2. file doesn't exist, overwrite on
3. file exists, overwrite off
4. file exists, overwrite on

You get to this line in cases 1, 2, and 4. And you only have to warn with case 4, so no need to check whether the file exists again.

cansav09: ^ Cool. This is a good way to lay out the options.

cansav09:
Okay, made the suggested changes, @jashapiro and things are working fine. What do you think?

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -0,0 +1,291 @@
...
+# Print out completion message
+message(paste("Filtered metadata file saved to: \n", metadata_file))
+
+# Make sure that we have metadata for all these samples.
+if (!all(unique(maf_df$Tumor_Sample_Barcode) %in% metadata$Tumor_Sample_Barcode)) {
+  stop("There are samples in this MAF file that are not in the metadata.")
+}
+
+################## Calculate VAF and set up other variables ####################
+# If the file doesn't exist or the overwrite option is being used, run this.
```
jashapiro: Fix comments to reflect change in logical order. Repeat below.

cansav09: Oh good point!

cansav09:
I'm temporarily removing VarDict from the analysis until #135 is unresolved. This way the Circle CI tests will pass and then this can be merged. 

VarDict needs to be added back to this analysis after #135 is resolved. This issue is [tracked here.](https://github.com/cansav09/openpbta-notebook-concept/issues/2)

PR #137, opened 2019-09-28
Update the txome dimensionality reduction module now that files are split

jaclyn-taroni:
#### Purpose/implementation

Now that the RNA-seq gene expression files are split based on library selection, the dimensionality reduction plotting module needed to be updated to take that into account. In this PR, I am removing the references to combined files and any downstream results/data/plots. I'm also adding shell scripts to run the entire series of steps (`dimension-reduction-plots.sh`) and a subset of steps in CI with changes to the parameters (`ci-dimension-reduction-plots.sh`).  

#### Issue

#9 

#### Directions for reviewers

* It is likely that this module will need to be tweaked prior to publication (i.e., different subsets of samples, changing variable to color points, different PCs plotted). Is there enough flexibility in the way the module is structured and enough information in the README to make this easy to revisit?
* Do you agree with my take on the results?

#### Results

No overall/general changes to results -- I don't see a discernible pattern according to broad histology in these plots.


analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+This module runs dimension reduction—namely PCA, UMAP, and t-SNE—on gene expression data from kallisto and RSEM.
+
+To the run the whole pipeline and produce 4 plots with panels for all three approaches
+
+1. RSEM stranded data
+2. RSEM poly-A data
+3. kallisto stranded data
+4. kallisto poly-A data
+
+with points colored by broad histology, simply run:
```
cansav09:
```suggestion
To run the whole pipeline and produce 4 plots with panels for all three approaches with points colored by broad histology, simply run:
```

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
+# Dimension reduction: gene expression data
+
+This module runs dimension reduction—namely PCA, UMAP, and t-SNE—on gene expression data from kallisto and RSEM.
+
+To the run the whole pipeline and produce 4 plots with panels for all three approaches
```
cansav09: What are the three approaches being referred to? This itemized list doesn't make it clear to me.

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+3. kallisto stranded data
+4. kallisto poly-A data
+
+with points colored by broad histology, simply run:
+
+```
+./dimension-reduction-plots.sh
+```
+
+## Notes on further applications
```
cansav09:
```suggestion
## Notes on further applications
```
Seems like this is the whole README, so feels weird to call it "Notes" maybe just *General Usage* or something?

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+    --metadata ../../data/pbta-histologies.tsv \
+    --filename_lead kallisto_polyA_${seed} \
+    --output_directory results \
+    --seed ${seed} \
+    --skip_tsne
+done
+
+```
+
+Unfortunately, this approach will also run PCA 5 times.
```
cansav09: Should it have an option to not do this?

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+##### R
+
+It's also possible to repeat this UMAP step 5 times within R:
+
+```R
+# load the required custom functions
+source("util/dimension-reduction-functions.R")
+
+# read in kallisto poly-A data and transpose
+expression_file <- "../../data/pbta-gene-expression-kallisto.polya.rds"
```
cansav09: Should you use/encourage the use of `file.path`?

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+# Read in expression data
```
cansav09: Sometimes you capitalize your comments and sometimes you don't. This doesn't matter but I'm just stating this because it's something my brain notices.

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+
+```
+
+Unfortunately, this approach will also run PCA 5 times. 
+
+##### R
+
+It's also possible to repeat this UMAP step 5 times within R:
+
+```R
```
cansav09: I think technically most markdowns want this to be a lower case `r` but I am unsure if this matters for anything.

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+The PCA plots are limited to PC1 and PC2.
+
+### Plot additional principal components
+
+If we wanted to plot additional PCs for the RSEM stranded data in R, we can take advantage of the custom `plot_dimension_reduction` function and use the following R code:
+
+```R
+# load the required custom functions
+source("util/dimension-reduction-functions.R")
+# read in the data.frame that has the PCA info and the metadata
```
cansav09: Could use some free lines between these comments and the code.

analyses/transcriptomic-dimension-reduction/README.md
```diff
@@ -0,0 +1,130 @@
...
+# plot itself - PC7 and PC10
+plot_dimension_reduction(aligned_scores_df = pca_df,
+                         point_color = "broad_histology",
+                         x_label = "PC7",
+                         y_label = "PC10",
+                         score1 = 7,
+                         score2 = 10)
+```
+
+The `point_color` argument of `plot_dimension_reduction` can be changed to any variable that is in the TSV file read in.
```
cansav09:
```suggestion
The `point_color` argument of `plot_dimension_reduction` can be changed to any variable that is a column in the original TSV file.
```

.circleci/config.yml
```diff
@@ -28,24 +28,12 @@ jobs:
...
-          command: ./scripts/run_in_ci.sh Rscript --vanilla analyses/transcriptomic-dimension-reduction/scripts/get-plot-list.R  --input_directory analyses/transcriptomic-dimension-reduction/results --filename_lead rsem_all --output_directory analyses/transcriptomic-dimension-reduction/plots --color_variable RNA_library
-          
+      
       - run:
-          name: Unsupervised Transcriptomic Analysis - Multipanel Plots      
-          command: ./scripts/run_in_ci.sh bash analyses/transcriptomic-dimension-reduction/04-multipanel-plots.sh
-          
-        #### Add your analysis here ####
+          name: Transcriptome dimensionality reduction
+          command: ./scripts/run_in_ci.sh ./analyses/transcriptomic-dimension-reduction/ci-dimension-reduction-plots.sh
```
cansav09: Also, it is still passing, but why don't you have to say `bash` in front of this?

analyses/transcriptomic-dimension-reduction/README.md
jaclyn-taroni: PCA, UMAP, t-SNE

.circleci/config.yml
jaclyn-taroni: Because this file is [executable](https://askubuntu.com/questions/229589/how-to-make-a-file-e-g-a-sh-script-executable-so-it-can-be-run-from-a-termi).

analyses/transcriptomic-dimension-reduction/README.md
jaclyn-taroni: I think that use case is covered in the R section below. The way to do this would be to include logic to skip each of the PCA, UMAP, and t-SNE (or enable each step depending on how you did it), which is easy enough to do if it looks like it will come up often. I think a more likely scenario is that someone picked one of the three methods for display and is running it multiple times or tweaking the plots.

analyses/transcriptomic-dimension-reduction/README.md
jaclyn-taroni: I think technically @cbethell capitalizes comments and I don't 😁

analyses/transcriptomic-dimension-reduction/README.md
jaclyn-taroni: It does not, I just checked.

analyses/transcriptomic-dimension-reduction/README.md
jaclyn-taroni: I think in this setting this is easier to _read_. That was my rationale.

jaclyn-taroni:
> I do have one question however. Would you argue that a diagram of this module's folder structure is not necessary to include in README.md given the complexity of the README as is?

I don't view the folder structure in the README as a must have.

PR #141, opened 2019-10-02
FIX: data directory for CI txome dimension reduction

jaclyn-taroni:
#### Purpose/implementation

The CI script for `transcriptomic-dimension-reduction` used the incorrect data directory, but still passed.

#### Issue

This is something I noticed as part of #139 

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,6 +299,9 @@ The contents of `analyses/gene-expression-clustering/run-gene-expression-cluster
 # This script runs the gene-expression-clustering analysis
 # Author's Name 2019
 
+set -e
+set -o pipefail
+
 Rscript --vanilla analyses/gene-expression-clustering/01-filter-samples.R
 Rscript --vanilla analyses/gene-expression-clustering/02-cluster-heatmap.R
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -299,6 +299,9 @@ The contents of `analyses/gene-expression-clustering/run-gene-expression-cluster
 # This script runs the gene-expression-clustering analysis
 # Author's Name 2019
 
+set -e
+set -o pipefail
```
cbethell: This would exit the script immediately and return the exit status of the failed command, am I correct?

jaclyn-taroni: `set -e` should cause immediate exit when a non-zero exit code is returned. As for `set -o pipefail`:

> The bash shell normally only looks at the exit code of the last command of a pipeline. This behavior is not ideal as it causes the `-e` option to only be able to act on the exit code of a pipeline’s last command. This is where `-o pipefail` comes in. This particular option sets the exit code of a pipeline to that of the rightmost command to exit with a non-zero status, or to zero if all commands of the pipeline exit successfully.

from https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

jaclyn-taroni: You can see this in action failing on the 2nd step - dimension reduction for poly-A: https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/511?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

Issue #128, opened 2019-09-25
Proposed Analysis: Copy number consensus calls


jharenza:
#### Scientific goals

_What are the scientific goals of the analysis?_
Create consensus calls from ControlFreeC and CNVKit

#### Proposed methods

_What methods do you plan to use to accomplish the scientific goals?_
Breakpoints will not perfectly overlap between algorithms, so the analyst will likely have to define a window for overlap of copy number alterations to deem consensus calls.

#### Required input data

_What input data will you use for this analysis?_
SEG files

#### Proposed timeline

_What is the timeline for the analysis?_
2 weeks

#### Relevant literature

_If there is relevant scientific literature, put links to those items here._

xiehongbo:
We will use additional methods to get CNV calls with a balanced sensitivity and specificity for the cohort.

jaclyn-taroni:
Are you planning on tackling this @jharenza and @xiehongbo? If so, I will mark as `in progress`.

fingerfen:
We have a pipeline, count me in as well!

xiehongbo:
Yeah, we are tackling this.

jharenza:
> We have a pipeline, count me in as well!

@fingerfen - great! What is the pipeline and what inputs do you need? We may have to set you up on CAVATICA to run this.

xiehongbo:
let talk about it today during our meeting.

jaclyn-taroni:
Hi @jharenza @xiehongbo @fingerfen,

Do you have an idea of when we should expect the first pull request for this issue? I am also wondering if we know what the format of the output of this analysis will be now that the two callers have different file formats. This information will help us in development for issues like #6 and #186.

jharenza:
Hi @fingerfen and @xiehongbo - we were able to finish the data releases to include the new CNVkit and ControlFreeC files, so now you are able to submit a pull request with your analysis. Is it possible to do this next week? @fingerfen can you also list here the output columns of your final file for @jaclyn-taroni ? Thanks!

hongboxie:
We assume, that sample QC has been done by the sample noisy level.
here is what we did for summarizing consensus CNVs from three predictors of somatic CNs

1) We define two CNVs are the same event if the CNVs overlapping each other >50% reciprocally. We typically use much higher threshold for Germline CNVs (60-80%)
2) We took any CNVs that are identified by >=2 approaches (cnvkit,freec, etc).
[...]

Issue #8, opened 2019-07-12
Planned Analysis: Copy number plot showing recurrently amplified/deleted regions in different PBTA cancers


PichaiRaman:
The idea would be to visualize trends of deletions / amplifications in each of the cancer types (where relevant). Pointing out highly amplified/deleted regions with oncogenes/tumor suppressors may be interesting.

cgreene:
@jharenza : is this still blocked? It seems like something that could be done now and if the CNV calls get improved, then this would just get re-run.

jharenza:
> @jharenza : is this still blocked? It seems like something that could be done now and if the CNV calls get improved, then this would just get re-run.

I think I may have added it that way thinking consensus calls @gonzolgarcia creates would go into this, but perhaps consensus for CNV and SV would be separate issues?

cbethell:
I have begun to tackle this issue. Thus far, I believe the package `GenVisR` would be useful for this issue based on the literature I found [here](https://genviz.org/module-03-genvisr/0003/05/01/cnFreq_GenVisR/). If there are any suggestions for the implementation of an alternative package, please feel free to express them here.

jaclyn-taroni:
Related: #128

jharenza:
Hi @cbethell ! Thanks for working on this! I also just created an issue #128 for creation of consensus calls for CN to improve our accuracy for downstream analytics like this issue. I previously have used [GISTIC](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3218867/), but it does not work well for small cohorts (ie you need a large N to call recurrence within a caner). We may want to focus this issue on histologies with a specified >= N so that we have enough power to detect recurrent changes already known in the literature. For example, we have a large cohort of low-grade gliomas, we have a good number of medulloblastomas (could see how many of each subtype we have since we know each subtype has different CNAs - [Figure 2](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3306779/)), high-grade gliomas. I found [this recent paper](https://journals.plos.org/plosone/article/file?id=10.1371/journal.pone.0041082&type=printable) describing `TAGCNA`, which may work similar to GISTIC.

jaclyn-taroni:
**Summarizing _my understanding_ of the status of this issue:**

* The mention of GISTIC and TAGCNA = I understand we're looking for significant amplifications or deletions and therefore some metric and test for significance. 

* As far as visualization is concerned, @jharenza has requested

[...]

PR #142, opened 2019-10-02
Addition of R script to compare CNV caller output

cbethell:
#### Purpose/implementation

The purpose of this draft PR is to produce a copy number plot showing the recurrently amplified/deleted regions detected by CNVkit and Control-FREEC. 

The initial plan was to implement the [TAGCNA](https://journals.plos.org/plosone/article/file?id=10.1371/journal.pone.0041082&type=printable) workflow, but I encountered some issues when attempting to format the data to fit their pipeline. 

The [`GenVisR` package](https://bioconductor.org/packages/release/bioc/vignettes/GenVisR/inst/doc/Intro.html) was used as a temporary alternative to TAGCNA and the plots produced via this package include:
`analyses/cnv-comparison/plots/compare_cnv_output_frequency.pdf`
`analyses/cnv-comparison/plots/compare_cnv_output_proportion.pdf`

The `ggplot2` package was used to produce the remaining plots found in the `plots` directory of this module. 

#### Issue

This PR addresses issue #8 

#### Directions for reviewers

Per [the comment made] on the relevant issue #8, I attempted to implement the TAGCNA workflow. However, it was unsuccessful due to the specificity of the input format it expects. 

I am still looking into the best way to produce figures like those in the [TAGCNA documentation](https://journals.plos.org/plosone/article/file?id=10.1371/journal.pone.0041082&type=printable), but felt it necessary to show what has been produced thus far. 

1. Given the purpose of this analysis, I chose to write it in a R script. I used the `sample-distribution-analysis` module as a guide in this sense. That being said, would anyone argue for the need of separate scripts and a shell script to sequentially execute said scripts?

2. I also began to include `optparse` functions and decided to file a draft PR of the script as is so that I can receive a second opinion on whether or not command line options are necessary for this module (again, I used the `sample-distribution-analysis` module as a reference).

#### Results

This script produces five plots as follows:
`analyses/cnv-comparison/plots/compare_cnv_output_barplot_aberration.pdf`
`analyses/cnv-comparison/plots/compare_cnv_output_barplot_histology.pdf`
`analyses/cnv-comparison/plots/compare_cnv_output_boxplot.pdf`
`analyses/cnv-comparison/plots/compare_cnv_output_frequency.pdf`
`analyses/cnv-comparison/plots/compare_cnv_output_proportion.pdf`

#### PR Checklist
* [x] Run a linter (not done as yet)
* [x] Double check your paths


analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,200 @@
...
+# This script is intended to be run via the command line from the top directory 
+# of the repository as follows:
+#
+# Rscript analyses/cnv-comparison/util/cnv-comparison-functions.R
+
+read_in_cnv <- function(input_directory, file_path){
+  # Given the file path of the CNV data, read in the data.
+  #
+  # Arg:
+  #   input_directory: file path of input directory
```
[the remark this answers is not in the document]
cbethell: Good point, I do see how this can be confusing. I am sure this was better suited for my original, more complicated script, but making it one `file.path` to the data does seem to make the most sense now.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,200 @@
...
+
+  return(dataframe)
+
+}
+
+filter_segmean <- function(dataframe){
+  # Given the data.frame containing CNV caller output, return a data.frame with
+  # a column with the values 0 and 1 to represent loss and gain, respectively.
+  #
+  # Args:
```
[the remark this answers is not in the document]
cbethell: I like this idea.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,200 @@
...
+  #
+  # Return:
+  #   dataframe: the data frame containing the input data
+
+  # Read in cnv data
+  dataframe <-
+    read.table(gzfile(file.path(input_directory, file_path)), header = TRUE)
+
+  # Rename the columns
+  dataframe <-
```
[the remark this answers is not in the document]
cbethell: I now realize that my comment above this line was misleading as this was meant to rearrange the columns, rather than simply rename them. That being said, the function `colnames` won't be feasible here. I am thinking of using `dplyr::select` as the replacement here unless either of you may have a better solution.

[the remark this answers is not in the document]
cansav09: Yeah, do `dplyr::select/rename`

cbethell:
@jashapiro I'll wait on your review before changing this script into a R notebook or running it from a bash script with `optparse` options. I have committed changes that address all of the other comments @cansav09 made.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,200 @@
...
+  # Return:
+  #   cnv_df: the data frame containing the input data
+
+  # Read in cnv data
+  cnv_df <-
+    read.table(gzfile(file_path), header = TRUE)
+
+  # Rearrange the columns
+  cnv_df <- cnv_df %>%
+    dplyr::select("chrom", "loc.start", "loc.end", "ID", "num.mark", "seg.mean")
```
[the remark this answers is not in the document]
cbethell: Good point, will do.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,200 @@
...
+  filtered_cnv_df <- cnv_df %>%
+    dplyr::mutate(aberration = dplyr::case_when(segmean < -segmean_cutoff ~ 0,
+                                  segmean > segmean_cutoff ~ 1)) %>%
+    dplyr::filter(!is.na(aberration))
+
+  return(filtered_cnv_df)
+
+}
+
+plot_cn_freq <-
```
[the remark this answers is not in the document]
cbethell: I also see your point here, however, the motivation behind making this a function was to keep the calling of functions in the `01-cnv-comparison-plotting.R` script uniformed and the number of lines of code to a minimum. 

That being said, would your recommendation here be to call the `GenVisR` function in the `01` script nonetheless?

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,171 @@
...
+  # Read in cnv data
+  cnv_df <-
+    read.table(gzfile(file_path), header = TRUE)
+
+  # Rearrange the columns
+  cnv_df <- cnv_df %>%
+    dplyr::select("chrom", "loc.start", "loc.end", "ID", "num.mark", "seg.mean")
+  
+  # Reorder chromosome levels 
+  levels(cnv_df$chrom) <- c(paste0("chr", c(1:22, "X", "Y")))
```
jashapiro: The above doesn't actually convert to a factor, but makes it a kind of strange hybrid character with a levels attribute:
`dplyr::mutate(chrom = factor(chrom, levels = paste0("chr", c(1:22, "X", "Y"))`
should work

jashapiro: If you look at the before and after plots, you can see that the order didn't actually change but the labels did. Which would be bad...

cbethell: Ahh, I thought it was strange when I initially tried to convert to a factor, what R claimed to already be a factor. Thanks, I'll give that a shot now.

jashapiro: oooh... add stringsAsFactors=False to your read.table (or use readr...)

cbethell: I also made that realization a minute ago. That'll also be in this upcoming commit 👍

analyses/cnv-comparison/01-cnv-comparison-plotting.R
```diff
@@ -0,0 +1,123 @@
...
+
+#### Install packages ----------------------------------------------------------
+if (!("GenVisR" %in% installed.packages())) {
+  install.packages("BiocManager")
+  BiocManager::install("GenVisR")
+}
+
+if (!("cowplot" %in% installed.packages())) {
+  install.packages("BiocManager")
+  BiocManager::install("cowplot")
```
jashapiro: Is there a reason not tin install this from cran?

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,169 @@
...
+  # Return:
+  #   barplot: barplot depicting the proportion of aberrations detected across
+  #            chromosomes, annotated by the `broad_histology`` variable in the
+  #            metadata
+
+  # Create a data.frame with the filtered dataframe joined with the metadata
+  meta_joined <- filtered_cnv_df %>%
+    dplyr::inner_join(metadata, by = c("sample" = "Kids_First_Biospecimen_ID"))
+
+  # Create barplot where the y-axis represents the size of aberration
```
jashapiro: On second look, I am pretty confused by these plots. What is actually being plotted here? I had thought it was the number of each type in each chromosome, but the `stat = "identity"` is confusing me. Would the default `stat = "count"` make more sense? 

On some investigation, it looks like these ultimately produce the same plots as the ones that I was expecting, but if you leave off the `y` aesthetic, and use the default `geom_bar()` (no stat) you get a much smaller file size and better y axis labels.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,169 @@
...
+  #   filtered_cnv_df: data.frame filtered by cutoff segmean value
+  #
+  # Return:
+  #   barplot: barplot depicting the proportion of aberrations detected across
+  #            chromosomes
+
+  # Create barplot where the y-axis represents the size of aberration
+  barplot <- ggplot2::ggplot(filtered_cnv_df,
+                             ggplot2::aes(x = chromosome,
+                                          y = as.factor(aberration),
```
jashapiro: See earlier comment about the y axis/stat

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,169 @@
...
+    dplyr::inner_join(metadata, by = c("sample" = "Kids_First_Biospecimen_ID"))
+
+  # Create barplot where the y-axis represents the size of aberration
+  barplot <- ggplot2::ggplot(meta_joined,
+                             ggplot2::aes(x = chromosome,
+                                          y = broad_histology,
+                                          fill = broad_histology)) +
+    ggplot2::geom_bar(stat = "identity") +
+    ggplot2::theme_bw() +
+    ggplot2::theme(axis.text.y = ggplot2::element_blank())
```
jashapiro: Add y axis label?

analyses/cnv-comparison/01-cnv-comparison-plotting.R
cbethell: Good question, I cannot pinpoint my initial motivation here so I can change this to install from cran.

jharenza:
> This looks pretty good now, though I am very much confused by the results: the two callers are kind of shockingly different, and I find it very strange that ControlFREEC is calling more on the smaller chromosomes. That result makes me somewhat suspicious about what is going on, and I would like to hear more about the data there.
> 
> There is one major change that I think would be good for the ggplot-based plots, and that is to combine the two data frames with an extra column designating whether the sample is from CNVkit vs ControlFREEC, then using faceting to split the two, rather than relying on cowplot. This would have the advantage of putting the two plots on the same scale. You can do this pretty easily with `dplyr::bind_rows()` using the `.id` option to specify the new column name.
> 
> I would also like to see the results for the histology plot separated by aberration type, especially since the two methods have such different proportions of gain/loss calls.
> 
> On what I think was one of the original questions, I do think this script might probably be better off converted to a notebook, to give some space to comment a bit on these results, rather than just producing the raw plots.

@cbethell and @jashapiro - would love to see this in a notebook. Of note, we are still benchmarking  (nearing the end) the CN algorithms (and have a caveat on this in the README). We had a few samples show genome-wide gain and after tweaking parameters and adding options for BAF adjustment, we have much better results. Of additional note, I thought I mentioned this, but forgive me if not, I think we would ideally use results from #128 for this ticket, though I think @cgreene and I thought this ticket could still be worked on to get the code ready for a later plug and chug. We have not created a PR for #128 yet because we realized we had to use different CN files from what we had released (containing predicted gain/loss info without the need to threshold LRR) in order to create consensus (these data will be added, hopefully in #146), but the early data is only showing ~50% consensus. We are now working on integrating Manta as a 3rd algorithm and doing a 2/3 consensus and want to toy with degree of overlap based on size of CNV, rather than 2/2. Long story short, I can help you review the results by histology if in a notebook based on what we know about certain diseases and CNV expectations!

jaclyn-taroni:
>  Of additional note, I thought I mentioned this, but forgive me if not, I think we would ideally use results from #128 for this ticket, though I think @cgreene and I thought this ticket could still be worked on to get the code ready for a later plug and chug. 

From https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/8#issuecomment-535018323: 
> I imagine the output would still be a SEG file, such that you can start with any of the SEGs we provide.

This appears to have the functionality to read in files in that format:  https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/142/files#diff-fd9d53007cdfd11cf0ae23e3eede7eabR12

> Long story short, I can help you review the results by histology if in a notebook based on what we know about certain diseases and CNV expectations!

What are you looking for in a notebook @jharenza? Do you want to see the plot here: [`cnv-comparison/plots/compare_cnv_output_barplot_histology.pdf`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6222c9589bcc68153b54d2598930707644e0d47f/analyses/cnv-comparison/plots/compare_cnv_output_barplot_histology.pdf) or are you looking for additional analyses? My inclination is to, once checked for correctness, get this merged and iterate on it.

jharenza:
@jaclyn-taroni ahh thanks for the link, i didn't readily see the plot - the notebook idea was just to see everything gathered together for quick visuals. My question for this analysis - is this more of a QC or do we want to turn this into a manuscript figure? If the latter, I think having this plotted in GISTIC-type format for the different histologies or as an IGV-type figure, grouped by histology, would be most effective for showing recapitulation of previous findings.

cbethell:
> @jaclyn-taroni ahh thanks for the link, i didn't readily see the plot - the notebook idea was just to see everything gathered together for quick visuals. My question for this analysis - is this more of a QC or do we want to turn this into a manuscript figure? If the latter, I think having this plotted in GISTIC-type format for the different histologies or as an IGV-type figure, grouped by histology, would be most effective for showing recapitulation of previous findings.

@jharenza I can work on plotting this as an IGV-type figure. My initial attempt to apply GISTIC here was unsuccessful due to compatibility issues with my OS. However, I can give it a more thorough attempt or explore other avenues to produce this figure. I will mention you on this issue for further review once produced.

jharenza:
OK - I can run GISTIC quickly on the v5 data and potentially add this to the data release, will send you the files so you can get started. I also noticed maftools updated some of their CN plotting abilities, so producing those per histology could be nice, too.

cbethell:
@jashapiro I addressed the comments you made above. Does this look ready to merge now? The changes involving GISTIC can be made and filed in a separate PR.

analyses/cnv-comparison/01-cnv-comparison-plotting.Rmd
```diff
@@ -0,0 +1,178 @@
...
+# Create the output directory if it does not exist
+if (!dir.exists(output_directory)) {
+  dir.create(output_directory, recursive = TRUE)
+}
+
+# List of file paths to the CNV data 
+cnv_list <-
+  list(
+    file.path(input_directory, "pbta-cnv-cnvkit.seg.gz"),
+    file.path(input_directory, "pbta-cnv-controlfreec.seg.gz")
```
jashapiro: Name here to save work later.
```suggestion
    cnvkit = file.path(input_directory, "pbta-cnv-cnvkit.seg.gz"),
    controlfreec = file.path(input_directory, "pbta-cnv-controlfreec.seg.gz")
```

analyses/cnv-comparison/01-cnv-comparison-plotting.Rmd
```diff
@@ -0,0 +1,178 @@
...
+
+```{r}
+# Filter CNV data by cutoff segmean using custom `filter_segmean` function
+cnv_filtered <-
+  lapply(cnv_data, filter_segmean, segmean_cutoff = 0.5)
+
+# Bind rows of dataframes in cnv_filtered
+combined_cnv_filtered <- dplyr::bind_rows(cnv_filtered, .id = "cnv_caller") 
+combined_cnv_filtered$cnv_caller <- gsub("1", "cnvkit", combined_cnv_filtered$cnv_caller)
+combined_cnv_filtered$cnv_caller <- gsub("2", "controlfreec", combined_cnv_filtered$cnv_caller)
```
jashapiro: With the named list above, this should be correct automatically, so you should be able to delete these two lines.

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,179 @@
...
+  meta_joined <- filtered_cnv_df %>%
+    dplyr::inner_join(metadata, by = c("sample" = "Kids_First_Biospecimen_ID"))
+
+  # Create barplot where the y-axis represents the size of aberration
+  barplot <- ggplot2::ggplot(meta_joined,
+                             ggplot2::aes(x = chromosome,
+                                          fill = broad_histology)) +
+    ggplot2::geom_bar() +
+    ggplot2::theme_bw() +
+    ggplot2::facet_grid(cnv_caller~aberration) +
```
jashapiro:
```suggestion
    ggplot2::facet_grid(cnv_caller ~ aberration) +
```

analyses/cnv-comparison/util/cnv-comparison-functions.R
```diff
@@ -0,0 +1,179 @@
...
+  #   barplot: barplot depicting the proportion of aberrations detected across
+  #            chromosomes
+  
+  # Create barplot where the y-axis represents the size of aberration
+  barplot <- ggplot2::ggplot(filtered_cnv_df,
+                             ggplot2::aes(x = chromosome,
+                                          fill = aberration)) +
+    ggplot2::geom_bar() +
+    ggplot2::theme_bw() +
+    ggplot2::facet_grid(cnv_caller~aberration) +
```
jashapiro:
```suggestion
    ggplot2::facet_grid(cnv_caller ~ aberration) +
```

analyses/cnv-comparison/01-cnv-comparison-plotting.Rmd
jaclyn-taroni: Use `<-` assignment though please 🙂

jashapiro: not assignment, naming!

jaclyn-taroni: Nope I'm wrong!

jaclyn-taroni: 🎢

cbethell:
> Looks good! (Except for ControlFreec...)

Thank you and agreed!

PR #143, opened 2019-10-03
Test ways to work with variables specific to CI

cgreene:
#### Purpose/implementation

I am trying to figure out a generalizable way we can support reasonable defaults for general purpose use but also select parameters that follow the same code paths with the fewer samples available in CI.

#### Issue

No existing issue.

#### Directions for reviewers

PLEASE DO NOT MERGE THIS. IT WILL BREAK EVERYTHING. I JUST NEEDED TO CREATE A PR TO TEST SOME THINGS IN CI.

#### Results

TBD

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
 
       - run:
           name: env sh hold my env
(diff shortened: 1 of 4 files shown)


cgreene:
I'm going to close this since it would be dangerous if merged. #147 has a solution based on what I learned from this PR.

PR #133, opened 2019-09-27
Ss gsea hallmark

PichaiRaman:
#### Purpose/implementation

GSVA analysis of PBTA Transcriptomic data to identify if any histologies are uniquely associated with certain pathways. The main idea here is to see if certain histologies have significantly higher pathway activity (Hallmark Pathways) than the rest of the histologies in the PBTA. 

**The method here was to**
* Perform a GSVA to get Pathway scores
* Run ANOVA to determine if a pathway was differentially expressed across PBTA histologies
* Use Tukey HSD to perform pairwise comparisons of histologies across differentially expressed pathways from the previous step

#### Issue

This is an analysis at the pathway level which will be used in conjunction with gene level analysis to make inferences as to mechanisms of tumorigenesis in certain histologies.

#### Results

* Table of pathways significantly associated with particular histologies. 
* Heatmap of significant pathways across the PBTA and in particular histologies. 
* Boxplots with specific examples. 

#### Docker and continuous integration

_Check all those that apply or remove this section if it is not applicable._

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

PR #147, opened 2019-10-04
Add CI-only variable instructions & support

cgreene:
#### Purpose/implementation

This PR does not have a scientific question.

At times, it is helpful to be able to set a variable only in CI. This occurs most frequently when we want to test the full set of code paths, but the small test dataset on CI will block or break certain code paths. This implements an approach using environment variables that allows one to do certain things only in CI, while also allowing the shell scripts to have reasonable defaults for the full set. This is important because we will want appropriate values used when we run the final set to produce the results release that is used in the manuscript.

#### Issue

There's not an issue for this per se, but discussion around PR #133 led me to test some things in #143 to come up with a potential solution to this recurring challenge.

#### Directions for reviewers

This should be ready to go live in terms of implementation. If someone has an example where they could easily implement this strategy to test it, it'd be much appreciated. You can see my testing in #143 and the corresponding CircleCI runs.

#### Results

This does not produce scientific results.

N/A. Instructions have been added to README and functionality has been added to `run_in_ci.sh`. It should not introduce breaking changes, as when no environment variables are specified nothing is altered.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,13 +177,13 @@ The most recent version of the project Docker image, which is pushed to Docker H
 docker pull ccdlopenpbta/open-pbta:latest
 ```
 
-**If you are a Mac or Windows user, the default limit for memory available to Docker is 2 GB. 
-You will likely need to increase this limit for local development.** 
+**If you are a Mac or Windows user, the default limit for memory available to Docker is 2 GB.
+You will likely need to increase this limit for local development.**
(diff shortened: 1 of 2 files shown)


cgreene:
FYI we are interviewing a postdoc candidate today, so I will be more disconnected than usual. If someone wants to make edits/changes and get this merged so that this approach can be more widely used, please feel free. Otherwise, I'll try to address changes either this weekend or early next week.

Issue #135, opened 2019-09-28
Issues reading in VarDict file


jaclyn-taroni:
#### File(s)

`pbta-snv-vardict.vep.maf.gz`

#### Release

`release-v5-20190924`

#### Link to [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/)

N/A

#### Question/issue

I'm having trouble reading the VarDict file into R in the same manner that I handle all other MAF files, specifically using:

```R
    snv_file <- data.table::fread(filename,
                                  skip = 1,  # skip version string
                                  data.table = FALSE)
```

I get the following message:

```
Avoidable 31.758 seconds. This file is very unusual: it ends abruptly without a final newline, and also its size is a multiple of 4096 bytes. Please properly end the last row with a newline using for example 'echo >> file' to avoid this  time to copy.
````

It also seemed to be hanging when I tried to `gunzip` the file. I have not fully investigated this -- I'm going to prioritize #124 with skipping this file to unblock review.

**Comment by jharenza** [comment]
@bmennis - can you recreate this file ? see above error. Thanks!

**Comment by jharenza** [comment]
@jaclyn-taroni it looks like when we download the file from S3 or CAVATICA and gunzip, it works fine and MD5sum matches, however, when I just checked MD5sum of my download from the release, it did not match, so I think the file is truncating upon download in the release. Can you check yours? Also, the uncompressed file is 29GB.

**Comment by jaclyn-taroni** [comment]
Hm, my md5sum did match what was reported in `md5sum.txt` but if yours got truncated it may explain your gunzip error.

> Also, the uncompressed file is 29GB.

Good to know. That is larger than I was expecting and will probably take a long time! I'll give it a try on my desktop tomorrow.

**Comment by jharenza** [comment]
If it cannot load locally, let us know and we can create a CAVATICA project where you can run interactive R notebooks using your code to create the file subsets (and/or consensus calls, cc: @cansav09) and we can add those results to a data release.

**Comment by jaclyn-taroni** [comment]
I got:
```
gzip: pbta-snv-vardict.vep.maf.gz: decompression OK, trailing garbage ignored
```
just now

cansav09:
[...]

PR #149, opened 2019-10-07
Use uncompressed VarDict file to generate file for CI

jaclyn-taroni:
#### Purpose/implementation

There is something off about the end of the compressed VarDict file. Currently, we don't have a subset file for VarDict on CI. This is a temporary fix for that.

#### Issue

#135 

#### Directions for reviewers

Directions for @cansav09 to run this since she is digging into the MAF files at the moment:
* Download v5 release
* `gunzip` the VarDict file, keeping the original compressed file with `-k`
* Run `./analyses/create-subset-files/create_subset_files.sh`


jaclyn-taroni:
I updated the files from running this on S3 and am rerunning the workflow now. It has passed the data download step, which is the most pertinent step.

PR #154, opened 2019-10-09
Edits to Circle CI option instructions

cansav09:
#### Purpose/implementation

Quick clarifying note about the `-` in the example for Circle CI option instructions. I was confused when I came across this. 

#### Issue
NA

#### Directions for reviewers

Merge it if you think it helps!

### DIFF
--- a/README.md
+++ b/README.md
@@ -317,14 +317,14 @@ We would remove the runs `Filter Samples` and `Cluster Samples and Plot Heatmap`
 #### Passing variables only in CI
 
 The analyses run in CI use only a small portion of the data so that tests can be run quickly.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -336,6 +336,7 @@ option_list <- list(
 ```
 
 Then we would create a shell script (perhaps `run_pathway_sig.sh`) that uses a default environment variable. If `OPENPBTA_PATHSIG` is defined, it will be used. Otherwise, a value of 0.05 is used.
+Note: the `-` before the `0.05` below is necessary notation and *not* designating a negative 0.05.
```
cgreene:
```suggestion
Note: the `-` before the `0.05` below is necessary notation for a default parameter and *not* designating a negative 0.05.
```

PR #159, opened 2019-10-18
Draft PR: SNV Caller Comparison Notebook

cansavvy:
### Purpose/implementation

After my initial series of PRs with scripts to assess snv callers individually, this PR adds the notebook which takes that initial analysis and compares the snv callers to each other. 

The eventual goal is to decide on a set of mutations from these callers that we can move forward with. 

*Two things to keep in mind about the current notebook/PR:*
1) This is only the subset of the data being shown currently, so it will have a lot more data later. This will affect aesthetics of the plots as well as memory requirements to run it. 
2) I haven't written up results/conclusion yet since we are still missing data and I have only run this on a subset of the data. 

### Issue

For SNV caller comparison #161, #103 and Tumor Mutation Burden #3 and sort of #11

### Directions for reviewers

_I'm looking for an initial "broad strokes feedback" about the plots and analyses:_
You can see the output notebook here to evaluate my initial questions: https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers.nb.html

**Some questions for this PR:** 
Which plots are useful? Which are not? 
What do you think of the current analyses? 
Are there other analyses you would like to see added? 
How do you feel about the plots' aesthetics? 

### Results

See the rough draft of this notebook output from this notebook here: https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers.nb.html

**Preliminary conclusion:**
From just the subset data it looks like VarDict should be dropped and we could move forward with mutations that are found in Mutect2, Strelka2, and Lancet. 

But, we should see how this looks after the v6 data come out and the rest of the samples are added. As of now, we do not have WXS samples for Lancet or VarDict.

#### Docker and continuous integration

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
Main _RPackages added to the Dockerfile:_
- UpSetR
- GGally 
- ggupset
- [x] This analysis has been added to continuous integration.

* [x] Run a linter
* [ ] Set the seed- I didn't, but let me know if I should.
* [ ] Comments and/or documentation up to date *- I will add more comments but after we figure out which plots we are keeping*
* [x] Double check your paths
* [x] Spell check any Rmd file or md file
* [x] Restart R and run all notebooks fresh and save

PR #163, opened 2019-10-23
Quick edits to speed up Circle CI tests

cansavvy:
#### Purpose/implementation

CircleCI tests take ~11 min in part due to some of the snv caller tests and largeness of the VarDict file (even when it is subset). 

Vaf filter has been upped from 0.10 to 0.25, this should cut down the VarDict file some more and speed up some of the analysis steps. 

#### Issue

NA

#### Results

NA

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
 
       - run:
           name: SNV Caller Analysis 
-          command: OPENPBTA_VAF_CUTOFF=0.25 ./scripts/run_in_ci.sh bash analyses/snv-callers/run_caller_evals.sh    
+          command: OPENPBTA_VAF_CUTOFF=0.5 ./scripts/run_in_ci.sh bash analyses/snv-callers/run_caller_evals.sh    
 
       - run:
           name: Sex prediction from RNA-seq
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
This still takes over 8 minutes. Is there anything you think you would not be thoroughly testing if you raised the cutoff further?

jashapiro:
One other speedup that seems worth doing... `hg38_genomic_region_annotation.rds` is not too big for github, and generating it takes a good couple minutes every time. It sonly 40MB, and won't change often, so it seems reasonable to include it in the repo. Thoughts?

cansavvy:
That sounds good, @jashapiro  I will do that change.

jaclyn-taroni:
I would still try increasing the cutoff further, unless that would prevent something from being tested.

cansavvy:
> I would still try increasing the cutoff further, unless that would prevent something from being tested.

I've upped it to 0.5. 

Another way to decrease the run time for snv caller bit is to turn off the regional analysis which I made an option for on another branch. Although I think it is a good analysis to do, I think it would be better to do it after we reach a consensus. The results of that analysis haven't added to the consensus mutation decisions. Thoughts?

jaclyn-taroni:
You're referencing https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/159#issuecomment-545006308 correct? Let's have that discussion over there.

cansavvy:
> Was removing `wxs_bed_regions_all.tsv` intentional?

Yes it is an old development remnant that was a place holder before v5 data was released and had the actual BED regions files. It should be removed.

Issue #161, opened 2019-10-22
Proposed Analysis: Getting a consensus set of SNV mutations 


cansavvy:
#### Scientific goals

From the results of the four mutation callers, determine a set of mutations that we believe are reliable and real. 

#### Proposed methods

**Overall steps:** 
1) Get an initial evaluation of each caller (implemented by #117, #118, #125, #126, and  #144). 
2) Compare the callers with each other (in progress but set up by #159)

The [snv-caller  README](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers) explains aspects of this analysis this in more detail. 

#### Required input data

pbta-snv-lancet.vep.maf.gz
pbta-snv-mutect2.vep.maf.gz
pbta-snv-strelka2.vep.maf.gz
pbta-snv-vardict.vep.maf.gz

#### Proposed timeline

The set up for this will be mostly wrapped up this week. If the data are ready, we can run this next sprint. 

#### Relevant literature/information

https://software.broadinstitute.org/cancer/cga/mutect
https://github.com/AstraZeneca-NGS/VarDict
https://github.com/Illumina/strelka
https://github.com/nygenome/lancet

cansavvy:
From in person talk with @jashapiro, plan is to: 

1) Turn distribution sina plots into density plots 
2) Do experiment with increasing VAF cutoffs -> recalculate TMB and determine if this improves TMB correlations across callers. Can we recover VarDict by increasing VAF cutoff? 
3) Turn TMB distribution into facet wrapped plot so it is more easily readable.

cansavvy:
Ran the full v6 data. Will upload and add a report. VarDict could not be run because of out-of-memory issues. Will need to optimize some memory intensive steps and figure out an acceptable VAF cutoff for VarDict to actually be run. That being said the consensus made without VarDict is probably more reliable.

cansavvy:
[consensus_mutation.maf.tsv.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3797553/consensus_mutation.maf.tsv.zip) I will add a more detailed summary of this to the README.

jaclyn-taroni:
Subsumed by #207? What do you think @cansavvy ?

jharenza:
Should we close this one since this was released with #273 ?

jaclyn-taroni:
This is probably better tracked with #275 as well

PR #164, opened 2019-10-23
Retire the `mutect-vs-strelka` analysis

cansavvy:
#### Purpose/implementation

The `mutect-vs-strelka` analysis was the original attempt to find consensus mutation calls between the two callers. The results of this analysis led to adding [VarDict and Lancet as a callers](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/103). Since the addition of these callers, the snv caller analysis had too be overhauled to accommodate and compare 4 callers. 

This means `mutect-vs-strelka` is out of date but should still be kept around for record keeping purposes. To this end, I've added a README to explain the evolution of this analysis and have also removed it from CircleCI tests.  

A minor but related change in this PR, I changed the plots from being saved to a PDF to a png so they could be added to the README. 

#### Issue

This pertains to #30 and #103 which have been superseded by the issue #161 

#### Directions for reviewers

Does the README make sense? Is there any information that should be added that I am missing? 

#### Results

NA. There are no new results, I am just retiring old ones. 

- [x] This analysis has been removed from continuous integration.


jaclyn-taroni:
Can you tell me about the `combined_results.tsv.zip` addition? I didn't spot any changes in the code such that I would expect to see this file added.

analyses/mutect2-vs-strelka2/README.md
```diff
@@ -0,0 +1,54 @@
...
+to break these variants into their single nucleotide changes.
+In these analyses, these base changes have been grouped together and collectively
+called `long_changes`.
+The higher base resolution of Strelka2, and its ability to parse apart the SNVs
+from each other, is more useful to us for this particular analyses, as the larger
+structural variants are better detected in the Manta or LUMPY analyses.
+
+### Gene Summary
+
+![](/plots/gene_summary_cor_strelka-vs-mutect2.png)
```
jaclyn-taroni: I think you may want 
```suggestion
![](plots/gene_summary_cor_strelka-vs-mutect2.png)
```
It's not showing up for me in the rich diff.

cansavvy: Strange. My Atom "GitHub Markdown" preview shows it correctly.

cgreene: The leading `/` makes it an absolute path, so you'll definitely want to remove that as @jaclyn-taroni suggests. Is this in the root directory of some Atom project, so Atom interprets `/` as equivalent to a relative path? I don't know why it'd be set that way, but it's the only thing that comes to mind.

cansavvy:
> Can you tell me about the `combined_results.tsv.zip` addition? I didn't spot any changes in the code such that I would expect to see this file added.

The code for this file creation was always there, I just hadn't uploaded it to GitHub previously. I am unsure why. It could be the past version of the file were too large for GitHub.

jaclyn-taroni:
> The code for this file creation was always there, I just hadn't uploaded it to GitHub previously. I am unsure why. It could be the past version of the file were too large for GitHub.

Okay, sounds good!

analyses/mutect2-vs-strelka2/README.md
cansavvy: Fixed it.

Issue #155, opened 2019-10-11
Analyses including multiple samples from the same individual


jashapiro:
#### Question/issue

Many of the analyses that have been proposed are sensitive to the fact that we have multiples tumor samples from different time points or tumors from the same individual. For example, biospecimens `BS_K07KNTFY` and `BS_AQMKA8NC` are both tumor WGS data (initial and recurrence, respectively) from participant `PT_00G007DM`. While this is extremely useful data, it presents questions for many particular analyses, which I would like to discuss in this issue.

In particular, analyses of mutation prevalence, variant allele frequency distributions, classification accuracy, etc. are likely to be affected by these non-independent samples. In some cases, a simple awareness of the issue will be sufficient, and analyses can be written to account for or take advantage of the redundancy in the data. However for many analyses, decisions of which samples to include or exclude will need to be made, and it would be good to have an agreed upon set of standards and procedures.

For a specific example, in the analysis of mutation co-occurrence (#13), including all samples would result in many spurious reports of co-occurrence, as it is quite common for two samples from the same individual to have the same sets of mutations. Similarly, analyses of recurrent fusions (#10), distribution of tumor mutation burden (#3), etc. will likely be affected. 

One potential solution is to use only primary tumors and/or the earliest sampled tumor from each individual in analyses such as this. However, this would miss some potential co-occurrence patterns that may be important in progression and recurrence, which might suggest that the latest tumor from each individual would be better. Doing both is of course an option as well, but I am curious to hear what others think is most appropriate. Ultimately, we may want to add a recommendation to the documentation for future analyses.

jharenza:
@jashapiro I recently had this on my mind as I saw some of the mutation analyses you described, so I am glad you are bringing this up in an issue. As you allude to, I think how we use the data also depends upon how data are being used/plotted. For example, I think it makes sense to limit to one primary tumor per patient when doing mutual exclusivity/co-occurrence analyses (#13), but we could calculate TMB (#3) for all patients and when these are plotted, plot with diagnosis and relapse separated. In fact, we expect a higher TMB in relapse tumors compared to primary, so this would be good validation. For fusions (#10), I think this is also important to discuss. We also have cell lines and some of the fusions may be lost in culture or at relapse, so these would be important things to note and I would think could be something within the scope of another issue. 

[...]

PR #167, opened 2019-10-23
Independent sample analysis (1 of 2?)

jashapiro:
#### Purpose/implementation

The purpose of this PR is to add exploration of the situations where we have multiple specimens from the same individual, particularly in the case of whole genome sequences. Included in PR is a notebook that explores the extent of the issue (some of which appeared in comments on #155). The end of the notebook, and the accompanying R function, propose a solution of random selection among the earliest samples from each participant. 

The next PR will include a standalone script that runs this function after some prefilters, generating a list of specimen IDs that can be used in any analysis where multiple WGS from the same individual would create bias. My current plan is to run this in two ways: including only primary tumors in one list, and in the other a list of primary tumors for the participants where one is available, but can include non-primary samples when primary tumor data is not available.

#### Issue

This pull addresses issue #155 

#### Directions for reviewers

Please look at the filtering steps and 

#### Results

Currently only [this notebook](http://shapbio.me/OutputShare/00-repeated-samples.nb.html)

Future PR will include files with lists of selected specimen ids for future analysis.


analyses/independent-samples/00-repeated-samples.Rmd
```diff
@@ -0,0 +1,221 @@
...
+
+
+## Select independent samples
+
+Using the information described above, I created a function `independent_samples()`
+to select the earliest biospecimen from each participant, or in the case of more 
+than one with the same date, randomly select among the available samples.
+
+An example of this function is below, applied to the WGS Solid Tissue tumor samples.
+Note that in this case, only tumors designated as primary will be included,
```
jaclyn-taroni: Can you note what primary means here at the moment (e.g., in v5) in terms of what fields and values you use?

analyses/independent-samples/independent-samples.R
```diff
@@ -0,0 +1,56 @@
...
+                                      "age_at_diagnosis_days"))
+   
+  # Choose randomly among specimens from the same participant 
+  early_ind <- early_samples %>%
+    dplyr::group_by(Kids_First_Participant_ID) %>%
+    dplyr::summarize(Kids_First_Biospecimen_ID = sample(Kids_First_Biospecimen_ID, 1)) %>%
+    dplyr::pull(Kids_First_Biospecimen_ID)
+  
+  return(early_ind)
+}
```
jaclyn-taroni: New line here.

analyses/independent-samples/independent-samples.R
```diff
@@ -0,0 +1,56 @@
...
+#' @param sample_df A data frame of samples, with columns corresponding to those
+#'   in `pbta-histologies.tsv`
+#' @param tumor_types Designates which types of tumors will be included. Options
+#'   are "primary" to include only primary tumores, "prefer_primary" to include 
+#'   primary tumors when available, but fall back to other types, or "any" to 
+#'   randomly select among all available specimens.
+#' 
+#' @return a vector of biospecimen ids
+independent_samples <- function(sample_df, 
+                           tumor_types = c("primary", "prefer_primary", "any")){
```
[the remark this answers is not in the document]
jashapiro: Easy enough to do. I don't have a strong preference on this, but I would generally feel like keeping it in the parent scripts should be sufficient. Though as I type this I can imagine some situations where the function might be called multiple times to test sensitivity to gene sets, which could potentially benefit from passing in the seed explicitly. I'll add it.

analyses/independent-samples/00-repeated-samples.Rmd
```diff
@@ -0,0 +1,223 @@
...
+seq_samples %>%
+  filter(Kids_First_Participant_ID %in% notissue_participants)
+```
+
+Since this is only one sample it seems like the best move will be to remove all derived cell lines.
+
+## Unlabeled composition samples
+
+9 samples are listed as "Not Reported" for `composition`. 
+What are these, and do we have correponding samples for which composition is known?
```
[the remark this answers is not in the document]
jharenza: Update: I will change all of the Tumor Not Reported to Tumor Solid Tissue as I confirmed these are not cell lines. However, we cannot change the Normals since this information did not exist at sample collection. However, if it was really imperative, we could probably do GSEA for tissue type and predict :P! Since some composition data is missing, we can use `sample_type` once removing cell lines, from which to select our dataset.

analyses/independent-samples/independent-samples.R
```diff
@@ -0,0 +1,62 @@
...
+   dplyr::group_by(Kids_First_Participant_ID) %>%
+   dplyr::summarize(age_at_diagnosis_days = min(age_at_diagnosis_days)) %>%
+   dplyr::left_join(sample_df, by = c("Kids_First_Participant_ID", 
+                                      "age_at_diagnosis_days"))
+   
+  # Choose randomly among specimens from the same participant 
+  early_ind <- early_samples %>%
+    dplyr::group_by(Kids_First_Participant_ID) %>%
+    dplyr::summarize(Kids_First_Biospecimen_ID = sample(Kids_First_Biospecimen_ID, 1)) %>%
+    dplyr::pull(Kids_First_Biospecimen_ID)
```
jharenza: Question about this - will this PR result in a sample set for everyone to use or will everyone generate their own sample set, because this randomization will then be different for everyone if the latter is the case.

jaclyn-taroni: I believe the plan is generating the sample set for everyone in part 2. I have a strong preference for including the sample set for everyone in the data download for single point of truth reasons.

jashapiro: Yes, the plan was to have a single sample set in part 2, with a small script that just does that. I just wanted to include the thinking and prelim analysis in this notebook.

analyses/independent-samples/00-repeated-samples.Rmd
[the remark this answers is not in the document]
jashapiro: For this analysis, I am already looking at only tumors, so that should clear everything up. Thanks for looking into it!

analyses/independent-samples/00-repeated-samples.Rmd
```diff
@@ -0,0 +1,223 @@
...
+  filter(experimental_strategy == "WGS") %>%
+  pull(Kids_First_Participant_ID) %>%
+  unique() 
+
+No_WGS <- WXS_participants[!(WXS_participants %in% WGS_participants)]
+
+length(No_WGS)
+```
+
+So there are currently 16 samples with missing WGS data, but that should change in a future release.
```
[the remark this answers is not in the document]
jashapiro: I'm okay with this... I'll probably go ahead and produce sample files with and without WXS.

analyses/independent-samples/independent-samples.R
jashapiro: And when I say single sample set, I was actually thinking there might be up to 4: With and without WXS, and with and without non-primary tumors. Thoughts?

jaclyn-taroni: I like that idea @jashapiro. Good to start thinking about how to document those in the README.

PR #169, opened 2019-10-24
Move SNV caller analysis to last step in CI

jaclyn-taroni:
#### Purpose/implementation

The SNV caller step's run time has been shortened with #168 and #163, but it still is ~3x longer than any other step. Here I'm moving it to the last step (per a suggestion from @jashapiro yesterday) and (hopefully) making it more obvious where folks should add their analyses.

As a result, people submitting PRs to see if they need to make changes more quickly than if they added their analyses after this SNV step.

Alternative idea: add the "add your analysis here" after 

```
      - run:
          name: List Data Directory Contents
          command: ./scripts/run_in_ci.sh ls data/testing
```

#### Issue

N/A - contributor friendliness

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,10 +43,6 @@ jobs:
           command: ./scripts/run_in_ci.sh ./analyses/transcriptomic-dimension-reduction/ci-dimension-reduction-plots.sh 
 
       - run:
-          name: SNV Caller Analysis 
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
Yea I agree that it could have some unintended consequences but is something maintainers should keep an eye on in general. I will leave as is for now. Everything but the SNV caller step is relatively quick, so it should not cause too much friction this way.

Issue #162, opened 2019-10-23
Update: sample distribution plots accounting for multiple samples from the same individual


jaclyn-taroni:
In light of https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/155, the plots from #5 that are in the [`sample-distribution-analysis`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/70879095fd79d7be5bbb21acf6b743eb96552ae6/analyses/sample-distribution-analysis) should probably be updated to remove "duplicates" when getting an overall picture of the cohorts (e.g., distribution of histologies).  It is also of interest to display the breakdown of the `tumor_descriptor` of `pbta-histologies.tsv`, perhaps with some smoothing.

jaclyn-taroni:
Related: https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/38

jaclyn-taroni:
I addressed part of this with #170 and can continue to work on this.

PR #170, opened 2019-10-24
Sample distribution plots: account for multiple samples from same individual

jaclyn-taroni:
#### Purpose/implementation

The sample distribution plots did not account for multiple samples from the same individual (#155) or the fact that there are multiple experimental strategies in the `pbta-histologies.tsv` file. 

Here, I'm limiting the plots to tumor tissue only and to distinct pairs of participant ids and whatever is being plotted.

Note it is not the case that a single participant identifier will map to a single `disease_type_new`/`broad_histology`/`short_histology` value. So, we have a higher number of samples in these plots than number of unique individuals. This seems appropriate for these figures.

#### Issue

#162 - does not look at the `tumor_descriptor` breakdown

#### Directions for reviewers

* Does it seem reasonable to limit these plots to tumors and exclude cell lines for these plots? These will probably be included in an overview figure (not the interactive ones so much).
* Does my filtering/group scheme seem consistent with what we'd like to present?

#### Results

Same overall picture, which is what I would expect, but the numbers are more consistent with the number of participants.


jaclyn-taroni:
Funnily enough I was looking at what you added with #167 and making this probably overly complicated for its specific purpose.

PR #172, opened 2019-10-25
SNV caller analysis part 1 revamp

cansavvy:
### Purpose/implementation

This PR implements a few smaller changes that were initially discussed and implemented on a retired PR #159.

#### Changes in this PR: 

1) Makes the regional analysis optional throughout the pipeline. 
    - Adds as an `optparse` option in both `01-calculate_tmb_vaf.R` and in `02-run_eval.R`
    -  Employs the `--no_region` option in `run_caller_evals.sh`
2) Fixes a chromosome labeling problem with the COSMIC file. 
   - Chromosome's X and Y were labeled as 23 and 24 though our data has them as X and Y. 

### Issue

For SNV caller comparison #161, #103 and Tumor Mutation Burden #3 and sort of #11

### Directions for reviewers

Are these changes implemented well? 
Does everything seem to be in order? 

Is there a better way for me to have the non-regional report as opposed to having a whole other copy of that template? I do not like having the whole thing repeated but am unsure as to an alternative? 

### Results

No new results from these changes. 

*These changes do not require new packages or new CircleCI tests. *

* [x] Run a linter
* [x] Set the seed- I didn't, but let me know if I should.
* [x] Double check your paths


analyses/snv-callers/README.md
```diff
@@ -12,14 +12,15 @@ The GDC has [good documentation on the fields](https://docs.gdc.cancer.gov/Data/
   * [Genomic regional analyses](#genomic-regional-analyses)  
   * [Tumor mutation burden](#tumor-mutation-burden-calculation)
   * [COSMIC mutation overlap](#cosmic-mutation-overlap)  
-* Comparison analysis - in development
+* [Comparison analysis](#comparison-of-callers)
  * [Mutation IDs](#mutation-ids)
+ * [Combinations of callers]()
```
jashapiro: No link?

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -97,29 +99,48 @@ option_list <- list(
     metavar = "character"
   ),
   make_option(
-    opt_str = "--vaf_filter", type = "numeric", default = 0,
+    opt_str = "--vaf_filter", type = "character", default = "0",
```
jashapiro: Is there a reason you changed this to character? I prefer numeric when that is what you need.

analyses/snv-callers/util/plot_functions.R
```diff
@@ -34,18 +34,23 @@ base_change_plot <- function(vaf_df, exp_strategy = "BOTH", filter_cutoff = 0) {
 
   # Count the number of each type of base change
   base_count_df <- vaf_df %>%
-    dplyr::mutate(change = as.factor(change),
+
+    dplyr::mutate(
+      change = as.factor(change),
       # Change factor level order so ins and del are at the end
-      change = factor(change, c(grep("ins|del", levels(change), value = TRUE, invert = TRUE), "ins", "del"))) %>% 
+      change = factor(change, c(grep("ins|del", levels(change), value = TRUE, invert = TRUE), "ins", "del"))
```
jashapiro: Is this where the long_change snuck in? If so, move it to the back too?

Easier way to to this is with `forcats`
`forcats::fct_relevel(change, "ins", "del", "long_change", after = Inf)`

cansavvy: Cool beans. I didn't like how messy my method was but wasn't sure of a cleaner alternative.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansavvy: With the `optparse` options and other upstream things were inconsistent so this was my attempt to make it all the same. Let me take another look and figure out why I thought this was simpler when on face value it doesn't seem that way.

analyses/snv-callers/README.md
cansavvy: This section shouldn't be here until a later PR, so I will remove it from the TOC.

PR #174, opened 2019-10-25
Getting consensus mutations list  by comparing callers

cansavvy:
### Purpose/implementation

This is a side analysis that originated from [this original PR and analysis](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/159#issuecomment-545096465). 

*Note:* This PR depends on changes made in #172 and shouldn't be reviewed or tested until that is merged. 

After my initial series of PRs with scripts to assess snv callers individually, this PR adds the notebook which takes that initial analysis and compares the snv callers to each other. 

### Issue

For SNV caller comparison #161, #103 and Tumor Mutation Burden #3 and sort of #11

### Directions for reviewers

See the rough draft of this notebook output from this notebook here: https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers.nb.html

**Some questions for this PR:** 
What do you think of the current analyses? 
Are there other analyses you would like to see added? 
How do you feel about the `setdiff` step for ignoring the VAF filter mutation files? Is there a preferred work around? Should I reorganize the files so this isn't necessary?

Is there a better, more elegant way to make the data `rbind` together seamlessly? 

### Results

See the rough draft of this notebook output from this notebook here: https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers.nb.html

**Preliminary conclusion:**
VarDict makes too many calls and at variants that are very low VAF.  The other three callers are fairly reasonable and inline with each other. I propose we drop VarDict and move forward with mutations called by all three of Lancet, Strelka2, and Mutect2. 

But, we should see how this looks after the v6 data come out and the rest of the samples are added. As of now, we do not have WXS samples for Lancet or VarDict.

Main _RPackages added to the Dockerfile:_
- UpSetR
- GGally 
- ggupset

#### PR Checklist
* [x] Run a linter
* [x] Set the seed- I didn't, but let me know if I should.
* [x] Double check your paths


cansavvy:
Using `library(grid)` was my solution for now for the CircleCI error of :
```
Error in grid.newpage() : could not find function "grid.newpage"
```
For some reason `UpSetR` is not calling the `grid::grid.newpage` function properly. I don’t want to follow too much of a rabbit hole beyond this, but my hunch is that this is a very specific problem tied to a specific version of `UpSetR` and I do see some changes in their [code regarding this, but it looks like their changes may be incomplete. ](https://github.com/hms-dbmi/UpSetR/commit/8a7c8e224c9fba12eebbdbfe1ce83fa30cb81be8) I'm going to roll `UpSetR` back to a version that is prior to this changes and see if that fixes it. But the temporary fix is just `library(grid)`.

cansavvy:
That worked! Okay. This is ready for your review whenever you are ready, @jaclyn-taroni

analyses/snv-callers/compare_snv_callers.Rmd
```diff
@@ -0,0 +1,484 @@
...
+
+  # Summarize by the number of times each category shows up
+  count_df <- df %>%
+    dplyr::group_by(caller, eval(parse(text = var))) %>%
+    dplyr::summarise(count = dplyr::n())
+
+  # Calculate sum totals for each caller
+  totals <- count_df %>%
+    dplyr::group_by(caller) %>%
+    dplyr::summarise(total = sum(count, na.rm = TRUE))
```
jashapiro: You can simplify these with some use of count() and add_count(). I *think* the below does the same as these plus the next few lines.

```
count_df <- df %>%
    dplyr::count(caller, eval(parse(text = var)), name = "count") %>%
    dplyr::add_count(caller, wt = count, "total") %>%
    dplyr::mutate(percent = count/total)
```

analyses/snv-callers/compare_snv_callers.Rmd
```diff
@@ -0,0 +1,484 @@
...
+# Save this plot
+png(file.path(consensus_plots_dir, "upset_plot.png"), width = 900, height = 900)
+upset
+dev.off()
+
+# Print this out
+upset
+```
+
+## VAF for combination sets of callers
```
jashapiro: I feel like this calculation & the actual combination of these callers should be in a separate script. While we hope not to have to run this repeatedly, it might be nice to run the consensus generation separately from the full notebook.

analyses/snv-callers/compare_snv_callers.Rmd
```diff
@@ -0,0 +1,484 @@
...
+Let's get the mutation ids for these. 
+
+```{r}
+consen_mutations <- callers_per_mutation %>%
+  dplyr::filter(caller_combo == "lancet-mutect2-strelka2") %>%
+  dplyr::select(-caller_combo) %>%
+  dplyr::pull(mutation_id)
+```
+
+Isolate the mutations to only these mutations, use the Strelka2 stats.
```
jashapiro: I know we said use strelka stats, but we already calculated median VAF? Should we add that on?

jaclyn-taroni:
> Using `library(grid)` was my solution for now

Weird! Taking a look now.

analyses/snv-callers/compare_snv_callers.Rmd
cansavvy: Will try to use this!

jaclyn-taroni:
Is there a reason you need a development version (i.e., one not on CRAN) of `UpSetR`?

cansavvy:
> Is there a reason you need a development version (i.e., one not on CRAN) of `UpSetR`?

I installed the commit from the same version as the one on CRAN originally (version 1.4.0), and it wasn't installing right. I can go back and try it again since there were other package installation problems (GGally) that also were happening at the same time.

cansavvy:
CRAN version worked this time! This now installs UpSetR from CRAN.

analyses/snv-callers/compare_snv_callers.Rmd
```diff
@@ -0,0 +1,484 @@
...
+  # Calculate percentages based on totals
+  count_df <- count_df %>%
+    as.data.frame() %>%
+    dplyr::mutate("percent" = count_df$count / totals$total[match(
+      count_df$caller,
+      totals$caller
+    )])
+
+  # Rename the column as what it was before
+  colnames(count_df)[2] <- var
```
jaclyn-taroni: Can maybe use `dplyr::rename` as part of the pipe suggested by @jashapiro instead?

jaclyn-taroni:
It looks like splitting things has resulted in more lines of code. In the interest of getting this reviewed thoroughly, I think we want to split this up and spread review out across multiple people to move things along efficiently (sorry @cansavvy 😕).

cansavvy:
> It looks like splitting things has resulted in more lines of code. In the interest of getting this reviewed thoroughly, I think we want to split this up and spread review out across multiple people to move things along efficiently (sorry @cansavvy 😕).

Yes I realized this and will plan to split it up some more. I do think these changes will make it better overall though.

cansavvy:
I'm going to close this PR because it will be replaced by ~3 smaller PRs which I will tie a reference back to here.

PR #175, opened 2019-10-25
Add an issue template specifically for updating an analysis

jaclyn-taroni:
#### Purpose/implementation

As new data releases come out or we come to appreciate new issues with the data (looking at you #155  👀 ), certain analysis modules will need to be updated. 

I filed an issue about updating a specific analysis (#162) because better in an issue than solely in my head!

It occurred to me that it would be useful to increase visibility into what updates need to be made and I think an issue template could help with that. I don't expect that we will always know for sure what data release was used for a current version of an analysis, but where we do that could be helpful information.

#### Issue

N/A - contributor/organizer friendliness

#### Directions for reviewers

Is there anything that should be added or removed?


cgreene:
This is a good idea and I approve this. I have looked on my phone at the
changes.

Right now the GitHub mobile interface isn't working great on my phone but I
approve. If wifi ever worked on Amtrak I'd formally do it from my laptop,
but no such luck.

Issue #19, opened 2019-07-12
Planned Analysis: Molecularly subtype all tumors


jharenza:
Without methylation data, the goal of this analysis is to best molecularly subtype all brain tumors based on defining features.
- [x] Medulloblastomas (MBs) into SHH, WNT, Group 3, and Group 4 - @PichaiRaman has created a classifier [here](https://github.com/PichaiRaman/MedulloClassifier) and classified the samples with RNA-Seq.
- [ ] ATRTs into SHH, MYC, TYR. See: [1-s2.0-S1535610816300356-main.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3387143/1-s2.0-S1535610816300356-main.pdf)
- [ ] Ependymomas into ST-RELA, ST-YAP1, PF-A, and PF-B. Will require some genome-wide copy number examination. See: [0D556E30-D832-4D6F-9F2F-49554F68D56F.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3387093/0D556E30-D832-4D6F-9F2F-49554F68D56F.pdf)
- [ ] High-grade gliomas, according to brain region (if known, eg: midline vs hemispheric) and defining lesion (eg H3F3A K27/28M, G34/5R/V, HIST1H3B K27/28M, IDH-mutant, ACVR1-mutant, PDGFRA, and mesenchymal subtypes). See: [WHO 2016 classifications](https://braintumor.org/wp-content/assets/WHO-Central-Nervous-System-Tumor-Classification.pdf) 
- [ ] Non-MB embryonal tumors (old PNET classified tumors) into ETMRs and CNS embryonal NOS
- [ ] Oligodendrogliomas as IDH-mutant and 1p/19q codeleted
- [ ] Chordomas - subtype into poorly differentiated (INI1/SMARCB1 loss)

jharenza:
I prefer to use [this](https://docs.google.com/spreadsheets/d/1Sa0hNX1lje40HdBpWiLUBY6-53UzvQczFLYb3uIm2jw/edit?usp=sharing) as a working list of tumor histologies which we can molecularly subtype. I have gone through and broadly classified each histology based on the WHO 2016 classifications and also classified "Other" tumors based on internal CBTTC data.

linneacripe:
Medulloblastoma- need five subtypes, SHH p53 mutated and SHH p53 wild type

jharenza:
[...]

PR #176, opened 2019-10-25
Addition of script to produce oncoprint

cbethell:
### Purpose/implementation

This draft PR adds the notebook which produces oncoprints given the output of strelka2, mutect2, lancet, and vardict. 

The goal is to produce an oncoplot including cnv and fusion information, in addition to the snv information already displayed. For reference, here is a plot produced using a workflow authored by @jharenza:
2019-10-24-brain-oncoprint18-1
In an attempt to adapt said workflow to our data, I discovered that the format of our data files are different from the input files used to produce the above plot. Therefore, I was unable to successfully plug our input files into the code. That being said, it seems that this may require some amount of data wrangling which is still in progress. However, please feel free to offer your relevant suggestions at any point. 

### Issue

This addresses issue #6 on producing an oncoprint showing the landscape of genetic lesions across PBTA.

### Directions for reviewers

1. Is there a filtering step you feel should be added to this notebook?
2. Does the first plot where the data from all four callers are combined seem beneficial? 
3. Do you feel the aesthetics of the plots are appropriate?
4. The features of the metadata I chose to annotate the plots further include `broad_histology`, `short_histology`, `reported_gender`, and `tumor_descriptor`. Do these features seem to be the most appropriate annotations for these plots?

The plots can be found in the `plots` directory of this module, and in the html output of this notebook [here](https://cbethell.github.io/open-pbta-output/01-plot-oncoprint.nb.html).

### Results

Oncoprints displaying each snv callers’ output across this dataset has been produced thus far. 
The output files are:
- `analyses/oncoprint-landscape/plots/combined_maf_oncoprint.png`
- `analyses/oncoprint-landscape/plots/strelka2_oncoprint.png`
- `analyses/oncoprint-landscape/plots/mutect2_oncoprint.png`
- `analyses/oncoprint-landscape/plots/lancet_oncoprint.png`
- `analyses/oncoprint-landscape/plots/vardict_oncoprint.png`

* [x] Run a linter
* [ ] Set the seed - Not needed.
* [x] Double check your paths


jharenza:
@cbethell Thanks for working on this! This will probably be the first or second figure and most critiqued from a biological standpoint to represent the tumors landscape accurately, so I have a few thoughts:
- Yes, there was some data wrangling for adding copy number and fusions to these plots. For fusions, I appended them to the MAF file, so once #166 is completed, we can use that final filtered file to add high-confidence fusions to the MAF to be integrated into the plot. This will also require you to specify `Fusion` as a variant classification upon reading the MAF.
- For copy number, once #128 is completed, these can also be added. This may be tricky, as we are dealing with NGS tools instead of array, and in general, it makes more sense to add to an oncoprint focal amplifications and homozygous deletions, rather than any genes with CN changes due to broad gain/loss. For these, I had to threshold a bit, but once we see the results from the consensus calls, I think we can better tackle what this means.
- Regarding the multiple MAFs, I would suggest just working with one for now (as we will use consensus SNV calls from a final MAF from #174 as input to this figure). This can be used as proof of principle. 
- One thing that is tricky about MAF files is that using the files as they are results in many private variants coming through into the MAF. The _MUC_ and _TTN_ genes, for example, are some of the largest genes in the genome and by default are most mutated. These are generally not considered cancer driver genes, so oftentimes, we will come up with a gene list for subsetting the MAF. Top N (25-50?) brain tumor driver genes, for example. For this, perhaps we create an issue to create a driver gene list, and we can potentially add this to the data release - @jaclyn-taroni - thoughts? I had created two gene driver lists for pediatric brain tumors - [long](https://github.com/marislab/create-pptc-pdx-oncoprints/blob/master/data/brain-goi-list-old.txt) and [short](https://github.com/marislab/create-pptc-pdx-oncoprints/blob/master/data/brain-goi-list.txt) based on the literature. We could also try using these to start and see what the plots look like?
- Finally, I think that we will also want to add `molecular_subtype` to these plots. Currently, this is pretty sparse (and I realize MB subtypes are only denoted for RNA because that's what the classifier used to predict them, but should be added at the patient level), but will be addressed with #19.

jaclyn-taroni:
> Regarding the multiple MAFs, I would suggest just working with one for now (as we will use consensus SNV calls from a final MAF from #174 as input to this figure). This can be used as proof of principle.

Let's move to using a script where you accept a MAF file as a command line argument @cbethell.

For these three points:

> * Yes, there was some data wrangling for adding copy number and fusions to these plots. For fusions, I appended them to the MAF file, so once #166 is completed, we can use that final filtered file to add high-confidence fusions to the MAF to be integrated into the plot. This will also require you to specify Fusion as a variant classification upon reading the MAF.
> * For copy number, once #128 is completed, these can also be added. This may be tricky, as we are dealing with NGS tools instead of array, and in general, it makes more sense to add to an oncoprint focal amplifications and homozygous deletions, rather than any genes with CN changes due to broad gain/loss. For these, I had to threshold a bit, but once we see the results from the consensus calls, I think we can better tackle what this means.
> * Finally, I think that we will also want to add molecular_subtype to these plots. Currently, this is pretty sparse (and I realize MB subtypes are only denoted for RNA because that's what the classifier used to predict them, but should be added at the patient level), but will be addressed with #19.

We can get this code working and through review and make these updates as they come in. As far as the driver lists are concerned, let's try adding the download of the driver lists @jharenza linked to a shell script that includes the following steps:

1. Download driver lists.
2. Run the oncoplot R script.

So to accomplish step one, you could do something like (untested):
```sh
mkdir driver-lists
wget https://github.com/marislab/create-pptc-pdx-oncoprints/raw/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list-old.txt --directory-prefix=driver-lists -O brain-goi-list-long.txt
wget https://github.com/marislab/create-pptc-pdx-oncoprints/blob/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list.txt --directory-prefix=driver-lists -O brain-goi-list-short.txt
```

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,210 @@
...
+
+# Install maftools
+if (!("maftools" %in% installed.packages())) {
+  install.packages("BiocManager")
+  BiocManager::install("maftools")
+}
+library(maftools)
+
+# Define a color vector for plots 
+colores = c("Missense_Mutation" = "#35978f",
```
jaclyn-taroni:
```suggestion
colores <- c("Missense_Mutation" = "#35978f", 
```

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,210 @@
...
+      "Fusion",
+      "Multi_Hit",
+      "Multi_Hit_Fusion"
+    )
+  )
+
+#### Specify genes -------------------------------------------------------------
+
+# Read in gene list
+goi_list <-
```
jaclyn-taroni: I would also make this a command line option.

analyses/oncoprint-landscape/run-oncoprint.sh
```diff
@@ -0,0 +1,20 @@
...
+# This script should always run as if it were being called from
+# the directory it lives in.
+script_directory="$(perl -e 'use File::Basename;
+  use Cwd "abs_path";
+  print dirname(abs_path(@ARGV[0]));' -- "$0")"
+cd "$script_directory" || exit
+
+mkdir driver-lists
+wget --directory-prefix=driver-lists -O driver-lists/brain-goi-list-long.txt https://github.com/marislab/create-pptc-pdx-oncoprints/raw/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list-old.txt 
+wget --directory-prefix=driver-lists -O driver-lists/brain-goi-list-short.txt https://github.com/marislab/create-pptc-pdx-oncoprints/blob/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list.txt
```
jaclyn-taroni: I thought the `blob` -> `raw` was necessary and remembered to change the first one but perhaps not
```suggestion
wget --directory-prefix=driver-lists -O driver-lists/brain-goi-list-short.txt https://github.com/marislab/create-pptc-pdx-oncoprints/raw/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list.txt
```

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,210 @@
...
+            "Gain" = "#D55E00",
+            "High_Level_Gain" = "#FF0000",
+            "Multi_Hit_Fusion" = "#CD96CD")
+
+#### Functions -----------------------------------------------------------------
+
+# Get `magrittr` pipe
+`%>%` <- dplyr::`%>%`
+
+plot_oncoplot <- function(maf_file, filename){
```
jaclyn-taroni: Also think this is now a MAF object and not a MAF file.

cbethell: Good point, this change will be in the next commit.

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+option_list <- list(
+  optparse::make_option(
+    c("-m", "--maf_file"),
+    type = "character",
+    default = NULL,
+    help = "file path to MAF file that contains snv information",
+  ),
+  optparse::make_option(
+    c("-c", "--cnv_file"),
+    action = "store_true",
```
jaclyn-taroni: My understanding is that you only want to use `action = "store_true"` if this is a logical ([`make_option` docs](https://www.rdocumentation.org/packages/optparse/versions/1.6.4/topics/make_option))

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+    )
+  )
+
+#### Specify genes -------------------------------------------------------------
+
+if (!is.null(opt$goi_list)) {
+  # Read in gene list
+  goi_list <-
+    read.delim(
+      file.path("driver-lists", "brain-goi-list-long.txt"),
```
jaclyn-taroni: This should be the command line option, not hardcoded.

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+      sep = "\t",
+      header = F,
```
jaclyn-taroni:
```suggestion
      header = FALSE,
```

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+      as.is = T
```
jaclyn-taroni:
```suggestion
      as.is = TRUE
```

analyses/oncoprint-landscape/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+_The oncoprint can be viewed below:_
+![Oncoprint](plots/maf_oncoprint.png)
+
+## Folder structure 
+
+The structure of this folder is as follows:
+
+```
+oncoprint-landscape
```
jaclyn-taroni: This should probably be the output of `tree`: https://rschu.me/list-a-directory-with-tree-command-on-mac-os-x-3b2d4c4a4827

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+# Path to output directory for plots produced
+plots_dir <-
+  file.path(root_dir, "analyses", "oncoprint-landscape", "plots")
+
+if (!dir.exists(plots_dir)) {
+  dir.create(plots_dir)
+}
+
+genes <- NULL
+cnv_file <- NULL
```
jaclyn-taroni: Leftover from development?

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,237 @@
...
+
+# Install maftools
+if (!("maftools" %in% installed.packages())) {
+  install.packages("BiocManager")
+  BiocManager::install("maftools")
+}
+library(maftools)
+
+# Source the color palette for plots
+source(file.path("util", "oncoplot-palette.R"))
```
jaclyn-taroni: This should utilize `root_dir` so it's looking in the right directory no matter where this script is called from.

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,254 @@
...
+data_dir <- file.path(root_dir, "data")
+
+# Path to output directory for plots produced
+plots_dir <- file.path(root_dir, "analyses", "oncoprint-landscape", "plots")
+
+if (!dir.exists(plots_dir)) {
+  dir.create(plots_dir)
+}
+
+genes <- NULL
```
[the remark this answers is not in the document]
jaclyn-taroni: You have the logic down below now: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/176/files#diff-f1245b275299730475f8b3d538c277f6R204 so I think this can come out.

analyses/oncoprint-landscape/util/oncoplot-palette.R
```diff
@@ -0,0 +1,32 @@
...
+                   "Start_Codon_Del" = "#56B4E9",
+                   "Fusion" = "#7B68EE",
+                   "Multi_Hit" = "#f46d43",
+                   "Hom_Deletion" = "#313695",
+                   "Hem_Deletion" = "#abd9e9",
+                   "Amplification" = "#c51b7d",
+                   "Loss" = "#0072B2",
+                   "Gain" = "#D55E00",
+                   "High_Level_Gain" = "#FF0000",
+                   "Multi_Hit_Fusion" = "#CD96CD")
```
jaclyn-taroni: New line here

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,241 @@
...
+  # TODO: Filter and set up `cnv_file` to be in the column format -
+  # "Hugo_Symbol, Tumor_Sample_Barcode, Variant_Classification" as required by
+  # the `read.maf function`
+}
+
+# Read in fusion file
+if (!is.null(opt$fusion_file)) {
+  fusion_file <- data.table::fread(opt$fusion_file, stringsAsFactors = FALSE)
+
+  #### Incorporate Fusion Data -------------------------------------------------
```
jaclyn-taroni: Can you add a `#TODO` here as well? Specifically I think this might need to change once we get the consensus calls (I'm not sure you'll need to do the separation in quite the same way).

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,241 @@
...
+if (!is.null(opt$cnv_file)) {
+  cnv_file <- data.table::fread(opt$cnv_file, stringsAsFactors = FALSE)
+  # TODO: Filter and set up `cnv_file` to be in the column format -
+  # "Hugo_Symbol, Tumor_Sample_Barcode, Variant_Classification" as required by
+  # the `read.maf function`
+}
+
+# Read in fusion file
+if (!is.null(opt$fusion_file)) {
+  fusion_file <- data.table::fread(opt$fusion_file, stringsAsFactors = FALSE)
```
jaclyn-taroni:
```suggestion
  fusion_file <- data.table::fread(opt$fusion_file, data.table = FALSE, stringsAsFactors = FALSE)
```
probably will make things a bit easier

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -0,0 +1,241 @@
...
+    dplyr::group_by(ID) %>%
+    dplyr::summarise(Sample = dplyr::n()) %>%
+    as.data.frame()
+
+  reformat_multi$Variant_Classification <-
+    ifelse(reformat_multi$Sample == 1, "Fusion", "Multi_Hit_Fusion")
+  reformat_multi <- reformat_multi %>%
+    tidyr::separate(ID, c("Tumor_Sample_Barcode", "Hugo_Symbol"), sep = ";")
+  reformat_multi$Sample <- NULL
+  reformat_multi$Variant_Type <- "OTHER"
```
jaclyn-taroni: I think this section would benefit from using more R-friendly column names and it could be simplified by using additional dplyr functionality:

```R
fus_sep <- fusion_file %>%
  # can add a note here about 5' and 3'
  tidyr::separate(FusionName, c("Gene1", "Gene2"), sep = "--") %>%  
  dplyr::select(Sample, Gene1, Gene2)

reformat_fusion <- fus_sep %>% 
  # here we want to tally how many times the 5' gene shows up as a fusion hit 
  # in a sample
  dplyr::group_by(Sample, Gene1) %>%
  dplyr::tally() %>%
  # if the sample-5' gene pair shows up more than once, call it a multi hit 
  # fusion
  dplyr::mutate(Variant_Classification = 
                  dplyr::if_else(n == 1, "Fusion", "Multi_Hit_Fusion"),
                # required column for joining with MAF
                Variant_Type = "OTHER") %>%
  # correct format for joining with MAF
  dplyr::rename(Tumor_Sample_Barcode = Sample, Hugo_Symbol = Gene1) %>%
  dplyr::select(-n)
```

Should be equivalent. Then you're ready to bind `reformat_fusion`.

PR #179, opened 2019-10-25
Make SNV analysis have a save to RDS file option 

cansavvy:
#### Purpose/implementation

Not a scientific question but a data handling one. This has to do with the SNV Caller evaluation scripts. Because of the format of the MAF columns, sometimes things get messed up when I write and then read TSV files. So instead, to transfer from one analysis to another more easily, I'm adding an option to allow me to save VAF and TMB stats to RDS files. 

#### Issue

Related to the code that is working on #161 

#### Directions for reviewers

Are the RDS options implemented well?
Is the documentation for the added option clear? 
Any potential red flags you see with this change? 

#### Results

No new results besides being able to save to RDS files instead of TSV

No new packages or tests needed. The old tests will test if this works.


analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -249,9 +276,14 @@ if (file.exists(vaf_file) && !opt$overwrite) {
 
   message(paste(nrow(vaf_df), "mutations left after filter and merge"))
 
-  # Write this to a TSV file
-  vaf_df %>% readr::write_tsv(vaf_file)
-
+  # Write the vaf file
+  if (opt$file_format == "rds") {
```
jaclyn-taroni: If you flip the order of these, that would be more consistent with RDS being the default, i.e., save a TSV if you asked for a TSV, otherwise save an RDS. Also seems to be what you do in the other file.

cansavvy: Will do.

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
```diff
@@ -131,6 +141,15 @@ opt$bed_wgs <- file.path(root_dir, opt$bed_wgs)
 opt$bed_wxs <- file.path(root_dir, opt$bed_wxs)
 opt$cosmic <- file.path(root_dir, opt$cosmic)
 
+# Bring along the file suffix. Make to lower.
+file_suffix <- tolower(opt$file_format)
+
+# Check that the file format is supported
+if (!(file_suffix %in% c('rds', 'tsv'))) {
+  stop("Option used for file format (-f) is not supported. Only 'tsv' or 'rds'
```
jaclyn-taroni:
```suggestion
  warning("Option used for file format (-f) is not supported. Only 'tsv' or 'rds'
```
And then say "defaulting to RDS" ?

analyses/snv-callers/scripts/02-run_eval.R
```diff
@@ -106,6 +115,15 @@ option_list <- list(
 # Parse options
 opt <- parse_args(OptionParser(option_list = option_list))
 
+# Bring along the file suffix. Make to lower.
+file_suffix <- tolower(opt$file_format)
+
+# Check that the file format is supported
+if (!(file_suffix %in% c('rds', 'tsv'))) {
+  stop("Option used for file format (-f) is not supported. Only 'tsv' or 'rds'
```
jaclyn-taroni: Same comment about changing this to a warning

analyses/snv-callers/scripts/01-calculate_vaf_tmb.R
cansavvy: Okay. Will do.

Issue #13, opened 2019-07-12
Planned Analysis: Co-Occurence / Mutual Exclusivity 


PichaiRaman:
Determining genetic lesions (Mutation, CNV, Fusion) and/or pathways which co-occur or are mutually exclusive across the PBTA. This could help associate lesions with pathways or define potential synthetic lethality relationship.

cansav09:
Would we be looking as an end goal to make some figure like this where we would likely have disease type on the bottom? ![10 1038_ng1975-f2_large_2](https://user-images.githubusercontent.com/23458084/61900827-67e3ad80-aeec-11e9-82ea-a959d85b4761.jpg) (from Rotika et al, 2019).

cansav09:
Would we be looking to have something like this figure for this analysis, but with the disease type along the bottom? ![Screen Shot 2019-07-25 at 3 01 05 PM](https://user-images.githubusercontent.com/23458084/61901156-23a4dd00-aeed-11e9-9d69-d17c405bf5fb.png) This is from Rotika et al, 2019 doi: http://dx.doi.org/10.1101/566455.

cansav09:
And after #19 is done, the molecular subtypes could also be added.

cansav09:
One of my initial questions is how we'd like to combine results that measure similar things. 

For example, we have two fusions results. Do we want only take the common fusions from both files?  
```
arriba.fusions.tsv
star-fusion.fusions.tsv
```
And if I understand correctly, these all have results that overlap: 
```
strelka2.maf
manta-sv.maf
mutect2.maf
```
Note that I'm still digging into all these files and determining what's here, so after I do some initial analyses I may come back with a suggestion on how to handle this, but if someone knows this data better and has an idea of what we want to do first, then I'm of course open to those suggestions.

jharenza:
> One of my initial questions is how we'd like to combine results that measure similar things.
> 
> For example, we have two fusions results. Do we want only take the common fusions from both files?
> 
> ```
> arriba.fusions.tsv
> star-fusion.fusions.tsv
> ```
> 
> And if I understand correctly, these all have results that overlap:
> 
> ```
> strelka2.maf
> manta-sv.maf
> mutect2.maf
> ```
> 
> Note that I'm still digging into all these files and determining what's here, so after I do some initial analyses I may come back with a suggestion on how to handle this, but if someone knows this data better and has an idea of what we want to do first, then I'm of course open to those suggestions.

[...]

PR #181, opened 2019-10-29
Initial Co-occurence plots  

jashapiro:
#### Purpose/implementation

Which pairs of genes are repeatedly mutated across tumors?

This script compiles lists of genes mutated in each sample, then looks for pairs of genes that are mutated together across samples more (or less) often than would be expected by chance. 
Analysis is on the gene level, so a single mutation in a gene counts the same as many mutations in the gene.
Also, current analysis excludes synonymous mutations and intron mutations, but includes other non-coding mutations.

Currently looks at all tumor types, and choses which genes to analyze based on finding the most mutated genes.

This is the basic outline of the analysis, but updates are expected to add tumor type based analysis, as well as specific analysis of driver genes.

#### Issue

#13 

#### Directions for reviewers

* Overall code organization? I could make things more modular, but wanted to get a first pass analysis out.
* Do the statistics seem correct and reasonable? 
* Should the co-occurence calculations be updated to include q-values in addition to raw p-values?

* For future work: 
  * Which gene sets should be evaluated?
  * Which types of mutations? Should this be run with only coding mutations?
  * Which Tumor groups?

#### Results

Table of co-occurence results and resulting figure that plots relationships between genes based on log10(p-values), with positive scores indicating co-occurence of mutations and negative being mutual exclusivity (which is not apparent in this initial analysis).


analyses/interaction-plots/scripts/01-process_mutations.R
```diff
@@ -0,0 +1,209 @@
...
+    opt_str = "--out", type = "character", 
+    default = file.path("analyses", "interaction-plots", "results", "cooccurence.tsv"),
+    help = "Relative file path (from top directory of 'OpenPBTA-analysis')
+            where output table will be placed.",
+    metavar = "character"
+  ),
+  make_option(
+    opt_str = "--specimen_list", type = "character", 
+    default = file.path("analyses", "independent-samples", "results", "independent-specimens.wgs.primary.tsv"),
+    help = "Relative file path (from top directory of 'OpenPBTA-analysis')
```
jaclyn-taroni: Looks like a copy paste from the MAF option.

analyses/interaction-plots/scripts/01-process_mutations.R
```diff
@@ -0,0 +1,209 @@
...
+  ),
+  make_option(
+    opt_str = "--out", type = "character", 
+    default = file.path("analyses", "interaction-plots", "results", "cooccurence.tsv"),
+    help = "Relative file path (from top directory of 'OpenPBTA-analysis')
+            where output table will be placed.",
+    metavar = "character"
+  ),
+  make_option(
+    opt_str = "--specimen_list", type = "character",
```
jaclyn-taroni: Are you planning on leveraging this option to filter to disease types or would you split up the MAF data.frames based on something in the metadata? This is not important yet, but trying to understand where this is headed.

analyses/interaction-plots/scripts/02-plot_interactions.R
```diff
@@ -0,0 +1,82 @@
...
+# Command line example:
+#
+# Rscript analyses/interaction-plots/02-plot_interactions.R \
+#   --infile 
+
+#### Initial Set Up
+# Establish base dir
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+script_root <- file.path(root_dir, "analyses", "interaction-plots")
+
```
jaclyn-taroni: You have a `root_dir` below and `script_root` is not used anywhere else.

analyses/interaction-plots/scripts/02-plot_interactions.R
```diff
@@ -0,0 +1,82 @@
+# Create plot of co-occurence and mutual exclusivity
+#
+# JA Shapiro for ALSF - CCDL
+#
+# 2019
+#
+# Option descriptions
```
jaclyn-taroni: Documentation's incomplete here and in the command line example.

analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -0,0 +1,131 @@
...
+  # filter on vaf and consequence (for single Consequence cases)
+  maf_df <- maf_df %>% 
+    dplyr::filter(vaf >= min_vaf) %>%
+    dplyr::mutate(consequence_list = stringr::str_split(Consequence, ",")) %>% 
+    dplyr::rowwise() %>%
+    dplyr::filter(!all(unlist(consequence_list) %in% exclude_consequence))
+    
+  if (length(include_consequence) > 0){
+    maf_df <- maf_df %>% 
+      dplyr::filter(any(unlist(consequence_list) %in% exclude_consequence))
```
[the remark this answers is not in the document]
jashapiro: You think correctly. That's what I get for not testing...

analyses/interaction-plots/scripts/01-process_mutations.R
jashapiro: I hadn't yet decided on this. I think maybe generating the specimen lists in a separate script, then using them here with little change may be the way to go. Different types may also have different gene lists of interest, but I haven't plugged that in as a script level option yet. The main calculation function can handle arbitrary gene lists though, so it should be easy to add that.

analyses/interaction-plots/scripts/02-plot_interactions.R
jashapiro: Will fix here, but does the command line example need all options?

jaclyn-taroni: Normally I would say no, but in this case it's an input file and an output file.

Issue #180, opened 2019-10-26
Generate v6 files for continuous integration


jaclyn-taroni:
After the v6 release tracked in #146 happens, we will need to update the subset files we use in continuous integration to reflect that new release. This has several parts:

- [x] Tackling the TODO related to the VarDict file provided #135 is addressed in that release
  https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/c8a2399fb31e1024938c24693ed5fddc6343756f/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L155
- [x] Updating the default release in the shell script that generates the subset files
  https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/c8a2399fb31e1024938c24693ed5fddc6343756f/analyses/create-subset-files/create_subset_files.sh#L10
- [x] Running `./analyses/create-subset-files/create_subset_files.sh` to get the new files
- [x] Uploading the files in `data/testing/release-v6-2019XXXX` to S3

We should keep an eye on _possibly_ enriching for WXS samples the way we do for the poly-A experiments, but let's see if not doing that poses a problem for anything like #3 or #11 before we iterate on that.

jaclyn-taroni:
@cansavvy will tackle this when #146 lands

cansavvy:
Only thing I will need more direction on is the `Uploading the files in data/testing/release-v6-2019XXXX to S3` component. I think last time I just sent you, @jaclyn-taroni, the updated test files. Do you want me to do the same this time?

jaclyn-taroni:
Yes that would be fine.

jaclyn-taroni:
If the files from https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples and GISTIC files get added this release, this might be slightly more involved than described herein. However, since these would be new, no one will have developed against them yet. That means we can take a two step approach where we first do what is listed here and then deal with the new files in a separate PR.

jaclyn-taroni:
@cansavvy with #190 landing, this is ready to go. We do not need the further steps I described in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/180#issuecomment-547841652. So you will need to run `download-data.sh` to get the new version (v6) and then carry out the steps above.

cansavvy:
I'm on it! 🚀

jaclyn-taroni:
This has been resolved for now.

PR #192, opened 2019-10-30
Updates to create subset for v6 data release

cansavvy:
#### Purpose/implementation

These are updates having to do with the release of v6 data. This includes that the v5 VarDict file had some problems we implemented work arounds for and now those can be undone.

These were as prescribed from #180 
- [x] Tackling the TODO related to the VarDict file provided #135 is addressed in that release
- [x] Updating the default release in the shell script that generates the subset files.

#### Issue

#180 

#### Directions for reviewers

Everything look in order?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -79,10 +79,11 @@ jobs:
          ################################
          #### Add your analysis here ####
          ################################
-         
-      - run:
-          name: SNV Caller Analysis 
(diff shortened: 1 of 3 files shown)


cgreene:
Data download failed the first time, but I just restarted the job and it has now made it past that point.

jaclyn-taroni:
I was probably in the middle of putting the new stuff on S3.

PR #196, opened 2019-10-31
V6 snv pipe updates

cansavvy:
#### Purpose/implementation

Due to changes in clinical data columns.

#### Issue

A combination of #161 and the implementation of #190.

#### Directions for reviewers

Do you see any shoot in the foot opportunities with these fix/unfixes? 

#### Results

No new results yet.


jaclyn-taroni:
Oops I forgot. This step needs to get un-commented in CI. I commented it out to get things to pass as part of #192.

PR #200, opened 2019-10-31
V7 Release

yuankunzhu:
#### Purpose/implementation

data release updates for v7 :jack_o_lantern:
- changes:
  - Add ControlFreeC CNV merged TSV file. [data format](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/format/controlfreec-tsv.md)
  - Add ControlFreeC ploidy to clinical file.
  - Add sample lists for analyses requiring unique patient specimens. [issues/155](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/155)
    - independent-specimens.wgs.primary-plus.tsv    
    - independent-specimens.wgswxs.primary-plus.tsv
    - independent-specimens.wgs.primary.tsv         
    - independent-specimens.wgswxs.primary.tsv
    ​
#### Directions for reviewers
changed on release notes and download script. 
- file content updated for (added ControlFreeC ploidy)
  - `pbta-histologies.tsv`
- add file
  - `independent-specimens.wgs.primary-plus.tsv`
  - `independent-specimens.wgs.primary.tsv`
  - `independent-specimens.wgswxs.primary-plus.tsv`
  - `independent-specimens.wgswxs.primary.tsv`
  - `pbta-cnv-controlfreec.tsv.gz`

PR #202, opened 2019-10-31
Revert "V7 Release"

jharenza:
Reverts AlexsLemonade/OpenPBTA-analysis#200

reverting this, as I just realized molecular_subtype is now missing from the clinical file. will see if @yuankunzhu can update this on S3 tonight.


jharenza:
thanks, sorry about that! a part of my code removed it when I was doing a test.

Issue #16, opened 2019-07-12
Planned Analysis: Primary vs Relapse


PichaiRaman:
Compare primary samples to relapse samples to determine genomic and transcriptomic differences and identify trends. See (https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5172533/).

PichaiRaman:
I thought I may work on this. @cgreene Let me know if you prefer we wait and do other analyses first.

cgreene:
I believe that the Docker/CI stuff is in a reasonable enough place that it's possible for folks to jump into things more. Getting a set of analyses done will help us continue to improve the instructions for new contributors. I'd be happy to see what comes of this! 👍 

Can you take a look at the new issue template for proposed analyses, and maybe write brief blurbs about this? I'm hoping they will help us also clearly convey the findings in the manuscript:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/.github/ISSUE_TEMPLATE/propose-an-analysis.md

jaclyn-taroni:
@PichaiRaman are you still working on this issue? It would be great to have you flesh this out a bit more using the [template for proposed analyses](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/.github/ISSUE_TEMPLATE/propose-an-analysis.md) as a guide. This could be particularly helpful if someone else works on the issue, as it will give us more information for what you were upon filing.

jharenza:
@PichaiRaman - from our discussion in person, it sounded like you are not working on this anymore. Are you able to add additional information, per @jaclyn-taroni's request? Thank you!

jaclyn-taroni:
The last table in [this section](https://alexslemonade.github.io/OpenPBTA-analysis/analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.nb.html#disease_type_-_descriptors_pairs) of the notebook added in #213 tells us that there are few instances where we have primary tumor data and recurrence data:

| disease_types | descriptors | n |
|---------------|-------------|---|
Ependymoma | Initial CNS Tumor, Recurrence | 5|
| Low-grade glioma;astrocytoma (WHO grade I/II) | Initial CNS Tumor, Recurrence | 4 |
| Ganglioglioma | Initial CNS Tumor, Recurrence | 2 |
| Low-grade glioma;astrocytoma (WHO grade I/II) | Initial CNS Tumor, Progressive, Recurrence | 2|
[...]

PR #213, opened 2019-11-03
Additional tables for sample distribution: breakdown by tumor descriptor

### LINKED ISSUE #162: Update: sample distribution plots accounting for multiple samples from the same individual

jaclyn-taroni:
#### Purpose/implementation

To my knowledge, we don't have a tables that:
* Just list the number of assays (with the paired normal removed)
* Look at the breakdown by `tumor_descriptor` (e.g., primary vs. other) within assay type (WGS/WXS vs. RNA-seq)
* Look at the breakdown by `tumor_descriptor` by histology (useful for things like planning #16)

Here I'm adding those tables in a single notebook and using the `flextable` package to display things nicely. 

You can see the rendered notebook [here](https://jaclyn-taroni.github.io/openpbta-notebook-concept/03-tumor-descriptor-and-assay-count.nb.html).

#### Issue

Closes #162 

#### Directions for reviewers

* Please check for correctness, i.e., the code accomplishes what the prose describes.
* What do you think of the `flextable` display?
* Are there any additional tables that you think it'd be useful to have in this notebook?

#### Results

One thing that struck me -- within a histology (as tracked in `disease_type_new`), there are not that many instances of: has primary and progressive data, has primary and recurrence data. This is good to know!


jharenza:
I like this!

analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd
```diff
@@ -0,0 +1,263 @@
...
+tumor_df %>%
+  filter(tumor_descriptor == "Initial CNS Tumor") %>%
+  distinct(Kids_First_Participant_ID, disease_type_new) %>%
+  group_by(disease_type_new) %>%
+  tally() %>%
+  arrange(desc(n)) %>% 
+  regulartable() %>%
+  fontsize(size = 12, part="all")
+```
+
```
jashapiro: I have questions about some of these descriptors, and the meaning of delimiters. For example, we have: `Ganglioglioma;Low-grade glioma/astrocytoma (WHO grade I/II)`, as well as `Ganglioglioma` on its own, and `Low-grade glioma;astrocytoma (WHO grade I/II)` on its own. Is the first some kind of combination of the latter two? Is there a semantic difference between the use of `;` and `/` in these cases?  This PR is probably not the right place for this discussion...

analyses/sample-distribution-analysis/results/primary_sites_counts.tsv
```diff
@@ -1,21 +1,21 @@
 primary_site	number_of_types	max_type	second_max_type
```
jashapiro: This seems broken?

jaclyn-taroni: Yep definitely broken https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/214

analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd
```diff
@@ -0,0 +1,263 @@
...
+  arrange(desc(n)) %>% 
+  regulartable() %>%
+  fontsize(size = 12, part = "all")
+```
+
+The same holds for RNA-seq.
+
+### Paired genomic, transcriptomic assays
+
+How many participants have paired genomic and transcriptomic samples for the same `tumor_descriptor` values?
```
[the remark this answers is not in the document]
jaclyn-taroni: I can take out looking at `paired_df` in line 134. I think it's more confusing than helpful.

analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd
jharenza: Hmm, these are all derived from the way they were added to the database - seems the `\` should not be between LGG/astrocytoma, but that should be a `;`.  Ganglioglioma and LGG/astrocytoma are different and only shown together since they were both on the pathology report. If possible, through genomic analyses, we could further dial into the diagnoses, we can possibly replace with one or the other.

jaclyn-taroni:
The table that is broken is now tracked in #220, so I will go ahead and merge this.

Issue #186, opened 2019-10-30
Proposed Analysis: map from SEG file to genes (and broader segments)


jaclyn-taroni:
#### Scientific goals

We currently have copy number information in the form of SEG files. The goal of this analysis is to generate focal copy number files for consumption in the oncoprint (#6) and co-occurrence/mutual exclusivity modules (#13). In essence we want to know what genes are amplified or deleted in each sample for downstream analysis. ~This will also have an element of determining what thresholds are appropriate to make the focal copy number calls.~ See: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/186#issuecomment-548345704

The output of this analysis should be data in tabular format:

| gene_symbol | biospecimen_id | label |  seg.mean  |
|:-------------:| :---------------:|:----:|:------------:|
| SMARCB1  | BS_XXXXXXXX | Hem_Deletion | some float |
| ... | ... | ... | ... |
| ... | ... | ... | ... |
| ... | ... | ... | ... |

As @jharenza pointed out [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/182/files#r340587968) on #182, it is not known at this time if we can make homozygous/hemizygous calls.

#### Proposed methods

Here's how we think this work should be broken up into a series of steps (as discussed with @cbethell and @jashapiro in person):

1. Put together code, adapted from [this code](https://github.com/marislab/pptc-pdx-copy-number-and-SVs/blob/0117f5c5ea89f0c364178ab9423807bb1362e66e/R/focal-CN-revision.R) linked by @jharenza, that takes a SEG file as input and produces the tables outlined above. This will most likely make use of the [`GenomicFeatures`](https://bioconductor.org/packages/release/bioc/html/GenomicFeatures.html) package. ~It will use the same, somewhat arbitrarily chosen thresholds introduced in #182.~
~2. An analysis that specifically examines _SMARCB1_ deletions in ATRTs. This will help with choosing thresholds.~
3. An analysis that examines RNA-seq expression levels for genes that are called as homozygous deletions. This could potentially help with cutoffs as well.
4. Moving from single genes to broader segments of the genome. We can use tabular format above + information about how genes are ordered on chromosomes to get an idea of what broader amplifications or deletions are recurrent. Whiteboard sketch capturing the overall idea:
![gene-to-broader](https://user-images.githubusercontent.com/19534205/67873613-17db3f00-fb0a-11e9-9a5e-e192981aae00.jpg)

[...]

PR #217, opened 2019-11-03
SMARCB1 deletions in ATRT with current SEG to gene mapping

jaclyn-taroni:
#### Purpose/implementation

In #182, @jharenza suggested we look at _SMARCB1_ deletions in ATRT. This analysis is crossed out on #186 because we were no longer looking to set LRR thresholds, but I was curious and thought it was worth looking at whether these show up given the SEG file to gene symbol mapping introduced in #195.

#### Issue

#186 

#### Directions for reviewers

Rendered notebook is [here](https://jaclyn-taroni.github.io/openpbta-notebook-concept/seg-to-gene-mapping-atrt.nb.html).

#### Results

We don't pick up _SMARCB1_ deletions in ATRT and we don't capture _SMARCB1_ at all. This doesn't mean that the SEG file to gene symbol mapping is wrong, but it is worth being aware of. A possible update to this is to look at the genomic location of _SMARCB1_ in the package(s) we used for the mapping step.


jaclyn-taroni:
@jashapiro figured out that the `GenomicFeatures::genes` step was filtering out _SMARCB1_ specifically because it was also mapped to some other alternative chromosome. We added filtering to chromosomes 1:22 with https://github.com/jaclyn-taroni/OpenPBTA-analysis/commit/d0850811cc8cd312536d96c3f3d24effb9282b92. I've now run this notebook on the `annotated_cn.tsv` file that included that filtering step and these results look more consistent with what we would expect.

Here's the updated rendered notebook: https://jaclyn-taroni.github.io/openpbta-notebook-concept/seg-to-gene-mapping-atrt-chrom-filter.nb.html

cc @jharenza

jaclyn-taroni:
I forgot to note: these results still are using CNVkit with no notion of ploidy, so the contents of `annotated_cn.tsv` will still probably change but the overall picture looks better.

jharenza:
Yay!! Nice work @jashapiro and @jaclyn-taroni!. One note: we should keep X&Y in if we can, as _ATRX_ is on the X chr. I think, however, the sex chromosomes _may_ have to be handled with caution. When we ran the CN tools, we used the `germline_sex_estimate` to correctly assign sex and therefore ploidy of those chromosomes, and I haven't looked into it, but for males, this _may_ mean X&Y each show a loss of one copy as baseline and therefore. To do that, we can check the tools to assess ploidy for X/Y in males vs females.

jaclyn-taroni:
I've made some updates now that #253 is filed. You can see the rendered notebook here:  https://jaclyn-taroni.github.io/openpbta-notebook-concept/seg-to-gene-mapping-atrt-both-formats.nb.html

To my non-expert eye, this still looks good (cc @jharenza).

jharenza:
Looks good, but are we missing 4 samples from ControlFreeC?

jaclyn-taroni:
For the files I've used here, I chose to filter out `neutral` status so perhaps it could be that step?

https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a5f8a6be13db24b3f45ed7e5f6d4b8736e11ed61/analyses/focal-cn-file-preparation/01-prepare-cn-file.R#L144

jharenza:
Ahh ok. I will be interested to see the results of #128 to see if these 4 samples come back with Manta results. If not, I may be concerned that we may miss them. cc: @hongboxie and @fingerfen - something we can revisit once we get consensus calls.

analyses/focal-cn-file-preparation/seg-to-gene-mapping-atrt.Rmd
```diff
@@ -0,0 +1,64 @@
...
+
+From [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/182#discussion_r340587968) on #182:
+
+> may advise starting with ATRT's and SMARCB1 deletions
+
+Here, I'm looking at the output of `01-prepare-cn-file.R` to see if we recover that information in the segment to gene mapping.
+We're going to get the data from a pull request/branch that is currently open.
+This will get whatever the current files are for that branch.
+
+```{sh}
```
jashapiro: I don't know if `results='hide'` works in shell blocks, but if not, maybe throw a `--quiet` on these?

```suggestion
{sh, results = 'hide'}
```

jaclyn-taroni: Let's try it!

jaclyn-taroni: Doesn't look that way, will do `--quiet`

jashapiro: 😞

jaclyn-taroni:
3 samples are in CNVkit but not ControlFreeC and also ControlFreeC finds one gain which CNVkit marks as a loss (via the choices made over on #253): https://jaclyn-taroni.github.io/openpbta-notebook-concept/seg-to-gene-mapping-atrt-short-comparison.nb.html

jaclyn-taroni:
This is now well covered via the PRs for #244

PR #227, opened 2019-11-04
Co-occurence plot by disease

jashapiro:
#### Purpose/implementation

Create co-occurence plots separated by disease type

#### Issue

#13 

#### Directions for reviewers

Are these choices of disease types reasonable? 

Given the lack of significant results in many cases, how should these results be presented? (This may wait for later). 

Are the cutoffs for number and mutation status of genes to plot reasonable?

#### Results

This produces co-occurence plots foe the most common disease types in the dataset. Note that for many diseases, there are not enough commonly mutated genes to provide meaningful results. I now filter genes by the number of mutated samples (currently > 5), given that genes with only a small number of mutated samples are unlikely to have significant interactions, and when they do, they are more likely to be false positives.


analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -191,19 +282,21 @@ gene_sample_counts <- maf_filtered %>%
...
+    mutant_samples = dplyr::n(),
+    total_muts = sum(mutations),
+    muts_per_sample = mean(mutations)
+  ) %>%
+  dplyr::arrange(
+    desc(mutant_samples),
+    desc(muts_per_sample)
+  ) %>%
+  dplyr::filter(mutant_samples >= opts$min_mutated |
+    dplyr::row_number() <= 2) # keep at least 2 genes
```
jaclyn-taroni: Based on this comment, I would expect the logic to be `>= 2` but I might be missing something

analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -44,66 +49,100 @@ library(ggplot2)
...
     metavar = "character"
   ),
   make_option(
-    opt_str = "--seg", type = "character", 
+    opt_str = "--cnv",
+    type = "character",
     default = file.path("data", "pbta-cnv-cnvkit.seg.gz"),
-    help = "Relative file path (from top directory of 'OpenPBTA-analysis') 
+    help = "Relative file path (from top directory of 'OpenPBTA-analysis')
             to CNV file to be analyzed in seg format. Can be .gz compressed.",
```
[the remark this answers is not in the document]
jaclyn-taroni: This actually is probably a discussion for a different pull request.

analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -137,27 +175,28 @@ if(exists("specimen_file")){
...
+    Variant_Classification,
+    Variant_Type,
+    Reference_Allele,
+    Tumor_Seq_Allele1,
+    Tumor_Seq_Allele2,
+    Tumor_Sample_Barcode,
+    t_depth,
+    t_ref_count,
+    t_alt_count,
+    Consequence,
```
jaclyn-taroni: Does the trailing `,` not cause issues here?

analyses/interaction-plots/scripts/02-process_mutations.R
jaclyn-taroni: Or is this to make sure that there are no problems with the call to `coocurrence` ?

.circleci/config.yml
```diff
@@ -74,7 +74,7 @@ jobs:
                 
       - run:
           name: Interaction plot
-          command: ./scripts/run_in_ci.sh bash analyses/interaction-plots/01-create-interaction-plots.sh     
+          command: OPENPBTA_ALL=0 ./scripts/run_in_ci.sh bash analyses/interaction-plots/01-create-interaction-plots.sh
```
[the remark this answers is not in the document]
jashapiro: Will do.

analyses/interaction-plots/scripts/02-process_mutations.R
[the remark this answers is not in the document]
jashapiro: Yeah, right now it is SEG, but I expect it to change when I update for actually handling CNVs.

analyses/interaction-plots/scripts/02-process_mutations.R
jashapiro: Yes, this is to prevent errors with coocurrence when there aren't enough genes with the minimal # of mutants (because of few samples or subsetting)

analyses/interaction-plots/scripts/02-process_mutations.R
jashapiro: apparently not? I'll kill it anyway.

Issue #226, opened 2019-11-04
CI: Handle the case where only a few files get updated between releases


jaclyn-taroni:
Only 6 files changed between v7 and v8, no entirely new files were added, and the biospecimen identifiers should be unchanged (see: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md#release-v8-20191104). Given the way [`create-subset-files`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/3f1819d4a4018daf1eb1c551f89ff157849957e9/analyses/create-subset-files) is currently set up, we'd have to rerun every file through and select IDs. This is time-consuming and requires lots of RAM. 

We should come up with an easier-but-still-automated way to use the [list of biospecimen IDs](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/3f1819d4a4018daf1eb1c551f89ff157849957e9/analyses/create-subset-files/biospecimen_ids_for_subset.RDS) generated from an earlier release (here `release-v7-20191031`) to subset the new files that changed between releases (here `release-v8-20191104`).

jaclyn-taroni:
The RDS file with the biospecimen IDs is not up to date: https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/d210b46f9429aae64856dc52aec03040c0657b0e/analyses/create-subset-files

So this is not immediately useful, but I'm going to implement it soon anyway.

PR #230, opened 2019-11-05
WIP: use existing IDs or existing subset files for CI

jaclyn-taroni:
#### Purpose/implementation

Right now we have to rerun every file through and select IDs when generating subset files for continuous integration. This is time- and RAM-intensive. Sometimes, only a handful of files change between releases and the biospecimen IDs should be the same. Here I'm making changes to save time in the following cases:

* We have the older releases CI files available and we want to generate new subset files for only the ones that changed between releases (Method 2 in the README).
* We don't have the older releases CI files, but we can use the same list of biospecimen IDs we have around

#### Issue

#226 

#### Directions for reviewers

This is a work in progress and I haven't functionally tested this yet. I wanted to get some thoughts out there and solicit feedback on the overall idea. We very well might be able to put all of this together in a single bash script, but I'm not very good at bash and needed to take this interim step first 😄 

cc: @jashapiro @cgreene for shell scripting thoughts and @cansavvy often ends up running it


jaclyn-taroni:
Closing for now.

PR #238, opened 2019-11-06
PR 2 of 3: Mutation Signature: Functions

cansavvy:
#### Purpose/implementation

This second PR adds the functions that will be used by the main notebook of this analysis. 

#### Issue

#11 

#### Directions for reviewers

What do you think of the functions? 
Do you see any pitfalls or things that should be optimized? 

#### Results

Still no results in this PR, just set up. 

- [ ] This analysis has not yet been added to continuous integration. (This will happen in the third PR).


analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+#
+# C. Savonen for ALSF - CCDL
+# 2019
+
+# Load this library
+library(deconstructSigs)
+
+################################################################################
+sample_mut_sig_plot <- function(which_sig_list, label = "none", output_dir = getwd(),
+                                sample_ids) {
```
jashapiro: `sample_ids` is required, right? If so, keep it in the first set of arguments, and put the ones with default values later.

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+  # Given a list of `deconstructSigs::WhichSignature` output, make plots for each sample using
+  # deconstructSigs::plotSignatures
+  #
+  # Args:
+  #
+  #   which_sig_list: a list of `whichSignature` output for each sample
```
jashapiro: Personal preference here, but I kind of like named lists, rather than separate list items and an id vector.  For me, it does a better job of keeping things in sync. It also prevents possible errors if ids are repeated when they shouldn't be (which in this case would result in overwriting files...)

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+  #
+  # Returns:
+  #   A data.frame that is samples x signatures and has the number of mutations
+  #   per mb for each signature x sample combo
+
+  # Count the total number of signature mutations for each sample
+  total_muts <- apply(sigs_input, 1, sum)
+
+  # Pull out the signature weights and make into matrix
+  sig_weights <- do.call(
```
jashapiro: I *think* dplyr::bind_rows() will do the right thing here, with an `.id="Tumor_Sample_Barcode"` argument. This is assuming that which_sig_list is named with the tumor sample barcodes?

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+    "rbind.data.frame",
+    lapply(which_sig_list, function(sample_data) sample_data$weights)
+  ) %>%
+    tibble::rownames_to_column("Tumor_Sample_Barcode")
+
+  # Calculate the number of mutations contributing to each signature
+  # Here the weight is multiplied by the total number of signature mutations. 
+  sig_num_df <- apply(sig_weights[, -1], 2, function(row_weights) row_weights * total_muts) %>%
```
jashapiro: Save yourself some work! Try 

```
sig_weights  %>% dplyr::mutate_at(-Tumor_Sample_Barcode, ~ . * total_muts) %>%
``` 
(I think I got the syntax right)

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+    dplyr::inner_join(dplyr::select(
+      maf_df,
+      "Tumor_Sample_Barcode",
+      "short_histology",
+      "experimental_strategy"
+    ),
+    by = "Tumor_Sample_Barcode"
+    ) %>%
+    # Get rid of Panel samples
+    dplyr::filter(experimental_strategy != "Panel") %>%
```
jashapiro: This is fine, but as always, filter before join is nicer.

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,209 @@
...
+
+    # Reformat for plotting
+    reshape2::melt() %>%
```
jashapiro: Give the variable column a good name?

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: Ah good point. Will change this.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: If `$weights` list were the only list within each entry for each sample, I would do this, but there are some other outputs from `deconstructSigs::whichSignatures` so I have to select the weights out. I don't know if `dplyr::bind_rows` has arguments that allow me to do that?

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: AH! Right!

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: Yes! I ended up doing that but hadn't merged it yet!

analyses/mutation-signatures/util/mut_sig_functions.R
jashapiro: Probably not. :hammer:

cansavvy: womp womp.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: I just remembered why its in this order, `sig_num_df` doesn't have this information, but both data.frames have Panel samples that need to be taken out.

cansavvy:
Ready for another look, @jashapiro !

analyses/mutation-signatures/util/mut_sig_functions.R
jashapiro: And you still haven't? Oh, you did below, but you can do it here:

```
reshape2::melt(varnames = c("Tumor_Sample_Barcode ", "signature"),
               value.name="num_mutations") 
```

either way...

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,204 @@
...
+    ) %>%
+
+    # Get rid of Panel samples
+    dplyr::filter(experimental_strategy != "Panel") %>%
+    
+    # Reformat for plotting
+    reshape2::melt() %>%
+
+    # Only keep distinct
+    dplyr::distinct(Tumor_Sample_Barcode, variable, .keep_all = TRUE) %>%
```
jashapiro: When you keep distinct here, are the values the same for every pair?  I don't see why there should be duplicates, but if there are, are we losing info here?

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,204 @@
...
+  #               each sample x signature combo
+  #   label: a character string for the title of the plot to be passed to ggplot2::ggtitle
+  #
+  # Returns:
+  #   A bubble matrix plot with the number of mutations per Mb and proportion of 
+  #   tumors with a non-zero weight for all samples summarized by histology.
+  #
+  # Summarize the mut_per_mb column by histology
+  grouped_sig_num <- sig_num_df %>%
+    dplyr::arrange(signature) %>%
```
jashapiro: This is what you will presumably need to fix to make these numerically ordered, rather than alphabetical. Though you also shouldn't need to order the df until after grouping and summarizing.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: It is not.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: I like this better. Will do this!

Dockerfile
```diff
@@ -143,5 +144,14 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
 # Check to make sure the binaries are available by loading the bedr library
 RUN Rscript -e "library(bedr)"
 
+# Install for mutation signature analysis
+RUN R -e "BiocManager::install(c('BSgenome.Hsapiens.UCSC.hg19', 'BSgenome.Hsapiens.UCSC.hg38'))"
```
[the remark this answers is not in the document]
cansavvy: It is, this has to do with a faulty dependency check of `deconstructSigs` it only looks for hg19 being installed, and will not let you install it if hg19 is not installed. (hg38 doesn't suffice for it).

analyses/mutation-signatures/util/mut_sig_functions.R
```diff
@@ -0,0 +1,204 @@
...
+  sig_num_df <- do.call(
+    "rbind.data.frame",
+    lapply(which_sig_list, function(sample_data) sample_data$weights)
+  ) %>%
+    tibble::rownames_to_column("Tumor_Sample_Barcode") %>%
+
+  # Calculate the number of mutations contributing to each signature
+  # Here the weight is multiplied by the total number of signature mutations. 
+  dplyr::mutate_at(dplyr::vars(-Tumor_Sample_Barcode), ~ . * total_muts) %>%
+    as.data.frame() %>%
```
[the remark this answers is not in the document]
cansavvy: Good point. Will delete.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: We are not losing info, it has to do with the information being joined. `maf_df` has more rows per sample, so I put this up earlier so its more clear that its because of `maf_df` we are doing this for.

analyses/mutation-signatures/util/mut_sig_functions.R
cansavvy: Now that you mention it, I am unsure what this `arrange` is here for.

Dockerfile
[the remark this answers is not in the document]
jashapiro: Well that’s dumb.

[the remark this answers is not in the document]
cansavvy: Agreed. I may file an issue with them later if someone hasn't filed an issue with them already.

PR #240, opened 2019-11-06
Focal CN RNA expression 

cbethell:
#### Purpose/implementation

This PR addresses the below comment found on issue #186 :
> Look at RNA-seq expression for loss calls (use RSEM data) - we probably don't want to rely on this to make definitive total loss calls (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/216/files#r341875649) but let's see what this looks like, generally speaking

#### Issue

This PR addresses a portion of issue #186.

#### Directions for reviewers

1. Do you have any recommendations for a refactoring method that may decrease the time it takes to run this script? 
2. How do you feel this data would be better visualized? 
3. Are there genes that should be filtered out (in this case for the purpose of better visualization)?

Note: I am thinking it would be more useful to convert this into a notebook once some refactoring is done. 

#### Results

This PR produces two plots depicting the RNA expression values (separated by selection strategy: polyA and stranded) distributed across genes that have been labeled with having focal CN losses: 
- `analyses/focal-cn-file-preparation/plots/cn_loss_expression_polyA.png`
- `analyses/focal-cn-file-preparation/plots/cn_loss_expression_stranded.png`

The plots can be seen in the `display-plots.md` file [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/eda5ea6d4c61c319954831640706970781fd7ce5/analyses/focal-cn-file-preparation/display-plots.md).

#### PR Checklist
* [x] Run a linter
* [x] Set the seed- Not needed.
* [x] Double check your paths


analyses/focal-cn-file-preparation/display-plots.md
```diff
@@ -0,0 +1,7 @@
+This markdown displays the plots produced in the `rna-expression-validation.R` script of this module.
```
jashapiro: I really like this little display markdown file. I'm going to have to use this idea in my analyses!

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,166 @@
...
+  dplyr::select(
+    gene_symbol,
+    biospecimen_id,
+    label,
+    copy_number,
+    Kids_First_Participant_ID,
+    tumor_descriptor,
+    tumor_ploidy
+  ) %>%
+  dplyr::filter(label %in% c("Hom_Deletion", "Hem_Deletion"))
```
jashapiro: Move this to the first step (line 109). In general, you always want to filter before joining, as join gets more expensive the more rows it has to process.

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,166 @@
...
+    #   copy_number_df: focal copy number data.frame annotated with the metadata
+    #   expression_df: RNA-seq expression data.frame
+    #   metadata: the relevant metadata data.frame
+    #
+    # Returns:
+    #   combined_df: data.frame with information from the focal CN, the
+    #                RNA-seq expression, and metadata data.frames
+    
+    # Change expression data.frame to long tidy format
+    long_expression <- expression_df %>%
```
jashapiro: I'd be curious if switching the order of `melt` and `inner_join` here improved speed. Try joining the metadata first, then melting to long format. That would reduce the number of comparisons needed for the join. I'm not sure it will make a huge difference, as the big expense is presumably joining the two long data frames.

cbethell: I attempted this, but the code was a bit messier this way and it did not significantly improve speed so I left this section in the `melt` then `inner_join` order. I did however address all of your other comments above.

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,175 @@
...
+          "gene_symbol" = "gene_id",
+          "tumor_descriptor"
+        )
+      ) %>%
+      dplyr::rename(biospecimen_id_cn = biospecimen_id.x, 
+                    biospecimen_id_expression = biospecimen_id.y) %>%
+      dplyr::mutate(log_expression_value = (log2(expression_value) + 1)) %>%
+      dplyr::mutate(expression_class = dplyr::case_when(
+        expression_value == 0 ~ "0",
+        expression_value > 0 & expression_value <= 10 ~ "0-10",
```
jashapiro: How did you choose these values? These feel pretty arbitrary, and it is hard to say what the correct comparative values should be without knowing what the overall distribution is. I think having a few more breaks would not be bad either. Perhaps we want to do these plots with z-scores as well?

Really, we would like to compare this to non-deleted samples, but that might be slow to compute. But if we were to attempt this, it might be nice to see a scatter plot of genes that have deletions, comparing mean expression for samples with deletions to those without.

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,175 @@
...
+        expression_value > 10 & expression_value <= 100 ~ "10-100",
+        expression_value > 100 ~ ">100"))
+
+    # Save results
+    readr::write_tsv(combined_df, file.path(results_dir, filename))
+
+    # Subset data.frame for plotting
+    combined_subset_df <- combined_df %>%
+      dplyr::filter(!(expression_class == ">100"))
```
jashapiro: Why are you filtering these off? We do want to know about high expression samples that are supposed to be deleted, right?

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,175 @@
...
+
+png(
+  file.path(plots_dir, "cn_loss_expression_polyA.png"),
+  width = 1500,
+  height = 900,
+  res = 120
+)
+ggplot2::ggplot(rsem_combined_polyA_df,
+                ggplot2::aes(x = label, fill = expression_class)) +
+  ggplot2::geom_bar()
```
jashapiro: Since the counts between these are so different, we probably want to use proportions on the y axis. Also this will make them order from bottom to top, which I think makes logical sense.
```suggestion
  ggplot2::geom_bar(position =  position_fill(reverse = TRUE)) +
  ylab("Proportion of called genes")
```

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,175 @@
...
+  merge_expression(cn_df_metadata,
+                   rsem_expression_stranded,
+                   metadata,
+                   "annotated_cn_stranded.tsv.gz")
+
+#### Plot and Save -------------------------------------------------------------
+
+png(
+  file.path(plots_dir, "cn_loss_expression_polyA.png"),
+  width = 1500,
```
jashapiro: For these stacked bar plots, lets make smaller plots (and really, you could do pdfs for optimal scalability since there are not many elements)

analyses/focal-cn-file-preparation/rna-expression-validation.R
cbethell: You are correct, this is fixed in the latest commit.

analyses/focal-cn-file-preparation/rna-expression-validation.R
cbethell: I was able to change the plots to pdfs but this affected their appearance in the `display-plots.md` file. They show up in this file locally but not on github. Should I perhaps convert the script into a notebook for this purpose?

jashapiro: I guess then leave them as pngs for now. Easy fix for publication if needed.

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,189 @@
...
+  dplyr::select(
+    gene_symbol,
+    biospecimen_id,
+    status,
+    copy_number,
+    tumor_ploidy,
+    Kids_First_Participant_ID,
+    tumor_descriptor
+  )
+
```
jashapiro: Here: calculate log2 expression for each data set & z scores. This does not depend on cn data. It may be more efficient to calcuate these while expression is still in matrix format, but you will probably want long format before merging with CN. Something like: (with other metadata as well, as needed)

| specimen_id | Gene_id | expression | log2_exp | z_score |
|-------------|---------|------------|----------|---------|
|             |         |            |          |         |
|             |         |            |          |         |
|             |         |            |          |         |

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,189 @@
...
+cn_df_non_loss_metadata <- cn_df %>%
```
jashapiro: We actually don't want this as non-loss. What we really want is all the genes/specimen combos that are not in in the CN data *at all*. So that is going to have to be a different, anti-merge function later.

analyses/focal-cn-file-preparation/rna-expression-validation.R
```diff
@@ -0,0 +1,189 @@
...
+    "util",
+    "rna-expression-functions.R"
+  )
+)
+
+#### Join data -----------------------------------------------------------------
+
+# Add metadata to focal CN data.frame
+cn_df_loss_metadata <- cn_df %>%
+  dplyr::filter(status %in% c("loss")) %>%
```
jashapiro: Don't filter to loss here, just include all the CN data.

cbethell:
I am going to close this PR and open a new PR once @jashapiro's most recent comments are addressed (this PR is getting a little messy and once changes are made the comments on specific parts of the code won't make as much sense as they do now).

Issue #201, opened 2019-10-31
Possible refactor: `snv-callers`


jaclyn-taroni:
We should take a moment to re-evaluate the overall structure of the [`snv-callers`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers) module to see if there are improvements that can be made and what the scope for those would be. We can take a look early next week.

cansavvy:
**Things I'm noticing so far:**
- Some of the plots should be restructured for accommodating the full dataset. 
   - Turn VAF correlation plot into a hexplot
   - Save histology legend separately
-  VAF of caller combinations doesn't save correctly to its png
- Some output messages are printing out the wrong file.path though they are saving to the correct file path. 

**Optimization question*
- Would be nice to have it print out an overall numbers report somewhere. As opposed to digging through the progress messages. 
- How do we avoid saving and re-reading the really big VAF files? i.e. VarDict and the master VAF file. Maybe make option where 03-merge_caller can source 04 directly and skip the save/load steps.

cansavvy:
### Changes to make to SNV caller pipeline

**Plot Aesthetic Changes:** 
- Some of the plots should be restructured for accommodating the full dataset. 
   - Turn VAF correlation plot into a hexplot
   - Save histology legend separately for `compare_snv_callers.Rmd` plot. 
   -  VAF of caller combinations doesn't save correctly to its png

**Serious flaw**
- Get rid of `opt$output` 2x normalization flaw. 

**Memory saving changes: **
01-calculate_tmb_vaf.R and 03-merge_callers are the two biggest memory hogs look for opportunities here. 
- Make option where 03-merge_caller can source 04 directly and skip the save/load steps.  
- Change merging and joining steps to use `dbplyr` so its using SQL and uses less memory. 
- As opposed to making mutation_id it's own column, can use a join function that functionally matches callers by the columns that make up `mutation_id.`
- Reduce amount of intermediary files that have to be written and read in again (this is a big use of memory).

cansavvy:
@jashapiro Let me know if you have further ideas for reducing the memory usage of this pipeline.

jashapiro:
As you note, a big memory hog is going to be reading and writing large tables. One way to handle this may be to use the [arkdb package](https://github.com/ropensci/arkdb) (Note that despite instructions, it is available on CRAN). 

[...]

PR #242, opened 2019-11-07
Draft PR for SNV refactoring

cansavvy:
#### Purpose/implementation

Trying to reduce the memory burden by using `dbplyr::`. I don't know that thus far I'm succeeding so I need other eyes to tell me how to design the structure of this better. 

#### Issue

#201 

#### Directions for reviewers

I am not happy with the current structure/flow of this pipeline. How do I improve it and its memory usage? As of now this works, but it isn't faster, its slower. 

I would advise skipping the other changes for now, and looking at the `wrangling_functions.R` file first to get the idea. 

How should `arkdb` be used to reduce the memory burden? 

Can you give me an idea of what a good structure would be for these functions in their use of dbplyr and arkdb? (Perhaps in sketch or whatever format you prefer?). 

#### Results

No new results yet. 

I haven't updated these thing since this is still at the testing stage.


cansavvy:
Going to close this, since through my discussions with @jashapiro , it sounds like we are going to try a different method than `dbplyr`.

PR #258, opened 2019-11-11
Install Pandas and other data science python libs

jashapiro:

Add pandas and other related packages to the docker.

None yet, but needed for later.

#### What was your approach?

pip install with explicit versions for all installed and dependency packages.

N/A

Does the install method used seem reasonable?


Dockerfile
```diff
@@ -157,5 +157,15 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
...
+RUN pip3 install "numpy==1.17.3" && \
+   pip3 install "six==1.13.0" "setuptools==41.6.0" && \
+   pip3 install "cycler==0.10.0" "kiwisolver==1.1.0" "pyparsing==2.4.5" "python-dateutil==2.8.1" "pytz==2019.3" && \
+   pip3 install "matplotlib==3.0.3" && \
+   pip3 install "scipy==1.3.2" && \
+   pip3 install "pandas==0.25.3"
+
 #### Please install your dependencies here
-#### Add a comment to indicate what analysis it is required for
+#### Add a comment to indicate what analysis it is required for
```
cgreene:
```suggestion
#### Add a comment to indicate what analysis it is required for

```

Dockerfile
```diff
@@ -157,5 +157,15 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
...
+# Install python3 data science basics (pandas)
+# using pip to get more current versions
+RUN apt-get update -qq && apt-get -y --no-install-recommends install python3-pip 
+RUN pip3 install "numpy==1.17.3" && \
+   pip3 install "six==1.13.0" "setuptools==41.6.0" && \
+   pip3 install "cycler==0.10.0" "kiwisolver==1.1.0" "pyparsing==2.4.5" "python-dateutil==2.8.1" "pytz==2019.3" && \
+   pip3 install "matplotlib==3.0.3" && \
+   pip3 install "scipy==1.3.2" && \
+   pip3 install "pandas==0.25.3"
+
```
cgreene: Did you want to set the default python to python3 or should we remind folks to use python3 in their shell scripts/CI sections if they want to run with it?

cgreene: Actually - now that I'm typing this out I think just reminding folks to use the `python3` command is the way to go.

jashapiro: I decided not to change it; I was worried about breaking other things in strange ways.

PR #261, opened 2019-11-12
Create and sqlite database from MAF files to facilitate merging

jashapiro:

Speed up/require less memory for making consensus files and other analyses that would otherwise require repeated parsing of large MAF files.

#### What was your approach?

Create an sqlite database with separate tables for each SNV algorithm (and one for sample/histology metadata).

For two (strelka and lancet) these are full data tables. Strelka because that is the default, and lancet as a backup because it is smallest. (If we select consusenus as all calls that are in 3 of 4 tables, this will include at least one of the two.)

For other MAF files, only a subset of data are included, to keep the database size down.

Each table also has an added `VAF` column calculated as `t_alt_count/(t_ref_count  + t_alt_count)`

The tables are indexed for quicker search and filtering by SNV (defined by the combination of `Chromosome`, `Start_Position`, `Reference_Allele`, `Allele`) and by `Tumor_Sample_Barcode`.

#201 

Does the database as made provide a good balance of size/utility. Are there other columns that should be added to the `needed_cols` list?

Documentation is still in progress, but is being added. I wanted to get this pushed for early review.

N/A

An sqlite database, stored in `scratch` by default.


analyses/snv-callers/run_db_setup.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -e
```
cansavvy: I'm going to presume I can eventually merge this with `run_caller_analysis.sh` so only one bash script needs to be run?

analyses/snv-callers/scripts/01-setup_db.py
```diff
@@ -0,0 +1,282 @@
...
+# Create caller list
+callers = [
+    ('strelka', args.strelka_file),
+    ('mutect', args.mutect_file),
+    ('lancet', args.lancet_file),
+    ('vardict', args.vardict_file)
+]
+
+for table_name, maf_file in callers:
+    if not maf_file:
```
cansavvy: I don't understand this part? Can you explain it in the context of this loop?

analyses/snv-callers/scripts/01-setup_db.py
```diff
@@ -0,0 +1,282 @@
...
+    'TEXT': 'object',
+    'INTEGER': 'Int64',
+    'REAL': 'float64'
+}
+maf_dtypes = {col: dtypes[sql_type] for col, sql_type in maf_types}
+
+# connect to/create the db
+con = sqlite3.connect(args.db_file)
+
+chunksize = 1e5
```
cansavvy: Is there any reason this should be a parse option?

analyses/snv-callers/scripts/01-setup_db.py
cansavvy: To be more specific, I don't understand lines 244- 249? Why does it look like lancet and strelka being treated differently than the other two?

analyses/snv-callers/run_db_setup.sh
jashapiro: That is the idea.

analyses/snv-callers/scripts/01-setup_db.py
jashapiro: There's a comment there! I want to have some full tables, but some with only the `needed_cols`. Since we will later plan to pull from strelka, that one is complete. But as a backup, I picked lancet.

jashapiro: The specific line there (`if not maf_file`) skips the rest of the loop if there was no specified maf file for that caller.

analyses/snv-callers/scripts/01-setup_db.py
jashapiro: Arguably yes, but this seemed a good default, and it makes almost no speed difference I could discern for reasonable values.

PR #267, opened 2019-11-15
Updates: run merged script

kgaonkar6:

Run fusion annotation/standardization in the run script (updated with changes in 02,01 scripts )

fusion annotation/standardization/filtering by running 01 (standardization) +02(general filtering )+03 (GTEX zscore annotation) for file list of general QC-ed fusion calls that need to go through project specific filtering as step 04.

#### What was your approach?
01. standardize starfusion and arriba format (update here strand information is not kept in the breakpoints) 
02. filter artifact from read throughs(update here is that fusions called by mitellman database are kept) , if fusion is found in normal databases, if expression is less than 1 FPKM in expression matrix
fusion annotation in terms of genelist and fusionlist and 
03. GTEx zscore annotation per gene and columns to specify if differentially expressed in samples for fusions filtered in 02 and 

Run script was previously reviewed by @jaclyn-taroni updated the script to run merged standardized files to 02 script.

Do we still want to run polya and stranded expression as different runs?

No, needs project specific filtering

RDS files for polya and standed general QC-ed fusion calls that will be inputs for 04 project specific filtering

NA

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,20 +39,8 @@ jobs:
           command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render(input = 'analyses/collapse-rnaseq/02-analyze-drops.Rmd', params = list(polya.annot.table = 'pbta-gene-expression-rsem-fpkm-collapsed-table.polya.rds', stranded.annot.table = 'pbta-gene-expression-rsem-fpkm-collapsed-table.stranded.rds'), clean = TRUE)"
 
       - run:
-          name: Fusion standardization - Arriba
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
> Do we still want to run polya and stranded expression as different runs?

The filtering step doesn't seem to be particularly intensive. (I'm looking at run time CI for that information, though.) What is the plus side of running them together?

kgaonkar6:
> > Do we still want to run polya and stranded expression as different runs?
> 
> The filtering step doesn't seem to be particularly intensive. (I'm looking at run time CI for that information, though.) What is the plus side of running them together?

I was just wondering since there isn't any difference in the expressionFilter parameter used specifically for filtering polya or stranded .. yeah but you are right since it doesn't take too long we can keep it as is.

PR #268, opened 2019-11-15
SNV Refactor with sqlite database (PR 3 of n): Comparison Plots Notebook

cansavvy:

Adding the comparison plots notebook back to the refactored pipeline. 

The plots are essentially the same as before: https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers_plots.nb.html

Not a new scientific question, but a refactoring of the methods for #161 

#### What was your approach?

Building off of @jashapiro 's database creation script (PR #261), I'm now taking that database and converting the rest of my scripts to use that as is described in [201 comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/201#issuecomment-552600590 )

#201 

- Do you have suggestions on DRYing this up? This feel rather long and complicated to follow.
- I tried to find a way to make the `GGally::ggpairs` plot into a hex plot, there is no straightforward way I found, but let me know if you have suggestions. 

### Results

The results look similar to before: 
https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers_plots.nb.html

Looks the same so far, but this is still the subset data and not the full datasets. 

## Pull Request Check List:
* [x] Run a linter
* [x] Set the seed (if applicable) - Unneeded. 
* [x] Double check your paths
* [x] Spell check any Rmd file or md file - No new .md files. 
* [x] Restart R and run all notebooks fresh and save - NA


jashapiro:
Not a review yet, but after some discussion/looking last night, and in light of what I think is going on with #269, can you also add in a version of (some of) these plots where you filter to coding/translated regions? You can do this on the `Variant_Classification` column, removing `IGR`, `5'Flank`, `3'Flank`, and `Intron`.  This will make it so we are potentially penalizing lancet less for not looking at the whole genome.

cansavvy:
Sounds good!

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -72,69 +63,148 @@ if (!("GGally" %in% installed.packages())) {
...
+              "Reference_Allele",
+              "Allele",
+              "Tumor_Sample_Barcode")
+```
+
+Set up tables from the database but only call the columns we need.
+
+```{r}
+strelka <- dplyr::tbl(con, "strelka") %>% 
+  dplyr::select(join_cols, "VAF")
```
jashapiro: Per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/268#issuecomment-554415946, add 'Variant_Classification' here.

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -72,69 +63,148 @@ if (!("GGally" %in% installed.packages())) {
...
-                                    "all_callers_tmb.rds"))
+```{r}
+# Set up the data as a data.frame
+vaf_df <- all_caller %>% 
+  dplyr::select("VAF_lancet", "VAF_strelka", "VAF_vardict", "VAF_mutect") %>%
+  as.data.frame() %>%
+  dplyr::mutate(VAF_consensus = VAF_strelka)     
+ 
+# Which should be part of the consensus
+consensus_index <- which(apply(!is.na(vaf_df[, -3]), 1, sum) < 3)
```
jashapiro: This may be a bit clearer:
```suggestion
consensus_index <- vaf_df %>%
  dplyr::select(-c(VAF_vardict, VAF_consensus)) %>%
  rowSums(is.na(.)) > 0
```

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -72,69 +63,148 @@ if (!("GGally" %in% installed.packages())) {
...
-# Callers per mutation data.frame
-callers_per_mutation <- readr::read_rds(file.path(consensus_results_dir, 
-                                                  "callers_per_mutation.rds"))
+```{r}
+# Take out the consensus VAF for this combination
+long_caller_df <- long_vaf_df %>% 
+  dplyr::filter(caller != "consensus")
+
+# Determine the combination of callers for each mutation
+callers_per_mutation <- tapply(
```
jashapiro: Just out of curiousity, is `tapply` faster than `group_by %>% summarize` here? I feel like we had this discussion earlier, but since you are doing the tapply twice, and you could do both callers and median in the same `summarize` call and not have to worry about conversions, etc, this would DRY it up a bit.

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -157,27 +227,49 @@ callers_per_mutation %>%
...
-Set up the summarized form of the base change information. 
+Summarize base change information per caller. 
 
 ```{r}
 # Summarize by the number of times each base change shows up in each category.
-perc_change_df <- vaf_df %>%
+perc_change_df <- all_caller %>% 
+  dplyr::mutate(base_change = paste0(Allele, ">", Reference_Allele)) %>%
+  # Case when only works for df not sql, so we will convert now
+  as.data.frame() %>%
```
jashapiro: If we are pulling all_caller into memory ever, then we should probably do it up at the top (the first time we create it), and use the in-memory functions  for subsequent steps. If we select only the columns we will need for this analysis, it might not even be too big.

The key is we don't want to have to execute the big joins more than once, if we can avoid it.

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -157,27 +227,49 @@ callers_per_mutation %>%
...
+  dplyr::mutate(
+  # From the base_change variable, summarize insertions, deletions, and
+  # changes that are more than one base into their own groups.
+  change = dplyr::case_when(
+    grepl("^-", base_change) ~ "ins",
+    grepl("-$", base_change) ~ "del",
+    nchar(base_change) > 3 ~ "long_change",
+    TRUE ~ base_change
+  )) %>% 
+  dplyr::select(-Allele, -Reference_Allele, -Chromosome, -base_change,
```
jashapiro: Drop most of these columns before pulling from the database (before `as.data.frame()`) to save memory (if we are still pulling from the db here, and not much earlier).

analyses/snv-callers/compare_snv_callers_plots.Rmd
cansavvy: I kept writing this, being like there is definitely a better way to do this but I don't know why I couldn't figure it out. This is good. Thanks.

analyses/snv-callers/compare_snv_callers_plots.Rmd
cansavvy: It is much faster which is why I have it done this way. This is something we talked about when I originally made these analyses.

cansavvy: But I agree, this is ugly.

analyses/snv-callers/compare_snv_callers_plots.Rmd
cansavvy: Ah. Good point. Will change.

analyses/snv-callers/compare_snv_callers_plots.Rmd
jashapiro: You might be able to speed it up a bit by making one call to `tapply` with a custom function that calculates both `med_vaf` and assembles the `caller` names. Not sure it is worth it, though.

cansavvy: Oh. True. I will see how clunk/not clunky that is. Right now I feel like this whole notebook is pretty clunky, so I'm trying to cut down on the clunk.

jashapiro: 🤣

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -236,21 +238,21 @@ Summarize base change information per caller.
...
+    # From the base_change variable, summarize insertions, deletions, and
+    # changes that are more than one base into their own groups.
+    change = dplyr::case_when(
+      grepl("^-", base_change) ~ "ins",
+      grepl("-$", base_change) ~ "del",
+      nchar(base_change) > 3 ~ "long_change",
+      TRUE ~ base_change
+    )
+  ) %>%
+  # Widdle down to necessary columns
```
jashapiro: I'll allow it.

analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -327,6 +328,42 @@ GGally::ggpairs(vaf_df %>% dplyr::select(-VAF_consensus),
...
+
+Make a barplot illustrating the percent of the mutations for each caller that 
+that are each type of change. 
+
+```{r}
+perc_var_df %>%
+  ggplot2::ggplot(ggplot2::aes(x = Variant_Classification, y = percent)) +
+  ggplot2::theme_classic() +
+  ggplot2::geom_bar(
+    position = "dodge", stat = "identity",
```
jashapiro: Trying to decide if these will look better as stacked plots. Your call

cansavvy: How would they be stacked?

jashapiro: You should just be able to drop the `position="dodge"`

cansavvy: Ah, so same variables just stacked.

Issue #84, opened 2019-08-28
Proposed Analysis: RNA Expression-based Prediction of Sex


cgreene:
#### Scientific goals

In https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/73 it's noted that the reported sex of the participants in the study didn't align with information from germline sequencing in 11 cases. It could be interesting to understand how accurately this can be called from gene expression data. For some studies, gene expression data is all that is available. If it can also be accurately called directly from gene expression data, these studies will be better positioned to evaluate their metadata.

#### Proposed methods

* Construct an elastic-net logistic regression classifier using gene expression values as features and reported sex as the labels. I suggest elastic net because the signal is expected to be relatively linear, most genes are expected to play little role, but a few sets of genes are likely to be predictive and highly correlated, and it makes sense to spread weights across them to produce a robust predictor.
* Evaluate using cross-validation with reported labels across the full set.
* Evaluate using cross-validation with germline-based labels across the full set
* Evaluation prediction accuracy using germline sequencing within each histology.

#### Required input data

For classifier construction:
* The reported sex in the PBTA histologies file.
* Gene expression estimates from kallisto, RSEM, or both.

For evaluation:
* The germline-based sequencing calls.
* Histologies, so that performance can be broken out by histology as well.

#### Proposed timeline

I am proposing this analysis but don't have time to do it, so I would leave this as an estimate for someone who decides to take this on.

#### Relevant literature

There are a number of reports of this being readily discoverable, even with unsupervised methods. These are two from our group where we noticed this, but there are also others from other groups:
* Unsupervised VAE -> feature aligned with sex: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5728678/
* Multiple unsupervised methods -> feature aligned with sex: https://www.biorxiv.org/content/10.1101/573782v1

bill-amadio:
I would like to give this a try.  pbta-gene-expression-kallisto.rds looks like a good first source of gene expression estimates (1028 samples of 200,000+ transcript abundance values).  pbta-histologies.tsv has an entry for each of these samples.  Model construction will be with glmnet and glmnetUtils packages which allow tuning, through cross validation, of the regularization weight (lambda) and elastic net weight (alpha) simultaneously.

jaclyn-taroni:
Welcome @bill-amadio - thanks for getting this analysis started! We can use this issue to discuss strategy and address any questions around the analysis you have.

[...]

PR #276, opened 2019-11-15
03 evaluate model

bill-amadio:

This is the third of a three-script sequence to develop and evaluate a predictive model for sex prediction from RNASeq expression data.

How to evaluate the predictive model described above.

#### What was your approach?

The model is elastic net logistic regression, built in script 02-train_elasticnet using expression data of the user's choice and the reported gender column of the OpenPBTA histologies file.  Calls for each sample correspond to the Male/Female value that has a model probability greater than 50%.

Evaluation of call accuracy was done using the caret::confusionMatrix and caret::twoClassSummary functions.  The output is displayed and saved to the user specified output directory.

A third file, containing observed and predicted values along with the model probabilities for each sample is also saved.  This can be used to experiment with call thresholds other than 50% and examine erroneous calls for model tuning.

84: RNA Expression Based Prediction of Sex.

Are other kinds of evaluation useful?  All output at this stage is tabular.  Would one or more plots be useful?

caret also provides lift calculations.  I've seen lift used in marketing analyses to evaluate a model's ability to identify potential buyers vs. non-buyers.  This might be useful if someone were interested in the most likely samples from just one sex.

Any chance the sequence of an object got corrupted?

Predictive accuracy of reported_gender from the top 25% median absolute deviation transcripts is 98.5%.  Even using 0.5% of the transcripts yields 80% accuracy.

### Results

Confusion Matrix and Statistics
kallisto_stranded expression data, reported_gender as target
70/30 train-test partition; top 25% transcripts by MAD

                   Reference
Prediction Female Male
    Female    112    4
    Male        0  142

               Accuracy : 0.9845
                 95% CI : (0.9608, 0.9958)
    No Information Rate : 0.5659
    P-Value [Acc > NIR] : <2e-16

                  Kappa : 0.9686

 Mcnemar's Test P-Value : 0.1336

            Sensitivity : 1.0000
            Specificity : 0.9726
         Pos Pred Value : 0.9655
         Neg Pred Value : 1.0000
             Prevalence : 0.4341
         Detection Rate : 0.4341
   Detection Prevalence : 0.4496
      Balanced Accuracy : 0.9863

       'Positive' Class : Female

      ROC      Sens      Spec
0.9984100 1.0000000 0.9726027

caret::confusionMatrix and twoClassSummary tables.

textbook prediction accuracy given the number of transcripts in the expression data set.


jaclyn-taroni:
Hi @bill-amadio - I am going to fix the CI file and add this script to CI

bill-amadio:
Is that something I should be doing?

jaclyn-taroni:
Sure, I can walk you through it! What is causing the error is `HEAD` in line 54 of the `.circleci/config.yml` file: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/276/files#diff-1d37e48f9ceff6d8030570cd36286a61R54

And to add this script to CI you will need to add a run step that runs this script after the cleaning and model training steps that are at line 63 and line 67, respectively. Here are the instructions we have in the README: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/README.md#adding-analyses-to-ci Let me know if you have questions!

jaclyn-taroni:
Hi @bill-amadio, thank you for adding this to CI. The continuous integration is currently failing with:

```
Error in library(caret) : there is no package called ‘caret’
Execution halted
Exited with code 1
```

You can see that [here](https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/1346) or by clicking `Details` by `ci/circleci: run_analyses — Your tests failed on CircleCI`.

That tells us that the `caret` package is missing from the project Docker image. We add new libraries to the Docker image by altering the [`Dockerfile`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/Dockerfile) that is in the root of the repository/project (see [this section](https://github.com/AlexsLemonade/OpenPBTA-analysis#docker-image) of the README as well). 

Because we use [Rocker Project images](https://www.rocker-project.org/) as our base, we [can use the `install2.r`](https://www.rocker-project.org/use/extending/) script to install packages from CRAN. We can do this by adding `caret` to lines 74-79 of the `Dockerfile` to become:

```
# Required for sex prediction from RNA-seq data
RUN apt-get update -qq && apt-get -y --no-install-recommends install \
    && install2.r --error \
    --deps TRUE \
    glmnet \
    glmnetUtils \
    caret
```

If you're familiar with installing things on Linux, this will look familiar. The `RUN` statement is Docker-specific syntax.

This branch `03-evaluate_model` is currently behind the AlexsLemonade `master` branch, so I would recommend doing the following:

1. Press the `Update Branch` button on this pull request.
2. Use `git pull` to update the local copy of the `03-evaluate_model` branch.
3. Add the `caret` to the Docker image via the method I laid out above to the `Dockerfile` on the `03-evaluate_model` branch and commit those changes.
4. Push your changes and they will show up here on this pull request.

If `caret` installs properly, it will be on the Docker container that's being used in continuous integration and we should no longer see that error.

Let me know if you have any questions about adding things to Docker.

analyses/sex-prediction-from-RNASeq/03-evaluate_model.R
```diff
@@ -0,0 +1,163 @@
...
+model_transcripts <- readRDS(transcript_index_file_name)
+
+#--------
+
+#-------- predict from model and evaluate sex predictions with caret::confusionMatrix
+
+predict_class <- predict(best_fit, newx = test_expression[, model_transcripts], type = "class", s = best_fit$lambda.1se)
+predict_prob <- predict(best_fit, newx = test_expression[, model_transcripts], type = "response", s = best_fit$lambda.1se)
+
+cm_set <- data.frame(obs = test_targets[, "reported_gender"])
```
jaclyn-taroni: We are going to want to evaluate performance in predicting the labels `germline_sex_estimate` at some point. There are a few ways we could approach this -- we could accept the target column name for what we want to use as the reference for the confusion matrix as a command line argument. So if you wanted to evaluate performance on both `reported_gender` and `germline_sex_estimate`, you would run this script twice. Alternatively you could hardcode the evaluation of both in this script. If we think about this as a general framework that will be adapted/built upon for future work (e.g., predicting disease labels), I think the first way lends itself to being a bit more flexible?

analyses/sex-prediction-from-RNASeq/03-evaluate_model.R
```diff
@@ -0,0 +1,163 @@
...
+cm_set$Male = unname(predict_prob)
+cm_set$Female = 1 - unname(predict_prob)
+cm_set$pred <- factor(ifelse(cm_set$Male >= .5, "Male", "Female"))
+cm_set$sample <- test_targets[, "Kids_First_Biospecimen_ID"]
```
jaclyn-taroni: Why not use `predict_class` here?

analyses/sex-prediction-from-RNASeq/03-evaluate_model.R
bill-amadio: I agree, and I am able to make this change today (Wednesday).

analyses/sex-prediction-from-RNASeq/03-evaluate_model.R
bill-amadio: Using predict_class for cm_set$pred is cleaner.  I can make this change today.

analyses/sex-prediction-from-RNASeq/03-evaluate_model.R
bill-amadio: Don't we need to specify the target column in the 02 script so that the model will be trained properly?

jaclyn-taroni: My understanding of what is laid out on #84 is that we'll always train on `reported_gender`, but you could also change the 02 script to accept the correct column as a command line argument and set the default to `reported_gender`?

bill-amadio:
We set up the command line arguments according to the initial specs you wrote for the 01 02 03 scripts, so the organization of inputs and outputs that you describe is possible with no modifications.  We do not create any folders that do not exist.  Is that something you want to add?

I'll work on the shell script.  Should I create a new branch?

Because we save only the best fit out of glmnet, the model object files are not large.  Only about 1 MB for the model trained on the top 25% MAD transcripts.  Is setting up the tracking a git administration task, or must something be done to the code?

jaclyn-taroni:
> We set up the command line arguments according to the initial specs you wrote for the 01 02 03 scripts, so the organization of inputs and outputs that you describe is possible with no modifications. 

Okay that's what I thought, but wanted to be sure!

> We do not create any folders that do not exist. Is that something you want to add?

Yes I think this would be the modification then. Something like:

```r
dir.create(opt$output_directory, recursive = TRUE, showWarnings = FALSE)
```

or if you want to be a bit more explicit:

```r

if (!dir.exists(opt$output_directory)) {
  dir.create(opt$output_directory, recursive = TRUE)
}
```

Any time you'll be writing output.

> I'll work on the shell script. Should I create a new branch?

Yes, I'll get this branch merged in shortly after I update the branch and CI passes.

>  Because we save only the best fit out of glmnet, the model object files are not large. Only about 1 MB for the model trained on the top 25% MAD transcripts. Is setting up the tracking a git administration task, or must something be done to the code?

It is a matter of adding and committing those objects in the same way that you do your code.

Issue #10, opened 2019-07-12
Planned Analysis: Analysis of recurrent fusions


PichaiRaman:
This should probably be partitioned by cancer type and/or molecular subtype. This could be a **stacked** bar chart (one bar per fusion) for instance with colors representing different cancer types.

jharenza:
Krutika and I would like to work on this analysis. We have already created a workflow for annotating and identifying high-confidence fusions from the two fusion algorithms.

jharenza:
@kgaonkar6 - for this recurrent analysis, we should pay special attention to samples in which we have diagnosis/relapse pairs (#155) or patient/cell line matched data and should separate/plot differently. For example, we can look at cell line results separately (or exclude from this analysis). For patients with more than one RNA sample, I think we should look into how similar/different they are/if this differs by phase of therapy and think about how to describe this data.

jharenza:
@kgaonkar6 - in writing up some comments for @cansavvy for #18, I realize we could use the recurrent fusion list here. Can you possibly create a quick PR to create a **table** of recurrently-fused genes and recurrent fusions by broad cancer type? I think @cansavvy can perhaps start with that for survival analysis.

daveagbo733kweku:
Hi

jaclyn-taroni:
Closing all planned analysis tickets in favor of opening new proposed analysis/updated analysis tickets as needed.

PR #277, opened 2019-11-16
04 project specific filtering

kgaonkar6:
project specific filtering to capture recurrent fusion and oncogene annotated gene fusions

#### Putative Driver:
Filtering for general cancer specific genes

#### Scavenge back filtered fusions to add to putative oncogenic fusions:
1. In-frame/frameshift fusions is called in atleast 2 samples per histology OR
2. In-frame/frameshift fusions is called in atleast 2 callers 
AND

Filtering for recurrent fusions and oncogenic fusion in all histologies

#### What was your approach?
Mostly summarizing data to identify 
 * if annotated as oncogenic (from genereferencelist and fusionreferencelist in reference folder)
 * if inframe/frameshift found in both callers/called in more than 2 samples per histology ;these fusions should also be uniquely found in a histology and not multifused ( We call a gene multifused if we identify more than 5 different partner genes for a gene in a sample )

We want to  retain unique calls for oncoprint so I've only kept columns that I think that don't duplicate the fusion calls. 

For example each FusionName may have duplicate rows because of different so I've removed the following columns.
-Caller,-annots,-JunctionReadCount,-SpanningFragCount,-Confidence,-LeftBreakpoint,-RightBreakpoint in the final fusion calls
Can we review further if Fusion_Type also needs to be aggregated for oncoprint or any other column that needs to be removed to keep only unique.

I think #10 refers to a figure similar to 
[recurrent_Fusion_and_Genes_combined.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3853620/recurrent_Fusion_and_Genes_combined.pdf) I can add that as another script if the figure looks applicable.  

Yes this table is ready for oncoprint:
results/PutativeDriverFusion.tsv 

### Results
results/PutativeDriverFusion.tsv 

table

3328 fusions are identified as putative oncogenic fusions for PBTA


analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+limitMultiFused<-params$limitMultiFused
+
+print("Raw calls from STARfusion and Arriba for PBTA")
+table(fusion_calls$Caller)
+```
+
+
+```{r}
+
+# aggregate caller
```
jaclyn-taroni: This particular step had a problem in the earlier PR #166 (see https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/166/files#r340102734) where you used `group_by(FusionName, Sample )` instead of `group_by(FusionName, Sample ,Fusion_Type)`.

Generally, in earlier steps I think you use `group_by(FusionName, Sample)`, e.g.,

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/54a8a1ae9d91881aed0360df1ac47df8d940c139/analyses/fusion_filtering/02-fusion-filtering.R#L221

Is not including `Fusion_Type` going to be problematic in those steps?

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+  dplyr::filter(!grepl("LOCAL_REARRANGEMENT|LOCAL_INVERSION",annots))
+
+#to add aggregated caller from fusion_caller.summary
+fusion_calls<-fusion_calls %>% 
+  dplyr::filter(Fusion_Type != "other") %>% dplyr::select(-Caller,-annots) %>%
+  left_join(fusion_caller.summary,by=(c("Sample","FusionName","Fusion_Type"))) %>%
+  dplyr::select(-JunctionReadCount,-SpanningFragCount,-Confidence,-LeftBreakpoint,-RightBreakpoint) %>% unique()
+
+#merge with histology file
+fusion_calls<-merge(fusion_calls,clinical,by.x="Sample",by.y="Kids_First_Biospecimen_ID")
```
jaclyn-taroni: Are you using anything besides `broad_histology` from this file? I am wondering for a few reasons -- it's memory efficient to get rid of the columns you don't need and I noticed in `results/PutativeDriverFusion.tsv` that that contains lots of the clinical metadata. 

I can also imagine a situation where someone is working with this version of the fusion files and then doesn't need to join with the histologies file. That could be convenient, but the histologies file might be updated in a later release and this could be rendered out of date. I think you want to design the output of this keeping in mind that things are consistently updated.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+  #                 !Gene2A %in% fusion_recurrent5_per_sample$GeneSymbol |
+  #                 !Gene1B %in% fusion_recurrent5_per_sample$GeneSymbol | 
+  #                 !Gene2B %in% fusion_recurrent5_per_sample$GeneSymbol) %>%
+  dplyr::filter(!is.na(Gene1A_anno) | !is.na(Gene1B_anno) | !is.na(Gene2A_anno) | !is.na(Gene2B_anno)) %>%
+  unique()
+  
+# merge putative annotated oncogenic and scavenged back non-oncogenic annotated, recurrent fusions
+putative_driver_fusions<-rbind(QCGeneFiltered_recFusionUniq,putative_driver_annotated_fusions)  
+
+write.table(putative_driver_fusions,"PutativeDriverFusion.tsv",sep="\t",quote=FALSE,row.names = FALSE)
```
jaclyn-taroni: I would go with
```suggestion
write.table(putative_driver_fusions, file.path(outputfolder, "PutativeDriverFusion.tsv"),sep="\t",quote=FALSE,row.names = FALSE)
```
and remove `setwd(outputfolder)` at line 245

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+
+
+
+```
+
+
+```{r}
+
+
+outputfolder<-"results"
```
jaclyn-taroni: I would move this into `params` because that is where I would expect this information to be based on the rest of the notebook. I don't think you need the working directory of this chunk to be in `results`.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+setwd(outputfolder)
+# filter QCGeneFiltered_recFusion to remove fusions found in more than 1 group 
+
+QCGeneFiltered_recFusionUniq<-QCGeneFiltered_recFusion %>%
```
jaclyn-taroni: Personally I might opt for:
```r
recurrent_symbols <- fusion_recurrent5_per_sample$GeneSymbol
```

and then this block becomes
```r
QCGeneFiltered_recFusionUniq<-QCGeneFiltered_recFusion %>%
  dplyr::filter(!FusionName %in% group.count$FusionName) %>%
  dplyr::filter(!Gene1A %in% recurrent_symbols |
                  !Gene2A %in% recurrent_symbols |
                  !Gene1B %in% recurrent_symbols | 
                  !Gene2B %in% recurrent_symbols) %>%
  unique()
```
but that is just a personal style preference.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+
+
+
+
+```{r}
+setwd(outputfolder)
+
+# filter for putative driver genes and mutifused genes per sample
+putative_driver_annotated_fusions <- fusion_calls %>%
+  # dplyr::filter(!Gene1A %in% fusion_recurrent5_per_sample$GeneSymbol |
```
jaclyn-taroni: I would remove these lines that are commented out since you do this step above.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,269 @@
...
+  #                 !Gene2A %in% fusion_recurrent5_per_sample$GeneSymbol |
+  #                 !Gene1B %in% fusion_recurrent5_per_sample$GeneSymbol | 
+  #                 !Gene2B %in% fusion_recurrent5_per_sample$GeneSymbol) %>%
+  dplyr::filter(!is.na(Gene1A_anno) | !is.na(Gene1B_anno) | !is.na(Gene2A_anno) | !is.na(Gene2B_anno)) %>%
+  unique()
+  
+# merge putative annotated oncogenic and scavenged back non-oncogenic annotated, recurrent fusions
+putative_driver_fusions<-rbind(QCGeneFiltered_recFusionUniq,putative_driver_annotated_fusions)
```
jaclyn-taroni: Should the `unique` step go here instead?

analyses/fusion_filtering/04-project-specific-filtering.Rmd
jharenza: @kgaonkar6 and @jaclyn-taroni I suggest outputting only PT_ID and BS_ID as metadata with these results so it can be merged with future clinical later, as @jaclyn-taroni  suggested

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: in the previous steps in 02-fusion-filtering.R we are using the grouping to get gene [1A/1B/2A/3B] expression per Sample so the group_by(FusionName,Sample), the Fusion_type will not be a useful group to groupby there. 

But in the aggregation step here we wanted unique [inframe/frameshift] fusion calls per Sample called by 1 or N callers so I'm using the Fusion_Type here as well so that if a fusion has 1 breakpoint that is inframe and another that is frameshift these will get a unique row to itself.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: ok done!

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: yes I did add a unique here as well to remove the overlaps between QCGeneFiltered and putative_driver_annotated. Thanks for pointing that out :)

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: done

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: yeah that does look cleaner.

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: done

analyses/fusion_filtering/04-project-specific-filtering.Rmd
kgaonkar6: Im only keeping Kids_First_Biospecimen_ID, Kids_First_Participant_ID and broad_histology now. The broad_histology is used to check for n samples per group etc. Should that be ok?

jharenza: In this case, it is not necessary since we are only producing a table with the final fusions - only useful for plotting. We are just talking about the final TSV for use in oncoprints, correct? As a supplemental table, we probably would want to add that information. It is possible `broad_histology` would change if we find samples mis-diagnosed...

jaclyn-taroni: I am inclined to leave `broad_histology` out of the final table as well. The "final" TSV may change if samples have `broad_histology` labels change, but this is all set up to be easily rerun!

analyses/fusion_filtering/04-project-specific-filtering.Rmd
jaclyn-taroni: Okay, sounds good 👍

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+Putative Driver:
+Filtering for general cancer specific genes
+Fusions with genes in either onco
+
+This notebook assumes you are in OpenPBTA-analysis project folder structure.
+
+
+```{r}
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+setwd(root_dir)
```
jaclyn-taroni:
```suggestion
```

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+#load required packages
+suppressPackageStartupMessages(library("readr"))
+suppressPackageStartupMessages(library("tidyverse"))
+suppressPackageStartupMessages(library("reshape2"))
+suppressPackageStartupMessages(library("qdapRegex"))
+
+#read filtFusion files
+strandedQCGeneFiltered_filtFusion<-readRDS(params$dataStranded)
```
jaclyn-taroni:
```suggestion
strandedQCGeneFiltered_filtFusion<-readRDS(file.path(root_dir, params$dataStranded))
```

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+polyaQCGeneFiltered_filtFusion<-readRDS(params$dataPolya)
```
jaclyn-taroni:
```suggestion
polyaQCGeneFiltered_filtFusion<-readRDS(file.path(root_dir, params$dataPolya))
```

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+
+QCGeneFiltered_filtFusion<-rbind(strandedQCGeneFiltered_filtFusion,polyaQCGeneFiltered_filtFusion)
+
+write.table(QCGeneFiltered_filtFusion,"analyses/fusion_filtering/results/FilteredFusion.tsv",sep="\t",quote=FALSE,row.names = FALSE)
```
jaclyn-taroni:
```suggestion
write.table(QCGeneFiltered_filtFusion, file.path(outputfolder, "FilteredFusion.tsv"),sep="\t",quote=FALSE,row.names = FALSE)
```

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+fusion_calls<-unique(QCGeneFiltered_filtFusion)
+# remove distance from intergenic fusions
+fusion_calls$FusionName<-unlist(lapply(fusion_calls$FusionName,function(x) rm_between(x, "(", ")", extract = F)))
+
+
+# get grouping column id
+group<-params$group
+
+# get histology file
+clinical<-read.delim(params$histology,stringsAsFactors = FALSE)
```
jaclyn-taroni:
```suggestion
clinical<-read.delim(file.path(root_dir, params$histology), stringsAsFactors = FALSE)
```

analyses/fusion_filtering/04-project-specific-filtering.Rmd
```diff
@@ -0,0 +1,262 @@
...
+numSample<-params$numSample
+
+# Max number of groups 
+numGroup<-params$numGroup
+
+# Max number of times gene can be fused per sample
+limitMultiFused<-params$limitMultiFused
+
+# results folder
+outputfolder<-params$outputfolder
```
jaclyn-taroni: Move this up to before when you write `QCGeneFiltered_filtFusion` to file.

kgaonkar6: done!

kgaonkar6:
done!

Issue #275, opened 2019-11-15
Updated analysis: snv-callers


jashapiro:
#### What analysis module should be updated and why?
snv-callers: The generation of the SNV consensus

#### Why should the module be updated?

The current SNV consensus strategy will lose some multinucleotide mutations due to differences between strelka2 and other callers.  As noted in https://www.biorxiv.org/content/biorxiv/early/2019/04/30/623702.full.pdf, strelka does not make any multinucleotide (MNV) calls, instead calling as strings of SNVs, whereas other callers make `DNP`, `TNP` and `ONP` calls for di-, tri- and quad+ nucleotide variants, respectively.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

There seem to be two main options: 

1. Split all MNVs from mutect and lancet before they hit the database that the consensus list is built from in `scripts/01-setup_db.py`. This may be a bit simpler to implement at the outset, but does lose haplotype information.
2. Build the SNV consensus as we currently do in `scripts/02-merge_callers.R`, then write a second query to pull just the MNVs from callers that create them, split those up, and look for confirmation in the strelka data. This will make the consensus creation code a bit uglier, but may end up quicker overall, and can also be made to preserve haplotypes more easily.

I will be pursuing option 2 to start.

#### What input data should be used? Which data were used in the version being updated?

v10, when it arrives

#### When do you expect the revised analysis will be completed?

Early the week of 11/18

#### Who will complete the updated analysis?

@jashapiro, with assistance as needed from @cansavvy

jashapiro:
Note that this was previously also discussed in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/30#issuecomment-519263233

migbro:
Hi all,
   I was asked to post my attempt within the DRC to tackle this issue.  It's in cwl, but with the meat of the work done injected as python code, so shouldn't be too hard to suss out.
The steps I go through are first, normalize each vcf following this protocol:
https://github.com/kids-first/kf-somatic-workflow/blob/master/tools/normalize_vcf.cwl

Then I build the mnps using this tool:
https://github.com/kids-first/kf-somatic-workflow/blob/master/tools/prep_mnp_variants.cwl

Then I take those results and run the consensus caller on them:
https://github.com/kids-first/kf-somatic-workflow/blob/master/tools/bcbio_variant_recall_ensemble.cwl

These are all tied together, using this workflow script:
[...]

PR #279, opened 2019-11-18
Jashapiro/mnv consensus

jashapiro:

Recovering missing variants in the consensus SNV set that are missed due to strelka2 not making multinucleotide calls.

#### What was your approach?

Convert MNVs from mutect2 and lancet into sets of SNVs, then merge those with the strelka calls.

Currently, these are not converted back to MNVs, but are lest as consecutive SNVs. This conversion could be fixed later.

#275

A comparison to the previous consensus files would be useful to see exactly how many are recovered.

Should we add a flag to SNVs that may be part of MNVs, or leave that as an exercise for the downstream user (to look for consecutive calls).


analyses/snv-callers/scripts/02-merge_callers.R
```diff
@@ -67,6 +67,69 @@ option_list <- list(
...
+#'
+#' @param mnv_tbl a table containing MNVs (may be from an sql connection)
+#'
+#' @return a data frame of SNVs derived from the MNVs
+#'   (does not contain any of the original SNVs)
+split_mnv <- function(mnv_tbl) {
+  mnv_df <- mnv_tbl %>%
+    dplyr::filter(Variant_Type %in% c("DNP", "TNP", "ONP")) %>%
+    as.data.frame() %>%
+    # temp_id is not currently used, but will be handy for reconstituting MNVs
```
cansavvy: This comment is confusing to me, because it looks like you use `temp_id` to group by below. Can you elaborate on what you mean by "temp_id is not currently used"?

analyses/snv-callers/scripts/02-merge_callers.R
```diff
@@ -67,6 +67,69 @@ option_list <- list(
...
+    dplyr::mutate(temp_id = dplyr::row_number()) %>%
+    # lancet adds a base to the start of MNV Allele Calls
+    dplyr::mutate(
+      Allele = dplyr::case_when(
+        stringr::str_length(Allele) == stringr::str_length(Reference_Allele) ~
```
cansavvy: Can you add a comment or two here to explain the logic of this `case_when` a bit more?

analyses/snv-callers/scripts/02-merge_callers.R
```diff
@@ -67,6 +67,69 @@ option_list <- list(
...
+          Allele,
+        stringr::str_length(Allele) == stringr::str_length(Reference_Allele) + 1 ~
+          stringr::str_sub(Allele, 2)
+      )
+    )
+  # separate into individual rows
+  if ("Match_Norm_Seq_Allele1" %in% colnames(mnv_df)) {
```
cansavvy: What's the rationale of using Match_Norm_Seq_Allele if it exists?

jashapiro: I'm not using it, just preserving it if it exists. Otherwise every site would still have the multibase calls in this column. Added a more detailed comment

analyses/snv-callers/scripts/02-merge_callers.R
jashapiro: I did... its a hack because lancet is strange.

analyses/snv-callers/scripts/02-merge_callers.R
jashapiro: You are right, I use it to give position #s which I then use to calculate new start positions. So This comment is not correct.

PR #285, opened 2019-11-21
Add snakemake to Docker

fingerfen:
This is adding Snakemake into the Dockerfile

It is addressing issue #128 

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -186,5 +186,20 @@ RUN pip3 install "numpy==1.17.3" && \
 # Add curl
 RUN apt-get update && apt-get install -y --no-install-recommends curl
 
-#### Please install your dependencies here
-#### Add a comment to indicate what analysis it is required for
+
+# Add virtualenv to install Snakemake for CNV_consensus_call Pipeline
+
+RUN pip3 install virtualenv
+
+# Complete install of header files and static libraries for python-dev
+
+RUN sudo apt-get install python3-dev -y
+
+# Install Snakemake
+
+RUN git clone https://github.com/snakemake/snakemake.git && \
+   cd snakemake && \
+   virtualenv -p python3 .venv && \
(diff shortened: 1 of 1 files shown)


cgreene:
Is there a reason to setup things this way instead of just using pip3 to install snakemake as was done for the other python dependencies a few lines above this?

fingerfen:
Dear Dr. Greene,

Pip3 snakemake does not work due to a problem with Datrie. 
Then I tried to pip the newest version of Datrie but the pip failed. It seems that there is something wrong with Datrie's pip installation. 
Due to time constraints, I had to forgo the pip3 install and use this approach instead. 

I tried every single installation method suggested in the Snakemake documents and the only ones that work are conda install and what I have done in the Dockerfile. 
Conda is the most straightforward way but from past experience, I do not think it is feasible to add conda into the docker at this point. 

Thank you,

Nhat Duong

cgreene:
Interesting! I created #286 to investigate while my local Docker is updating. Thanks for noting this. I'll see what I can nail down, and if we can't get pip3 to go then we can go this route.

jashapiro:
It looks like the datrie dependency problem may have been fixed: 
> datrie 0.8 has been released on pypi, resolving this issue
https://bitbucket.org/snakemake/snakemake/issues/934/installation-failed-in-python-37

But it also appears that this problem may have been restricted to python 3.7, and the image we are using has python 3.5.3, so this bug may not be relevant to this install.

It does look like the pip3 install in #286 was successful, so that may be the way to go for installation if it can work for your application.

cgreene:
Ok! You should be able to pull upstream master @fingerfen and be good to go. You may already have it set up, but just in case you don't:
https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/syncing-a-fork

Thanks for your contributions!

PR #286, opened 2019-11-21
test snakemake + some whitespace stuff

cgreene:

No scientific question. Testing a pip3 install of snakemake. My locker docker is currently updating so trying via CI.

#### What was your approach?

Added snakemake to the pip3 section

N/A

Does it pass CI?

Not unless it works.

N/A

N/A

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -175,7 +175,7 @@ RUN R -e "BiocManager::install(c('TCGAbiolinks'), update = FALSE)"
 
 # Install python3 data science basics (pandas)
 # using pip to get more current versions
-RUN apt-get update -qq && apt-get -y --no-install-recommends install python3-pip
+RUN apt-get update -qq && apt-get -y --no-install-recommends install python3-pip  python3-dev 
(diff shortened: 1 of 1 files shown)


jashapiro:
Looks like this works, but there are many dependencies.. do we want to specify versions for all of them?

fingerfen:
I just tested my analysis on it. It worked. Seems like this is the best way to install Snakemake. Thank you for the help :) 
Should I close my pull request and ask you to pull this solution here instead?

cgreene:
Sure - I can merge after it gets reviewed. I'll request a review from @jashapiro.

@jashapiro : my thinking on that question is that at this point what we want to avoid is having things break, but that numeric reproducibility isn't yet important (though we will want to preserve the container that we use for the analyses associated with the publication that we produce). We could define the specific versions to avoid that, but there's a lot of packages here so I'm not sure if that'd be the best way to go or if it'd be more brittle in the case of adding additional packages. Let me know what you think.

PR #289, opened 2019-11-25
Minor changes to snv analysis

jashapiro:

Running `analyses/snv-callers/run_caller_consensus_analysis.sh resulted in some warnings and errors, including not actually writing out the TMB files.

#### What was your approach?

Fix misc bugs

None

The three main changes are:

1. writing TMB files
2. Not including strand in generated interval objects in `wrangle_functions.R`
3. changing OPENPBTA_PLOT variable in shell script

Of these:

1. should be non-controversial
2.  I think is correct, but I may be missing a use case where strand is required. The ones I could see were MAF files, which are all on +, and the WGS and WXS bed files don't have strand info. Let me know if I missed anything, in which case this might be best to revert to the version where strand was explicitly present or absent (as indicated in the comment I removed)
3. If there is a better way to handle this, I'm open to suggestions.

### Results
N/A


analyses/snv-callers/util/wrangle_functions.R
```diff
@@ -96,7 +95,7 @@ maf_to_granges <- function(maf_df) {
       start = maf_df$Start_Position,
       end = maf_df$End_Position
     ),
-    strand = maf_df$Strand,
+    # strand = maf_df$Strand,
```
jaclyn-taroni: I might put a brief comment here that contains some of the information that is included in the text of the PR.

PR #290, opened 2019-11-25
01 02 03 script

bill-amadio:

This branch contains a new shell script to run scripts 01-clean_split_data.R, 02-train_elasticnet and 03-evaluate model in succession.  We also begin tracking the output files from 02-train_elasticnet: the model object, the filtered transcripts used to build the model (necessary for predicting new cases) and the model coefficients.

Can we predict gender from RNASeq expression?

#### What was your approach?

Train an elastic net logistic regression model from samples with reported_gender.

84 - RNA expression-based prediction of sex.

Accuracy, completeness, operability.

In the shell script, I supplied values for all arguments that did not have default values.  Is that correct?

I received the following warning when adding the model files to my local branch:
$ git add analyses/sex-prediction-from-RNASeq/models/
warning: LF will be replaced by CRLF in analyses/sex-prediction-from-RNASeq/models/kallisto_stranded_36354_0.25_model_coefs.tsv.
The file will have its original line endings in your working directory

Probably not relevant for you, but FYI: when switching from running Rscript commands from the Windows command line to bash run-sex-prediction-from-RNASeq.sh, I got "There is no package called..." errors.  https://stackoverflow.com/questions/27673000/rscript-there-is-no-package-called identified the error as a local one, and their fix worked.  FYI.

Yes

### Results

From 70/30 train/test split:

Confusion Matrix and Statistics

          Reference
Prediction Female Male
    Female    112    4
    Male        0  142

               Accuracy : 0.9845
                 95% CI : (0.9608, 0.9958)
    No Information Rate : 0.5659
    P-Value [Acc > NIR] : <2e-16

                  Kappa : 0.9686

 Mcnemar's Test P-Value : 0.1336

            Sensitivity : 1.0000
            Specificity : 0.9726
         Pos Pred Value : 0.9655
         Neg Pred Value : 1.0000
             Prevalence : 0.4341
         Detection Rate : 0.4341
   Detection Prevalence : 0.4496
      Balanced Accuracy : 0.9863

       'Positive' Class : Female

      ROC      Sens      Spec
0.9984100 1.0000000 0.9726027

confusion matrix and two-value tables out of caret package

textbook accuracy predicitng reported_gender and germline_sex_estimate on unseen cases.

- [ X] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ X] This analysis has been added to continuous integration.
@jaclyn-taroni : The HEAD statement around line 50 of config.yml appeared again.  We removed it in the last pull request.  I deleted it, and also deleted it from config.yml in my origin master branch.


jaclyn-taroni:
I'm going to remove the new line in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/290/files#diff-3254677a7917c6c01f55212f86c57fbf because it looks like is causing the Docker image to rebuild and also update the branch.

jaclyn-taroni:
@bill-amadio I would expect to see the contents of the `results` directory (output of `03-evaluate_model.R`) added here as well, is there any reason to not include them like file size?

analyses/sex-prediction-from-RNASeq/run-sex-prediction-from-RNASeq.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# This script runs the sex-prediction-from-RNASeq analysis
+# Author's Name Bill Amadio 2019
+
+set -e
+set -o pipefail
+
```
jaclyn-taroni: The reason this fails in continuous integration (CI) currently is because the current directory when it is being run in CI is the root of the directory. You can add the following here:
```sh
# This script should always run as if it were being called from
# the directory it lives in.
script_directory="$(perl -e 'use File::Basename;
  use Cwd "abs_path";
  print dirname(abs_path(@ARGV[0]));' -- "$0")"
cd "$script_directory" || exit
```
And this should fix the directory issues with CI and make sure it runs the same way no matter where it's called from.

jaclyn-taroni: Another thing you might want to try in this shell script is using variables for the arguments you are making to the various R scripts. For example, the data where the test and training data lives is constant across the 3 scripts. 
If you wanted to change the directory name for the processed data, you would have to change it in three places instead of one place if you used a variable. You could add the following to the top of the script, along with any other arguments it'd be good to set as variables (e.g., model directory, filename lead):

```sh
# use a variable for the directory that contains the training/testing data
PROCESSED=processed_data
```
and then to use it, the lines below become:

```sh
Rscript --vanilla 01-clean_split_data.R --expression ../../data/pbta-gene-expression-kallisto.stranded.rds --metadata ../../data/pbta-histologies.tsv --output_directory $PROCESSED --filename_lead kallisto_stranded

Rscript --vanilla 02-train_elasticnet.R --input_directory $PROCESSED --output_directory models --filename_lead kallisto_stranded

Rscript --vanilla 03-evaluate_model.R --test_set_input_directory $PROCESSED --model_input_directory models --output_directory results --filename_lead kallisto_stranded
```

Also worth a thought as you revisit this - if you use a bash script to run everything, do you need `03-evaluate_model.R` to 'construct' the input file names using `paste`?

analyses/sex-prediction-from-RNASeq/run-sex-prediction-from-RNASeq.sh
```diff
@@ -0,0 +1,13 @@
...
+Rscript --vanilla 01-clean_split_data.R --expression ../../data/pbta-gene-expression-kallisto.stranded.rds --metadata ../../data/pbta-histologies.tsv --output_directory processed_data --filename_lead kallisto_stranded
+
+Rscript --vanilla 02-train_elasticnet.R --input_directory processed_data --output_directory models --filename_lead kallisto_stranded
+
+Rscript --vanilla 03-evaluate_model.R --test_set_input_directory processed_data --model_input_directory models --output_directory results --filename_lead kallisto_stranded
```
jaclyn-taroni: Can you add running this again with `--target_column germline_sex_estimate` and commit the results? So I'm proposing two `03-evaluate_model.R` steps.

bill-amadio: I pushed changes that

1) fix script directory issue
2) eliminate all file name constructions by paste
3) run 03-evaluate_model twice, once for reported_gender and once for germline_sex_estimate
4) track both models and results directories

git did not record changes to the file kallisto_stranded_36354_0.25_model_transcripts.RDS in the models directory.  The file time stamp (today) is the same as the other two files in the models directory, and I am sure it is new since I delete all output files and directories before a test run of the scripts.

jaclyn-taroni:
Wanted to give you a heads up @bill-amadio -- our current process for generating files for testing doesn't take into account `reported_gender` (see: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/303#issuecomment-560671684). Until we address that, I have commented out the training and evaluate steps in CI with https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/709887e3555bdb9f0d634ff561e41d1c4ada96d5 and we'll have to wait to merge this. All of that needs to be done on our end, but wanted to let you know about a potential delay.

bill-amadio:
Thanks, @jaclyn-taroni.

jaclyn-taroni:
Okay @bill-amadio, I fixed this with https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/308! I'm going to merge the latest changes from master into your branch so we can get the changes in this pull request tested.

jaclyn-taroni:
@bill-amadio since all of your R scripts have gone in, I'm going to remove the `sex-prediction-from-RNASeq.Rmd` notebook from the repository and the commented out step from the CI `config.yml`. I'm aiming to review the shell script today!

jaclyn-taroni:
In making those changes, I identified a way to use a bash array for the target columsn to cut down on the lines of code (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/290/commits/e808f1b7eb221ba47ceb86164b8473078a848123). I need to put more comments in the shell script with that change, but I wanted to make sure this ran in CI.

jaclyn-taroni:
It looks like for this split of the data the `reported_gender` and `germline_sex_estimate` labels are the same, is that also what you saw locally @bill-amadio ?

jaclyn-taroni:
@cbethell since I'm the last person to work on this, can you take a look at the file paths, etc. and make sure we are not doing something like accidentally evaluating on the training set instead of the test set please?

bill-amadio:
@jaclyn-taroni Thanks for all of Friday's work on the script.

@jaclyn-taroni: On Friday, you asked

For presenting results, having an R notebook (04 in your pipeline) that reads in the results from 03-evaluate_model.R is something that comes to mind. Interested to hear your thoughts on what else might go in such a notebook.

Certainly depends on the intended audience, but I think we can present the confusion matrix and a visualization of the distribution of probabilities out of the model.  The calls are quite confident

For example
Strength_of_Calls

We might also include deciles of the Probability Female values.  The median is 0.14, and the 60th percentile is 0.80, so only 24 calls with probability Female between 0.14 and 0.80.

If I remember correctly,test set accuracy was 80% using the top 1% MAD transcripts.  We might plot test set accuracy as a function of percentage of transcripts used, and we might identify transcripts with the most predictive power.  

Again, depends on the intended audience.  Can you share some details on how sex prediction from RNASeq will be used going forward?

@jaclyn-taroni In the test set for seed 36354, the reported_gender and germline_sex_estimate columns are identical.  There are four samples in the training set with reported_gender != gremline_sex_estimate.

@jaclyn-taroni I can start the README today.

jaclyn-taroni:
Hi @bill-amadio, quoting @cgreene from #84 here:

> It could be interesting to understand how accurately [germline sex estimate] can be called from gene expression data. For some studies, gene expression data is all that is available. If it can also be accurately called directly from gene expression data, these studies will be better positioned to evaluate their metadata.

My take on the _long term view_ -- imagine given any (human) RNA-seq expression data set, you want to be able to predict genetic sex. Down the road, that may require training on a different data set but the analysis here can serve as a proof of concept and road map for that. In the shorter term, this could be used on pediatric brain tumor data sets that have incomplete metadata. I view the audience as anyone who might want to improve the metadata for their RNA-seq cohort, so I think the big picture of performance is what is most important to display but any "model diagnostics" that could be helpful when moving to the longer term are also valuable.

In service of both the long- and the short-term, we may want to know the following:

1. How many features are required?
2. Does prediction performance vary across biological contexts (e.g., cancer type)?
3. Does the library strategy matter a lot for prediction?

So, to the first point, totally agree that it would be interesting to understand performance as a function of the number of transcripts used. To the second point, @cgreene mentioned looking at performance broken out by histology on #84. I like your idea of plotting the distribution of probabilities in general and for this specific purpose. I would use the `disease_type_new` column in `pbta-histologies.tsv` to break things out. These two points (1 & 2) are where I would focus the next pull request. 

Regarding the library strategy (number 3), I think we've done a good job setting up the scripts in the module to make it relatively straightforward to train on stranded data and predict on poly-A data (and the other way around, though the sample size for poly-A is much smaller). If performance is poor when the library strategies are different, that's important for an end user to know. I view this as something we might want later after we present and document what you've done so far.

I'm going to get this merged, but please feel free to comment here! I'm looking forward to hearing your thoughts.

bill-amadio:
Regarding 2. Does prediction performance vary across biological contexts (e.g., cancer type)?

We need to think about how we define performance.  We have a model that was trained on a 70/30 split of our gene expression data set.  The 70/30 model produced 4 errors out of 258 calls in the test set.  disease_type_new values for the four discordant calls are 

disease_type_new values for discordant calls
[1] "Medulloblastoma"                                    "Brainstem glioma- Diffuse intrinsic pontine glioma"
[3] "High-grade glioma;astrocytoma (WHO grade III/IV)"   "Medulloblastoma"                                   

There are 40 different values in the disease_type_new column for the 258 samples in the test set.  Medullo has 34 samples, High_grade glioma has 30, and Brainstem glioma has 4; many have just one sample.

I don't think we have enough samples to use accuracy of the 70/30 model as the performance metric for breakouts by disease_type_new.  Training a model on, say, the Medullo samples only (100 total) introduces additional reasons for performance variance beside disease_type_new.

We might consider distribution of call probabilities as a performance metric.  I ran two-sample Kolmogorov-Smirnov tests for Medullo v. non-Medullo and High Grade glioma v. not).  Results and visualizations appear below.  The Medullo test showed a significant difference; High Grade glioma did not.

	Two-sample Kolmogorov-Smirnov test

data:  rg_details_expanded[which(rg_details_expanded$disease_type_new == "Medulloblastoma"), "Female"] and rg_details_expanded[which(rg_details_expanded$disease_type_new !=     "Medulloblastoma"), "Female"]  
D = 0.26628, p-value = 0.02427
alternative hypothesis: two-sided

StrengthOfCalls_MedulloHighlight

	Two-sample Kolmogorov-Smirnov test

data:  rg_details_expanded[which(rg_details_expanded$disease_type_new ==  "High-grade glioma;astrocytoma (WHO grade III/IV)"), "Female"] and rg_details_expanded[which(rg_details_expanded$disease_type_new != "High-grade glioma;astrocytoma (WHO grade III/IV)"), "Female"]    
D = 0.10439, p-value = 0.9046
alternative hypothesis: two-sided

StrengthOfCalls_HighGradeGliomaHighlight

Basically, we see some of the less confident calls were for Medullo samples.  Perhaps a heat map might be more comprehensive and informative.  I'll give that a try.

Am I missing something in what @cgreene meant when he wrote about breaking out performance by histologies in [#84](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/84)?

cgreene:
@bill-amadio : this is exactly what I was wondering about! I thought that perhaps in some histologies silencing might be breaking down, and this could potentially show up with changes in X inactivation (which your classifier might then struggle to classify). It looks like performance is so good (and so concordant with the genetic sex calls) that it probably isn't revealing this. The good news, is that there are a lot of datasets without annotated sex information, so this classifier could be helpful to predict that metadata in such studies (as referred to by @jaclyn-taroni in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/290#issuecomment-563319254 )!

PR #293, opened 2019-11-26
V11 Release

yuankunzhu:
#### Purpose/implementation

data release updates for v10
- changes:
  - Add putative oncogenic fusions from [fusion filtering and prioritization analysis](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering)
  - Update consensus SNV/indels to contain [consensus MNVs](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/275)
  - Add RNA-Seq [collapsed matrices](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/287) - wrong files (tables of transcripts removed) were included with [V10](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/273)
#### Directions for reviewers
changed on release notes and download script. 

- release file added:
  - pbta-fusion-putative-oncogenic.tsv
  - pbta-snv-consensus-20191125.zip
  - pbta-gene-expression-rsem-fpkm-collapsed.polya.rds
  - pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
  - pbta-snv-consensus-mutation.maf.tsv.gz
  - pbta-snv-consensus-mutation-tmb.tsv

- release file removed:
  - pbta-gene-expression-rsem-fpkm-collapsed_table.polya.rds
  - pbta-gene-expression-rsem-fpkm-collapsed_table.stranded.rds
  - pbta-snv-consensus_11122019.zip

Issue #287, opened 2019-11-22
Planned data release: V11


jharenza:
#### What data file(s) does this issue pertain to?
V10 collapsed files were tables of genes removed, not the collapsed matrices, #248 

#### What release are you using?

V10

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
Planned data to release:
- [ ] RNA-Seq collapsed FPKM files, #248 
- [ ] Putative oncogenic fusion TSV, #39

jashapiro:
As long as there is an update planned, I will be able to provide updated consensus SNV files. (With MNVs included)

jaclyn-taroni:
Related: #275

jaclyn-taroni:
New consensus files from @jashapiro: https://open-pbta.s3.us-east-1.amazonaws.com/data/snv-consensus/snv-consensus-20191125.zip

jharenza:
Re: putative oncogenic fusion (above), I am thinking we release the final prioritized list for the oncoprints, which would also become a supplemental table (I guess we add to the manuscript doc later?). cc: @jaclyn-taroni

jaclyn-taroni:
To clarify, you are referring to this file: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/49acc98f5ffd86853fc70f220623311e13e3ca9f/analyses/fusion_filtering/results/PutativeOncogenicFusion.tsv correct @jharenza?

What are you going to name this file? I lean towards something like `pbta-fusion-putative-oncogenic.tsv`.

Regarding the new consensus mutation files, I think since we will probably put the `pbta-fusion-putative-oncogenic.tsv` directly in the release folder, we should make some changes to how the consensus mutation files are distributed.

The file structure of `snv-consensus-20191125.zip` is as follows:

```
.
├── README.md
├── consensus_mutation.maf.tsv
└── consensus_mutation_tmb.tsv
```

Could we rename the contents to:
`consensus_mutation.maf.tsv` -> `pbta-snv-consensus-mutation.maf.tsv` (+ possibly compress this file)
`consensus_mutation_tmb.tsv` -> `pbta-snv-consensus-mutation-tmb.tsv`

[...]

PR #294, opened 2019-11-26
04 project specific filtering

kgaonkar6:
project specific filtering to capture recurrent fusion and oncogene annotated gene fusions. 
UPDATE: the genereference list was updated 11/25 to remove curated and predicted TF gene list.

#### Putative Driver:
Filtering for general cancer specific genes

#### Scavenge back filtered fusions to add to putative oncogenic fusions:
1. In-frame/frameshift fusions is called in atleast 2 samples per histology OR
2. In-frame/frameshift fusions is called in atleast 2 callers 
AND

Filtering for recurrent fusions and oncogenic fusion in all histologies

#### What was your approach?
Mostly summarizing data to identify 
 * if annotated as oncogenic (from genereferencelist and fusionreferencelist in reference folder)
 * if inframe/frameshift found in both callers/called in more than 2 samples per histology ;these fusions should also be uniquely found in a histology and not multifused ( We call a gene multifused if we identify more than 5 different partner genes for a gene in a sample )

We want to  retain unique calls for oncoprint so I've only kept columns that I think that don't duplicate the fusion calls. 

For example each FusionName may have duplicate rows because of different so I've removed the following columns.
-Caller,-annots,-JunctionReadCount,-SpanningFragCount,-Confidence,-LeftBreakpoint,-RightBreakpoint in the final fusion calls

Yes this table is ready for oncoprint:
results/PutativeDriverFusion.tsv 

### Results
results/PutativeDriverFusion.tsv 

table

3213 Fusions are identified as putative oncogenic fusions for PBTA


jaclyn-taroni:
@kgaonkar6 is this just a matter of updating the reference lists and rerunning without changes to `04-project-specific-filtering.Rmd`?  `04-project-specific-filtering.Rmd` is showing up as entirely new in the `Files Changed` tab and this branch has conflicts that are too complex to be handled via the web interface. I think this may have come about as a result of adding these changes to the same branch `kgaonkar6:04_project_specific_filtering` as the changes that went in as part of #277. 

I would recommend syncing your master branch with the upstream AlexsLemonade master, creating a new branch (e.g., `update_fusion_references`), adding and committing the changes to the reference list, rerunning the pipeline, and then adding a committing the changes to the output (e.g., `results/FilteredFusion.tsv`, `results/PutativeDriverFusion.tsv`).

Side note -- are these the results that are part of the v11 release (#293 / #287), i.e., those use the updated references?

jharenza:
Hi @jaclyn-taroni - I can answer the last question. Yes, this update was already performed for the table in the V11 release (only two fusions were removed `KAT2B--LINC02008` and `SF1--MTA2` and I looked into them - no evidence in the literature. `SF1` had previously been reported in a leukemia, but with an oncogenic partner, `PDGFRA`. In our pipeline, if we saw this, it would come out because it has an oncogenic partner, so this made me comfortable removing these not so validated lists of TFs).

To explain further, as I was writing the [methods](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#fusion-prioritization), I actually did not realize @kgaonkar6 was collecting all TFs from TRANSFAC lists (N = 158 from predicted TFs, 201 from curated TFs) plus the Cell paper cited with a more validated list. I thought it was just the latter, so that was some miscommunication on our end. I inspected the differences and many of the added "TF" from TRANSFAC are really cofactors (eg: LMO2, RB1, NF1 and not TFs; though, as they are oncogenes, they would be retained via that list), so we opted for just the list from the Cell paper, which were deemed YES TFs (N = 1639). So, we updated this to only use the validated list and in an effort to not do multiple releases of this, we updated the V11 file already. 

Hope this helps. Happy Thanksgiving! :)

jaclyn-taroni:
Thank you @jharenza, that's helpful in thinking about documentation around v11. Happy Thanksgiving!

kgaonkar6:
Thanks @jaclyn-taroni @jharenza , I was trying to checkout master after the fetch upstream command but it would ask me to commit the changes I already had made to the reference and results folder so I thought of committing it to this branch itself would fix that. I'll try to sync the master to upstream again and follow your suggestions to create a new branch.

PR #299, opened 2019-11-27
V11 consensus mutation downstream module updates

jaclyn-taroni:

I know the following modules are going to fail in CI with v11:
* `oncoprint-landscape`
* `mutational-signatures`
* `tmb-compare-tcga` (in upcoming commit)

Filing this WIP pull request with file path changes + rerun results so we are in good shape for v11 on Monday.


analyses/oncoprint-landscape/00-map-to-sample_id.R
```diff
@@ -166,8 +166,13 @@ if (!is.null(opt$independent_specimens)) {
 
 message("Preparing MAF file...")
 
-# MAF files already capture sample IDs
+# join the sample_id information to the MAF file and then set as the tumor
+# sample barcode
 maf_df <- maf_df %>%
+  inner_join(select(histologies_df,
```
cansavvy: I am not as familiar with the oncoprint stuff, so @cbethell should look at this since she'll have a better idea of what's going on in this context.

jaclyn-taroni:
@cansavvy there are some lines being removed https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/299/files#diff-38d104937067e561a698f9d963629126L1111 is that ok?

cansavvy:
> @cansavvy there are some lines being removed https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/299/files#diff-38d104937067e561a698f9d963629126L1111 is that ok?

Yes. I re-ran the TCGA set up again locally to double check. I think those samples got mis-matched when I was reconfiguring things when I added in the MC3 data, but now it is fixed. The RDS file just never got re-ran (a flaw of having that `if` statement).

jaclyn-taroni:
Closing because this is now included in #293 - https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/293/commits/20c3df60c3c2f2679d8a660ef88263f2cbadebe9

Issue #296, opened 2019-11-27
Updated analysis: `snv-callers` documentation


jaclyn-taroni:
#### What analysis module should be updated and why?

The `snv-callers` documentation is out of date.

#### Why should the module be updated?

As far as I can tell, the README doesn't account for the changes as part of the refactor (#201 and related PRs): https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/49acc98f5ffd86853fc70f220623311e13e3ca9f/analyses/snv-callers/README.md

We are also changing the release process with v11 such that the consensus mutation files are no longer included as a zip file and therefore there will not be a README available at `data/snv-consensus-<date>` (see https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/287#issuecomment-558653481 and https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/293#issuecomment-558782828).

#### What changes need to be made? Please provide enough detail for another participant to make the update.

- [x] `snv-callers/README` should be updated to reflect the current state of the analyses
- [x] The main repository README should include information on the consensus files under the `Data Format` header.
  - [x] @jashapiro has documented `pbta-snv-consensus-mutation.maf.tsv.gz` in https://github.com/yuankunzhu/OpenPBTA-analysis/pull/1/files
  - [x] The TMB files need similar documentation

#### What input data should be used? Which data were used in the version being updated?

N/A I am requesting a documentation change.

#### Who will complete the updated analysis?

I am assigning @cansavvy.

cansavvy:
I believe this has been addressed with the merging of #333. Is there anything left that needs more documentation?

jaclyn-taroni:
👍

PR #304, opened 2019-12-02
Update Documentation to Reflect SNV refactoring

cansavvy:
#### Main changes
- Revamped the whole README to reflect the refactoring of the SNV scripts. 
- Got rid of the 00-setup.R and its being called in the bash script. 

#296 

- Does this make the analysis and it methods more clear? 
- Would you feel confident in re-running this analysis by looking at this documentation? 
- Is there anything in particular that is missing and should be documented more? 
- Is there anywhere that should have a link to a source some place that doesn't currently have a link? 

### Results
 No new results just updating the documentation

These items were taken care of previously. 


analyses/snv-callers/README.md
```diff
@@ -1,213 +1,75 @@
...
-  * [Variant allele fraction calculation](#variant-allele-fraction-calculation)  
-  * [Genomic regional analyses](#genomic-regional-analyses)  
-  * [Tumor mutation burden](#tumor-mutation-burden-calculation)
-  * [COSMIC mutation overlap](#cosmic-mutation-overlap)  
-* [Comparison analysis](#comparison-of-callers)
- * [Mutation IDs](#mutation-ids)
- * [Consensus mutation calls](#consensus-mutation-calls)
-* [Overall file structure](#overall-file-structure)
-* [Summary of functions](#summary-of-custom-functions)
+This consensus mutation file is [MAF-like](#consensus-mutation-call) meaning it is TSV file that contains many of the fields of a [MAF file](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) but also has [VAF](#variant-allele-fraction-calculation)
```
jashapiro: I am pretty sure we are including the full set of fields now, as they are all in the database. 
```suggestion
This consensus mutation file is [MAF-like](#consensus-mutation-call) meaning it is TSV file that contains the fields of a [MAF file](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) but adds [VAF](#variant-allele-fraction-calculation), but does not contain a starting comment line with a version number.
```

analyses/snv-callers/README.md
```diff
@@ -1,213 +1,75 @@
...
-                                             <caller_name>_vaf.<file_format>
-                                             <caller_name>_tmb.<file_format>
- --file_format: What type of file format were the vaf and tmb files saved as? Options are
-               "rds" or "tsv". Default is "rds".
- --output : Where you would like the output from this script to be stored.
- --overwrite : If TRUE, will overwrite any reports of the same name. Default is
-               FALSE
-```
+- `consensus_mutation.maf.tsv` - Mutations that were called by all three of these callers for a given sample are saved to this file.
+This file is MAF-like meaning it is TSV file that contains many of the fields of a [MAF file](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) but also some added calculations like [Variant Allele Fraction](#variant-allele-fraction-calculation) and some sample metadata information.
```
jashapiro: Do we want to mention again the absence of a comment line?
```suggestion
This file is MAF-like meaning it is TSV file that contains the fields of a [MAF file](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) but adds [Variant Allele Fraction](#variant-allele-fraction-calculation).
```

analyses/snv-callers/README.md
```diff
@@ -1,213 +1,75 @@
...
-together deletions, insertions, and long (more than a SNV) as their own groups.
+### Mutation Comparisons
 
-*Output for this analysis*
-* `results/<caller_name>/<caller_name>_vaf.rds`
-* `plots/<caller_name>/<caller_name>_<strategy>_base_change.png`
+In order to compare mutations across callers, the data tables for each caller
+were indexed by: `Chromosome` `Start_Position` `Reference_Allele` and `Allele`.
+Meaning that across callers, if all these fields were identical, they were considered to be the same mutation.
+This was done in the `01-setup_db.py` script using the function.
```
jashapiro: The indexing is an implementation detail for speed, it speeds comparisons later, which also end up including sample for matching (but we get enough speed without specifically indexing that as well).

So I think this section needs to just describe which callers were used for the consensus, what was considered the same mutation (the fields mentioned, but also sample).  and how we handle MNVs.

analyses/snv-callers/README.md
```diff
@@ -1,213 +1,75 @@
...
-             is given from top directory of 'OpenPBTA-analysis'
- --overwrite : If TRUE, will overwrite any reports of the same name. Default is
-             FALSE
-```
-
-# Individual Caller Evaluation
-
-The first step in this analysis is an individual evaluation of each MAF from each caller.
-This analysis prints out a report that is further subdivided by the WGS and WXS samples.
+It is "MAF-like" file because it has many of the same fields as a MAF file but..
```
jashapiro: Okay, this the third time we define MAF-like... we should be less repetitive.

analyses/snv-callers/README.md
```diff
@@ -223,144 +85,65 @@ Where genome size is calculated from the respective BED file as:
...
+- By default, the scripts will not overwrite existing files of the same name. However,
+this can be overridden with `--overwrite` option.
 
-After running an initial evaluation and set up of each of the callers' MAF files,
-the `compare_snv_callers.Rmd` notebook can be run to get final results on consensus
-mutation calls.  
+### 01-setup_db.py
 
-### Mutation IDs  
+Creates and/or fills a database of variant calls that will be used by subsequent calls.
```
jashapiro: Probably worth mentioning that the maf tables are complete for strelka and lancet, but not for mutect and vardict, to save space. Here you might mention the indexing by position and mutation, to facilitate later searches and comparisons.
```suggestion
Creates and/or fills an SQLite database of variant calls that will be used by subsequent steps to find consensus mutations.
```

analyses/snv-callers/README.md
```diff
@@ -223,144 +85,65 @@ Where genome size is calculated from the respective BED file as:
...
 
-`mutation_id` is a concatenation of:  
-* `Hugo_Symbol`  
-* [`change`](#base-change-analysis)  
-* `Start_Position`  
-* `Tumor_Sample_Barcode` (the sample ID)  
+### 02-merge_callers.R
 
-If mutation_id's are identical among MAF files, they are considered the same.
+Using the database created by `01-setup_db.py`, merge callers' data files into consensus MAF-like file.
```
jashapiro: ... as described at (link)

analyses/snv-callers/README.md
cansavvy: Fair enough! ha.

analyses/snv-callers/README.md
cansavvy: Would you be able to add a few lines about how you did the MNVs? Or do you have elsewhere where that is summed up nicely?

jashapiro: I will add it. I think the best summary is in #275, but it isn't quite as clear as it could be. I'll write something better tomorrow morning.

cansavvy: Sounds good. You can always add it as a PR suggestion and then I can just commit it.

analyses/snv-callers/README.md
```diff
@@ -1,213 +1,63 @@
...
+- A zip file containing:
+  - `consensus_snv.maf.tsv` - is  [MAF-like file](#consensus-mutation-call) that contains the snvs that were called by all three of these callers for a given sample are saved to this file.
+  These files combine the [MAF file data](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) from 3 different SNV callers: [Mutect2](https://software.broadinstitute.org/cancer/cga/mutect), [Strelka2](https://github.com/Illumina/strelka), and [Lancet](https://github.com/nygenome/lancet).
+  See the methods on the callers' settings [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#somatic-single-nucleotide-variant-calling) and see [the methods of this caller analysis and comparison below](#summary-of-methods).  
+  - `consensus_snv_tmb_coding_only.tsv` - Tumor Mutation burden calculations using *coding only* mutations use the consensus of Lancet, Mutect2, and Strelka2. 
+  - `consensus_snv_tmb_all.tsv` - Tumor Mutation burden calculations using *all* mutations use the consensus of Mutect2, and Strelka2. (Lancet was excluded because it has a coding region bias).
```
jashapiro:
```suggestion
  - `consensus_snv_tmb_all.tsv` - Tumor Mutation burden calculations using *all* mutations use the consensus of Mutect2, and Strelka2. (Lancet was excluded because it was run with settings that bias its calls towards coding regions).
```

analyses/snv-callers/README.md
```diff
@@ -1,213 +1,63 @@
...
+  These files combine the [MAF file data](https://docs.gdc.cancer.gov/Data/File_Formats/MAF_Format/) from 3 different SNV callers: [Mutect2](https://software.broadinstitute.org/cancer/cga/mutect), [Strelka2](https://github.com/Illumina/strelka), and [Lancet](https://github.com/nygenome/lancet).
+  See the methods on the callers' settings [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#somatic-single-nucleotide-variant-calling) and see [the methods of this caller analysis and comparison below](#summary-of-methods).  
+  - `consensus_snv_tmb_coding_only.tsv` - Tumor Mutation burden calculations using *coding only* mutations use the consensus of Lancet, Mutect2, and Strelka2. 
+  - `consensus_snv_tmb_all.tsv` - Tumor Mutation burden calculations using *all* mutations use the consensus of Mutect2, and Strelka2. (Lancet was excluded because it has a [coding region bias in the way it was ran](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#snv-and-indel-calling)).
```
jashapiro:
```suggestion
  - `consensus_snv_tmb_all.tsv` - Tumor Mutation burden calculations using *all* mutations use the consensus of Mutect2, and Strelka2. (Lancet was excluded because it has a [coding region bias in the way it was run](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#snv-and-indel-calling)).
```

PR #309, opened 2019-12-03
Check that files in data are unchanged in CI

jaclyn-taroni:

Here I'm making a change to our download script such that the `data` directory will no longer have write permissions once it's done running in CI. This helps us ensure that all modules are being tested on the same set of files.

N/A

Is this the best way to implement this?


cgreene:
Removing execute on directories makes them invisible but traversable. I think you just want the -w. Let's see what this does on CI :)

jaclyn-taroni:
Hopefully it fails!

cgreene:
But on the writing step and not the cd step 😁

jaclyn-taroni:
Same exit status @cgreene

cgreene:
> ^@^@cd: usage: cd path prog...
😕

jashapiro:
Nope. ???

jashapiro:
Since this seems to be failing in mysterious ways, maybe we should change approach? How about checking that the files in data are unchanged as a final step? I would think checking modification times could be sufficient... Thoughts?

jaclyn-taroni:
> How about checking that the files in data are unchanged as a final step? 

To clarify - The last step in CI _not_ the last step in the download step, correct?

jaclyn-taroni:
#317 is the preferred approach, so I am closing this.

Issue #165, opened 2019-10-23
Proposed Analysis: apply TP53 & NF1 classifiers to PBTA data


jaclyn-taroni:
#### Scientific goals

Identify samples that have _TP53_  or _NF1_ inactivation using gene expression data. 

#### Proposed methods

* This would use a pre-existing classifier (trained on TCGA data) that was described in Knijnenburg et al. 
* It seems appropriate to keep the poly-A and stranded samples separate for this.

#### Required input data

Gene expression data. I believe relative abundance data at the gene-level is what was used to train the classifier. In this cohort that would correspond to:

`pbta-gene-expression-rsem-fpkm.polya.rds`
`pbta-gene-expression-rsem-fpkm.stranded.rds`

#### Proposed timeline

I think this could be accomplished in 2 weeks.

#### Relevant literature

* Initially described in [Knijnenburg et al. _Cell Reports._ 2018.](https://doi.org/10.1016/j.celrep.2018.03.076)
* Applied in [Rokita et al. _bioRxiv._ 2019.](https://doi.org/10.1101/566455 )

cgreene:
From a conversation with @gwaygenomics and @PichaiRaman I have some additional info that may be helpful:
```
Hi Pichai and Casey,

Probably the easiest way to apply the TP53 classifier is by swapping in PBTA data in this script: https://github.com/marislab/pdx-classification/blob/master/1.apply-classifier.ipynb

This is the analysis done for the PPTC PDX paper.

Thanks!
Greg
```

jharenza:
Updated to add _NF1_ since that classifier also works well in pediatric data!

jaclyn-taroni:
Relevant pub for _NF1_:  [Way et al. _BMC Genomics._ 2017.](https://doi.org/10.1186/s12864-017-3519-7)

jharenza:
@kgaonkar6 will work on this!

gwaygenomics:
perhaps important to note distinction here:

> Relevant pub for NF1: Way et al. BMC Genomics. 2017.

This demonstrated proof of concept for NF1 loss classification in Glioblastoma specifically.

> Initially described in Knijnenburg et al. Cell Reports. 2018.

This trained an NF1 loss classifier using pancancer data (so many more cancer types than just glioblastoma)

> Applied in Rokita et al. bioRxiv. 2019.

[...]

PR #312, opened 2019-12-04
add modules needed for classifier

kgaonkar6:
Add modules to dockerfile to run classifier from this environment yaml file:
https://github.com/marislab/pdx-classification/blob/master/environment.yml

How do samples classify using TP3/NF1 scores from https://github.com/marislab/pdx-classification

#### What was your approach?
I added modules to dockerfile to run classifier from this environment yaml file that were missing from the dockerfile:
https://github.com/marislab/pdx-classification/blob/master/environment.yml

Proposed Analysis: apply TP53 & NF1 classifiers to PBTA data #165

This PR is only to add the modules for the script https://github.com/marislab/pdx-classification/blob/master/1.apply-classifier.ipynb

I have some questions going forward with this analysis, should 01_setup.sh be git clone the repo above or should 01_apply-classifier.ipynb be I recreate a notebook similar to the one above for our dataset? ( I did git clone and edit the above script to run on polya set on an ec2 instance by setting up the conda environment and it runs successfully)

@jharenza seems like to get the scores we only need to run 1.apply-classifier.ipynb from the issues ticket. Should the 01_apply-classifier.ipynb be run with both polya and stranded merged into one input file or separate runs?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -188,7 +188,7 @@ RUN pip3 install "numpy==1.17.3" && \
 
 
 # pip install for modules Ras, NF1, and TP53 Classifiers
-RUN pip3 install "statsmodels==0.10.2" "plotnine==0.3.0" "scikit-learn==0.19.1" "rpy2==2.9.3" "seaborn==0.8.1" "jupyter==1.0.0" "ipykernel==4.8.1" "widgetsnbextension==2.0.0" "nb_conda_kernels==2.1.0"
+RUN pip3 install "statsmodels==0.10.2" "plotnine==0.3.0" "scikit-learn==0.19.1" "rpy2==2.9.3" "seaborn==0.8.1" "jupyter==1.0.0" "ipykernel==4.8.1" "widgetsnbextension==2.0.0"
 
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -186,6 +186,11 @@ RUN pip3 install "numpy==1.17.3" && \
    pip3 install "pandas==0.25.3" && \
    pip3 install "snakemake==5.8.1"
 
+
+# pip install for modules Ras, NF1, and TP53 Classifiers
```
cgreene: I think you'll want to install these packages into the previous `RUN` command with the multi-line formatting. The multi-line is for convenience/readability but keeping them in the same `RUN` command will put all the python installs on the same layer.

cgreene: Actually it looks like for R we're doing this this way (multiple RUN commands), so never mind on adding these to the above RUN command! I do think the multi-line formatting is easier to read though and would prefer that.

kgaonkar6: ah ok yeah, I can make the edit to make it more readable. Thanks!

PR #316, opened 2019-12-06
Change to subset of genome for CI

jashapiro:

This is a test to see if we run into problems by using only a subset of chromosomes in CI.

Slow CI

#### What was your approach?

Rather than exporting the full genome from BSgenome, I tried making a subset, using a couple of smaller chromosomes (but keeping sex chromosomes, in case they are used specially). Hopefully this won't break downstream analyses!

Are there particular chromosomes that should be sure to be included?


jaclyn-taroni:
I don't think anyone is using the FASTA yet, so I don't expect any issue with using a subset of chromosomes?

PR #317, opened 2019-12-06
Try making data unwritable in CI, take 2

jaclyn-taroni:

This is an alternative to the "check that nothing has changed" process that is _currently_ in #309. I'm wondering if our `cd` oddness was due to `run_in_ci.sh` and I want to test the method in #309 at the same time. Thus, I'm filing this draft PR.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           
       - run:
           name: Write md5sum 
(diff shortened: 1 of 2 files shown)


jaclyn-taroni:
@cgreene what do you think of this vs. #309? I bet it was an issue with `run_in_ci.sh`...

PR #322, opened 2019-12-09
Add shell script and README for `collapse-rnaseq`

jaclyn-taroni:

In this PR, I am:

* Adding documentation to the `collapse-rnaseq` module using https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/64 as a guide. I am also explicitly noting that `00-create-rsem-files.R` is upstream of this repository.
* Adding a shell script that runs the analyses per [**Adding Analyses with Multiple Steps**](https://github.com/AlexsLemonade/OpenPBTA-analysis#adding-analyses-with-multiple-steps)
* Making use of a module-specific `results` folder per [**Folder Structure**](https://github.com/AlexsLemonade/OpenPBTA-analysis#folder-structure)
* Adding a note about the collapsed matrices in the **Data Formats** section of the README.

Is this documentation sufficient?

### Results

N/A - this is a documentation + shell script change.


cgreene:
Does the stranded average correlation value show up for you in the HTML? I'm not seeing it, but it may also be because firefox is struggling with the tables.

Poly-A
Screen Shot 2019-12-09 at 9 11 22 AM


Stranded
Screen Shot 2019-12-09 at 9 11 09 AM

jaclyn-taroni:
I'm having the same problems in Firefox with what's in `master`... stand by

cgreene:
I think that's because the top of the table has NAs there. If I sort descending, things show up.

jharenza:
chrome is also very slow with this, however i am getting another odd issue - when I search for `MYCN`, it seems that some gene names are truncated, but not sure if it is due to slowness.
Screen Shot 2019-12-09 at 9 18 48 AM

jaclyn-taroni:
I think I'll need to take out the use of `DT` for the first table
```R
Output created: 02-analyze-drops.nb.html
Warning messages:
1: In instance$preRenderHook(instance) :
  It seems your data is too big for client-side DataTables. You may consider server-side processing: https://rstudio.github.io/DT/server.html
2: In instance$preRenderHook(instance) :
  It seems your data is too big for client-side DataTables. You may consider server-side processing: https://rstudio.github.io/DT/server.html
```

jaclyn-taroni:
That notebook used to display all genes. I'm now limiting it to instances where multiple Ensembl IDs map to the same gene symbol (those are the cases where things get dropped) and where correlation can be calculated. I also added a table that counts the number of `NA` values with some documentation for the _why_ (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/322/commits/940ad029deb8561ff709bb0529aebbef52fb5554).

PR #328, opened 2019-12-11
CNV consensus (3 of n): Filter bad segments

fingerfen:

Continue copy number consensus call

Merging consensus calls

Issue #128 

The new "get_rid_bad_segments.py" file

No


analyses/copy_number_consensus_call/Snakefile
```diff
@@ -2,9 +2,9 @@
 # Updated Dec 5, 2019
 
 ## Define the ending file(s) that we want
-ALL_FREEC= expand("../../scratch/interim/{sample}.freec.dup.bed", sample=config["samples"])
-ALL_CNVKIT= expand("../../scratch/interim/{sample}.cnvkit.dup.bed", sample=config["samples"])
-ALL_MANTA= expand("../../scratch/interim/{sample}.manta.dup.bed", sample=config["samples"])
+ALL_FREEC= expand("../../scratch/interim/{sample}.freec.dup.filtered.bed", sample=config["samples"])
+ALL_CNVKIT= expand("../../scratch/interim/{sample}.cnvkit.dup.filtered.bed", sample=config["samples"])
+ALL_MANTA= expand("../../scratch/interim/{sample}.manta.dup.filtered.bed", sample=config["samples"])
```
jashapiro: Similarly to what I did in the pull request I filed, we can simplify this a bit, taking advantage of snakemake features, and reduce redundancy: 

```suggestion
OUTPUT= expand("../../scratch/interim/{sample}.{caller}.dup.filtered.bed", 
               sample=config["samples"], 
               caller=["freec", "cnvkit", "manta"])
```

analyses/copy_number_consensus_call/src/scripts/bad_chromosomal_seg_updated_merged.txt
```diff
@@ -0,0 +1,3693 @@
+chr1	0	535988	telo,,,	length,,,
```
jashapiro: Where did this file come from? Is there a script that generated it? Or did it come from a different source? 

If the former, we would like to have that script included in the repository. If the latter, we should indicate where the file came from in a README in the base directory of this analysis.

analyses/copy_number_consensus_call/src/scripts/get_rid_bad_segments.py
```diff
@@ -0,0 +1,135 @@
+
+################## PURPOSE #################
+# This script is to get rid of unwanted/trouble regions like telomeric, centromeric, and known seg-dups regions
+# This script is to take in 2 files
+#      1st: A reference/list of bad segments that contains the bad segments and their start/end positions
+#      2nd: A list of CNVs that need the bad segments to be filtered out
+
+################# ASSUMPTION ###############
+# It is assumed that the reference file DO NOT have overlapping telomeric/centromeric/segments
+# The provided reference file "bad_chromosomal_seg_updated_merged.txt" DOES NOT have
```
jashapiro: Where is this provided from?

analyses/copy_number_consensus_call/src/scripts/get_rid_bad_segments.py
```diff
@@ -0,0 +1,135 @@
...
+
+    ## Get the start/end coordinates of the cnv
+    start_cnv = int(i[1])
+    end_cnv = int(i[2])
+
+    ## Make sure end > start (end downstream from start)
+    if start_cnv > end_cnv:
+        start_cnv, end_cnv = end_cnv, start_cnv
+
+    ## For each CNV, loop through the entire reference file
```
jashapiro: This step could make things slow. Are the CNVs sorted? If so, we can probably keep track of where we are in the reference file.  If the CNV is after our current reference position, we can search forward in the reference, but we won't need to go all the way to the end, and we won't need to go back to the start for the next CNV. I don't know how long this takes on the full data, but if it is slow, this might be a place to look for a speed boost.

analyses/copy_number_consensus_call/src/scripts/get_rid_bad_segments.py
```diff
@@ -0,0 +1,135 @@
...
+        ## Strip the trailing spaces for each line
+        stripped_content = [i.rstrip() for i in file.readlines()]
+
+        ## If the file is TAB delimited, split the file up by TABs
+        if stripped_content[0].find('\t') != -1:
+            final_content = [i.split('\t') for i in stripped_content]
+
+        ## If the file is space delimited, split the file up by the spaces
+        else:
+            final_content = [i.split() for i in stripped_content]
```
jashapiro: For these purposes, are we likely to get both tab and space separated files? I would have assumed that if these are bed files they should all be separated the same way.

hongboxie:
On how "the blacklist" is created. 
1) The blacklist primarily includes IGLL regions, centromeric and telomeric regions. We published a few CNV centered papers, this is standard practice how we do it. PMID: 25066379; PMID: 28398664;PMID: 31222980;PMID: 26742502;PMID: 25892112. 
2) This can also be viewed from my colleague Kai Wang at his PennCNV website : http://penncnv.openbioinformatics.org/en/latest/misc/faq/
3) Segmental duplication regions: it is also popular to remove segmental duplication regions from the final CNV detection. For instance: PMID: 24098321. This is also come from our observation that those regions, present most noisy results that cloud our conclusions.  After removing those, we have much cleaner results, especially for NGS (WGS/WES). For genotyping array, due to how probes are selected, the problem may not as obvious. For WGS/WES, we do see the difference.
4) We are actually want to remove one more class of regions : low mappability regions. But want to leave this open for now.

jashapiro:
Thank you @hongboxie, this is useful information! 

Is there a script or link that that we could include to document these decisions? I also just want to confirm that the regions are for hg38 (only because the PennCNV link seems only to include hg19).

hongboxie:
I will use PennCNV as the webPage link. 

Everything could be converted from hg19 to hg38 build by leftover program from UCSC utility tool set.

Nhat should provide a file of the blacklist which can be posted somewhere as a linked file.

I think that it should be adequate.

analyses/copy_number_consensus_call/Snakefile
fingerfen: I never knew that you could do this with Snakemake. I had so many questions when I was learning Snakemake. I am glad you are showing me this now. This DOES reduce so much redundancy. Thanks you :)

analyses/copy_number_consensus_call/src/scripts/bad_chromosomal_seg_updated_merged.txt
fingerfen: @xiehongbo provided me with the file so I will let him explain it. I took the file he gave to me, sorted and merged any overlapping segments.

analyses/copy_number_consensus_call/src/scripts/get_rid_bad_segments.py
fingerfen: For the purpose of this pipeline, the input file into this step is **ALWAYS** going to be space separated files. The reason why this is here is because when I made this script, I wanted to add a little flexibility into it in case someone wants to take this script out and use it on their tab separated data. I could take it out if you think this is not necessary.

cgreene: Would it be safe to use `re.split` and split on any whitespace of one or more characters with `'\s+'`?

cgreene: Also - don't bed files require tab spacing? Are there tools for bed files that output space separated ones?

analyses/copy_number_consensus_call/src/scripts/bad_chromosomal_seg_updated_merged.txt
fingerfen: @jashapiro I also would like to clarify something. From a glance, it seems that **bedtools subtract** can do what the "get_rid_bad_segments.py" is doing. 

Are you suggesting that we replace that script with a simple bedtools subtract line instead? It seem like bedtools subtract would do the same job and probably much faster.

jashapiro: It depends. I want this analysis to be merged quickly, so I don't really want to add more for you to do. But if it is a quick substitution and you can verify that your code and bedtools work the same way, then it might be a good idea. 

I would suggest that we get this merged in as is since we know it works, and you can think about updating to bedtools in a future PR if you have time.

I am more concerned for this PR with making sure we have all the steps required to generate the bad_chromosomal_seg file. That would include the origin of the files that @hongboxie gave you, as well as the scripts for liftover, sorting and merging, as required.

analyses/copy_number_consensus_call/src/scripts/get_rid_bad_segments.py
cgreene: @fingerfen : if `re.split` would let you simplify many lines to one line and reduce the potential for an error, I'm a fan of that.

analyses/copy_number_consensus_call/src/scripts/bad_chromosomal_seg_updated_merged.txt
fingerfen: I see. So I think right now I will stick with this and make future PR to update to Bedtools if time permits.

As for the bed_chromosomal_seg file. How do you want me to document my process of making this file? Do I comment my process in a certain file somewhere?
I didn't generate this file with a scrip. I did so manually but I could write script if you think it is necessary.

jashapiro: A script would be very much preferred, even if it is just a simple shell script with a series of steps.

fingerfen: Since the purpose of the script is to document my process, does it have to be integrated into the pipeline? Or can it just be a standalone file?

jashapiro: It does not have to be integrated into the snakemake file, but we might ultimately want to add it to the shell script, though that doesn't need to be done right now. I can also add it into the testing once we have it documented in a first phase.

fingerfen: I have made the changes and pushed them to this PR. The changes were:

1) Implementing `re.split` in the .py code
2) Add in the script to generate the black list.

analyses/copy_number_consensus_call/src/scripts/generate_black_list.sh
```diff
@@ -0,0 +1,21 @@
...
+
+# After having the IGLL_telo_centromeric_region.txt from the method described by Kai Wang, the following were performed to form the black list
+# Sort the segdup file and merge it. 
+sort -k1,1 -k2,2n segdup_95 | bedtools merge > segdup_95_merged
+
+
+# Put the IGLL_telo_centromeric_region.txt and the segdup file together,
+# take only the first 3 columns,
+# sort them and merge any overlaping segment
+cat IGLL_telo_centromeric_region.txt segdup_95_merged | awk -v OFS='\t' '{print$1,$2,$3}' | sort -k1,1 -k2,2n | bedtools merge > bad_chromosomal_seg_updated_merged.txt
```
jashapiro: The output here is an actual bedfile, correct? If so, lets go ahead and label it that for clarity.
```suggestion
cat IGLL_telo_centromeric_region.txt segdup_95_merged | awk -v OFS='\t' '{print$1,$2,$3}' | sort -k1,1 -k2,2n | bedtools merge > bad_chromosomal_seg_updated_merged.bed
```

analyses/copy_number_consensus_call/Snakefile
```diff
@@ -83,3 +85,22 @@ rule manta_filter:
         """ | sort -k1,1 -k2,2n > {output.manta_del} && """
         """awk '$6~/DUP/ {{if ($5 > {params.SIZE_CUTOFF}) {{print$2,$3,$4,$5,"NA","NA",$6}}}}' {input} """
         """ | sort -k1,1 -k2,2n > {output.manta_dup}"""
+
+
+rule filter_telomeres:
+    input:
+        ## Define the location of the input file and take the path/extension from the config file
+        script=os.path.join(config["scripts"], "get_rid_bad_segments.py"),
+        bad_list=os.path.join(config["scripts"], "bad_chromosomal_seg_updated_merged.txt"),
```
jashapiro: As per my other suggestion, should this be renamed?

fingerfen: Yes, I will make that change

analyses/copy_number_consensus_call/src/scripts/generate_black_list.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+
+# This script captures the steps to generate the bad_chromosomal_seg_updated_merged.txt 
+
+# There are two components to this file the "IGLL regions, centromeric and telomeric regions" and the "segmental duplication regions"
+# 1) The IGLL regions, centromeric and telomeric regions are generated from the practice described by Kai Wang at his PennCNV website http://penncnv.openbioinformatics.org/en/latest/misc/faq/
+# 2) The segmental duplication are downloaded from UCSC genome browser. The segmental duplication with 95% identity was downloaded and merged.
+     The tracked used for this is here https://genome.ucsc.edu/cgi-bin/hgTrackUi?hgsid=783211655_qnHWS1w9VWkUebtWb7482jKHTMF8&c=chr1&g=genomicSuperDups
```
jashapiro:
```suggestion
#     The track used for this is here https://genome.ucsc.edu/cgi-bin/hgTrackUi?hgsid=783211655_qnHWS1w9VWkUebtWb7482jKHTMF8&c=chr1&g=genomicSuperDups 
```

Issue #18, opened 2019-07-12
Planned Analysis: Survival Analysis across PBTA


PichaiRaman:
Comparison of genomic/transcriptomic features and derivates such as TMB, pathway activity, etc... to survival. Determine markers that could add prognostic values to different PBTA cancer types. KM Plots for significant and interesting features.

alexfelmeister:
I maybe able to help out here. I've done this with CBTTC data for demonstration purposes as well as done the back and forth about how to get this from that data structure.

jharenza:
@alexfelmeister - Thanks for volunteering! We have to work with @allisonheath to hammer down survival before adding it to the clinical file, so I have marked this issue as `blocked` for now.

jharenza:
survival has been added to the clinical file with `release-v4-20190909`

PichaiRaman:
I would like to potentially tackle this. I will work on not only categorical features such as mutations or fusions but was also going to look at RNA-Sequencing data and any other continuous measure. I have some experience with this and have some code pretty much ready to go 

https://www.sciencedirect.com/science/article/abs/pii/S2210776218304897

I was also thinking of potentially using other derived data such as Gene Set Scores, TMB, and immune cell type enrichment scores.

jaclyn-taroni:
@PichaiRaman do you have an idea of when you expect to file the first pull request for this? 

I also wanted to ask if the code you mentioned above was publicly available. If so, we could potentially have someone else get started on adapting it for this project and have you review. Let me know what you think!

jharenza:
@PichaiRaman should we un-assign you from this?

cansavvy:
I'm going to start looking into this and planning for it this week.

cansavvy:
A few questions for me to get this analysis going: 

### 1) What basic analyses and output plots/tables do we need for this issue? Can I get a list of needed plots/tables we need? 

It seems we will want some Kaplan-Meier survival probability curves like these [from Pajtler et al, 2015](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4712639/). Is this a good example to follow? I can build it in such a way that we can split it up by `short_histology` for now, but then when we get molecular sub-typing information, we can graph the data by their subtypes.

![Screen Shot 2019-12-04 at 1 46 55 PM](https://user-images.githubusercontent.com/23458084/70171855-8ed19d80-169d-11ea-9379-71d3d3194b73.png)

[...]

PR #331, opened 2019-12-12
Survival Analysis Examples Notebook

cansavvy:
This is a basic set up for survival analysis that is meant to be a sort of template that can be molded to fit the various data types we will want to test as predictors of survival. 

Here's a link to the html output for this notebook if you'd like to look at this in a rendered form: 
https://cansavvy.github.io/openpbta-notebook-concept/survival-analysis/survival-analysis_template.nb.html

#### What was your approach?

Following @jaclyn-taroni 's [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/18#issuecomment-562122042) I made this notebook have three different survival analyses representing different kinds of survival questions we might ask. 

- Kaplan-Meier - for illustrating probability of survival over time. 
- Log Rank - for testing the differences between two different curves
- Cox regression - for continuous variable testing

#18

I have never done survival analysis, and I don't know if my descriptions or use of the models is accurate. If this is to be an example analysis, I want to make sure we aren't having analysis that follow this model to be steered wrong. So if someone with more expertise in these analyses and statistical models can read over what I've written as well as the data handling, that is the main goal here. Also let me know where more explanation or documentation is needed. 

This is a very basic workflow on purpose. The thought is we can get this notebook example out there and as the data types and questions roll in (such as the molecular subtype labels), we can adapt this analysis for those data types and make separate notebooks or wrapper functions where it would be helpful. 

Two tables, and a plot. These are mostly examples of what kind of output these analyses can give. 

These are mostly example analyses, though initial TMB analysis does seem to be fairly predictive of survival. 

#### Pull Request Check List:
* [x] Run a linter
* [x] Double check your paths


jaclyn-taroni:
@cansavvy that notebook link 404s for me

sjspielman:
Overall this looks like a really nice start. The main "must fix" I see is that the plot saved in the chunk 140-142 is not complete. It seems the `surv_plot` object that `survminer::ggsurvplot()` returns is a list of plots, where the `surv_plot[[1]]` are the survival curves but what is saved is only `surv_plot[[2]]`. This probably happened because no object was specifically specified in line 141, so `ggsave` found the most recently plotted object which is only the 2nd index in the figure object. Probably save this as a grid of plots -use [`cowplot`](https://github.com/wilkelab/cowplot) or [`patchwork`](https://github.com/thomasp85/patchwork) (which is finally on CRAN!).

Some other coding aspects to consider (@jaclyn-taroni, any feedback on whether these notes are reasonable/excessive/etc?) - 
+ Naming R chunks for ease of reference going forward (eg, I could refer to the plot chunk more easily, and lines below)
+ Definitely worth looking into how flexible the `ggsurvplot` function is with manipulating further aspects of the plot (I can also come back to this). Or, clean up some of the variables first - captions are not ideal aesthestically, etc.  
+ As this comes together further, functions wrapping plotting could be useful to help others specify custom plot features (x-axis limits, etc).
+ When saving tables, it would be good practice to ensure that columns are usable without much manipulation. Eg, the headers for file `cox_regression_tmb.tsv` are going to be a **pain** to deal with later if anybody would like to import it.
+ Saving models, I have confirmed that `broom` functions will handle the output. Will be cleaner to replace the manual data frame creation in chunk starting line 172 with calls to `broom:tidy()` and `broom::glance()`, eg... (will require lib import)
```
> broom::tidy(logrank_fit)
# A tibble: 2 x 4
  reported_gender     N   obs   exp
  <chr>           <dbl> <dbl> <dbl>
1 Female           1026   284  285.
2 Male             1271   342  341.
> broom::glance(logrank_fit)
# A tibble: 1 x 3
  statistic    df p.value
      <dbl> <dbl>   <dbl>
1   0.00425     1   0.948
```
Similarly, for the `cox_table`:
```
> broom::tidy(cox_fit)
# A tibble: 1 x 7
  term  estimate std.error statistic   p.value conf.low conf.high
  <chr>    <dbl>     <dbl>     <dbl>     <dbl>    <dbl>     <dbl>
1 tmb     0.0506    0.0126      4.02 0.0000573   0.0260    0.0753
> broom::glance(cox_fit)
# A tibble: 1 x 15
      n nevent statistic.log p.value.log statistic.sc p.value.sc statistic.wald
  <int>  <dbl>         <dbl>       <dbl>        <dbl>      <dbl>          <dbl>
1   795    226          9.27     0.00233         21.7 0.00000313           16.2
# … with 8 more variables: p.value.wald <dbl>, r.squared <dbl>,
#   r.squared.max <dbl>, concordance <dbl>, std.error.concordance <dbl>,
#   logLik <dbl>, AIC <dbl>, BIC <dbl>
```

Some science aspects for someone w/ more/any survival analysis background - 
+ Unclear to me why `kap_fit` has NA's for an upper bound. How (un)expected output like this is?
+ How much does having balanced data matter for these analyses, eg number of `NA`s in overall survival (presumably those individuals did not pass away)? Are there specific statistical approaches to take when data is highly imbalanced, and are we in that circumstance?

komalsrathi:
@cansavvy In addition, can you also add the following:

1. Add the argument `pval = TRUE` to `survminer::ggsurvplot` to show the p-value on the plot.
2. Extract p.value from `survival::survdiff` using `pchisq(logrank_fit$chisq, df=1, lower=F)` and add to `logrank_table`

cansavvy:
>The main "must fix" I see is that the plot saved in the chunk 140-142 is not complete. It seems the surv_plot object that survminer::ggsurvplot() returns is a list of plots, where the surv_plot[[1]] are the survival curves but what is saved is only surv_plot[[2]]. This probably happened because no object was specifically specified in line 141, so ggsave found the most recently plotted object which is only the 2nd index in the figure object.

Thanks for this @sjspielman ! This was something I saw, tried to fix, but apparently didn't fix it entirely!

cansavvy:
> Definitely worth looking into how flexible the ggsurvplot function is with manipulating further aspects of the plot (I can also come back to this). Or, clean up some of the variables first - captions are not ideal aesthestically, etc. 

Yes I am concerned about this since it does *not* appear to take extra ggplot2 layers with a `+`.

cansavvy:
>When saving tables, it would be good practice to ensure that columns are usable without much manipulation. Eg, the headers for file cox_regression_tmb.tsv are going to be a pain to deal with later if anybody would like to import it.

Good call, didn't think about that the object's names are gonna be a pain later.

cansavvy:
@sjspielman Great call on the `broom`! It worked like magic! @komalsrathi and @sjspielman I tried to incorporate your suggestions let me know what you think. Would still like more feedback from someone on the scientific validity of this (including @sjspielman's [noted scientific questions above](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/331#issuecomment-565238772) in the last part of her comments.

komalsrathi:
@cansavvy We regularly use Kaplan Meier to account for categorical and Cox Regression for continuous covariates in survival analyses so your approach looks correct to me.

komalsrathi:
@cansavvy @jaclyn-taroni What's the difference between `Not Available` and `NA` in the gender column? Shouldn't they all be `NA`? The survival package automatically removes `NAs` and it is advisable to remove any rows corresponding to missing values before survival analysis.

jharenza:
@cansavvy - to answer the question from @sjspielman regarding `surv-fit`, it looks line NA == infinity in the upper bound, and this happens when the data is skewed. Some explanations [here](http://sphweb.bumc.bu.edu/otlt/MPH-Modules/BS/R/R7_LogisticRegression-Survival/R7_LogisticRegression-Survival4.html). 

Separate question: I think we should be using `germline_sex_estimate` for anything dealing with sex, rather than `reported_gender`. I realize these are performed on germline WGS, and you are using RNA-Seq here, but we should be able to map sex to the patient level.

cansavvy:
> @cansavvy @jaclyn-taroni What's the difference between `Not Available` and `NA` in the gender column? 

I don't know what the difference between `Not Available` and `NA` is. @jharenza do you know? 

> Separate question: I think we should be using germline_sex_estimate for anything dealing with sex, rather than reported_gender. I realize these are performed on germline WGS, and you are using RNA-Seq here, but we should be able to map sex to the patient level.

I am not using any RNA-seq data in these samples. Just TMB so far. Do we still prefer using the estimates over the `reported_gender`?

cgreene:
> Do we still prefer using the estimates over the reported_gender?

Yes - this is likely to be preferred here.

jharenza:
@cansavvy Ahhh, I think `Not Available` and `NA` come from KidsFirst ingestion. In my mind, I agree with @komalsrathi they should be the same, but I think if the field was left blank, it became `NA`, and otherwise, it was reported as `Not Available`. We can harmonize those.

Sorry, writing too fast - yes, I think we provided `germline_sex_estimate` as more of truth over reported data. Should we instruct people to use this in the README @jaclyn-taroni or should we provide some other document at a patient-level?

cansavvy:
Okay. Will streamline the `NA`s and change to `germline _sex_estimate`.

sjspielman:
I wonder whether it makes sense to rename this analysis to `survival_analysis_template` - the code is meant as a reference for forthcoming analyses, no?

@cansavvy, I also don't think cleaning up the ggplots is necessary at this stage - as long as we have a standard protocol for running a survival analysis, the plot can be prettied up later. Don't go down the `ggplot` rabbit hole yet ;)

jaclyn-taroni:
Does the formatting as of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/331/commits/64ef4c14397955ad0b29b809e26084306fd34e35 look better @sjspielman? Anything else you would like to see changed? I made a change to make this work in the context of CI with https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/331/commits/3ee5004b0c1adbeeb2bd05b0c3c3bb79a91ca062.

sjspielman:
Looks good to me, the only question is whether we want to extend to a function now per Issue #18 ? Or is this overkill at this time?

jaclyn-taroni:
> Looks good to me, the only question is whether we want to extend to a function now per Issue #18 ? Or is this overkill at this time?

I think a new pull request adding that change is more appropriate. Can you approve this pull request? It is still marked as changes requested.

jaclyn-taroni:
Thanks @sjspielman !

Issue #343, opened 2019-12-17
Data release documentation checklist


jaclyn-taroni:
We now have (#336) or are proposing (#342) multiple files that contain information about data files in addition to the release notes that are typically associated with a data release. I am proposing the following check list that should accompany data release pull requests to make it easier to get all of our ducks in a row for each data release

- [ ] Is the table in `doc/data-description.md` up to date?
- [ ] Is `doc/data-format.md` up to date?
- [ ] Is `doc/release-notes.md` up to date?

jharenza:
submitted PR #348 for this

jaclyn-taroni:
Closed via #350

PR #347, opened 2019-12-17
Release V12 data

jharenza:
### Purpose/implementation Section
#326 - Release V12 data

### release-v12-20191217
- release date: 2019-12-17
- status: available
- changes:
  - Add `data-file-descriptions.md` with data release to better track file types, origins, and workflows per [#334](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/334) and [#336](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/336)
  - Add stranded RNA-Seq for 23 PNOC samples and 21 CBTTC samples previously sequenced using a polyA library prep. Files updated:
    - pbta-fusion-arriba.tsv.gz
    - pbta-fusion-starfusion.tsv.gz
    - pbta-gene-expression-rsem-tpm.stranded.rds
    - pbta-gene-expression-rsem-fpkm.stranded.rds
    - pbta-isoform-expression-rsem-tpm.stranded.rds
    - pbta-isoform-counts-rsem-expected_count.stranded.rds
    - pbta-gene-counts-rsem-expected_count.stranded.rds
    - pbta-gene-expression-kallisto.stranded.rds
    - pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
  - Add recurrently-fused genes by histology and matrix of recurrently-fused genes by biospecimen from [fusion filtering and prioritization analysis](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering)
  - Update consensus TMB files and MAF [#333](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/333)
  - Add RNA-Seq [collapsed matrices](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/287) - wrong files (tables of transcripts removed) were included with [V10](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/273)
  - Rename `WGS.hg38.mutect2.unpadded.bed` to `WGS.hg38.mutect2.vardict.unpadded.bed` to better reflect usage
  - Update `pbta-histologies.tsv` to add new RNA-Seq samples listed above, [#222 harmonize disease separators](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/222), and reran [medulloblastoma classifier](https://github.com/d3b-center/medullo-classifier-package) using V12 RSEM fpkm collapsed files
[...]

PR #348, opened 2019-12-17
Data release checklist

jharenza:
Not a question, but improving PR checklist for data releases, suggested by @jaclyn-taroni 

#### What was your approach?
Update PR template

#343 

NA

Just added this at the end, is that ok?

Yes

NA

NA


doc/release-notes.md
```diff
@@ -1,5 +1,79 @@
 # release notes
 ## current release
+### release-v12-20191217
```
jaclyn-taroni: This appears to be a duplicate change from #347

doc/data-files-description.md
```diff
@@ -15,7 +15,7 @@ This document contains information about all data files associated with this pro
 
 
 
-### current release (release-v11-20191126)
+### current release (release-v12-20191217)
```
jaclyn-taroni: I think this is, too -- I'm going to revert these duplicate changes so that only the data release check list is being added.

jaclyn-taroni:
Actually think it's easier to close this and file a new one with just the checklist change. I will do that now.

PR #353, opened 2019-12-18
Survival Analysis: Functionalize it

cansavvy:
Survival analysis, but no actual change to the science is happening here, more changes to the way the example is shown and adding a wrapper function for future convenience. 

#### What was your approach?
I functionalized things and made the notebook explain how to use the function. 

Based on the discussion on #18  

What parts of the function do you like or not like? 
What parts of the function might be prone to breaking, shooting ourselves in the foot or just plain confusion? 

### Results
The results are the same as before, what is different is how the code examples are set up. 

This was done previously. 

## Pull Request Check List:
* [x] Run a linter
* [x] Set the seed (if applicable) I did not, but I can if we think that this is using random seeds. 
* [x] Double check your paths


analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,117 @@
...
+# C. Savonen for ALSF - CCDL
+#
+# 2019
+#
+# Install the packages we need
+if (!("survival" %in% installed.packages())) {
+  install.packages("survival")
+}
+
+if (!("survminer" %in% installed.packages())) {
```
sjspielman: Duplicated lines from 7-10

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,117 @@
...
+  install.packages("survminer")
+}
+
+# Attach this package
+library(survminer)
+
+# Magrittr pipe
+`%>%` <- dplyr::`%>%`
```
sjspielman: General question - is there differing behavior in `%>%` between packages? Curious as to the specific reasons for specifying this in all scripts, I'm unaware.

jaclyn-taroni: > General question - is there differing behavior in `%>%` between packages?

Not that I am aware of.

> Curious as to the specific reasons for specifying this in all scripts, I'm unaware.

We do this to avoid `library(dplyr)` because of the namespace collisions (although it doesn't _usually_ come into play).

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,117 @@
...
+  # A list with three objects: 1) the original model fit object,
+  #                            2) the summary table
+  #                            3) the original data.frame
+
+  # List the tests
+  supported_tests <- c("kap.meier", "log.rank", "cox.reg")
+
+  # Check that it is a supported test
+  if (!(test %in% supported_tests)) {
+    stop(paste0(test, "is not a supported test."))
```
sjspielman: Edit this sanity check as - `paste(test, "is not a supported test. Please specify one of the following:", paste(supported_tests, collapse=", "))`

analyses/survival-analysis/util/survival_models.R
sjspielman: Makes sense, better safe than sorry. Thanks

analyses/survival-analysis/util/survival_models.R
cansavvy: They are different packages. But I'm pretty sure `survminer` will install `survival` as a dependency so I'll delete `survival`

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  # Format the OS_status variable for the survival model functions
+  metadata$OS_status <- factor(metadata$OS_status, levels = c("LIVING", "DECEASED"))
```
sjspielman: This feels like a very strong assumption, to require specific column names in `metadata`. A more flexible solution will use tidyeval where users can provide the column names containing this information. I'd recommend using named arguments for this.
Depending on the `rlang` version we have associated with this project's docker image, there are two solutions:
1. If `rlang >=0.40`: [https://rdrr.io/github/tidyverse/rlang/f/NEWS.md](https://rdrr.io/github/tidyverse/rlang/f/NEWS.md). See section entitled "Interpolate function inputs with the curly-curly operator"
2. If earlier `rlang`, it's more annoying but still very doable (see bits about enquo) [https://tidyeval.tidyverse.org/dplyr.html](https://tidyeval.tidyverse.org/dplyr.html)

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,114 @@
...
+  metadata$OS_status_num <- as.numeric(metadata$OS_status)
+
+  # If other ind_data has been supplied, attempt to join it
+  if (!is.null(ind_data)) {
+    # Make it so the column names are automatically matching no matter what
+    colnames(ind_data) <- gsub("Tumor_Sample_Barcode|sample", "Kids_First_Biospecimen_ID", colnames(ind_data))
```
sjspielman: Again, a tidyeval scenario would be good. Any circumstance where column names are assumed gets dangerous. Users should specify column names as named input arguments to avoid any accidental conflict.

analyses/survival-analysis/util/survival_models.R
cansavvy: I agree that this is a strong assumption if we were using this for other data, but for the given project and metadata file I don't *think* these column names will be changing. That being said, I'm all for making things more generalizable/flexible so I'll checkout these sources!

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,114 @@
...
+  ind_var <- paste0(ind_var, collapse = "+")
+
+  # Piece together a model
+  model <- paste("survival::Surv(OS_days, OS_status_num)",
+    ind_var,
+    sep = " ~ "
+  )
+  
+  # We'll need to grab this later for ggsurvplot
+  assign("model", model, envir = .GlobalEnv)
```
sjspielman: So, is the expectation that a variable named `model` will be propagated to the global environment and then whoever uses the function should be able to go forth with their plotting? I would think it's safer to return this from the function - is that not possible?
If it's not possible to return the model object, we shouldn't send a variable called `model` into the global environment. Too much potential to interfere with existing objects, since it's such a general name.  I'd change the global name to `survival_analysis_model`? Or, could users pass in as a named argument what they want the output model to be named?

cansavvy: Yes. This is my janky work around for trying to still get `ggsurvplot` to work. I'm not happy with the current solution (for the list of problems you've noted) but I was trying to figure out what is a better solution that `ggsurvplot` will still be able to use. I can have it save a `model` as another item to the list that is returned if that is preferred and see if I can get `ggsurvplot` to work with that.

cansavvy: Alright. Got a better solution for this. Took a little more digging into what `survplot` was looking for me to figure out how to address it.

cansavvy:
@sjspielman !! Thanks for the tidyeval tips and tricks! I've incorporated what I think was your suggestion for increasing flexibility for column names. Let me know what you think regarding my changes.

analyses/survival-analysis/util/survival_models.R
```diff
@@ -0,0 +1,192 @@
...
+    
+    # If not all the columns are found, stop
+    if (!all(found_cols)) {
+      stop(cat("The following column names specified: \n", 
+               paste(needed_cols[which(found_cols)], collapse = "\n"), 
+               "\n ...were not found in the specified data.frames.", 
+               "Check your `metadata_sample_col`, `ind_data_sample_col`, and `ind_var` arguments."))
+    }
+    # Join this ind_data to the metadata
+    metadata <- dplyr::inner_join(metadata, ind_data,
```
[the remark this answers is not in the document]
sjspielman: Ok, looks you got it with a different strategy, works just as well!

PR #354, opened 2019-12-18
Make filenames consistent for analysis files included in data download

jaclyn-taroni:

The purpose of this PR is to change the filenames of the putative oncogenic fusion file and the consensus mutation MAF + both TMB files to match what is included in the data download. This both makes it easier for those involved in the data release process and is less likely to confuse contributors.


jaclyn-taroni:
See #359 😄

PR #356, opened 2019-12-18
CI files: Account for v12 release changes

### LINKED ISSUE #337: Generate CI subset files for v12
Depends on #326 - we'll need to generate new subset files for continuous integration. Because of the changes planned for v12 or _discussed_ for v12 (#334), there are a few considerations:

* The GISTIC files are brand new. We will likely need to write new code:
  - [ ] To extract biospecimen identifiers ([this function](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/09477cafa880856de001f4844aec39fb61c73f5e/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L34))
  - [ ] To subset them ([this function](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/09477cafa880856de001f4844aec39fb61c73f5e/analyses/create-subset-files/02-subset_files.R#L47))
* There may be a new Markdown that needs to get moved to the subset folder. https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/09477cafa880856de001f4844aec39fb61c73f5e/analyses/create-subset-files/create_subset_files.sh#L38
* We currently enrich for poly-A samples (starts [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/09477cafa880856de001f4844aec39fb61c73f5e/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L154) but is throughout [`analyses/create-subset-files/01-get_biospecimen_identifiers.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/09477cafa880856de001f4844aec39fb61c73f5e/analyses/create-subset-files/01-get_biospecimen_identifiers.R)). There will still be analysis modules that use poly-A files, so this should likely still be performed. However, we may want to keep an eye on including samples that are assayed with both strategies in the CI files.

jaclyn-taroni:

These changes to the CI subset file generation account for the following v12-related changes:

* Recurrently fused genes by histology table gets copied over
* Recurrently fused genes by sample file gets subset
* GISTIC zip file gets copied over
* Markdown files are tracked in `md5sum.txt`

Closes #337


jaclyn-taroni:
@cansavvy we need to do the following:

1. Checkout the  `v12-data-release` branch of `AlexsLemonade` which makes the required changes to the data download step. 
2. Run `bash download-data.sh` to obtain the v12 data while on that branch
3. Add remote `jaclyn-taroni` and checkout this branch `337-v12-ci` which contains the required changes to generate the v12 files for CI.
4. Run `./analyses/create-subset-files/create_subset_files.sh`.
5. Snag the files in `release-v12-20191217`; they are what we need for CI.
6. Commit + push the changes to `analyses/create-subset-files/biospecimen_ids_for_subset.RDS`

Because we'll be subsetting VarDict, this will need to be run with 128 GB of RAM.

jaclyn-taroni:
I've changed the files in S3, so I will now rerun this workflow to get an idea of what breaking changes may have been introduced.

analyses/create-subset-files/01-get_biospecimen_identifiers.R
```diff
@@ -185,7 +185,18 @@ files_to_subset <-
   files_to_subset[which(!(files_to_subset %in% c(polya_rsem_files[-1],
                                                  stranded_rsem_files[-1])))]
 
-# TODO: REMOVE THIS once you are no longer testing locally!
+# drop recurrently fused genes by histology file -- it is small enough to
+# include the entire thing
+files_to_subset <-
+  files_to_subset[-grep("fused-genes-byhistology", files_to_subset)]
```
cansavvy: Hard to tell without running it but I think you could also make this this way: 
```suggestion
grep("fused-genes-byhistology", files_to_subset, invert = TRUE, value = TRUE) 
```

analyses/create-subset-files/create_subset_files.sh
```diff
@@ -40,18 +40,26 @@ Rscript --vanilla 02-subset_files.R \
 # histologies file
 cp $FULL_DIRECTORY/pbta-histologies.tsv $SUBSET_DIRECTORY
 
+# recurrently fused genes by histologies file
+cp $FULL_DIRECTORY/pbta-fusion-recurrently-fused-genes-byhistology.tsv $SUBSET_DIRECTORY
```
cansavvy: It feels like there should be a hyphen between `by` and `histology` but if this is the file name, then this is what it is.

PR #357, opened 2019-12-19
CNV consensus (5 of 6):Consensus call

fingerfen:

Implement the consensus calling part of the pipeline

issue #128 

The compare_variant_calling_updated.py script. 

I think this script is the bottle neck of the pipeline. As discussed before, I could go back to update this to use Dictionary after the pipeline is in. A singular call only take 1-2 second to run, but it takes a while with 940 samples


analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,257 @@
...
+
+        ## Define a variable to store the list of consensus CNVs
+        consensus_list = []
+
+        ## For every CNVs in list1, we look through the ENTIRETY of list2 and perform actions if 2 CNVs overlaps
+        ## Could probably speed this up by using a dictionary.
+        for m in list1:
+
+            ## Assign the start_pos, end_pos, and copy number of list1 to variables
+            start_list1 = int(m[1])
```
jashapiro: I am suggesting we do this for clarity and parallelism
```suggestion
            chr_list1 = m[0]
            start_list1 = int(m[1])
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,257 @@
...
+            list2_start_list = []
+            list2_end_list = []
+            list2_chr_str_end = ''
+
+
+            ## For every of the CNV in list one, we look over ALL CNVs in list 2 and make comparisons
+            for n in list2:
+
+                ## Assign start_pos, end_pos, and copy number of list2 to variables
+                start_list2 = int(n[1])
```
jashapiro: see above
```suggestion
                chr_list2 = n[0]
                start_list2 = int(n[1])
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,257 @@
...
+                end_list2 = int(n[2])
+                cnv_list2 = n[3]
+
+                ## Make sure end > start
+                if start_list2 > end_list2:
+                    start_list2, end_list2 = end_list2, start_list2
+
+                ## For the current CNV in list1, if it overlaps with any CNV from list2
+                if m[0] == n[0] and start_list1 <= end_list2 and end_list1 >= start_list2:
```
jashapiro: using the changes suggested above to make what is going on here more transparent.
```suggestion
                if chr_list1 == chr_list2 and start_list1 <= end_list2 and end_list1 >= start_list2:
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,257 @@
...
+            sys.stderr.write('$$$ Created new file succesfully\n')
+
+        ## Open up the file to write in it
+        with open( list_of_output_files[i], 'w') as file:
+            ## Loop through the output file and print line by line
+            for k in ival:
+
+                ## Add the sample name and file name to each line
+                single_name = list_of_output_files[i].split('/')[-1]
+                sample_name = args.sample
```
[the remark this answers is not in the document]
fingerfen: My thought is that people might be using the final consensus file for different purposes. Maybe people put CNVs from different files together. So with this, even if different CNVs from different files are put together, we still know which CNVs are from which files/samples. It is just an extra column IN CASE anybody needs it. I can definitely take it out if you don't think it is necessary.

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
fingerfen: I agree! Thanks for the suggestion!

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,257 @@
...
+        ## Add the consensus list into a final list that hold the
+        ## content of 3 files - manta_cnvkit , manta_freec, and cnvkit_freec
+        fin_list += [consensus_list]
+
+## Print the output into files
+for i,ival in enumerate(fin_list):
+    try:
+        ## Attempt to create a the file, if the file already exists, throw ERROR
+        with open( list_of_output_files[i] , 'x') as file:
+            sys.stderr.write('$$$ Created new file succesfully\n')
```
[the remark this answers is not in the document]
fingerfen: I just made a few small files and did a test. It seems that Snakemake DOES delete the output file, then regenerate that file if the input is at a later time stamp than the output. 

I put this step in so that the script doesn't override any info that the user might not want to be deleted. But I think in the context of this pipeline, we could just add a step to override the output file. What do you think?

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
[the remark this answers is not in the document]
jashapiro: Makes sense. I just thought the combination of sample id and callers would probably be sufficient. But I am fine with leaving it as is.

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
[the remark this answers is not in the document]
jashapiro: In that case I don't think it matters much. I would personally have an option for overwriting, but I don't feel strongly.

fingerfen:
I am implementing the suggested changes and will make an update shortly. Sorry for the delay.

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -42,92 +42,55 @@
...
-list_of_list = []
-
-
-## Loop through the file paths and read in their contents to store in a list
-for i in list_of_files:
+########################## DEFINE FUNCTIONS ###############################
 
+## Function to read content of input files
+def read_input_file(input_file):
     ## Check to see if the CNVs bed file is empty, if so, make an empty variable
```
jashapiro: Before the comments that start to describe individual elements, you should add an overall description of the function here. The python convention is to use a triple quoted line or lines immediately below the def line that describes what the function does, along with expected inputs and outputs (and their types). See more here: https://www.python.org/dev/peps/pep-0257/

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -42,92 +42,55 @@
...
-## We have 3 items in list_of_list. Each item is the content of an input file
-## Make a list to store the final files
-fin_list =[]
+    return content_dict
 
-## Loop through list_index
-for j, jval in enumerate(list_index):
-    for k,kval in enumerate(list_index[j+1:]):
+## Function to get consensus CNVs.
+def generate_consensus(list1_name,list1,list2_name,list2):
```
jashapiro: Docstring here too

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -145,115 +108,183 @@
...
+                        overlap_chrom = [chr_list1,str(chrom_start),str(chrom_end),'NULL',str(cnv_list1).strip(','),list2_chr_str_end.strip(','),m[-1]]
+
+                    ## Add the formatted consensus CNV into a growing list of consensus CNVs
+                    consensus_list = consensus_list + [overlap_chrom]
+
+    return consensus_list
+
+
+## Function to save output files
+def save_to_file(output_file_content, output_path, sample_name):
```
jashapiro: One more docstring here!

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -145,115 +108,183 @@
...
+    ## Open up the file to write in it, the 'w' option overwrites the file if the file exists.
+    with open( output_path, 'w') as file:
+        ## Loop through the output file and print line by line
+        for k in output_file_content:
+
+            ## Add the sample name and file name to each line
+            single_name = output_path.split('/')[-1]
```
jashapiro: Move this line outside the loop, as the path never changes.
you can also replace it with the safer `os.path.basename(output_path)` to avoid separator variation.

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -145,115 +108,183 @@
...
+            file.write('\t'.join(k) + '\t'+ sample_name + '\t' + single_name +'\n')
```
jashapiro: Minor style point.. I would probably write this as, but your version is perfectly clear.
```
file.write('\t'.join(k.extend([sample_name, single_name]) + '\n')
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -145,115 +108,183 @@
...
+
+input_file_names = list(dict_of_input_content.keys())
+
+## Put the output file paths into their own lists that is to be iterated over
+## This order is important
+output_list = [args.manta_cnvkit, args.manta_freec, args.cnvkit_freec]
+
+## Make a list of iteraby index for the MAIN for-loop below
+## This give a list as followed: [0,1,2]
+list_index = list(range(0,len(dict_of_input_content)))
```
jashapiro: Rather than doing these lists and indexes, we could simplify this to something like this to keep it simpler
```
caller_pairs = [('manta', 'cnvkit'), ('manta', 'freec'), ('cnvkit', 'freec')]
```

Then you can use a single loop below, something like:
```
for caller1, caller2 in caller_pairs:
    list1 = dict_of_input_content[caller1]
    list2 = dict_of_input_content[caller2]
    ....
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,290 @@
...
+        list2 = dict_of_input_content[input_file_names[kval]]
+
+        ## Call the "generate_consensus" function to get consensus CNVs
+        consensus_list = generate_consensus(input_file_names[jval],list1,input_file_names[kval],list2)
+
+
+
+        ## Add the consensus list into a final list that hold the
+        ## content of 3 files - manta_cnvkit , manta_freec, and cnvkit_freec
+        fin_list += [consensus_list]
```
jashapiro:
```suggestion
        fin_list.append(consensus_list)
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+
+## Print the output into files
+for i,ival in enumerate(fin_list):
```
jashapiro: Python style is to avoid iterating with indexes when you can iterate on the actual items, and `zip` is very handy for that
```suggestion
for content, outfile in zip(fin_list, output_list):
    save_to_file(content, outfile, args.sample)
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
fingerfen: Thank you for the suggestion @jashapiro. The reason why I did it my way was because I was trying to avoid hardcoding as much as possible. My intention was that when someone want to change the code, they can go in and be able to change the code more easily. 

I guess what I am struggling with is when to hardcode this information and when not to. However, I do agree that your method is simpler and definitely more clear. I will change the script to reflect that.

cgreene: In this case it might be even better to try itertools.combinations. Does `itertools.combinations(['manta', 'cnvkit', 'freec'], 2)` do what you want in terms of getting caller pairs that you can iterate though?

I also wonder if you could use `itertools.combinations` over dict_of_input_content's keys to get something that's both clear and more general. I don't quite understand what jval and kval are so I can't provide feedback that's highly specific to this code.

cgreene: Link for convenience to itertools.combinations: https://docs.python.org/3/library/itertools.html#itertools.combinations

fingerfen:
Thanks @jashapiro and @cgreene for the helpful suggestions as always. Since it's winter break, I have more time to work on this so I want to get it to a point where the scripts are clean, straightforward, and that everyone is happy with. Thus I am happy to implement any changes that you think will make the code better. I am also learning a lot by doing it. 

With this new commit, I believe I have addressed all of your comments. Please have a look. 

And thanks @jashapiro for letting me know about your availability. I only have one more step to PR and then probably another one to add in a header(column names) for these bed files. After that, this should be done! I should be readily available to address any problems over this winter break. 

Happy holidays to everyone and safe travels! :)

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
jashapiro: The only potential issue I see with this is if the order is not as expected: some of the surrounding code (in other scripts/snakefile) does depend on filenames being as expected (with the expected content!).

jashapiro: > Thank you for the suggestion @jashapiro. The reason why I did it my way was because I was trying to avoid hardcoding as much as possible. My intention was that when someone want to change the code, they can go in and be able to change the code more easily.

I think this is a good goal, but given that there is already some hard-coding in here (notably the arguments), it is going to be hard to get rid of it all. 

If removing hard coding were the full goal, I would probably simplify this script to take in two files and output the merged file, then move the combinations logic to snakemake. That would lose some efficiency because you would have to read in data more than once, but it is easier to extend. Having already moved the main merge logic to a function, this would be easy to do in the future if desired.

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,315 @@
...
+freec_dict = read_input_file(args.freec)
+
+## Put the input dictionaries into a bigger dictionary
+dict_of_input_content = {'manta':manta_dict, 'cnvkit':cnvkit_dict, 'freec':freec_dict}
+
+input_file_names = list(dict_of_input_content.keys())
+
+## Put the output file paths into their own lists that is to be iterated over
+## This order is important
+output_list = [args.manta_cnvkit, args.manta_freec, args.cnvkit_freec]
```
jashapiro: Since the order is important so that file names line up with the combinations, lets move the caller pairs up, then use that to make sure the output list is in the same order we will generate later.

```suggestion
## Make a list for pairs of callers to make consensus for
caller_pairs = list(itertools.combinations(['manta', 'cnvkit', 'freec'],2))

# generate output list in the same order
output_files = [getattr(args, "_".join(callers) for callers in caller_pairs]
```

analyses/copy_number_consensus_call/src/scripts/compare_variant_calling_updated.py
```diff
@@ -0,0 +1,315 @@
...
+args = parser.parse_args()
+
+
+## Read in the input files as dictionaries
+manta_dict = read_input_file(args.manta)
+cnvkit_dict = read_input_file(args.cnvkit)
+freec_dict = read_input_file(args.freec)
+
+## Put the input dictionaries into a bigger dictionary
+dict_of_input_content = {'manta':manta_dict, 'cnvkit':cnvkit_dict, 'freec':freec_dict}
```
jashapiro: Thinking a bit more on the idea of easy extensibility, how about something like this:
```
input_callers = ['manta', 'cnvkit', 'freec']
input_content = dict()
for caller in input_callers:
    input_content[caller] = read_input_file(getattr(args, caller))
```
Then you don't need to separately define `input_file_names` below (delete current line 276)

Issue #365, opened 2019-12-21
Updated analysis: Add README files to any analysis without a README


sjspielman:
#### What analysis module should be updated and why?

The following analysis modules are missing `README.md` documentation files:

+ `copy_number_consensus_call/`
+ `sex-prediction-from-RNASeq/`
+ `comparative-RNASeq-analysis/`
+ `focal-cn-file-preparation/`
+ `sv-analysis/`
+ `ssgsea-hallmark/`
+ `survival-analysis/`
+ `fusion_filtering/`
+ `selection-strategy-comparison/`

#### Why should the module be updated?

Modules should contain clear documentation about specific goals, input, output, and pipelines to ensure reproducibility _and_ to facilitate use by other contributors.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

For all **completed** modules, `README.md` files should explain:
+ The source (authorship) of the analysis
+ The purpose of the analysis
+ The analysis pipeline, if multiple scripts are run for the analysis
+ The output/outcome of the analysis

For all **in progress** modules, `README.md` files should explain, at a minimum, analysis goals and expected output/outcomes.

#### What input data should be used? Which data were used in the version being updated?

N/A

#### When do you expect the revised analysis will be completed?

Pending availability of module authors, as long as completed before publication. This is not strictly urgent, unless these analyses are specifically required for future analyses and added documentation will strongly facilitate.

#### Who will complete the updated analysis?

I have compiled, for each directory missing a `README`, who I _believe_ the authors are, based either on PRs or code comments. In each case it may be more appropriate for others who handled the associated PR (such as @jaclyn-taroni , @jharenza , @jashapiro , @cgreene , etc)  to assist in compiling these details:

+ `copy_number_consensus_call/` : @fingerfen 
+ `sex-prediction-from-RNASeq/`: @bill-amadio 
+ `comparative-RNASeq-analysis/`: @GeoffLyle
+ `focal-cn-file-preparation/`: @jaclyn-taroni 
+ `sv-analysis/`: @yangyangclover 
+ `ssgsea-hallmark/`: @PichaiRaman (although some comments in code indicate maybe @cbethell?)
+ `survival-analysis/`: @cansavvy 
+ `fusion_filtering/`: @kgaonkar6 
+ `selection-strategy-comparison/`: @jashapiro

jaclyn-taroni:
A few notes on a few of the more mature modules:
* Some of these are tracked individually: #319 #320
[...]

PR #366, opened 2019-12-21
Rerun transcriptomic-dimension-reduction and selection-strategy-comparison with v12 data

### LINKED ISSUE #360: Updated analysis: transcriptome dimension + selection strategy for v12

The following two modules should be updated for the same reason:

* [`transcriptomic-dimension-reduction`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/transcriptomic-dimension-reduction)
* [`selection-strategy-comparison`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/selection-strategy-comparison)

#### Why should the module be updated?

We have new stranded RNA-seq samples as of v12.

The filenames have not broken anything, so it is a matter of rerunning these two modules and pushing the changes to GitHub.

This is likely a 1-2 hour fix at most.

@jaclyn-taroni

jaclyn-taroni:

[`release-v12-20191217`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md#release-v12-20191217) includes about 40 stranded RNA-seq samples that were previously sequenced using poly-A library prep. I attempted to rerun 

#### What was your approach?

* For `transcriptomic-dimension-reduction`, I simply reran the module (with the addition of a new color in the color palette to account for the changes in `broad_histology`).
* For `selection-strategy-comparison`, I added a little bit of background to the top of the notebook and added a README per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/365#issuecomment-568191591. I also added a UMAP step looking at only the matched samples (using the `sample_id` column) for untransformed data because it appears that the new stranded samples group with the poly-A samples. That notebook should probably be gone through more carefully before being merged.

Closes #360 - perhaps a separate issue for revisiting the selection strategy comparison should be opened and I should revert https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/d40097c3ad4b65708ba1404391868ca6a8217161.

Can you think of any issues with identifier mapping that could be the underlying issue rather than the data themselves? I didn't dig into this exhaustively, I figured I would get other eyes on it first.

### Results

If we take a look at the UMAP plot for untransformed FPKM data it looks like:

umap_raw

where we can see some stranded samples grouped with the poly-A. And if we look just at the matched samples:

umap_matched_raw

There doesn't appear to be a big difference based on _strategy_ for these samples. And if we look at the dimension reduction plots for the log transformed stranded RSEM data:

Screen Shot 2019-12-21 at 1 39 22 PM

There's a cluster of `Diffuse astrocytic and oligodendroglial tumor`  samples (this is by `broad_histology`; `High-grade glioma` and `Brainstem glioma- Diffuse intrinsic pontine glioma` in `disease_type_new` both map to this) that I suspect are the same stranded samples that cluster with the poly-A samples.


jashapiro:
That is disheartening. I will have a closer look on Monday, but I doubt there is anything particularly wrong there. I would guess that the remaining differences are due to [extraction and library prep](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#nucleic-acids-extraction-and-library-preparation), which I assume the new stranded data did not repeat.

jaclyn-taroni:
I went ahead and reverted https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/d40097c3ad4b65708ba1404391868ca6a8217161 now that we have https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/374 open.

.circleci/config.yml
```diff
@@ -44,7 +44,7 @@ jobs:
 
       - run:
           name: Selection Strategy Comparison
-          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/selection-strategy-comparison/01-selection-strategies.rmd', clean = TRUE)"
+          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/selection-strategy-comparison/01-selection-strategies.rmd', params = list(neighbors = 2), clean = TRUE)"
```
cansavvy: Question does the low number of neighbors make this quick to run?

analyses/selection-strategy-comparison/01-selection-strategies.rmd
```diff
@@ -5,6 +5,9 @@ output:
   html_notebook:
     toc: yes
     toc_float: yes
+date: Sys.Date()
+params:
+  neighbors: 15
```
cansavvy: I'm assuming this is overridden by the command line option. I'm pretty sure this is what's happening but I'm voicing my thoughts so you can correct me if I'm wrong.

analyses/selection-strategy-comparison/README.md
```diff
@@ -0,0 +1,15 @@
+## Explore RNA-seq selection strategies
+
+**Module author:** Joshua Shapiro ([@jashapiro](https://github.com/jashapiro))
+
+This analysis was originally undertaken prior to [`release-v5-20190924`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md#release-v5-20190924), where samples from the two library strategies (stranded, poly-A) were split into separate files.
```
cansavvy: The wording is slightly ambiguous to me so I've written what I think this means. Feel free to edit the wording itself, just wanted clarification on what lead to what. 
```suggestion
This analysis was originally undertaken prior to [`release-v5-20190924`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md#release-v5-20190924). 
Data releases prior to v5 had samples from the two library strategies (stranded, poly-A) in the same file, but due to the findings of this analysis, libraries of different strategies were split into separate files in the subsequent data releases. 
```

analyses/selection-strategy-comparison/README.md
```diff
@@ -0,0 +1,15 @@
...
+This split was necessary because differences were apparent in this analysis.
```
cansavvy: If you take the wording above you can probably delete this sentence. 
```suggestion
```

analyses/transcriptomic-dimension-reduction/util/dimension-reduction-functions.R
```diff
@@ -12,7 +12,7 @@ perform_dimension_reduction <- function(transposed_expression_matrix,
   # Given a data.frame that containes the transposed expression matrix and the
   # name of a dimension reduction method, perform the dimension reduction
   # technique on the transposed matrix
-  # 
+  #
   # Args:
   #   transposed_expression_matrix: Transposed `RSEM` or `Kallisto` expression
   #                                 matrix
```
cansavvy: I can't comment on it directly because it's not a part of the current PR, but the comment says that seed is set to a default of 2019 but this function has no default set.

cansavvy:
There's also a place in an unchanged file where the comment says `parmeter` instead of `parameter`. Not a big deal but I noticed it in `run-dimension-reduction.R` line 148.

.circleci/config.yml
jaclyn-taroni: It makes it possible to run without failure in the poly-A set case.

PR #374, opened 2019-12-24
Updates on selection strategy, and v12 added "stranded" files.

jashapiro:

A quick look at what is going on with the stranded samples added in v12, to try to determine why the new saples were clustering with the poly-A samples from the earlier data.

#### What was your approach?
 I incorporated the additions and changes  into a new file (02-selection-strategies-update.rmd) from https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/d40097c3ad4b65708ba1404391868ca6a8217161 that are part of #366. I then looked for genes that were correlated with cluster assignments based on the UMAP data. 

As an intermediate analysis, I do not expect this to be a long term analysis, so the quality of the plots should be sufficient. But if there is a more refined analysis that people would like to see, this can be extended.

We should discuss the next steps, given that the data do not seem to be what we had hoped to obtain with the re-sequencing. We should definitely seek clarification on the precise methods that were employed in generating these data.

n/a

A notebook and some figures.

The main result is this figure showing expression levels of individual genes, divided by UMAP cluster and colored by library preparation method.

image

The main takeaway is that histones and noncoding RNAs have much lower expression in the cluster that contains poly-A samples, which is as expected. However, the new stranded samples have the same biases, indicating that they may have been subjected to poly-A selection during preparation.


cgreene:
This does look bad. I agree with your assessment that the most likely explanation appears to be that BGI sequenced the RNA with poly-A selection instead of the requested protocol.

If we wanted to confirm, someone from the D3B side with access to the raw data could examine the coverage across transcripts for a few longer genes. I'm guessing it'll be high on the 3' side and drop from there.

PR #382, opened 2019-12-31
Add recurrently fused files to 'modules at a glance' table

jaclyn-taroni:

Here I'm adding two derivative fusion files to the "Modules at A Glance" table because I forgot to ask for that as part of #375. I also corrected some links in the table so they point to the AlexsLemonade `master` branch.


jaclyn-taroni:
This is an approved documentation change, so I will not wait for CI.

PR #380, opened 2019-12-30
run script for classifier

kgaonkar6:
#### What scientific question is your analysis addressing?
Bash script to run classifer for polya and stranded expression data separately

#### What was your approach?
Run polya and stranded collapsed rsem FPKM as input for analyses/tp53_nf1_score/01-apply-classifier.py

#### What GitHub issue does your pull request address?
#165

#### Which areas should receive a particularly close look?
analyses/tp53_nf1_score/01-apply-classifier.py has been reviewed by @gwaygenomics  and @jaclyn-taroni just adding run script for polya and stranded separately

#### Is there anything that you want to discuss further?
All suggestions welcome :)

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What types of results are included (e.g., table, figure)?
tables
 analyses/tp53_nf1_score/results/pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv : 58 polya RNAseq samples
analyses/tp53_nf1_score/results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv : 1014 stranded RNAseq samples

#### What is your summary of the results?
TP53/NF1 inactivation classifier score from RNAseq samples.

#### Reproducibility Checklist

- [X] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [X] This analysis has been added to continuous integration.
- [X] This analysis is recorded in the table in `analyses/README.md`.

PR #385, opened 2019-12-31
Evaluation step TP53 classifier

kgaonkar6:
Evaluation of TP53/NF1 classifier scores 

#### What was your approach?
Use TP53/NF1 inactivation scores generated for polya and stranded (run script for classifier #380)
Identify potential true TP53/NF1 losses from SNV (Get tp53 nf1 alt #381)
Run `02-evaluate-classifier.py` to 
* format SNV calls to binary status of TP53/NF1 mutation status
* use clinical sample_id column to match RNAseq from classifier score dataframe with WGS/WES mutation dataframe
* plot ROC curve for true calls and shuffle data  

#165

In 02-evaluate-classifier.py I'm using sample_id to match RNAseq from the classifier score and WGS/WES from SNV data (TP53/NF1 loss binary status)  does that sound good?

Other than that I've made minimal changes to fit our data and folder structure in get_roc_plot() to read in for stranded OR polya samples datasets and plot polya_NF1.png & polya_TP53.png OR stranded_NF1.png and stranded_TP53.png

Any other plots/tables required here ? We are only plotting ROC for TP53 and NF1 is this PR

Yes

figure

We have ~0.8 AUROC for TP53 classifier in both stranded and polya. 
0.7 and 0.92 AUROC for NF1 in stranded and polya respectively. 

The count table for TP53 loss status in stranded data is (1= has TP53 coding SNV alteration, 0 = doesn't have coding SNV alteration)
0    933
1     81
The count table for NF1 loss status in stranded data is (1= has NF1 coding SNV alteration, 0 = doesn't have coding SNV alteration)
0    978
1     36

The count table for TP53 loss status in polya data is (1= has TP53 coding SNV alteration, 0 = doesnt have coding SNV alteration)
0    30
1    28
The count table for NF1 loss status in polya data is  (1= has NF1 coding SNV alteration, 0 = doesn't have coding SNV alteration)
0    53
1     5


kgaonkar6:
> @kgaonkar6 this looks much better, however still concerned about NF1 stranded, and think the selection of mutations will improve this, see #381.

I've added the newly generated TP53_NF1_snv_alteration.tsv and merged the clinical file here with only the IDs and no clinical information as suggested in #381 

> I also noticed _ras_ scores and _ras shuffle_ scores still being produced from the apply classifier step, and we should remove those, as we will not be using them - i think we should modify the code from this PR: #380
> 
Oh yes I'll remember to edit 01 to remove Ras in another PR and add all the files from this analysis to run script as well.

analyses/tp53_nf1_score/02-evaluate-classifier.py
```diff
@@ -0,0 +1,216 @@
...
+Evaluate Classifier Predictions
+Modified from PDX PPTC Machine Learning Analysis
+https://github.com/marislab/pdx-classification
+Rokita et al. Cell Reports. 2019.
+https://doi.org/10.1016/j.celrep.2019.09.071
+
+Gregory Way, 2018
+Modified by Krutika Gaonkar for OpenPBTA, 2020
+
+In the following notebook I evaluate the predictions made by the NF1 and TP53 classifiers in the input PDX RNAseq data.
```
gwaygenomics: No longer PDX right?

gwaygenomics: probably worth revising this docstring

analyses/tp53_nf1_score/README.md
```diff
@@ -8,6 +8,9 @@ Now published in [Rokita et al. _Cell Reports._ 2019.](https://doi.org/10.1016/j
 In brief, _TP53_ inactivation, _NF1_ inactivation, and Ras activation classifiers are applied to the stranded and polya OpenPBTA RNA-seq data.
 The classifiers were trained on TCGA PanCan data ([Way et al. _Cell Reports._ 2018](https://doi.org/10.1016/j.celrep.2018.03.046), [Knijnenburg et al. _Cell Reports._ 2018.](https://doi.org/10.1016/j.celrep.2018.03.076)).
 See [`01-apply-classifier.py`](01-apply-classifier.py) for more information about the procedure.
+To evaluate the classifier scores [`02-evaluate-classifier.py`](02-evaluate-classifier.py) uses SNV data to identify true TP53/NF1 loss samples and compares scores of shuffled data to true calls and plots ROC curves.
```
gwaygenomics:
```suggestion
To evaluate the classifier scores, we use [`02-evaluate-classifier.py`](02-evaluate-classifier.py) and input SNV data to identify true TP53/NF1 loss samples and compare scores of shuffled data to true calls and plots ROC curves. 
```

jaclyn-taroni:
Thanks for the review @gwaygenomics! To answer your methodological questions:

> Holy smokes! The NF1 classifier is working extremely well for poly-a samples. Perhaps suspiciously well... Is there documentation in this repo how poly-a samples were collected (vs. stranded?). Is this just a matter of sample prep? All other ROC curves are behaving as expected :)

Here's the nucleic acids prep section of the manuscript: https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#nucleic-acids-extraction-and-library-preparation

> Was there any difference in how these lesions were defined poly-a vs. stranded?

How aberrant lesions were defined were the same. Specifically, we filtered the MAF file to only include SNV in coding regions  
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/acc66dc18d5ee41105ca8d1ea6c0edc17831de38/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R#L58

And for _NF1_ specifically - we subsequently filter out any mutations that are classified as silent, classified as missense (because they are not in OncoKB: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/381#issuecomment-570748578), or in introns 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/acc66dc18d5ee41105ca8d1ea6c0edc17831de38/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R#L72

I dug into whether or not the difference could be due to the experimental strategy (e.g., WGS vs. WXS), but all samples with _NF1_ alterations that pass this filter are WGS. However, there are only 4 samples with NF1 alterations in the poly-A dataset. So perhaps that could be what's going on - thoughts?

jharenza:
Thanks for the comments @gwaygenomics! 

We did not use NF1 CNVs here, as we are concerned about using the calls in the absence of consensus calls - #128. Maybe we should hold on this for _NF1_ until that file is released and assess deep loss as an additional true alteration? 

One thing about the polyA/stranded discrepancies, in the current v12 release (#347), the `stranded` is not entirely a pure cohort, in that we discovered our latest batch of `stranded` was _both_ `polyA` and `stranded`. I wonder if this is having an effect...

jaclyn-taroni:
> One thing about the polyA/stranded discrepancies, in the current v12 release (#347), the stranded is not entirely a pure cohort, in that we discovered our latest batch of stranded was both polyA and stranded. I wonder if this is having an effect..

I think the class imbalance (low number of positives) for _NF1_ means that a ROC curve might be the wrong measure/display item here.

> Maybe we should hold on this for NF1 until that file is released and assess deep loss as an additional true alteration?

My preference would be to get this merged and document that we need to make this update.

Issue #251, opened 2019-11-08
Proposed Analysis: Molecularly subtype non-MB/ATRT embryonal tumors


jharenza:
#### Scientific goals

_What are the scientific goals of the analysis?_
1) Subtype non-MB embryonal tumors (non-medulloblastoma/ATRT; eg: old PNET-classified tumors) into ETMRs and CNS embryonal NOS. 
2) Assess whether any other tumors may have been mis-diagnosed but fit into these categories (eg: ETMR characteristics in other tumor diagnoses would make these newly ETMR cases).
3) Render the results in tabular form in a notebook.

#### Proposed methods

_What methods do you plan to use to accomplish the scientific goals?_
1) Review of copy number, mutation, and expression data.
2) Render results in tabular form in a notebook.

**ETMR, C19MC-altered**
- Embryonal tumor with multilayered rosettes (ETMR), C19MC-altered
- These tumors have focal amplification of the miRNA cluster on chr19 (denoted C19MC) and often have gene fusions involving _TTYH1_ and chr19 miRNA cluster genes
- These tumors have high expression of _LIN28A_ and serves as a biomarker for ETMRs

**ETMR, NOS**
- Embryonal tumor with multilayered rosettes (ETMR)
- These tumors have high expression of _LIN28A_ and serves as a biomarker for ETMRs, but do not show focal amplification of C19MC.

**CNS HGNET-MN1**
- CNS high-grade neuroepithelial tumor with MN1 alteration
- Likely previously diagnosed as PNET.
- Contain gene fusions involving 5' _MN1_. 3' fusion partners can include _BEND2_ and _CXXC5_. 
- Predominantly female patients.

**CNS HGNET-BCOR**
- CNS high-grade neuroepithelial tumor with BCOR alteration
- Tumors have internal tandem duplication of _BCOR_
- Median age of diagnosis less than 10 years

**CNS NB-FOXR2**
- Central nervous system (CNS) neuroblastoma with FOXR2 activation
- Over-expression and/or gene fusions in _FOXR2_

**CNS EFT-CIC**
- CNS Ewing sarcoma family tumor with CIC alteration
- Alterations in _CIC_, commonly fused with _ NUTM1_

**CNS Embryonal, NOS**
- CNS Embryonal tumor, not otherwise specified
- Tumors previously called PNET that do not fit into other groups above.

#### Required input data
_What input data will you use for this analysis?_
Copy number, RNA expression data, histologies file.

[...]

PR #352, opened 2019-12-18
PR 1 of n: Molecular Subtyping - HGG (Defining Lesions)

cbethell:
### Purpose/implementation Section

To molecularly subtype HGG samples.

#### What scientific question is your analysis addressing?

What are the samples in the OpenPBTA dataset that fit into the HGG molecular subtypes?

#### What was your approach?

I joined together the data that is relevant to molecular subtyping HGG samples, including the metadata, RNA expression, SNV, and CN data. 

I began by following the plan in the comment [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/249#issuecomment-564141215).

#### What GitHub issue does your pull request address?

This PR addresses issue #249.

#### Which areas should receive a particularly close look?

- Does this analysis appear to be correct? 
- Do the variables in the final data.frame seem suffice to molecularly subtype the HGG samples?
- Is there any obvious refactoring needed? 

#### Is there anything that you want to discuss further?

Note: A heatmap displaying the data in this PR is upcoming. 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

Yes, this PR is ready for review. 

#### What types of results are included (e.g., table, figure)?

A `tsv` file with the data in the final data.frame of the R notebook in this PR can be found in the `results` directory of this module at `results/HGG_molecular_subtypes.tsv`.

The table can also be viewed on the html output [here](https://cbethell.github.io/open-pbta-output/01-HGG-molecular-subtyping-data-prep.nb.html).

#### What is your summary of the results?

I have not yet developed a summary of the current results beyond the final data.frame

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### PR Checklist
* [x] Run a linter
* [x] Set the seed (NA)
* [x] Comments and/or documentation up to date 
* [x] Double check your paths
* [x] Spell check any Rmd file or md file
* [x] Restart R and run all notebooks fresh and save

PR #401, opened 2020-01-03
Molecular subtyping 251

sjspielman:

Continue subtyping embryonal samples, specifically as prompted Issue #251 

1. Can we identify molecular subtypes for unclassified embryonal tumors?

#### What was your approach?

Initial UMAP performed, with three different neighbor parameters (do not seem to matter) on _all embryonal_ tumors to examine whether current classifications/subtypes are apparent.

Address Issue #251 

I only see reasonable separation of ATRT, but MB samples are highly diffuse and the remaining embryonal tumors do not cluster well. Any thoughts on the most robust way to proceed?
Please additionally indicate if there are any issues with the UMAP implementation.

For incorporating additional data beyond RNAseq, will one want separate clustering analyses or a unified analysis? The scales of different quantities will obviously be quite different and difficult to simultaneously analyze using this type of approach.

Yes, please review the plot below (also displayed in the knitted html):

image

Figure above.

Larger circles are those which need subtyping. Plots appear highly robust to neighbor parameter. It is possible that several of the larger points are ATRT (there is nice clustering there), but all other samples are rather diffuse. Again, this only considers stranded expression.

**Question for reviewers** - when filling in the `analyses/README.md`, I have been writing a link to my fork since the analysis does not yet exist in ALSF's master. Correct/incorrect approach?

- [X ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [X ] This analysis has been added to continuous integration.
- [X ] This analysis is recorded in the table in `analyses/README.md`.


sjspielman:
Likely fails CI due to number of specified neighbors being larger than testing data, will address this soon.

sjspielman:
Hi @jharenza , I've begun updating the analysis to dig through expression, cnv, fusion rather than attempting a clustering. I'd like to run some of this by you before going further down the rabbit hole, as so far I'm not finding many samples to definitively subtype.  I'd appreciate some feedback on the updated knitted HTML for starting out with ETMRs!

The main "findings"  - only a handful of the unclassified samples have (eyeballing only, not very robust yet) notable LIN28A overexpression (fewer than 6), and merging up with information from fusions and/or C19MC gain, by `sample_id`

+`7316-3226` has TTYH1 and C19MC-region fusions, LIN28A log2 expression ~5.48
+`7316-4034` has C19MC Amplification, LIN28A log2 expression ~0.14 (fairly low)
+`7316-464` has C19MC Amplification, LIN28A log2 expression ~0.39 (still low, but getting higher!)
+`7316-784` has TTYH1 fusion, LIN28A log2 expression ~5.5 

Any thoughts so far? Thanks!

sjspielman:
Note, there are virtually no samples in the testing data that link up expression data with these samples, so this isn't going to pass CI under any circumstance I can think of where samples of interest are examined. The only way to pass CI is to not analyze any of these samples, which isn't going to make sense.

jaclyn-taroni:
@sjspielman I would recommended a step where you subset the files you need to only the relevant samples and then those should be small enough that you can commit them to the repository. The script or notebook that does the subsetting will not get tested with CI because there is no way it works.

This is what we did in https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-ATRT

And then you can set things up such that when someone runs the analysis locally it will use the most up to date files: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/4a3bb7e675513a6789e4cd8eb9597ee240f610ca/analyses/molecular-subtyping-ATRT/run-molecular-subtyping-ATRT.sh#L11

cgreene:
The main goal of CI is to make sure the code runs successfully as data/etc get updated. Is there an easy way to remove the subsetting to this set of samples and run over the full CI set using a CI environment variable?

sjspielman:
I think @cgreene's approach might be better here - if I pre-subset the files, I'll be saving empty TSV's no matter what. We just need a variable to not actually subset to the specific samples, I guess. Will do in next couple days.

jaclyn-taroni:
> if I pre-subset the files, I'll be saving empty TSV's no matter what.

I didn't follow this. To clarify, in the `molecular-subtyping-ATRT` module there is a folder that contains the relevant biospecimens that map to ATRT samples. Here I think you would subset to Embryonal tumors that are not MB or ATRT like you do here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/401/files#diff-77e96177f89b99ad2b1f0fe51ec5c22eR90

jharenza:
> Hi @jharenza , I've begun updating the analysis to dig through expression, cnv, fusion rather than attempting a clustering. I'd like to run some of this by you before going further down the rabbit hole, as so far I'm not finding many samples to definitively subtype. I'd appreciate some feedback on the updated knitted HTML for starting out with ETMRs!
> 
> The main "findings" - only a handful of the unclassified samples have (eyeballing only, not very robust yet) notable LIN28A overexpression (fewer than 6), and merging up with information from fusions and/or C19MC gain, by `sample_id`
> 
> +`7316-3226` has TTYH1 and C19MC-region fusions, LIN28A log2 expression ~5.48
> +`7316-4034` has C19MC Amplification, LIN28A log2 expression ~0.14 (fairly low)
> +`7316-464` has C19MC Amplification, LIN28A log2 expression ~0.39 (still low, but getting higher!)
> +`7316-784` has TTYH1 fusion, LIN28A log2 expression ~5.5
> 
> Any thoughts so far? Thanks!
> 
> > EDIT: I strongly suspect [this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/401/commits/f31ad7d17ac6589b725968235226b5bfdfda80a6) incidentally fixes CI problem, since there is no more UMAP and therefore no more neighbors.
> > EDIT: Nope.

Hi @sjspielman! I am not seeing `7316-3226` in the pbta-histologies file - was that a typo? I think this may be `7316-3326`. Do you know what level amplification of C19MC these samples have? I may want to manually spot check the plots we have as output from CNV algorithms, as these should be high-level focal gains. I presume the ones with the low-level _LIN28A_ expression may just be a broad gain, and not high-level amplification, and since 7316-4034 and 7316-464 do have have fusions, we may keep those as embryonal NOS (not otherwise specified) instead of definitive ETMRs. Since _LIN28A_ expression is a biomarker of this tumor, I think we can discard those with low-level expression as true ETMRs.

I was curious and just searched the fusion file for _TTYH1_ fusions and found some within other histologies I did not expect, making me think we should revisit the samples we are seeking. 
Screen Shot 2020-01-05 at 6 41 44 PM

For the ones with _TTYH1--DPRX/RNU6-698P_ fusions, I looked at where they were located in the genome, and they are adjacent to the C19MC cluster (see below), so I think these could be real ETMRs, misdiagnosed, and we should check _LIN28A_ expression. 
Screen Shot 2020-01-05 at 6 37 36 PM

Given the above, I think maybe we should approach this as we did with the start of the HGG subtyping: #352, in which we first looked for H3F3A or HIST1H1B SNVs. Here, maybe the first step for ETMRs should be looking for TTYH1 fusions in **_all_** samples, seeing if we pull any out, then step 2, take the histologies listed in the ticket, and subtype for ETMR/remaining embryonal subtypes.

I will checked out `7316-4034` and `7316-464` manually for C19MC amplification, but did not see it. We may be able to plot this if we plot just chr19 LRRs from one of the CNV files - we would expect a small blip on chr19. For reference, below is an example CNV image file for `7316-3326` showing positive C19MC amplification:
Screen Shot 2020-01-05 at 6 49 51 PM

sjspielman:
@jaclyn-taroni 
> I didn't follow this. To clarify, in the `molecular-subtyping-ATRT` module there is a folder that contains the relevant biospecimens that map to ATRT samples. 

Ah, I got confused so you were correct not to follow. In my weekend mind, the subsetting would have still been performed on testing data, but obviously this is not the case. What I should do is create a separate R script for subsetting the data which gets saved into the analysis `results` directory, and then don't run that script in CI.

jaclyn-taroni:
> What I should do is create a separate R script for subsetting the data which gets saved into the analysis results directory, and then don't run that script in CI.

We are on the same page 😁

sjspielman:
Taking a step back and searching _all_ samples for fusions involving either TTYH1 or C19MC-region genes. To identify the latter, I searched for any fusion where a) both breakpoints were on chr19, and b) at least ONE breakpoint fell within the C19MC region. The results have now been saved in **`results/samples_with_TTYH1_C19MC_fusions.tsv`** and are displayed in the updated Rmarkdown file. There are 30 samples with fusions in one of these regions, many of which are NOT typed as Embryonal CNS at all.

jharenza:
> Taking a step back and searching _all_ samples for fusions involving either TTYH1 or C19MC-region genes. To identify the latter, I searched for any fusion where a) both breakpoints were on chr19, and b) at least ONE breakpoint fell within the C19MC region. The results have now been saved in **`results/samples_with_TTYH1_C19MC_fusions.tsv`** and are displayed in the updated Rmarkdown file. There are 30 samples with fusions in one of these regions, many of which are NOT typed as Embryonal CNS at all.

Hi @sjspielman - sorry for the delay on this. I posted an email response from a domain expert [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/251#issuecomment-571807158) and never tied it into this thread. 

I think that we probably don't want an "either/or" here, but rather want to detect whether the fusion exists between _TTYH1_ and some gene(s) within/adjacent to the C19MC region. If this is the case, then we can confirm ETMR with _LIN28A_ as well as _DNMT3B_ expression, per the email correspondence. Does that help? 

From **`results/samples_with_TTYH1_C19MC_fusions.tsv`**, I see `BS_QWNBZ9RJ` and `BS_37GTVG4N` satisfy this criteria, but so do many below which we should include and test:

| Sample | FusionName | Include? |
|-------------|-------------------------|----------|
| BS_37GTVG4N | TTYH1--MIR515-2/MIR519C | Yes |
| BS_37GTVG4N | TTYH1--MIR512-2/MIR1323 | Yes |
| BS_65VFS9N1 | TTYH1--DPRX/RNU6-698P | Yes |
| BS_685DACMS | TTYH1--DPRX/RNU6-698P | Yes |
| BS_HWGWYCY7 | NPAS3--AC245884.4/TTYH1 | No |
| BS_N6N147BY | TTYH1--DPRX/RNU6-698P | Yes |
| BS_P39SQPTS | TTYH1--MIR512-2/MIR1323 | Yes |
| BS_QWNBZ9RJ | TTYH1--DPRX/RNU6-698P | Yes |
| BS_S5P8VTN5 | MACF1--TTYH1 | No |

Issue #398, opened 2020-01-03
Proposed Analysis: Fusion files specifically for consumption by molecular subtyping analyses


jaclyn-taroni:
Related to this thread: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/315#issuecomment-563425021

#### What are the goals of the analysis?

To generate files that contain information about the _presence or absence_ of specific fusions or genes participating in fusions to be used in generating subtype labels for:

* Ependymoma tumors #245 
* Non-MB/ATRT embryonal tumors #251 

#### What input data are required for this analysis?

We can use the putative oncogenic fusion file for this: `pbta-fusion-putative-oncogenic.tsv`

#### What is the expected output?

I think output here will be most useful if it is divided into two files in tabular format: 1) fusions relevant for ependymoma tumors and 2) non-MB/ATRT embryonal tumors.

Below I include what I think the files should contained based on my understanding of #245 and #251, where columns with a name that consists of a _single gene_ will indicate if that gene is included in _any_ fusion event.

##### Ependymoma

Toy example:

| Kids_First_Biospecimen_ID | RELA | C11orf95--YAP1 | YAP1--MAMLD1 | YAP1--MAMLD2 | YAP1--FAM118B | C11orf95--MAML2 |  PTEN--TAS2R1 | C11orf95--RELA | LTBP3--RELA |
|----|----|----|----|----|----|----|----|----|----|
| BS_XXXXXXXX | 1 | 0 | 0 | 0 | 0 | 0 | 1 | 1 | 1 |

##### Non-MB/ATRT embryonal tumors 

#251 states:

> [ETMR, C19MC-altered] tumors have focal amplification of the miRNA cluster on chr19 (denoted C19MC) and often have gene fusions involving TTYH1 and chr19 miRNA cluster genes.

Side note, I think this is the relevant publication for that: [Kleinman et al. _Nature Genetics_. 2014.](https://doi.org/10.1038/ng.2849)

Because it seems that this cluster would contain over 40 gene symbols ([Wikipedia](https://en.wikipedia.org/wiki/C19MC_miRNA_cluster)), I propose listing which, if any, fusions with _TTYH1_ are present in a sample.

Here's a toy example:

| Kids_First_Biospecimen_ID | TTYH1_fusions | MN1 | MN1--BEND2 | MN1--CXXC5 | FOXR2 | CIC-NUTM1 |
|----|----|----|----|----|----|----|
[...]

Issue #245, opened 2019-11-08
Proposed Analysis: Molecularly subtype ependymoma tumors


jharenza:
#### Scientific goals

_What are the scientific goals of the analysis?_
 Subtype ependymomas into ST-EPN-RELA, ST-EPN-YAP1, PF-EPN-A, and PF-EPN-B. Note: The publication listed below contains 9 subtypes of ependymomas, but only the 4 listed here are relevant to the OpenPBTA dataset (due to age at diagnosis) and will be discussed below.

#### Proposed methods

_What methods do you plan to use to accomplish the scientific goals?_
1) Review of copy number, mutation, and expression data.
2) Render results in tabular form in a notebook.

**ST-EPN-RELA (Supratentorial Ependymoma, _RELA_ fused)**
- Most commonly harbor _C11orf95-RELA_ fusions due to chromothripsis of chr 11q13.1 ([ref](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4050669/)). Only subgroup to exhibit chromothripsis.
- Other reported _RELA_ fusions include: _LTBP3-RELA_ 
- _PTEN-TAS2R1_  has been found in in _RELA_ negative samples (results in loss of _PTEN_): "Interestingly, in one case of the ST-EPN-RELA subgroup, negative for RELA fusion types 1 and 2, we detected a PTEN-TAS2R1 fusion leading to a frame shift and subsequent disruption of PTEN (Table S1)."
- Fusions result in activation of NF-κB target genes - See Figure 4c in above ref.
- Genes over-expressed relative to other EPN tumors: _RELA_ and _L1CAM_
- Common CN changes: _CDKN2A_ deletions. Chr9 or Chr9p loss

**ST-EPN-YAP1 (Supratentorial Ependymoma, _YAP1_ fused)** 
- Harbor _C11orf95-YAP1_ , _YAP1-MAMLD1_, _YAP1-MAMLD2_, and _YAP1-FAM118B_ fusions
- Other reported fusions include: _C11orf95-MAML2_
- Genes over-expressed relative to other EPN tumors: _ARLD4_ and _CLDN1_
- Focal 11q most frequent CNA, with rest of genome stable

**PF-EPN-A (Posterior Fossa Ependymoma, Type A)**
- Genes over-expressed relative to other EPN tumors: _CXorf67_ and _TKTL1_
- 1q gain most frequent CNA

[...]

PR #410, opened 2020-01-07
Added new fusion summary module

dmiller15:
updated anaylsis readme and ci config

In biospecimens from ependymoma and non-ATRT embryonal tumors, which ones have fusions of interest?

#### What was your approach?
Using the biospecimens that were identified to be in one or the other population, we filtered the fusions file. That filtered fusion file was simplified using a list of explicit and generic filters for fusions that have been shown to have relevance in the cancer of that population. The results were summarized in a TSV for each population.

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/398

The filters are of particular note since those values were mostly gleaned from the ticket. Perhaps there might be better ways of performing the filtering. Also the final summary tables aren't exactly set in stone. Particular for how we represent the generic fusions where we accept any fusion containing a particular gene. I simply left all the generic fusions found as unique columns. The ticket chose to aggregate these values, but the utility of that is up for discussion.

No.

Yes. I'm confident the tables are presenting the information requested in the ticket.

TSV tables.

C11orf95--RELA is nearly universal in Ependymoma tumors. Very few biospecimens have the fusions of interest. Almost none of the samples have more than one fusion.


sjspielman:
Hi @dmiller15 , thanks for getting started on this analysis. I see in the originating Issue #398, the analysis requested is to search _all_ samples for either classification or reclassification (my emphasis added):
> ... To generate files that contain information about the presence or absence of specific fusions or genes participating in fusions **to be used in generating subtype labels** ... 

It looks like you've subsetted the data to only certain histologies before searching for fusions of interest. Can you update the code to search all samples even those that have already been subtyped? Thanks!

dmiller15:
Thanks for the input @sjspielman. I've added files for each set of fusions that don't filter the biospecimens beforehand.

analyses/fusion-summary/01-fusion-summary.R
```diff
@@ -0,0 +1,167 @@
...
+demoEpend.bios <- filter(demo,
+                        short_histology == ependSH)$Kids_First_Biospecimen_ID
+demoEmbry.bios <- filter(demo,
+                        broad_histology == embryBH &
+                          !(short_histology %in% embrySH))$Kids_First_Biospecimen_ID
+
+#' Filter the fusion files for your two populations
+fuseEpend <- filterFusion(df = fuse, 
+                          bioid = demoEpend.bios, 
+                          fuses = ependFuses)
```
jaclyn-taroni:
```suggestion
                          fuses = ependFuses,
                          genes = "RELA")
```

analyses/fusion-summary/01-fusion-summary.R
```diff
@@ -0,0 +1,167 @@
...
+demo <- read.csv(opt$demographic_file, sep="\t")
+fuse <- read.csv(opt$fusions_file, sep="\t")
+
+#' Use the filters from above to get the biospecimen ids
+demoEpend.bios <- filter(demo,
+                        short_histology == ependSH)$Kids_First_Biospecimen_ID
+demoEmbry.bios <- filter(demo,
+                        broad_histology == embryBH &
+                          !(short_histology %in% embrySH))$Kids_First_Biospecimen_ID
+
```
jaclyn-taroni: I think we can remove these steps now that you have the files that are not filtered. It is straightforward for an analyst to filter the files using `pbta-histologies.tsv` if they need to downstream.

analyses/fusion-summary/01-fusion-summary.R
jaclyn-taroni: Alternatively - we can direct folks to the `pbta-fusion-recurrently-fused-genes-bysample.tsv` file because it contains information for `RELA`.

analyses/fusion-summary/01-fusion-summary.R
```diff
@@ -0,0 +1,167 @@
...
+#' 1: Used to filter short_histology to get Ependymoma tumors 
+ependSH = "Ependymoma"
+#' 2: Used for Embryonal tumors (short_histology)
+#'  not from ATRT or MB (broad_histology)
+embryBH = "Embryonal tumor"
+embrySH = c("ATRT", "Medulloblastoma")
+#' 
+#' *Fusions Filters*
+#' 1: Exact match a list of fusions common in Ependymoma tumors
+ependFuses = c(
```
jaclyn-taroni: @jharenza does the order of the partners matters for these? My read of #245 was yes

analyses/fusion-summary/01-fusion-summary.R
```diff
@@ -0,0 +1,167 @@
...
+  "YAP1--MAMLD1",
+  "YAP1--MAMLD2"
+)
+#' 2: Exact match a list of fusions common in Embryonal tumors
+#' as well as fusions containing a particular gene with any other gene
+embryFuses = c(
+  "CIC--NUTM1",
+  "MN1--BEND2",
+  "MN1--CXXC5"
+)
```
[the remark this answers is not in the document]
jharenza: Ahhh, thanks for this, @jaclyn-taroni. I did some digging into [this paper ](https://reader.elsevier.com/reader/sd/pii/S0092867416000556?token=2C44098DED338194711C1C3A175AA38CF218F4D166EDE7C2C5A03E4A6C9C50E099BB5EE04DFA1B9EFEE9197F6058B0B1) and it looks like _MN1_ is the 5' partner, so I can update #251 regarding this. I think when I wrote this up, it was not noted on the website I used.

analyses/fusion-summary/01-fusion-summary.R
jharenza: yes, correct, these are the fusions in the orientation that have been reported

analyses/fusion-summary/01-fusion-summary.R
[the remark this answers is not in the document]
jaclyn-taroni: 👍 thank you @jharenza. This is appropriately restrictive then!

analyses/fusion-summary/01-fusion-summary.R
dmiller15: Added an ependGene block with RELA here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/410/files#diff-508d314152253028799278a4a1540cf3R23-R25 and plugged the value in here.

analyses/fusion-summary/01-fusion-summary.R
dmiller15: Thanks for the input. Removed.

dmiller15:
> The files produced here have under 40 samples, and if I followed correctly, I believe this is due to the inclusion of only the samples (Kids_First_Biospecimen_ID) that have at least one of the fusions or genes that are being specified. We want to include all samples.

@sjspielman With regard to the above, I no longer drop the levels when making the table. You can inspect the newly generated outputs and see 787 samples, which is the number of unique biospecimens available in `pbta-fusion-putative-oncogenic.tsv`.

jaclyn-taroni:
Realized after approving that we no longer needed the demographic file - so I removed that in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/410/commits/bba4e786b38a3cac4b3cf1e32796b8b9170d81f1. I'll merge once CI finishes!

Issue #394, opened 2020-01-02
Proposed Analysis: chromosomal instability burden, recurrently altered genes


jaclyn-taroni:
I am breaking up #27 into more focused issues.

#### What are the scientific goals of the analysis?

Integrated analysis of CNV and structural variant data. This issue covers readouts/output described in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/27#issuecomment-516471524 that are not captured by #393, specifically (paraphrasing):

* Generating a measure of chromosomal instability (CIN) burden that can be plotted by tumor type
* Identify recurrently altered genes to be included in the OncoPrint (#6)
  - Amplification/tandem-duplication
  - Deep deletion/deletion
  - Other Structural variation: Inversion, translocations

#### What methods do you plan to use to accomplish the scientific goals?

From  https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/27#issuecomment-516471524:

> In order to integrate CNV calls and SV calls we focus on breakpoint co-locallization, more details in the manuscript: https://www.biorxiv.org/content/10.1101/572248v3

@gonzolgarcia do you have additional guidance you can provide outside of what you included on #27 such that someone else can complete this analysis? Links to source code would be ideal. Thank you!

#### What input data are required for this analysis?

`pbta-cnv-cnvkit.seg.gz` - though we ultimately may want to use a consensus filtered file (related: #128 #392)
`pbta-sv-manta.tsv.gz`

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

Unknown; the scope of this will need to be refined.

#### Who will complete the analysis (please add a GitHub handle here if relevant)?

@cansavvy is going to start digging into this a bit.

#### What relevant scientific literature relates to this analysis?

Lopez et al. Somatic structural variation targets neurodevelopmental genes and identifies SHANK2 as a tumor suppressor in neuroblastoma. _bioRxiv._ 2019. DOI: [10.1101/572248](https://doi.org/10.1101/572248)

cansavvy:
Does anyone know anything about [CINdex](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5761903/)?  I don't see any indication of it using SV data though, only CNV data.

jaclyn-taroni:
[...]

PR #413, opened 2020-01-08
Chromosomal Instability (PR 2 of 3): The functions

cansavvy:

This second PR has the functions that will be used for calculating and plotting chromosomal breakpoints. 

This was the task: 
> Generating a measure of chromosomal instability (CIN) burden that can be plotted by tumor type

I borrowed concepts from https://github.com/gonzolgarcia/svcnvplus for generating breakpoint density. I calculate breakpoint density and in a later PR, I will plot this data by tumor type. 

#### What was your approach?

These functions calculate breakpoint density for CNV, SV and combined CNV and SV and them plot them. I have an example of the output plot below. 

The first part of #394 (the recurrently altered genes part will be a different PR series that will undoubtedly use the functions this PR series uses as well as maybe additional ones). 

I have ideas about how to make these functions more generalizable, but before I get too far into the weeds, I need another pair of eyes to let me know how they look so far. 

 - Are there places I can make these functions more..
          - ...Efficient? 
          - ...DRYer? 
          - ...Easier to use? 
          - ...Better documented?

Keep in mind, I realize the comments and suggestions about these functions make change as you see them being used in context in the subsequent PRs. Don't worry about them being perfect in context at this stage, I'm more looking for places where my approach should be changed or optimized in general.

### Results
No results yet, that will be the next PR, just the functions. Here's an example plot of what these functions create: 

BS_ZZJF26C4_breaks

This will be added to circleCI in the third PR. There are no new packages that need to be added to the Dockerfile so that is set. 


analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
+# Functions for chromosomal instability calculations
+#
+# C. Savonen for ALSF - CCDL
+#
+# 2020
+
```
sjspielman: I wonder if it makes sense to load libraries here? Otherwise, the script(s) calling these functions fails to load libraries for whatever reason the code won't break.

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
...
+                          start_col_sv = "start",
+                          end_col_sv = "end") {
+  # For given breaks data.frame(s), calculate the breaks density for a tiled
+  # windows across the genome. Returns the data as a GenomicRanges object for
+  # easy mapping with `ggbio`. Where the density and counts are stored in
+  # @elementMetadata@listData.
+  #
+  # Args:
+  #   sv_breaks: a data.frame with the breaks for the SV data.
+  #   cnv_breaks: a data.frame with the breaks for the SV data.
```
sjspielman: Description identical for sv/cnv.

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
+# Functions for chromosomal instability calculations
+#
+# C. Savonen for ALSF - CCDL
+#
+# 2020
+
+make_granges <- function(break_df = NULL,
```
sjspielman: Is this intended to be, effectively, a private function? As in - do you foresee it being used outside of `break_density()`? Thinking about sanity checks where you say to check your `col` arguments, whether this will make sense to users who are calling `break_density()` directly. I think what you have is likely _fine_...

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
...
+                          end_col_sv = "end") {
+  # For given breaks data.frame(s), calculate the breaks density for a tiled
+  # windows across the genome. Returns the data as a GenomicRanges object for
+  # easy mapping with `ggbio`. Where the density and counts are stored in
+  # @elementMetadata@listData.
+  #
+  # Args:
+  #   sv_breaks: a data.frame with the breaks for the SV data.
+  #   cnv_breaks: a data.frame with the breaks for the SV data.
+  #   sample_id: sample ID for which the data needs to be extracted and made into
```
sjspielman: A bit more clarity here: My immediate intuition is that this refers literally to `sample_id` column in other files. If so, say explicitly it's the same as in histology file.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: All functions used used `::` so no libraries should be necessary.

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
...
+  #              a GenomicRanges object. If "all" is designated, all samples will
+  #              be kept. Multiple samples can be designated as a character vector.
+  #   chr_size_list: a named numeric vector of the sizes (bp) of the chromosomes.
+  #                  names of the chromosomes must match the format of the input
+  #                  break data. e.g. "chr1" or "1".
+  #   window_size: What size windows to calculate break density. Default is 1kb.
```
sjspielman: Your default is 1e6 which feels more than 1kb? Something needs to change, either description to mb or default to 1e3.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Whoops. Copy and paste remnant.

analyses/chromosomal-instability/util/chr-break-calculate.R
sjspielman: good point! great.

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,233 @@
...
+  # Check that a sample ID has been specified.
+  if (is.null(sample_id)) {
+    stop("No sample ID has been specified. Use the `sample_id` argument.")
+  }
+
+  # Determine how many samples are in the group
+  n_samples <- length(sample_id)
+
+  # Make CNV into GRanges
+  if (!is.null(cnv_breaks)) {
```
sjspielman: I _think_ this has to be changed to `if (!is.null(cnv_breaks) & is.null(sv_breaks))` and same for the next if. Otherwise, we have logic:
1. if not null CNV
2. if not null SV
3. if neither is null <- ends up being duplicative, right? you note this in your comments about overwriting. **Actually** it will probably be most clean to just rearrange: Make the last `if` come first (neither is null), and then some `else if`'s for the other two. Then a final else to `stop()` if both are null.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: The `sample_ID` argument can literally be any kind of sample ID, so long as the column specified in with `sample_col` arguments  for `sv_breaks` and `cnv_breaks` have a string that matches it.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Ah. This is a remnant I missed. Did have it 1kb, then switched it but forgot some of the labels.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Awesome. I knew I didn't like my logic set up here but wasn't sure how to make it better. This is great.

cansavvy: If I'm understanding your suggestion right after another read, I don't think it will work. I can't have the last if come first, because the last if uses the items created in the first two ifs.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: It will be used by another function I need to make for another part of this analysis besides `break_density` and it *could* be used in other scenarios, which is why it isn't all one function. Is there something you recommend I do to make the error message more easily interpretable whether in or outside another function?

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Basically if you just supply one of the datasets, that will be the only dataset used, but if you supply both datasets, we need to combine them.

analyses/chromosomal-instability/util/chr-break-calculate.R
sjspielman: As long as you will be able to interpret any potential error messages when called from another function you don't need to change anything. But, for example, you'd want to make sure that future functions that call this also have `cols` type arguments.

analyses/chromosomal-instability/util/chr-break-calculate.R
sjspielman: What I was thinking was this concept - you don't think it will work? 
if (both present)
else if (only CV)
else if(only SV)
else (everything is null)

Also, apologies, I can't test this in depth right now since data release 12 is actively downloading onto work computer that I haven't used in a month, but will look more closely interacting with the code tomorrow!! So if my suggestion makes no sense this is probably why.

analyses/chromosomal-instability/util/chr-break-plot.R
```diff
@@ -0,0 +1,118 @@
...
+  # For setting the scale later, need to get y's max
+  max_y <- max(
+    data.frame(granges@elementMetadata@listData) %>%
+      dplyr::pull(
+      !!rlang::sym(y_val)
+    ),
+    na.rm = TRUE
+  )
+  # Make the density plot
+  density_plot <- ggbio::autoplot(granges, ggplot2::aes(y = !!rlang::sym(y_val)),
```
sjspielman: Just FYI, there is another strategy for supplying args specifically to ggplot: `aes_string()` rather than `aes()`. You don't need it here, but for future reference it's an easier way to pass in columns if all you need is `aes()`'ing them. Facets....who can say....

analyses/chromosomal-instability/util/chr-break-plot.R
```diff
@@ -0,0 +1,118 @@
...
+    ggplot2::theme(axis.text.x = ggplot2::element_text(size = 3, angle = 45, hjust = 1)) +
+    ggplot2::ylab(y_lab) +
+    ggplot2::ggtitle(main_title) +
+    ggplot2::scale_y_continuous(breaks = seq(0, max_y, by = 2))
+
+  # Print out plot
+  density_plot@ggplot
+}
+
+chr_break_plot <- function(granges_list,
```
sjspielman: I'd change this something a bit more explicit, e.g. but not necessarily `make_chr_break_density_plots`? Adding a `make` at front and something indicating you're getting THREE plots in the figure.

analyses/chromosomal-instability/util/chr-break-plot.R
```diff
@@ -0,0 +1,118 @@
...
+  # Make the density plot
+  density_plot <- ggbio::autoplot(granges, ggplot2::aes(y = !!rlang::sym(y_val)),
+    geom = "line", scales = "free_x", space = "free_x",
+    colour = color
+  ) +
+    ggplot2::theme_classic() +
+    ggplot2::theme(axis.text.x = ggplot2::element_text(size = 3, angle = 45, hjust = 1)) +
+    ggplot2::ylab(y_lab) +
+    ggplot2::ggtitle(main_title) +
+    ggplot2::scale_y_continuous(breaks = seq(0, max_y, by = 2))
```
sjspielman: Are we confident `by=2` will be universally reasonable?

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Let me play around with a bit and see if I can do what you are suggesting here. I definitely like the logic of it better.

analyses/chromosomal-instability/util/chr-break-plot.R
cansavvy: Good to know. Thanks for that tip!

analyses/chromosomal-instability/util/chr-break-plot.R
cansavvy: So far it is. But this is a good question to keep an eye out for.

analyses/chromosomal-instability/util/chr-break-calculate.R
sjspielman: It just involves some more involved ifs, checking "if one of interest is not null **AND** other is null". The double negatives get trippy here, may be a good task for tomorrow morning..

analyses/chromosomal-instability/util/chr-break-plot.R
sjspielman: maybe just add a variable defining this as 2 at the top of the function, but i don't think it merits an argument.

analyses/chromosomal-instability/util/chr-break-plot.R
cansavvy: Is `make_multipanel_break_plot` too long of a name?

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,253 @@
...
+                         chrom_col = "chrom",
+                         start_col = "start",
+                         end_col = "end") {
+  # For a given breaks data.frame make a GenomicRanger object from it. Can
+  # filter to a single samples' data.
+  #
+  # Args:
+  #   break_df: for a data.frame with chromosomal coordinates and sample IDs.
+  #   sample_id: a character string that designates which data needs to be
+  #              extracted and made intoa GenomicRanges object by matching the
```
jaclyn-taroni:
```suggestion
  #              extracted and made into a GenomicRanges object by matching the
```

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,253 @@
...
+#
+# 2020
+
+make_granges <- function(break_df = NULL,
+                         sample_id = NULL,
+                         samples_col = "samples",
+                         chrom_col = "chrom",
+                         start_col = "start",
+                         end_col = "end") {
+  # For a given breaks data.frame make a GenomicRanger object from it. Can
```
jaclyn-taroni:
```suggestion
  # For a given breaks data.frame make a GenomicRanges object from it. Can
```

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -0,0 +1,253 @@
...
+  #                `elementData@listData$samples`
+  #   chrom_col: character string that indicates the column name with the
+  #              chromosome information. Default is "chrom".
+  #   start_col: character string that indicates the column name with the
+  #              start coordinate. Default is "start".
+  #   end_col: character string that indicates the column name with the
+  #            end coordinate. Default is "end".
+  #
+  # Returns:
+  # A Genomic Ranges formatted object for the particular sample that has the
```
jaclyn-taroni: Is this more appropriate?
```suggestion
  # A GRanges object for the particular sample that has the
```

jaclyn-taroni: Apparently it should be
```suggestion
  # A GenomicRanges object for the particular sample that has the
```

Issue #244, opened 2019-11-08
Proposed Analysis: Molecularly subtype ATRT tumors


jharenza:
#### Scientific goals

_What are the scientific goals of the analysis?_
Subtype ATRTs into SHH, MYC, TYR. 

#### Proposed methods

_What methods do you plan to use to accomplish the scientific goals?_
Suggestions:
1) In the absence of methylation data, perform t-sne or other clustering of expression data to determine whether the total N of ATRTs is high enough to enable subgroup clustering (See Supplemental Figure 1B in the manuscript below)
2) Review of histologies file, copy number, SV, mutation, and _gene expression data across all ATRT samples_.

Summarize results in a table, which can be added to a notebook, with molecular subtype designated. 

**ATRT TYR**
- most are infratentorial localized 
- broad _SMARCB1_ deletions (most have chr22q loss/monosomy 22)
- Major expression markers: overexpression of melanosomal genes (eg: _TYR_, _MITF_, _DCT_)
  - _TYR_ is over-expressed in every case in this subgroup but not expressed in any other subgroup (biomarker)
- overexpression of _VEGFA_
- overexpression of genes involved in ciliogenesis (eg: _DNAH11_ and _SPEF1_)
  - Explore additional TF targets in Figure 6A in the manuscript below for overexpression if needed

**ATRT SHH**
- infratentorial or supratentorial localized
- focal _SMARCB1_ deletions
- Major expression markers: overexpression of SHH pathway (can be accomplished by looking at single sample GSEA results or expression of SHH pathway genes - eg: _MYCN_, _GLI2_)
- overexpression of _CDK6_
- overexpression of _ASCL1_, _HES5/6_, and _DLL1/3_, indicating active _NOTCH_ signaling
  - _MYC_ is the most commonly overexpressed gene in this subgroup
  - Explore additional TF targets in Figure S6A in the manuscript below for overexpression if needed
- can have _SMARCA4_ loss, but not required
- this group has more total mutations than the other two

**ATRT MYC**
- most are supratentorial localized
- focal _SMARCB1_ deletions
- Major expression markers: overexpression of _MYC_ and _HOX_ gene cluster - eg: _MYC_, _HOTAIR_
  - Explore additional TF targets in Figure S6B in the manuscript below for overexpression if needed 

[...]

PR #414, opened 2020-01-08
Add `chr22q` loss variable to ATRT molecular subtyping

cbethell:

The purpose of this PR is to add chr 22 loss information to the ATRT subtype data.frame and tsv file.

This analysis is addressing the molecular subtyping of ATRT samples.

#### What was your approach?

I used the `broad_values_by_arm.txt` file from GISTIC to obtain the chr22q data for the ATRT samples.

This PR addresses issue #244.

Does this additional information appear to be accurately represented?

This PR includes changes to the heatmap:
- plots/atrt_heatmap.png

This plot can be viewed in the `README` [here](https://github.com/cbethell/OpenPBTA-analysis/blob/add-chr-22-atrt/analyses/molecular-subtyping-ATRT/README.md).

It also includes changes to the final output tsv file:
- results/ATRT_molecular_subtypes.tsv.gz


cbethell:
The html output with the final table displayed can be found [here](https://cbethell.github.io/open-pbta-output/01-ATRT-molecular-subtyping-data-prep.nb.html#gistic_data).

jaclyn-taroni:
It could be useful to organize the table columns such that the relevant information for each of the subtypes on #244 is grouped/presented next to each other. I think we at least want the _SMARCB1_ status and chr22q columns to be grouped and come earlier in the table.

cbethell:
> It could be useful to organize the table columns such that the relevant information for each of the subtypes on #244 is grouped/presented next to each other. I think we at least want the _SMARCB1_ status and chr22q columns to be grouped and come earlier (further to the left) in the table.

View the table reflecting the suggestion above [here](https://cbethell.github.io/open-pbta-output/01-ATRT-molecular-subtyping-data-prep.nb.html#save_final_table_of_results).

jaclyn-taroni:
@cbethell I noticed a couple instances where GISTIC data is missing but the focal data is not and vice versa (here are two examples): 

| sample_id | 	Kids_First_Biospecimen_ID |	Kids_First_Participant_ID | chr_22q_loss |	SMARCB1_focal_status |
|----|----|----|----|----|
| 7316-376 | BS_74A1TB03, BS_M4923M40 |	PT_MTE126WM | NA | loss, neutral |
| 7316-2187 | BS_53TV75NN, BS_850BAHH9 | PT_HVZTF42R | No | NA |

Do you know why that might be?

cbethell:
> @cbethell I noticed a couple instances where GISTIC data is missing but the focal data is not and vice versa (here are two examples):
> 
> sample_id	Kids_First_Biospecimen_ID	Kids_First_Participant_ID	chr_22q_loss	SMARCB1_focal_status
> 7316-376	BS_74A1TB03, BS_M4923M40	PT_MTE126WM	NA	loss, neutral
> 7316-2187	BS_53TV75NN, BS_850BAHH9	PT_HVZTF42R	No	NA
> Do you know why that might be?

@jaclyn-taroni  the `NA`'s are due to data missing for these samples from the gistic data and focal data, respectively.

jaclyn-taroni:
The GISTIC data and focal CN data you are using are both derived from the CNVkit data. That suggests to me that something else might be going on. Are all the samples that end up in that table (with WGS data) in the CNVkit file?

cbethell:
@jaclyn-taroni all of the samples are in the CNVkit file, but they do not all have data for `SMARCB1`. These samples are however not all in the GISTIC file.

jaclyn-taroni:
> all of the samples are in the CNVkit file, but they do not all have data for `SMARCB1`.
Ah, okay - so this suggests to me that these should be set as `neutral` or something else rather than `NA` because it's not that the data is missing but that we don't have evidence that it's a loss - do you agree?

cbethell:
> Ah, okay - so this suggests to me that these should be set as `neutral` or something else rather than `NA` because it's not that the data is missing but that we don't have evidence that it's a loss - do you agree?

Yes, that is a good point. I will implement this change now.

jaclyn-taroni:
@cbethell can you generate a list of `Kids_First_Biospecimen_ID` that are in the CNVkit data but not in the GISTIC data and [file a data issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/new?assignees=&labels=data&template=data-question.md&title=) please?

cbethell:
> @cbethell can you generate a list of `Kids_First_Biospecimen_ID` that are in the CNVkit data but not in the GISTIC data and [file a data issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/new?assignees=&labels=data&template=data-question.md&title=) please?

@jaclyn-taroni yes, I'm on it 👍

jharenza:
> > @cbethell can you generate a list of `Kids_First_Biospecimen_ID` that are in the CNVkit data but not in the GISTIC data and [file a data issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/new?assignees=&labels=data&template=data-question.md&title=) please?
> 
> @jaclyn-taroni yes, I'm on it 👍

Hi! Regarding this - I checked the `sample_seg_counts.txt` file that prints from the GISTIC run and found that the one sample that had `NA` in GISTIC but had CNVkit calls, `BS_74A1TB03`, actually had too many segments for GISTIC (>2500) for calls to be made and was excluded from GISTIC analyses. The `sample_seg_counts.txt` file has all sample segment counts and whether they were included or excluded, so that explains the discrepancies.

jaclyn-taroni:
@jharenza we figured that there was some GISTIC filtering step. We do not have much documentation around the GISTIC files at the moment: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#copy-number-files

jharenza:
OK - I think I will have to update that.

cbethell:
> Hi! Regarding this - I checked the `sample_seg_counts.txt` file that prints from the GISTIC run and found that the one sample that had `NA` in GISTIC but had CNVkit calls, `BS_74A1TB03`, actually had too many segments for GISTIC (>2500) for calls to be made and was excluded from GISTIC analyses. The `sample_seg_counts.txt` file has all sample segment counts and whether they were included or excluded, so that explains the discrepancies.

@jharenza thank you for this clarification. It indeed does explain the discrepancies as I cross checked said samples in a separate notebook [here](https://cbethell.github.io/open-pbta-output/GISTIC-missing-samples.nb.html#missing_samples).

I have also removed the duplicated columns in the final tsv file.

jaclyn-taroni:
My review of #410 made me think that a similar issue might be happening here with the CNVkit data. I have a few changes that I have not yet pushed because they conflict with the last commit and I should take another look. Because of how we filtered the subset files initially, we were inadvertently dropping a few samples without transcriptomic data so the table will have some additional samples but largely remain the same.

analyses/molecular-subtyping-ATRT/00-subset-files-for-ATRT.R
```diff
@@ -81,26 +81,31 @@ tmb_df <-
...
+# Read in GISTIC `broad_values_by_arm.txt` file
+gistic_df <-
+  data.table::fread(unzip(
+    file.path(root_dir, "data", "pbta-cnv-cnvkit-gistic.zip"),
+    files = file.path(
+      "2019-12-10-gistic-results-cnvkit",
+      "broad_values_by_arm.txt"
+    ),
+    exdir = file.path(root_dir, "scratch")
+  ))
```
jaclyn-taroni:
```suggestion
  ), data.table = FALSE)
```

analyses/molecular-subtyping-ATRT/00-subset-files-for-ATRT.R
```diff
@@ -142,8 +147,26 @@ readr::write_tsv(filtered_gsva_scores, file.path(results_dir, "atrt_gsva.tsv"))
...
+                   by = c("Tumor_Sample_Barcode" = "Kids_First_Biospecimen_ID")) %>%
   dplyr::filter(sample_id %in% atrt_df$sample_id)
 
 # Write to file
 readr::write_tsv(tmb_df, file.path(results_dir, "atrt_tmb.tsv"))
+
+#### Filter GISTIC data --------------------------------------------------------
+
+gistic_df <- gistic_df %>%
+  as.data.frame() %>%
```
jaclyn-taroni: Do you need to call `as.data.frame()` here to get the `tibble::column_to_rownames` step to work? Asking because you set this as a `data.frame` after the transpose. If it's necessary to do this twice, I think it may be because `gistic_df` is a `data.table` (see my comment above).

analyses/molecular-subtyping-ATRT/00-subset-files-for-ATRT.R
```diff
@@ -142,8 +147,26 @@ readr::write_tsv(filtered_gsva_scores, file.path(results_dir, "atrt_gsva.tsv"))
...
+  tibble::column_to_rownames("Chromosome Arm") %>%
+  t() %>%
+  as.data.frame() %>%
+  tibble::rownames_to_column("Kids_First_Biospecimen_ID") %>%
+  dplyr::left_join(select_metadata, by = "Kids_First_Biospecimen_ID") %>%
+  dplyr::filter(sample_id %in% atrt_df$sample_id) %>%
```
jaclyn-taroni: I think you can move up the filtering step such that you are removing non-ATRT biospecimens from the GISTIC data and then join with the metadata. In general if you can filter before joining, that is a better design pattern because you end up working with smaller `data.frame`.

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
```diff
@@ -168,14 +176,19 @@ metadata <- metadata %>%
     age_at_diagnosis_days,
     germline_sex_estimate,
     primary_site
-  )
+  ) %>%
+  dplyr::group_by(
+    sample_id,
+    Kids_First_Participant_ID
+  ) %>%
+  dplyr::summarize_all(function(x) paste(sort(unique(x)), collapse = ", "))
```
jaclyn-taroni: @cbethell if we do this collapse biospecimen IDs step up front, we can drop the biospecimen ID columns as we go since we join on `sample_id`. This means we don't have to collapse at the end and run the risk of having a bunch of duplicated columns.

analyses/molecular-subtyping-ATRT/00-subset-files-for-ATRT.R
```diff
@@ -81,26 +81,31 @@ tmb_df <-
...
 
-# Filter metadata for `ATRT` and define `location_summary` based on values in
-# `primary_site`
 atrt_df <- metadata %>%
   dplyr::filter(short_histology == "ATRT",
-                experimental_strategy == "RNA-Seq")
-
-# Write to file
-readr::write_tsv(atrt_df, file.path(results_dir, "atrt_histologies.tsv"))
+                sample_type == "Tumor",
```
jaclyn-taroni: Filtering by `experimental_strategy == "RNA-Seq"` also effectively drops normal samples, but it also made it such that any ATRT samples missing RNA-seq data got dropped. We don't need to write the ATRT subset histologies file as a TSV because the CI histologies file is always the same file as a data release: 

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2cbf526e24d92c11f093b33aa31663546b3fbb57/analyses/create-subset-files/create_subset_files.sh#L41

The reason why we can include the full `pbta-histologies.tsv` for testing is two-fold: 1) it's pretty small and 2) having "extra" biospecimens/samples/participants in the histologies file helps test for brittle code.

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
```diff
@@ -52,14 +52,18 @@ if (!dir.exists(results_dir)) {
...
 
+# Read in non-subsetted metadata
+metadata <- readr::read_tsv(file.path(root_dir, "data", "pbta-histologies.tsv"))
+
 # Read in ATRT subset metadata
-metadata <-
-  readr::read_tsv(file.path(input_dir, "atrt_histologies.tsv"))
+subset_metadata <- metadata %>%
+  dplyr::filter(short_histology == "ATRT",
+                sample_type == "Tumor",
```
jaclyn-taroni: Getting rid of the normal and cell line data here.

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
```diff
@@ -241,50 +254,60 @@ long_stranded_expression <- scale(t(filtered_expression),
...
   head(n = 15)
 
 # Join expression data with metadata filtered for `ATRT`
-atrt_expression_df <- metadata %>%
+atrt_expression_df <- collapsed_metadata %>%
   dplyr::left_join(expression_metadata,
                    by = "sample_id")
+
+# Remove data we no longer need
+rm(filtered_expression, long_stranded_expression, expression_metadata)
```
jaclyn-taroni: Removing from the workspace as we go makes it such that we don't have `data.frame` we no longer need in memory.

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
```diff
@@ -241,50 +254,60 @@ long_stranded_expression <- scale(t(filtered_expression),
...
-    )), collapse = ", ")
-  )
-
-# Display `cn_metadata`
-cn_df %>%
+  tidyr::spread(gene_symbol, status, fill = "neutral") %>%
+  dplyr::rename(SMARCB1_focal_status = SMARCB1,
+                SMARCA4_focal_status = SMARCA4)
+
+# add in the samples that did not have any copy number changes for either
```
jaclyn-taroni: Before, if a biospecimen did not have evidence for an alteration in either _SMARCB1_ or _SMARCA4_, it would look like it was "missing" in the final data.frame. This is likely because we filter out neutral events here: 

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2cbf526e24d92c11f093b33aa31663546b3fbb57/analyses/focal-cn-file-preparation/01-prepare-cn-file.R#L262

to save on memory for the annotation step and disk space when we save the prepped file. So here, we assume that if a sample appears _anywhere_ in the CNVkit file that we did assay it but that there was no evidence of a copy number change so we mark it neutral.

@jharenza you should be aware of this assumption as you look at the table! Not sure how common it is to essentially 'miss' segments when calling copy number from WGS and therefore _not_ assay  _SMARCB1_ and _SMARCA4_ for those samples...

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
```diff
@@ -329,44 +355,82 @@ tmb_df %>%
...
+rm(tmb_df, atrt_expression_cn_df)
+```
+
+### GISTIC data
+
+```{r}
+gistic_df <- gistic_df %>%
+  dplyr::select(-Kids_First_Biospecimen_ID) %>%
+  dplyr::group_by(sample_id) %>%
+  dplyr::mutate(chr_22q_loss = dplyr::case_when(`22q` == -1 ~ "Yes",
```
jaclyn-taroni: @jharenza we only have `-1` and `0` values in `22q` for these samples: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/7f012635616c171531b627606f6607874db04d30/analyses/molecular-subtyping-ATRT/atrt-subset/atrt_gistic_broad_values.tsv

Unless GISTIC itself incorporates the tumor ploidy information from `pbta-histologies.tsv`, we are not using any info about ploidy here. I think given that a `-1` would be a loss even when ploidy is 2, this might be fine but wanted to make you aware.

analyses/molecular-subtyping-ATRT/00-subset-files-for-ATRT.R
cbethell: Yes, you are correct.

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
jharenza: yes, GISTIC only uses -2, -1, 0, 1, 2 to my knowledge

analyses/molecular-subtyping-ATRT/01-ATRT-molecular-subtyping-data-prep.Rmd
jharenza: thanks for the heads up! The normal output from segmentation algorithms don't contain neutral segments, so this makes sense.

PR #406, opened 2020-01-06
PR 2 of n: Molecular Subtyping - Subset files for HGG script

cbethell:
### Purpose/implementation Section

The purpose of this PR is to subset the files for the HGG molecular subtyping scripts. 

#### What scientific question is your analysis addressing?

This analysis addresses the molecular subtyping of High-grade glioma samples question.

#### What was your approach?

I subset the histologies, focal copy number, fusion, and RNA expression files using the `sample_id` variable to match samples classified as HGG samples in the first notebook of this module `01-HGG-molecular-subtyping-defining-lesions.Rmd`. 

#### What GitHub issue does your pull request address?

This PR addresses issue #249.

### Directions for reviewers. Tell potential reviewers what kind of feedback you are soliciting.

- Does this analysis appear to be correct?
- Are there any areas that can be refactored?

#### Is there anything that you want to discuss further?

Note, this script will not be added to `.circleci`. 
Therefore, I have added a note to this module's `README` re the generation of the files in the `hgg-subset` directory. 
I have also added a shell script to run all the scripts in this module while accounting for the subset script in this PR (which would fail the `circleci` tests if ran).

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

Yes, this analysis is ready for review. 

#### What types of results are included (e.g., table, figure)?

The results of this PR are files subset for HGG samples as follow:
- hgg-subset/hgg_focal_cn.tsv.gz
- hgg-subset/hgg_fusion.tsv
- hgg-subset/hgg_histologies.tsv
- hgg-subset/hgg_zscored_expression.RDS

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.
- [x] This analysis is recorded in the table in `analyses/README.md`.

Issue #249, opened 2019-11-08
Proposed Analysis: Molecularly subtype high-grade glioma tumors


jharenza:
#### Scientific goals
_What are the scientific goals of the analysis?_
Subtype High-grade gliomas (high-grade astrocytic tumors, HGAT), according to defining histone lesion.

Note: historically, amino acid 1 was not counted in protein nomenclature, thus K27 and G34 mutations noted above really translate to K28 and G35. See:
[ActaNeuropathol.2018Leske.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3825553/ActaNeuropathol.2018Leske.pdf)
Our data maps to K28 and G35, consistent with COSMIC and pedcbioportal. 
- Fun fact: this was also the case with BRAF V600E (it was historically called V599E, but the community adapted the current nomenclature).

#### Proposed methods
_What methods do you plan to use to accomplish the scientific goals?_
1) Review of mutation, copy number, and expression data.
2) Render results in tabular form in a notebook.

Any sample that harbors _H3F3A_ K28M, _H3F3A_ G35R/V_, or _HIST1H3B_ K28M, if not classified as a high-grade glioma, should be, as this is a defining lesion. Eg: three former PNET tumors were re-classified as such.

**H3 K28 mutant**
- These tumors contain _H3F3A_ K28M or _HIST1H3B_ K28M mutations
- Co-occurring lesions include: _ACVR1_, _TP53_, _ATRX_ mutations; _PDGFRA_ amplification; _PTEN_ loss
- Mutually-exclusive lesions: _FGFR1_ mutations/fusions (thalamic); _IDH1_ mutations; _BRAF V600E_ (low-grade gliomas)
- Average age of 9 years
- Majority should be midline localized

**H3 G35 mutant**
- These tumors contain _H3F3A_ G35R/V mutations
- Co-occurring lesions include: _ATRX/DAXX_, _TP53_, _SETD2_ mutations, _NTRK_ fusions
- Mutually exclusive lesions: _IDH1_ mutations
- Average age of 20 years

**IDH mutant**
- These tumors contain _IDH1_ R132H mutations
- Co-occurring lesions include: _TP53_ mutations; P73-AS1  promoter methylation and downregulation
- High expression of _FOXG1_ and _OLIG2_
- Mutually exclusive lesions: chr7 gain and chr10 loss

**H3.3 and IDH wildtype**
[...]

PR #420, opened 2020-01-09
PR 3 of n: Molecular Subtyping - HGG (Data Prep)

cbethell:

The purpose of this PR is to wrangle data for the purpose of molecularly subtyping high-grade glioma samples. 

This analysis is addressing the question of which high-grade glioma samples fit into the high-grade glioma subtypes: H3 K28 mutant, H3 G35 mutant, IDH mutant, H3.3 and IDH wildtype, and 1p/19q co-deleted oligodendrogliomas.

#### What was your approach?

I joined together the relevant HGG subsetted files (from PR #406) and combined them in a final table by matching the `sample_id` values. 

This PR addresses issue #249.

- Does this analysis appear to be correct?
- Do the columns in the final table appear to represent the data in the best manner?

Yes, this analysis is ready for review. 

This PR produces a final tsv file as follows:
- `results/HGG_molecular_subtypes.tsv`

There are currently a significant amount of `NA`s in the final table due to data missing for certain `sample_id`s in the focal CN file. 
I am going over this more carefully to ensure that this is correct and not a matter of something going wrong in the workflow. 

- [x] This analysis is recorded in the table in `analyses/README.md` and is up to date.


jaclyn-taroni:
Okay - some observations! There are many (more than 10) instances of duplicate `sample_id`. In these cases, we can not use the `sample_id` to map between different experimental strategies. So instead, I propose use the `Kids_First_Biospecimen_ID` to join all the DNA-seq data together. That table is one output. We then drop all `sample_id` that are duplicated/ambiguous and then we can add the RNA-seq information. In essence, we will only have the "full set" of relevant data including fusions for those cases where the `sample_id` is not ambiguous.

The other thing that I'm noticing that there are going to be a lot of decisions around each data type. For example, when we talk about a _TP53_ mutation being mutually exclusive do we only mean _TP53_ coding mutations that are not silent? Does the orientation for some of the fusions matter? 

Unfortunately, I think that means that having the required scientific discussion is going to be difficult if we perform all of the processing in a single notebook. Instead - I propose that each data type (e.g., copy number, consensus mutation, fusion, etc.) is prepared in a single script and that those scripts go into individual pull requests followed by a final notebook that joins everything together.

Sorry this is considerably gnarlier than anticipated @cbethell - I will close this pull request now and I will plan to do the following this weekend:
- Open a new one with just the changes to defining lesions and the file subsetting
- Open a copy number pull request that will probably include the GISTIC data cleaning as well because I feel good about where those sections are at as of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/420/commits/0d2b08ce38e000e7852819982ec1333c4fabdd23
- Open a SNV pull request where I make the assumption that we only want to look at mutations we expect to be functional to get the conversation started

Then you and I can touch base on Monday about how to chunk out the rest of the work.

Issue #392, opened 2020-01-02
Discussion: consensus CN files for downstream analyses


jaclyn-taroni:
**Related to the generation of copy number consensus calls: #128**

Downstream analysis of copy number data, including analysis with GISTIC (#8), often relies on information contained in a SEG file outside of the region, e.g., `seg.mean` from CNVkit. @fingerfen is generating consensus calls from CNVkit, ControlFreeC, and Manta.

From @hongboxie https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/128#issuecomment-555521330:

> We will present two types of CNVs: Amplification (ploity>2) or Deletion (ploidy<2). We created 3 additional columns where we display original outcome of each method overlapping this region as evidence. The original output of each predictor will carry the information of the ploidy, if it has it.

My interpretation of this comment is that some of the required information for downstream analysis will be retained. 

I am filing this ticket to facilitate discussion around how we will use the copy number consensus results for downstream analysis. Because [GISTIC is currently run on the CNVkit data](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#copy-number-files) (and in my pretty limited experience there seems to be a decent amount of tooling for using CNVkit output), subsetting the CNVkit SEG file to only those regions that have support from the other methods and using that when `seg.mean` values are required is what comes to mind.

I am also tagging @jashapiro who has been reviewing the consensus code and @jharenza.

jharenza:
Hi @jaclyn-taroni - I think this sounds like a reasonable approach for CNVkit results/analyses requiring SEG files. Would be happy to rerun GISTIC once we have these consensus calls.

hongboxie:
@jaclyn-taroni I think it sounds reasonable. Do you have overall plan for downstream analysis of CNV data?

jaclyn-taroni:
@hongboxie broadly speaking, we'll want to run GISTIC (#8) and plot recurrent alterations in an OncoPrint (#6; though this does not require the segment mean values; also somewhat related to #394). I am assuming we probably want to integrate the CNV data with the SV data for visualization purposes. I've found this [UMCCR pipeline code](https://github.com/umccr/rock) that uses Manta and CNVkit output to create circos plots. I bet we can adapt it (or use it to guide something similar) to use the "consensus filtered" CNVkit data mentioned here, the AnnotSV-annotated Manta file, and [`circlize`](https://github.com/jokergoo/circlize).

[...]

PR #430, opened 2020-01-13
CNV consensus (8 of 6): Changing Step3 to use Bedtools Subtract

fingerfen:

Increase code run time and efficiency

#128 

The changes in the cnv_consensus.tsv. The Manta CNV order changed for one singular consensus CNV call. 

#### Documentation Checklist


fingerfen:
@jashapiro I just got rid of the `get_rid_bad_segments.py` file. Sad but a necessary step 😃 . 

As for the discussion around #392, I just had a look and it seems that, just like you said, we will have to make adjustment to the `awk` command by giving cnvkit its `seg.mean` column and give each of the other two callers a column of `NA`. We then have to change `first_merge` and `restructure_column.py`. With this, the format of any CNV in columns 4, 5, and 6 would be `chr:start:end:copy_number:seg.mean`. 

Sadly, my break ended last Monday and I have been back in school for a week now. That's why it took a while for me to make this PR. However, I would like to see my project until the end so I think I can handle the implementation of `seg.mean` this week and have you look over it. Will this be part (9 of 6)?

jashapiro:
> Sadly, my break ended last Monday and I have been back in school for a week now. That's why it took a while for me to make this PR. However, I would like to see my project until the end so I think I can handle the implementation of `seg.mean` this week and have you look over it. Will this be part (9 of 6)?

Okay, I will look forward to seeing your changes. You can just submit them as a new PR: no need to keep up the numbering if you don't want to!

PR #434, opened 2020-01-14
Add seg.mean (1 of 1)

fingerfen:

Carry the seg.mean column from Cnvkit calls throughout the entire analysis

#392 

#### Documentation Checklist


analyses/copy_number_consensus_call/Snakefile
```diff
@@ -83,10 +83,10 @@ rule manta_filter:
         ## The first awk looks at column 6 to filter out for loss/gain (DEL/DUP). Then it prints out 6 of the 7 columns above. Put NA for both p-value and copy number since MANTA results don't have these values.
         ## the first awk also filters out for CNV length
         ## The last pipe is to sort first digit of chromosome number numerically
-        """awk '$6~/DEL/ {{if ($5 > {params.SIZE_CUTOFF}) {{print "chr"$2,$3,$4,$5,"NA","NA",$6}}}}' {input} """
+        """awk '$6~/DEL/ {{if ($5 > {params.SIZE_CUTOFF}) {{print "chr"$2,$3,$4,$5,"NA","NA","NA",$6}}}}' {input} """
```
jashapiro: Is this the correct order? I would assume the added `NA` for manta should be at the end, as with freec?

jashapiro: Never mind... I see my error now.

analyses/copy_number_consensus_call/Snakefile
```diff
@@ -118,7 +118,7 @@ rule first_merge:
         ## Any condiders any segments within 10,000 bp to be the same CNV. 
         ## Merge but retain info from columns 2 (start pos), 3(end pos), 5(copy numbers), 7(CNV type)
```
jashapiro: Update comment here to include seg.mean

analyses/copy_number_consensus_call/Snakefile
```diff
@@ -33,11 +33,11 @@ rule freec_filter:
         ## The first awk looks at column 6 to filter out for loss/gain. Then it prints out 6 of the 7 columns above
         ## The pipe into the second awk filters the CNV length, freec pval, and add in the CNV type
         ## The last pipe is to sort first digit of chromosome number numerically
-        """awk '$6~/loss/ {{print "chr"$2,$3,$4,($4-$3 + 1),$5,$9}}' {input.events} """
+        """awk '$6~/loss/ {{print "chr"$2,$3,$4,($4-$3 + 1),$5,$9,"NA"}}' {input.events} """
         """ | awk '{{if ($4 > {params.SIZE_CUTOFF} && $6 < {params.FREEC_PVAL}){{print $0,"DEL"}}}}' """
```
jashapiro: For clarity, perhaps we should move the "DEL" addition here (and "DUP", similarly) to the first awk statement. So this could become:
```
"""awk '$6~/loss/ {{print "chr"$2,$3,$4,($4-$3 + 1),$5,$9,"NA","DEL"}}' {input.events} """
""" | awk '{{if ($4 > {params.SIZE_CUTOFF} && $6 < {params.FREEC_PVAL}){{print}}}}' """
```
The same simplification could be applied to the CNVkit awk statements.

analyses/copy_number_consensus_call/Snakefile
```diff
@@ -33,11 +33,11 @@ rule freec_filter:
         ## The first awk looks at column 6 to filter out for loss/gain. Then it prints out 6 of the 7 columns above
```
jashapiro: Line 32 comment should be updated to reflect the new output format and this line updated to match.
These comments can also be broken up so the lines are not so wide (100 characters is a good max, but 80-90 is better)

Similarly, update lines 57-58 and 82-83

analyses/copy_number_consensus_call/Snakefile
fingerfen: It is the correct order. I had the same question, :) . For Cnvkit and Freec, I need to add in `"DUP"` or `"DEL"` as the last column using the 2nd `awk` command, so you see the `"NA"` last. However, for Manta, the 6th column of the raw file is the DUP or DEL, so it should go like `"NA","NA","NA",$6`.

fingerfen:
@jashapiro Thanks for the suggestions. I have made all of the changes and updated the README.md file. I am now re-running the entire pipeline to see that it works.

PR #428, opened 2020-01-11
PR 7 of n - Molecular Subtyping - HGG (Cleaning mutation data)

jaclyn-taroni:
### Purpose/implementation Section

Per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/420#issuecomment-573257980, this pull request is adding a notebook that cleans the relevant mutation data for the purposes of subtyping HGG samples.

I've elected to use a notebook in the interest of documenting the decisions made here.

*View the rendered notebook here: https://jaclyn-taroni.github.io/openpbta-notebook-concept/04-HGG-molecular-subtyping-mutation.nb.html*

#### What GitHub issue does your pull request address?

#249

### Directions for reviewers. Tell potential reviewers what kind of feedback you are soliciting.

Because this depends on #426, you will see those changes in this branch. For the moment, here is where I would like you to direct your attention:

| File | Description | Relevant Commit(s) |
|-----|-------------|---------------------|
| `analyses/molecular-subtyping-HGG/04-HGG-molecular-subtyping-mutation.Rmd` | Notebook for cleaning the mutation data | https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/05551142036f8cecb8636f28673ccb0eec1ed8cb, https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/9fe7f972507116b8456fc7508ce40c9b528e3951, https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/49b7e4817a677c7aa1e5ec9c1870ccb3fd00dd9c, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/428/commits/71594e45db69d6aa7a0402001269a121358be7f1, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/428/commits/2df0f882af61418ca6525f3132c32e46140bca10 |
| `analyses/molecular-subtyping-HGG/run-molecular-subtyping-HGG.sh` | Shell script for running analysis module | https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/f31baace4abd7275f6d1bd62b53b65788d081935, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/428/commits/a6e39caa47eb640b34f8384b4b6e67b38f0b5d7f, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/428/commits/a66f2cd47581d3d6d58992ae0a8c5967af91c28e |
[...]

PR #427, opened 2020-01-11
PR 6 of n - Molecular Subtyping - HGG (Cleaning CNV data) 

jaclyn-taroni:
### Purpose/implementation Section

Per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/420#issuecomment-573257980, this pull request is adding a notebook that cleans the relevant copy number data for the purposes of subtyping HGG samples.

Here I've elected to use a notebook in the interest of documenting the decisions made here and in the file preparation that happens upstream of this notebook itself.

*View the rendered notebook here: https://jaclyn-taroni.github.io/openpbta-notebook-concept/03-HGG-molecular-subtyping-cnv.nb.html*

#### What GitHub issue does your pull request address?

#249 

### Directions for reviewers. Tell potential reviewers what kind of feedback you are soliciting.

Because this depends on #426, you will see those changes in this branch. For the moment, here is where I would like you to direct your attention:

| File | Description | Relevant Commit(s) |
|-----|-------------|---------------------|
| `analyses/molecular-subtyping-HGG/03-HGG-molecular-subtyping-cnv.Rmd` | Notebook for cleaning CNV data | https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/f846d20d0289b47904dd5f16f0ea5de2f537f8b3, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/427/commits/c3005bfeab3e263dae809f9de09716997418d9bc, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/427/commits/1408905bba5bf4eaa8dd5c50ceafc7539240ef51 |
| `analyses/molecular-subtyping-HGG/run-molecular-subtyping-HGG.sh` | Shell script for running analysis module | https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/9f7efde72abe531b025f265963d57771ff69f71f |
| `analyses/molecular-subtyping-HGG/results/HGG_cleaned_cnv.tsv` | Output table | https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/f846d20d0289b47904dd5f16f0ea5de2f537f8b3, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/427/commits/d2784aa9ecf1eb1965da6be25f8e50a9d62b3a69, https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/427/commits/614ca962dbad079d721d070f590c1527302bd188 |

#### Which areas should receive a particularly close look?

[...]

Issue #260, opened 2019-11-12
`age_at_diagnosis_days` column contained different information prior to v6


jaclyn-taroni:
#### What data file(s) does this issue pertain to?

`pbta-histologies.tsv`

#### What release are you using?

I am comparing `release-v5-20190924` to the current release, `release-v9-20191105`.

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.

There is nothing in the data harmonization section about this specifically: https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#clinical-data-harmonization yet, but it is related to this open PR: https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/55/files#diff-1958beef0777ea964ffe394a82c98903R163

#### Put your question or report your issue here.

For context, the [`independent-samples`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples
) module was developed using `release-v5-20190924`. @jashapiro used the `age_at_diagnosis_days` column to identify the _earliest_ samples:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/9c68671a4cba89681046c026eb1658794165e836/analyses/independent-samples/independent-samples.R#L48

I was looking to use a similar strategy as part of the participant-level merging in the Oncoprint pipeline (#243), but the information content of this column has changed. My intuition is that it used to contain something akin to age at timepoint and now it contains, as it's name suggests the age at diagnosis.

In the v9 histologies file, there are no instances of multiple ages being associated with the same participant ID:

```r
> v9_histologies_df <- readr::read_tsv("data/release-v9-20191105/pbta-histologies.tsv")
> v9_ages <- v9_histologies_df %>% 
    group_by(Kids_First_Participant_ID) %>% 
    summarize(ages = paste(sort(unique(age_at_diagnosis_days)), collapse = ", "))
> sum(grepl(",", v9_ages$ages))
[1] 0
``` 
In v5, that was not the case:

```r
> v5_histologies_df <- readr::read_tsv("data/release-v5-20190924/pbta-histologies.tsv")
> v5_ages <- v5_histologies_df %>% 
    group_by(Kids_First_Participant_ID) %>% 
    summarize(ages = paste(sort(unique(age_at_diagnosis_days)), collapse = ", "))
[...]

PR #435, opened 2020-01-15
PR 10 of n - Molecular Subtyping - HGG (Joining cleaned data)

jaclyn-taroni:

Here I am joining copy number and mutation data that was previously cleaned in #427 and #428 for the purpose of subtyping HGG samples. Some things to note:

* We seem to have gotten to the bottom of the duplicate `sample_id` issue mentioned [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/249#issuecomment-573438044) with the filtering changes in #426. I think that means we can deviate from the plan outlined in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/249#issuecomment-573438044 and join all the data together in this notebook. I am happy to finish that up with this pull request and report the results accordingly - what do you think @cbethell?
* There is not CNVkit data (and therefore GISTIC data) for the WXS samples, so we're missing copy number information for 32 out of 158 biospecimens. 

#249

#### Documentation Checklist


jaclyn-taroni:
Okay - I am going to update this to produce the final table!

jaclyn-taroni:
Here's the current version of the rendered notebook: https://jaclyn-taroni.github.io/openpbta-notebook-concept/07-HGG-molecular-subtyping-combine-table.nb.html

And this is what the final product currently looks like: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d00d79db625f8849d5ecd8db0073ffad8c1c1ade/analyses/molecular-subtyping-HGG/results/HGG_molecular_subtype.tsv

I made the following decisions about the `molecular_subtype` column:
* If there was an _H3F3A_ K28M or _HIST1H3B_ K28M mutation
  * No _BRAF_ V600E mutation -> `HGG_H3_K28`
  * _BRAF_ V600E mutation -> `BRAF_V600E_mutation`
* If there was an _H3F3A_ G35V or G35R mutation -> `HGG_H3_G35`
* If there was an _IDH1_ R132 mutation -> `HGG_IDH` 

All other samples were not include in the `HGG_molecular_subtype.tsv` file. The data for the other subtypes were less consistent with what was on #249, so I think that will require more expertise and possibly additional data if we would like to make those calls.

@jharenza can you take a look at the rendered notebook and the final table and give me your thoughts? Do you agree with my logic? Is the final table in a form you could use to add to `pbta-histologies.tsv`?

jharenza:
Hi @jaclyn-taroni! These look great. 

- One thing I realized when going through `HGG_defining_lesions.tsv` - we should probably also update the `short_histology` and `broad_histology` in addition to creating the `disease_type_reclassified` columns. That way, we can include in the next clinical file as updates. This is really only relevant for the few samples that completely changed disease type, but would be good to capture here.
- For `molecular_subtype` and `disease_type_reclassified`, it was recommended by Angela Waanders (disease expert), to use the terminology `DMG` for diffuse midline glioma for any sample with a histone mutation, rather than `HGG` for high-grade glioma, to go along with the WHO 2016 classifications.
- For IDH mutant, let's keep the `HGG_IDH` subtype for now - Angela wants to have a call to determine what to actually call this, so TBD.
- In the `HGG_molecular_subtype.tsv` table, I noticed the `BRAF_V600E_mutation molecular_subtype`, which I wrote should be _mutually exclusive_  of _H3F3A_ lesions. I dug into the literature some more and found in the [Jones cohort](https://pedcbioportal.kidsfirstdrc.org/results?cancer_study_list=phgg_jones_meta_2017&tab_index=tab_visualize&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=phgg_jones_meta_2017_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=phgg_jones_meta_2017_gistic&case_set_id=phgg_jones_meta_2017_all&Action=Submit&gene_list=H3F3A%0ABRAF), this combination was seen in 1.8% of DIPG samples, as well as [rarely](https://theoncologist.onlinelibrary.wiley.com/doi/pdf/10.1634/theoncologist.2017-0242) in some pediatric glioblastomas. I got an expert opinion on how to classify from Angela Waanders, and she suggests _nothing_ is mutually exclusive, just "rare". So, we should drop that consideration and instead of `BRAF_V600E_mutation` for molecular subtype, let's make that `DMG_H3_K28; BRAF_V600E" (or concatenate in some way).
- For `HGG_cleaned_all_table.tsv` and `HGG_cleaned_mutation.tsv`, we could add _SETD2_ mutations (co-occurring lesion of H3 G35) for completion.
- A few patients PT_3AWKWXEV, PT_P5HHJJPH, PT_T8V9ES93, and PT_Y8RNS9JE are not in the molecular subtype or the cleaned table, initially had dx of HGG, have no histone mutations, but have only _BRAF V600E_ mutations. These patients' biospecimens can possibly be LGGs. I think for these, we should note that they are exclusively `BRAF_V600E_ mutant` in the subtype, but they perhaps need a disease type rethink. Maybe we should see if their particular RNA samples cluster with LGG instead of HGGs to make that determination? I had not considered that when writing up the ticket.
- For H3.3 and IDH WT classification, I am going to have to dig into these more to assess what rules we should use to classify.
- I think I would have expected that for all DNA biospecimens in the `HGG_cleaned_all_table.tsv`, we would have one entry in the `HGG_molecular_subtype.tsv`, which would include the wildtype biospecimens. Am I missing something?
- Once those are completed, I think we may have a v2 of the `HGG_defining_lesions.tsv` table, based on the comment above in which there are both _H3 K28M_ and _BRAF V600E_ mutations in one sample, making it a unique, rare subtype.

jaclyn-taroni:
I'm going to likely end up responding to these points one by one.

> For HGG_cleaned_all_table.tsv and HGG_cleaned_mutation.tsv, we could add SETD2 mutations (co-occurring lesion of H3 G35) for completion.

We do include _SETD2_: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/fa214291713575be7fd20c92374b268870f4173f/analyses/molecular-subtyping-HGG/04-HGG-molecular-subtyping-mutation.Rmd#L109 

There are several instances where there is a coding mutation in _SETD2_. It is possible that the mutations may not be in the samples where such a mutation is expected. If we look at the `HGG_cleaned_mutation.tsv` table on GitHub: 
https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a7b97825b5be987cc31edae45a963bc1831c7ef9/analyses/molecular-subtyping-HGG/results/HGG_cleaned_mutation.tsv

We can see that `PT_MDWPRDBT` doesn't have any of the defining lesions, but does have a coding mutation in _SETD2_.

jharenza:
Re: 

```
1p/19q co-deleted oligodendrogliomas
This is not particularly straightforward, so I will leave that out for now. Possible next steps: only those with amplifications or gains in the two mentioned genes?
``` 

We expect very few, if any of these (maybe 1 when we looked by CNV), so I think maybe we can approach this cohort by starting to look for 1p/19q codeletions in all samples + IDH mutations, and if we see any, we may manually want to check the CNV plots just to confirm.

jharenza:
> I'm going to likely end up responding to these points one by one.
> 
> > For HGG_cleaned_all_table.tsv and HGG_cleaned_mutation.tsv, we could add SETD2 mutations (co-occurring lesion of H3 G35) for completion.
> 
> We do include _SETD2_:
> 
> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/fa214291713575be7fd20c92374b268870f4173f/analyses/molecular-subtyping-HGG/04-HGG-molecular-subtyping-mutation.Rmd#L109
> 
> There are several instances where there is a coding mutation in _SETD2_. It is possible that the mutations may not be in the samples where such a mutation is expected. If we look at the `HGG_cleaned_mutation.tsv` table on GitHub:
> https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a7b97825b5be987cc31edae45a963bc1831c7ef9/analyses/molecular-subtyping-HGG/results/HGG_cleaned_mutation.tsv
> 
> We can see that `PT_MDWPRDBT` doesn't have any of the defining lesions, but does have a coding mutation in _SETD2_.

ok great - I think I missed seeing that because it was not in `HGG_cleaned_all_table.tsv`.

jaclyn-taroni:
I think these two points may be related

> A few patients PT_3AWKWXEV, PT_P5HHJJPH, PT_T8V9ES93, and PT_Y8RNS9JE are not in the molecular subtype or the cleaned table, initially had dx of HGG, have no histone mutations, but have only BRAF V600E mutations. These patients' biospecimens can possibly be LGGs. I think for these, we should note that they are exclusively BRAF_V600E_ mutant in the subtype, but they perhaps need a disease type rethink. Maybe we should see if their particular RNA samples cluster with LGG instead of HGGs to make that determination? I had not considered that when writing up the ticket.

```r
> cleaned_all <- read_tsv("analyses/molecular-subtyping-HGG/results/HGG_cleaned_all_table.tsv")
Parsed with column specification:
cols(
  .default = col_character(),
  age_at_diagnosis_yr = col_double(),
  FOXG1_expression_zscore = col_double(),
  OLIG2_expression_zscore = col_double(),
  TP73_AS1_expression_zscore = col_double()
)
See spec(...) for full column specifications.

> all(c("PT_3AWKWXEV", "PT_P5HHJJPH", "PT_T8V9ES93", "PT_Y8RNS9JE") %in% cleaned_all$Kids_First_Participant_ID)
[1] TRUE
```
Is `HGG_cleaned_all_table.tsv` what you mean by cleaned table?

> I think I would have expected that for all DNA biospecimens in the HGG_cleaned_all_table.tsv, we would have one entry in the HGG_molecular_subtype.tsv, which would include the wildtype biospecimens. Am I missing something?

Regarding, the molecular subtype TSV from https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/435#issuecomment-574857277

> I made the following decisions about the molecular_subtype column:
> 
>     If there was an H3F3A K28M or HIST1H3B K28M mutation
>         No BRAF V600E mutation -> HGG_H3_K28
>         BRAF V600E mutation -> BRAF_V600E_mutation
>     If there was an H3F3A G35V or G35R mutation -> HGG_H3_G35
>     If there was an IDH1 R132 mutation -> HGG_IDH
> 
> All other samples were not include in the HGG_molecular_subtype.tsv file.

So there are some samples missing by design in this version of the file.

jaclyn-taroni:
> One thing I realized when going through HGG_defining_lesions.tsv - we should probably also update the short_histology and broad_histology in addition to creating the disease_type_reclassified columns. That way, we can include in the next clinical file as updates. This is really only relevant for the few samples that completely changed disease type, but would be good to capture here.

Can you be more specific? What would go in those columns?

jharenza:
> ```r
> PT_3AWKWXEV
> ```

Sorry I meant `HGG_molecular_subtype.tsv` - they are not in this table and based on your logic, it makes sense why - perhaps we should make another table of samples not included in the HGG subtype table so we can go through those and determine what they should be? 

For samples with histone mutations and BRAF mutations, `BRAF V600E mutation -> BRAF_V600E_mutation`, I was suggesting that we make that subtype `DMG_H3_K28M_BRAF_V600E` or some version of that, in which we can see both drivers in the subtype.

jharenza:
> > One thing I realized when going through HGG_defining_lesions.tsv - we should probably also update the short_histology and broad_histology in addition to creating the disease_type_reclassified columns. That way, we can include in the next clinical file as updates. This is really only relevant for the few samples that completely changed disease type, but would be good to capture here.
> 
> Can you be more specific? What would go in those columns?

What would go into those would be exactly what is in the clinical file now. For ex, there is an ependymoma that has changed to HGG/DMG, so its `short_histology` would go from `Ependymoma->HGAT` and `broad_histology` from `Ependymal tumor->Diffuse astrocytic and oligodendroglial tumor` (all of these will have HGAT for short, and Diffuse astrocytic* for broad).

jaclyn-taroni:
> For samples with histone mutations and BRAF mutations, BRAF V600E mutation -> BRAF_V600E_mutation, I was suggesting that we make that subtype DMG_H3_K28M_BRAF_V600E or some version of that, in which we can see both drivers in the subtype.

Yes I can do this and I'll also add samples (currently) without a subtype into `HGG_molecular_subtype.tsv` with `NA`.

I'll make the rest of the requested changes now.

jaclyn-taroni:
I think I've captured all the requested changes and I've filed #466 to look specifically at the 1p/19q co-deleted oligodendrogliomas

jaclyn-taroni:
Okay @jharenza, think I have finally gotten through all the v13 and consensus copy number related updates here! 

The notebook here should be current but may take a few more minutes to get published via GH pages: https://jaclyn-taroni.github.io/openpbta-notebook-concept/07-HGG-molecular-subtyping-combine-table.nb.html

Cleaned table is here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/92c2bc37b2c867e3dbfb951b1d5ee8d6af83f305/analyses/molecular-subtyping-HGG/results/HGG_cleaned_all_table.tsv

jharenza:
Hi @jaclyn-taroni, I did not catch this at the time I made the ticket, but just reading again - [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6562757/) suggests that there can be rare H3 K28M mutations in _HIST1H3B_, _HIST1H3C_, or _HIST2H3C_, so I think we should add the latter two to the analysis as well when you get a chance. We may not see anything here, but good to be comprehensive.

_"The K27M mutation primarily occurs in H3F3A (>70%), one of the two genes encoding H3.3, with a low frequency occurrence in H3.1 encoded by HIST1H3B or HIST1H3C, and rarely in H3.2 (HIST2H3C) [5,6,40,41]."_

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,529 @@
...
+The coding mutation list looks like what we would expect based on the description, with the exception of some BRAF mutations.
+There are no _IDH1_ mutations.
+
+```{r}
+h3_k28_df %>%
+  group_by(PTEN_focal_status, PDGFRA_focal_status) %>%
+  count()
+```
+
+The copy number information looks a bit different than what is described, but we are not yet using consensus calls.
```
jharenza: This looks fine - when I wrote in co-occurring, those are just _**some**_ of the lesions that had been reported as co-occurring, as opposed to lesions that must/usually co-occur.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,529 @@
...
+```{r}
+all_data_df %>%
+  filter(`1p` == "loss" | `19q` == "loss")
+```
+
+We don't see the _IDH1_ mutations we would expect and we do see an H3 K28M mutation, so we will leave this off as well for the moment and in a subsequent notebook we'll address [`AlexsLemonade/OpenPBTA-analysis#435 (comment)`](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/435#issuecomment-576898275):
+
+> We expect very few, if any of these (maybe 1 when we looked by CNV), so I think maybe we can approach this cohort by starting to look for 1p/19q codeletions in all samples + IDH mutations, and if we see any, we may manually want to check the CNV plots just to confirm.
+
+### Samples without histone mutations that have _BRAF_ V600 mutations
```
jharenza: Can we add a visualization to highlight these samples in a tsne or umap clustering of all of the data by broad histology? If they cluster with the LGAT, then I think we can re-categorize them. Maybe this is a separate notebook.

analyses/molecular-subtyping-HGG/README.md
```diff
@@ -54,6 +54,16 @@ This notebook produces a fusion results table found at `results/HGG_cleaned_fusi
...
+This notebook produces one table with the cleaned data found at `results/HGG_cleaned_all_table.tsv`.
+A table with the molecular subtype information for each HGG sample at `results/HGG_molecular_subtype.tsv` is also produced, where the subtype values in the `molecular_subtype` column are determined as follows:
+
+1. If there was an _H3F3A_ K28M or _HIST1H3B_ K28M mutation and no _BRAF_ V600E mutation -> `DMG_H3_K28`
+2. If there was an _H3F3A_ K28M or _HIST1H3B_ K28M mutation and a _BRAF_ V600E mutation -> `DMG_H3_K28_BRAF_V600E"`
+3. If there was an _H3F3A_ G35V or G35R mutation -> `HGG_H3_G35`
+4. If there was an _IDH1_ R132 mutation -> `HGG_IDH`
+5. If a sample was initially classified as HGAT, had no defining histone mutations, and a _BRAF_ V600E mutation -> `BRAF_V600E_mutation`
+6. All other samples that did not meet any of these criteria were marked as `NA`
```
jharenza: For 6, can we create a generic subtype - how about `HGG, H3 wildtype`?

jharenza:
I am still seeing 4 biospecimen IDs missing from `HGG_molecular_subtype.tsv` that are in `HGG_cleaned_all_table.tsv` - the patients are not missing, but below are the 4 BS_IDs:

>  "BS_EE73VE7V" "BS_6DT506HY" "BS_AEEKJ1QK" "BS_TYFD8ZMG"

jharenza:
Ok, I think after we address the above few changes, then we should be good to go!

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
jaclyn-taroni: I think this is a separate notebook and I'm most likely going to file a separate ticket that's focused just on this visualization.

jharenza:
Looks like those missing BS_IDs made it into the subtype table now. One note about the style changes of the subtypes - perhaps changing `DMG, H3 K28, BRAF_V600E` to `DMG, H3 K28, BRAF V600E` to be consistent with the plain `BRAF V600E`. Otherwise, I think the content is good!

jaclyn-taroni:
Good catch, I didn't intend for that `_` to be there - fixed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/435/commits/7e8e7917c002e11ab97312dc7f64c542fc04892b

jharenza:
Hi @jaclyn-taroni - in preparing the clinical file using a key of `Kids_First_Participant_ID;sample_id`, I found that this patient, `PT_KTRJ8TFY` has biospecimens classified as both DMG, H3 K28 (BS_5968GBGT, BS_AF5D41PD, BS_HYKV2TH9) and HGG, H3 wildtype (BS_EE73VE7V).   

I looked at whether this mutation was present in any of the MAF files and missed in consensus or not present at all. It looks like only VarDict caught this variant - something I was afraid of:

```
38f9d38f36c9:release-v13-20200116 harenzaj$ zmore pbta-snv-lancet.vep.maf.gz | awk '/BS_EE73VE7V/ && /H3F3A/ && /p.Lys28Met/'
38f9d38f36c9:release-v13-20200116 harenzaj$ zmore pbta-snv-strelka2.vep.maf.gz | awk '/BS_EE73VE7V/ && /H3F3A/ && /p.Lys28Met/'
38f9d38f36c9:release-v13-20200116 harenzaj$ zmore pbta-snv-mutect2.vep.maf.gz | awk '/BS_EE73VE7V/ && /H3F3A/ && /p.Lys28Met/'
38f9d38f36c9:release-v13-20200116 harenzaj$ zmore pbta-snv-vardict.vep.maf.gz | awk '/BS_EE73VE7V/ && /H3F3A/ && /p.Lys28Met/'
H3F3A	3021	.	GRCh38	chr1	226064434	226064434	+	Missense_Mutation	SNP	A	A	rs1057519903		BS_EE73VE7V	BS_SNRF1RKC	A	A								c.83A>T	p.Lys28Met	p.K28M	ENST00000366813	1/3	69	64	5	39	39	0	H3F3A,missense_variant,p.Lys28Met,ENST00000366814,;H3F3A,missense_variant,p.Lys28Met,ENST00000366813,;H3F3A,missense_variant,p.Lys28Met,ENST00000366815,NM_002107.4;H3F3A,missense_variant,p.Lys28Met,ENST00000366816,;AL512343.2,upstream_gene_variant,,ENST00000609423,;,regulatory_region_variant,,ENSR00000021067,;	T	ENSG00000163041	ENST00000366813	Transcript	missense_variant	458/1308	83/411	28/136	K/M	aAg/aTg	rs1057519903,COSM327928,COSM1961654	1		1	H3F3A	HGNC	HGNC:4764	protein_coding	YES	CCDS1550.1	ENSP00000355778	P84243	B2R4P9	UPI00000007B0		deleterious_low_confidence(0.01)	possibly_damaging(0.697)	1/3		Gene3D:1.10.20.10,PDB-ENSP_mappings:3av2.A,PDB-ENSP_mappings:3av2.E,PDB-ENSP_mappings:3muk.D,PDB-ENSP_mappings:3wtp.E,PDB-ENSP_mappings:4h9n.A,PDB-ENSP_mappings:4h9o.A,PDB-ENSP_mappings:4h9p.A,PDB-ENSP_mappings:4h9q.A,PDB-ENSP_mappings:4h9r.A,PDB-ENSP_mappings:4h9s.A,PDB-ENSP_mappings:4h9s.B,PDB-ENSP_mappings:4hga.B,PDB-ENSP_mappings:4n4i.B,PDB-ENSP_mappings:5ay8.A,PDB-ENSP_mappings:5ay8.E,PDB-ENSP_mappings:5b32.A,PDB-ENSP_mappings:5b32.E,PDB-ENSP_mappings:5b33.A,PDB-ENSP_mappings:5b33.E,PDB-ENSP_mappings:5dwq.F,PDB-ENSP_mappings:5dwq.G,PDB-ENSP_mappings:5dx0.F,PDB-ENSP_mappings:5dx0.G,PDB-ENSP_mappings:5dx0.H,PDB-ENSP_mappings:5dx0.I,PDB-ENSP_mappings:5kdm.A,PDB-ENSP_mappings:5x7x.A,PDB-ENSP_mappings:5x7x.E,Pfam_domain:PF00125,Prints_domain:PR00622,hmmpanther:PTHR11426,hmmpanther:PTHR11426:SF163,Superfamily_domains:SSF47113,mobidb-lite										likely_pathogenic	1,1,1						MODERATE	1	SNV	1		1,1,1		PASS	AAG	.	69													226064434
```

Because this patient has many samples and it has been verified from the PNOC003 trial and within these other samples that the tumor was H3 mutant, and we see this variant in VarDict, I think we should classify this patient as `H3, K28M` mutant, even though this was not in the consensus calls. Before this happened, I was thinking that if we encountered a case like this, we could potentially check whether these mutations were present in 1/4 callers and scavenge them for subtyping, however, that also defeats the purpose of using consensus calls. How do you think we should proceed here? Should we add this bit to the subtyping PR or should I add a note in the clinical file about it?

jaclyn-taroni:
VarDict was not used in the consensus calls because it was found to be very permissive, which is expected. From the [VarDict README](https://github.com/AstraZeneca-NGS/VarDict/tree/f0e0da09b7af7d3f18a3ce63d453e2cd2b975ad8#vardict):

> Due to the philosophy of VarDict in calling "everything", several downstream strategies have been developed to filter variants to for example the most likely cancer driving events. 

I don't like the idea of a call for `BS_EE73VE7V` based on VarDict alone. I think we should be consistent and use the strategy that is already included in this pull request and document _how_ these calls were generated and any "gotchas" -- e.g., only considered the consensus and biospecimens from the same participant can have different subtype labels given our methodology. The molecular subtype labels will be presented on a biospecimen basis anyway.

jharenza:
Right, I generally agree with that, but we have this subtype documented from the PNOC003 trial (table I attached above). So, perhaps I add it based on the PNOC clinical sequencing then, and not subtyping, for this specimen?

jaclyn-taroni:
The table you're referring to is on a per participant basis and we have no information about the rank order of samples (https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/260#issuecomment-564829290) - can you rule out that the biospecimen `BS_EE73VE7V`  is H3 wildtype?

jharenza:
Per discussion with @cgreene and @jaclyn-taroni, we will investigate variant calling parameters in D3b and leave as is with consensus from OpenPBTA. Note, the variant exists in IGV (below)- we can perhaps highlight in the paper:
Screen Shot 2020-02-03 at 8 53 51 AM

analyses/molecular-subtyping-HGG/03-HGG-molecular-subtyping-cnv.Rmd
```diff
@@ -48,6 +43,74 @@ if (!dir.exists(results_dir)) {
...
+```
+
+No, they're not.
+How many are missing?
+
+```{r}
+sum(!(included_wgs_samples %in% consensus_seg_ids))
+```
+
+We'd then expect `r sum(included_wgs_samples %in% consensus_seg_ids)` samples to have non-NA values in the cleaned CNV files.
```
cansavvy: These `NA`s are separate than those that eventually got switched to be called "wildtype" right? Just trying to follow what ended up happening to these samples and how it fits in with the final table.

analyses/molecular-subtyping-HGG/03-HGG-molecular-subtyping-cnv.Rmd
```diff
@@ -126,24 +183,37 @@ gistic_status_df <- gistic_df %>%
...
-head(final_hgg_cn_df, n = 10)
+final_hgg_cn_df <- full_join(cn_df_genes_status,
+                             gistic_status_df)
+
+nrow(final_hgg_cn_df)
+```
+
+`final_hgg_cn_df` includes all the samples that are in the consensus SEG file.
+So any missing CNA values at this stage come from the fact that neutral values are _not_ included in the annotated copy number files.
+We can fill any "missing values" with neutral calls as a result.
```
cansavvy: So this is referring to samples that did have CNV data, but didn't have a value for a particular gene because it was neutral?

analyses/molecular-subtyping-HGG/03-HGG-molecular-subtyping-cnv.Rmd
jaclyn-taroni: This is copy number data and the calls are ultimately based on SNV data.

analyses/molecular-subtyping-HGG/03-HGG-molecular-subtyping-cnv.Rmd
jaclyn-taroni: Yes, you are correct. If it's in the GISTIC file, that means it had copy number data and it is not included in the annotated file because neutral calls get filtered out.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,533 @@
...
+  )) %>%
+  select(-BRAF_V600E)
+```
+
+### H3 G35 mutant
+
+> * These tumors contain H3F3A G35R/V mutations
+> * Co-occurring lesions include: ATRX/DAXX, TP53, SETD2 mutations, NTRK fusions
+> * Mutually exclusive lesions: IDH1 mutations
+> * Average age of 20 years
```
cansavvy: You tested the age estimate in the previous section. Did you want to do that here as well? I don't know that it is completely necessary.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,533 @@
...
+No it doesn't.
+
+```{r}
+rm(mutation_df, cn_gistic_df)
+```
+
+### Join RNA-seq Data (Fusion, Gene Expression)
+
+We have two cleaned gene expression data files corresponding to the stranded and poly-A datasets.
+The (z-scored) distributions for the 3 genes of interest look pretty similar: https://alexslemonade.github.io/OpenPBTA-analysis/analyses/molecular-subtyping-HGG/06-HGG-molecular-subtyping-gene-expression.nb.html#density_plots
```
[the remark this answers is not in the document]
jaclyn-taroni: The expression genes of interest are separate from the additional genes that got added here, which were for mutation data.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,533 @@
...
+
+There are 2 instances where there are two stranded biospecimen IDs that map to the same `sample_id`.
+
+### All Data
+
+```{r}
+all_data_df <- left_join(dna_df, rna_df,
+                         by = c("Kids_First_Participant_ID",
+                                "sample_id")) %>%
+  rename(Kids_First_Biospecimen_ID_DNA = Kids_First_Biospecimen_ID.x,
```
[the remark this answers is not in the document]
jaclyn-taroni: 🆒 😎

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
```diff
@@ -0,0 +1,533 @@
...
+relevant_clinical_df <- histologies_df %>%
+  filter(sample_id %in% all_data_df$sample_id,
+         sample_type == "Tumor",
+         composition == "Solid Tissue") %>%
+  mutate(age_at_diagnosis_yr = 
+           floor(as.integer(age_at_diagnosis_days)/365)) %>%
+  group_by(Kids_First_Participant_ID, sample_id) %>%
+  summarize(glioma_brain_region = paste(sort(unique(glioma_brain_region)), 
+                                        collapse = ", "),
+            age_at_diagnosis_yr = paste(sort(unique(age_at_diagnosis_yr)),
```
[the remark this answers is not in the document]
jaclyn-taroni: I would probably keep this handling for `glioma_brain_region` -- that seems like there's more opportunity for discrepancies but yea I could presumably randomly select an age values since it is age at diagnosis.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
jaclyn-taroni: The table in the rendered notebook has few enough rows that it's easy to tell that the ages agree with what was described.

analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd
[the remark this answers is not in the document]
jaclyn-taroni: No discrepancies, which means that it's functionally the same as randomly selecting a value (there are no trailing `,`) - I think I will leave it as is because any discrepancies would be obvious in the final table?

Issue #436, opened 2020-01-15
CNVkit data only includes WGS samples


jaclyn-taroni:
Context is #435

#### What data file(s) does this issue pertain to?

`pbta-cnv-cnvkit.seg.gz` and `pbta-cnv-cnvkit-gistic.zip` which is downstream.

#### Put your question or report your issue here.

These files appear to only contain WGS samples. This is not recorded in the data formats documentation (https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#somatic-copy-number-variant-cnv-data) or in the data files descriptions included with the download (https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-files-description.md).

#### What release are you using?

`release-v12-20191217`

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.

https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#somatic-copy-number-variant-calling

It is not clear from this section whether not including the WXS and targeted panel samples was intentional or not. If it is not intentional, I would expect each of the experimental strategies to require different parameters which should be reported in the manuscript.

jharenza:
@jaclyn-taroni CNV and SV calling was only done on WGS samples - we should have specified this in the methods. Actually, until about two weeks ago, we did not create CNV workflows for WXS, and we have not benchmarked it, but I can talk to the team and see if we want to run this for the paper. We definitely would not run for the panel sample, as usually those algorithms are specifically developed by the clinical labs and I am not sure we have something that would work immediately.

jaclyn-taroni:
I don’t think it’s necessary to include the CNV calls for the WXS here. I was just aware that CNVkit could make such calls and it was not documented that this was only run for WGS samples in the three places I thought to look. If we remedy the documentation issue, I am happy to close this.

jaclyn-taroni:
I have filed #437 for the documentation changes in this repository and opened https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/69 for the manuscript changes. Once #437 goes in, I think we can close this.

jaclyn-taroni:
Closed via #437

PR #437, opened 2020-01-15
Document which data are WGS samples only

jaclyn-taroni:

It is possible to run CNVkit on exome data, but the CNV and SV OpenPBTA data are only run on WGS samples per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/436#issuecomment-574445530. 

Here I am documenting that in `doc/data-formats.md` and `doc/data-files-description.md` so folks know up front.

#436


jaclyn-taroni:
Documentation only change - merging without Circle.

PR #447, opened 2020-01-17
PR 2 of n - Molecular subtyping embryonal (Generate subset files)

jaclyn-taroni:

Molecular subtyping of non-MB, non-ATRT embryonal tumors.

#### What was your approach?

Here I am generating subset files for embryonal tumor subtyping. Specifically: 
* I am subsetting both collapsed expression datasets. I z-score the gene values _prior_ to subsetting the datasets since there are a low number of samples that are included. Thus, expression values will be in the context of the entire cohort. This differs from other subtyping tickets where there the sample size is larger.
* I am subsetting the annotated CNV kit files - both the autosome and sex chromosome files.

The other files required for subsetting (fusion summary, histology) are the same in CI as they are in the regular data download.

#251 

No areas in particular.

#### Documentation Checklist


analyses/molecular-subtyping-embryonal/02-generate-subset-files.R
```diff
@@ -0,0 +1,101 @@
...
+# This script subsets the files required for subtyping non-MB and non-ATRT
+# embryonal tumors. A sample will be included if either of the following
+# conditions are met: 1) the sample is labeled as an embryonal tumor
+# (broad_histology) but NOT an MB or ATRT tumor (disease_type_old) OR 2)
+# the sample contains a TTYH1 fusion (5' partner)
+
+library(tidyverse)
+
+#### Directories ---------------------------------------------------------------
+
```
cbethell: Maybe add a comment here about what the `rprojroot::find_root` function is doing. This may not be so intuitive for someone who has not been following the project.

PR #464, opened 2020-01-22
Changed link to arriba and STAR fusion output links

tkoganti:
Fix links for outputs of ARRIBA and STAR in docs


jaclyn-taroni:
Documentation change only - merging!

PR #466, opened 2020-01-22
PR 11 of n - Molecular subtyping - HGG (1p/19q co-deleted oligodendrogliomas only)

jaclyn-taroni:

From @jharenza on https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/435#issuecomment-576898275:

> We expect very few, if any of these (maybe 1 when we looked by CNV), so I think maybe we can approach this cohort by starting to look for 1p/19q codeletions in all samples + IDH mutations, and if we see any, we may manually want to check the CNV plots just to confirm.

Molecular subtyping 1p/19q co-deleted oligodendrogliomas

#### What was your approach?

Here I am using the GISTIC `broad_values_by_arm.txt` to identify samples with `1p` and `19q` loss (e.g., values less than zero). I then use that list of biospecimen IDs to filter the consensus mutation file and check for _IDH1_ mutations.

#249 

### Results

There was only one sample with losses in the GISTIC data for both of those arms. This sample did not have an _IDH1_ mutation. When I look at the consensus SEG file (note that the GISTIC output was run on the original CNVkit file), there's only one deletion in that sample and it is on chromsome 15.

Here's the rendered notebook: https://jaclyn-taroni.github.io/openpbta-notebook-concept/08-1p19q-codeleted-oligodendrogliomas.nb.html

#### Documentation Checklist


jaclyn-taroni:
There are no instances of 1p, 19q co-deletion when using the GISTIC output using the consensus SEG file.

jharenza:
once the README is checked off, I will approve.

.circleci/config.yml
```diff
@@ -30,6 +30,10 @@ jobs:
           name: Collapse RSEM
           command: ./scripts/run_in_ci.sh bash analyses/collapse-rnaseq/run-collapse-rnaseq.sh
 
+      - run:
+          name: Molecular Subtyping - HGG
+          command: OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-HGG/run-molecular-subtyping-HGG.sh
```
cbethell: The `OPENPBTA_SUBSET=0` argument here seems to be the reason why `circleci` is failing. This argument skips the `02` subset script in which the GISTIC `broad_values_by_arm.txt` file is extracted. 

Also, upon extraction, the folder `2019-01-28-consensus-cnv` that contains the GISTIC file is within another folder with the same name, hence when attempting to read the contents of the .txt file, we get an error. 

I _believe_ the reason for this is in the `02` script where the GISTIC file is extracted: 
```
download.file(url = "https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4123481/2019-01-28-consensus-cnv.zip",
              destfile = file.path(scratch_dir, "2019-01-28-consensus-cnv.zip"),
              quiet = TRUE)
unzip(file.path(scratch_dir, "2019-01-28-consensus-cnv.zip"),
      exdir = file.path(scratch_dir, "2019-01-28-consensus-cnv"),
      files = file.path("2019-01-28-consensus-cnv", "broad_values_by_arm.txt"))
gistic_df <- data.table::fread(file.path(scratch_dir,
                                         "2019-01-28-consensus-cnv",
                                         "broad_values_by_arm.txt"),
                               data.table = FALSE)
```
Here, the file path of the output directory is defined to be `scratch/2019-01-28-consensus-cnv` and the file path of the file we want is defined to be `2019-01-28-consensus-cnv/broad_values_by_arm.txt`. 
I think if we get rid of the "2019-01-28-consensus-cnv" in front of the file path to the file, then this should work.

PR #467, opened 2020-01-22
Generate CNV exclusion list

### LINKED ISSUE #438: Reproducing copy number excluded regions
#### What data file(s) does this issue pertain to?

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/copy_number_consensus_call/src/scripts/IGLL_telo_centromeric_region.txt

#### Put your question or report your issue here.

I am trying to reproduce the generation of the `IGLL_telo_centromeric_region.txt` file that is used for Copy number consensus generation #128. This file includes regions that are to be excluded from analysis due to high error rates. The telomeres and centromeres can be reproduced from UCSC data files, but I am confused by the immunoglobulin regions. The documentation points to http://penncnv.openbioinformatics.org/en/latest/misc/faq/, but it is not clear from where the enumerations of those regions is defined. Moreover, the regions defined there are for hg18: 
```
chr22:20715572-21595082
chr14:105065301-106352275
chr2:88937989-89411302
chr14:21159897-22090937
```

However, the regions in the `IGLL_telo_centromeric_region.txt` file do not seem to correspond to a liftOver of those regions to hg38

Applying liftOver hg18->hg38, I get the following regions:
```
chr22:22031174-22922910
chr14:105527919-106873021
chr2:88857361-89330430
chr14:21621904-22552154
```

The nearest equivalent regions in `IGLL_telo_centromeric_region.txt`  seem to be these:
```
chr22:21990603-22947816
chr2:88854372-89330679
chr14:21676543-22556766
```
(There are only three here, presumably because the other chr14 region falls near a telomere and is excluded that way?)

In addition, `IGLL_telo_centromeric_region.txt` includes the region `chr21:3100000-7000000` which is listed as a `stalk` (acrocentric arm) the UCSC cytoband file, but no other `stalk` regions are excluded, so I was not sure why this one was.

Can @fingerfen or @xiehongbo provide some additional information?

jashapiro:

The analysis of CNV consensus files takes advantage of a file of regions that are to be excluded due to expected (and previously observed) high levels of false positives. These include regions such as telomeres, centromeres, and known segmental duplications.

#### What was your approach?

I split up the various categories of excluded regions into separate files for ease of identification and modification. These are:

* `ref/centromeres.bed`
* `ref/heterochromatin.bed`
* `ref/immunoglobulin_regions.bed`
* `ref/segmental_dups.bed`
* `ref/telomeres.bed`

The origins of these files, with code where appropriate, are in `scripts/prepare_blacklist_files.sh`

The files are then merged with a new rule in the `Snakefile` to generate `ref/cnv_excluded.bed` which can be used downstream.

closes #438

This file differs slightly from the previously included file, provided by @fingerfen, mostly I think in the handling of segmental duplication regions. There seem to be some broader regions that are excluded, but I could not find references for why those had been excluded. 

Note however, that there do not appear to be major changes in the final CNV regions, though there are some effects at the margins.

Do we need to programmatically generate every region, or is it okay that the telomeres and IG regions are 
simply included as their own files?

How concerned should we be be about changes in the final set of calls resulting from this change? Can we adjust the generation to better match the previous set (included here as `ref/bad_chromosomal_seg_merged.bed`)

#### Documentation Checklist


jaclyn-taroni:
> Do we need to programmatically generate every region, or is it okay that the telomeres and IG regions are simply included as their own files?

It took me awhile to figure out that the origin of these files was described as comments in the shell script. So my vote would be yes let's programmatically generate these.

jashapiro:
> > Do we need to programmatically generate every region, or is it okay that the telomeres and IG regions are simply included as their own files?
> 
> It took me awhile to figure out that the origin of these files was described as comments in the shell script. So my vote would be yes let's programmatically generate these.

I can do the telomeres... I have no idea how to get the IG regions, unfortunately.

jashapiro:
In my work on #476, I have substantially updated the README to include more information on the creation of exclusion regions, making it hopefully easier to find. If those updates are a good start, it might make sense to make further changes in that branch?

Issue #455, opened 2020-01-18
Updated analysis: modules to be rerun with v13 stranded RNA-seq data


jaclyn-taroni:
#### What analysis module should be updated and why?

Below, I list what analyses need to be rerun as a checklist below using the [Modules at a glance table](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses#modules-at-a-glance) as my guide:

- [x] `collapse-rnaseq` - New collapsed files were included in the v13 release; the summary tables tracked in the repository have not been updated.
- [x] `gene-set-enrichment-analysis` - some other analysis modules consume the scores generated by this module, listed as subitems below
- [x] `immune-deconv` - uses both strategies' collapsed files; `results/deconv-output.RData`, which can be used downstream, has not been regenerated
- [x] `tp53_nf1_score` - uses both strategies' collapsed files; `*_classifier_scores.tsv` and ROC plots need to be regenerated
- [x] `transcriptomic-dimension-reduction` - seems pretty self explanatory 😄 

These do not need to be addressed as part of this ticket but I am including them for completeness
* `focal-cn-file-preparation` - this can wait on #387 
* `molecular-subtyping-embryonal` - this is actively being worked on and will get rerun _anyway_

- [x] `molecular-subtyping-HGG` - uses both strategies' collapsed files - will update with #435 
- [ ] `sex-prediction-from-RNASeq` - uses kallisto stranded data - will update as part of #470 
- [x] `molecular-subtyping-ATRT` - not only relies on the GSVA scores, also uses stranded collapsed file directly - will update with #462 

#### Why should the module be updated?

There were issues with poly-A stranded samples being included in ribodepleted stranded data set in v12, as well as some formatting inconsistencies.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

The scope of this ticket should include simply _rerunning things_ and all changes should be included in a single pull request. If any one module turns out to be problematic (e.g., require lots of changes to run properly or is a larger can of worms than anticipated), please file a new updated analysis ticket to track the work that needs to be done and move on to another module.

#### What input data should be used? Which data were used in the version being updated?

This applies to _any_ stranded RNA-seq data.

#### When do you expect the revised analysis will be completed?

Over the next ~1-2 week period.

#### Who will complete the updated analysis?

Likely someone from the CCDL (@cbethell @cansavvy @jashapiro or @jaclyn-taroni)

jaclyn-taroni:
[...]

PR #471, opened 2020-01-23
Update modules that use RNA-seq data to use release v13 data #455

jaclyn-taroni:

Some samples were removed from the stranded RNA-seq file in `release-v13-20200116`. These files should be _more or less_ the same as the files that were in `release-v11-20191126` (but there were some inconsistencies between file types, etc.)

Here I'm updating the following:

> 
> - [x] `collapse-rnaseq` - New collapsed files were included in the v13 release; the summary tables tracked in the repository have not been updated.
> - [x] `gene-set-enrichment-analysis` - some other analysis modules consume the scores generated by this module, listed as subitems below
> - [x] `immune-deconv` - uses both strategies' collapsed files; `results/deconv-output.RData`, which can be used downstream, has not been regenerated
> - [x] `tp53_nf1_score` - uses both strategies' collapsed files; `*_classifier_scores.tsv` and ROC plots need to be regenerated
> - [x] `transcriptomic-dimension-reduction` - seems pretty self explanatory 😄 

#455 - for the other modules that need to be updated, there are currently some open PRs so no need to rerun them as part of that.

### Results

The PCA/t-SNE/UMAP plots all look like what I would expect based on `release-v11-20191126`


cbethell:
LGTM!

Issue #453, opened 2020-01-18
Updated analysis: rerun GISTIC with consensus SEG file


jaclyn-taroni:
#### What analysis module should be updated and why?

_Note: Running GISTIC is upstream of this repository._

We now have consensus copy number calls (#128). Furthermore, we have a consensus SEG file that contains the required information to run GISTIC (discussed in #392 and implemented in #441): `analyses/copy_number_consensus_call/results/pbta-cnv-consensus.seg`  

#### What changes need to be made? Please provide enough detail for another participant to make the update.

I believe this is as straightforward as switching what `segfile` points to here: https://github.com/d3b-center/publication_workflows/blob/master/openPBTA/run-gistic.sh

#### What input data should be used? Which data were used in the version being updated?

Input: `analyses/copy_number_consensus_call/results/pbta-cnv-consensus.seg` 

In the GISTIC output that's included in the data download, the `pbta-cnv-cnvkit.seg` file was used as input. That file has not changed in some time according to the [release notes](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md).

#### When do you expect the revised analysis will be completed?

Unknown.

#### Who will complete the updated analysis?

Tagging @jharenza as this is upstream of the repo. The GISTIC output may be included in the v14 release (#432), but ideally this can be made available ahead of time so it can start getting used in downstream analyses (see: #8).

jharenza:
Hi @jaclyn-taroni! I set this up, but it quickly failed. 
`[harenzaj@c-2-02 PBTA]$ bash run_gistic_consensus.sh 
--- running GISTIC ---
GISTIC version 2.0.23
required parameters:
      base_dir: /mnt/isilon/maris_lab/target_nbl_ngs/PBTA/gistic/2019-01-18-consensus-cnv/
       segfile: /mnt/isilon/maris_lab/target_nbl_ngs/PBTA/pbta-cnv-consensus.seg
marker spacing: 10000
   refgenefile: /mnt/isilon/maris_lab/target_nbl_ngs/Gonzalo/install/GISTIC_2_0_23/refgenefiles/hg38.UCSC.add_miR.160920.refgene.mat
optional_params = 
         array_list_file: []
                cnv_file: []
                   t_amp: 0.1000
                   t_del: 0.1000
       join_segment_size: 2
                     ext: []
               qv_thresh: 0.2500
                remove_X: 0
                 markers: []
[...]

PR #476, opened 2020-01-24
Include neutral changes in cnv consensus .seg file

jashapiro:

GISTIC requires all assayed regions to be included for every sample, so the previous version of `pbta-cnv-consensus.seg` was failing to run. This set of changes is to update the creation of that file to include the full set of "neutral" changes (copy number 2) in the seg file.

#### What was your approach?

Many regions of the genome are excluded from CN calls in the consensus script, so it does not make sense to simply complement the CNV consensus calls to find the neutral regions. Instead, I first created a file of "callable" regions by removing any large segments that were excluded in the CNV consensus calling pipeline, such as telomeres, centromeres, and segmental duplications.

In addition, samples that were filtered out due to high numbers of segment calls are not included in the final .seg file. 

#392 
#453 

Is the process for generating the callable regions reasonable? The size cutoff of 200kb was selected somewhat arbitrarily, with the goal of not creating too many separate segments,  as some downstream analysis will filter samples based on segment number, even if most of those segments have copy number 2.

I currently exclude samples where only one caller passed filters, as those can not by definition have any consensus calls (2 callers must agree). However, the samples with *only* two callers (where one was excluded) are likely to have a different error profile (and probably fewer net calls). Is this something we should flag, and if so how? There does not seem to be much in the way of tumor type bias in this set, which is good.

Some of the changes here depend on file name changes in #467, so that PR should probably be merged in before this one.

#### Documentation Checklist


analyses/copy_number_consensus_call/scripts/bed_to_segfile.R
```diff
@@ -26,18 +26,29 @@ option_list <- list(
     c("-i", "--cnv_file"),
     type = "character",
     default = NULL,
-    help = "path to the cnv consensus file",
+    help = "consensus filetsv",
```
jaclyn-taroni:
```suggestion
    help = "consensus file tsv",
```

analyses/copy_number_consensus_call/scripts/remove_dup_NULL_overlap_entries.py
```diff
@@ -52,6 +52,8 @@ def remove_dup_null_outside(cell, start, end):
 )
 parser.add_argument('--file', required=True,
                     help='path to the file that needs duplicates and NULLs removed')
+parser.add_argument('--uncalled', required=False,
```
jaclyn-taroni: Is this used anywhere?

jashapiro: No it is not. I had thoughts that I might use it to annotate where there were only two possible callers, but abandoned that plan on some reflection. I'll remove it unless there is some great idea for how to use the information in this step.

Issue #397, opened 2020-01-03
Proposed Analysis: visualization of CNV and SV data with Circos plot


jaclyn-taroni:
Related to the following: #393, #394, https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/392#issuecomment-570332277, and https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/27#issuecomment-516471524

#### What are the scientific goals of the analysis?

To generate circular plots that integrate CNV and SV data. Specifically, it might be useful for the following cases:

* Display examples of chromothripsis per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/27#issuecomment-516471524
* There are some subtype tickets that would benefit from such a visualization (e.g., "[ST-EPN-YAP1] Focal 11q most frequent CNA, with rest of genome stable" #245)
* Likely most useful to examine translocations

The implementation should focus on _flexibility and extensibility_ because we don't know what "final" output will be required for the project.

#### What methods do you plan to use to accomplish the scientific goals?

Here I'm including links to code that could potentially be used to guide this analysis, rather than a specific _method_. 

From https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/392#issuecomment-570332277:

> I've found this [UMCCR pipeline code](https://github.com/umccr/rock) that uses Manta and CNVkit output to create circos plots. I bet we can adapt it (or use it to guide something similar) to use the "consensus filtered" CNVkit data mentioned here, the AnnotSV-annotated Manta file, and [`circlize`](https://github.com/jokergoo/circlize).

Relevant functions from @gonzolgarcia's package mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/394#issuecomment-570589012:

* https://github.com/gonzolgarcia/svcnvplus/blob/ea43a5dbd7ae38f1b26520a31af1daddadbaf5dd/R/circ.chromo.plot.r
* https://github.com/gonzolgarcia/svcnvplus/blob/ea43a5dbd7ae38f1b26520a31af1daddadbaf5dd/R/circ.wg.plot.r

#### What input data are required for this analysis?

* SEG file, most likely the "consensus filtered" file #392 (related: #128)
* `pbta-sv-manta.tsv.gz`

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

~1 week is my rough estimate

#### Who will complete the analysis (please add a GitHub handle here if relevant)?

[...]

PR #484, opened 2020-01-28
Circos plots

cansavvy:

**This is a tad large, let me know if I should break this up. Wasn't sure what the logical way to break it up was**. 

Visualization of CNV and SV data (#397). 

#### What was your approach?
Because the issue indicated we are not exactly sure what we need to be plotting, I made general purpose circlize plot functions and a notebook with examples of how to use the functions. Obviously there's a lot of parameters that can be tweaked, so as we hone in on what our plotting needs are, I can adjust what is here now. 

#397 

- Do the functions address the basic circos plotting needs discussed on #397?
- How do you feel about the option and logic handling of the functions? 
- I also am still confused on how the SV data translocation coordinates are formatted and where I should be pulling those coordinates from. If someone can comment on the validity or lack thereof of my example, that'd be great. They all have a width of 1 which seems wrong. 

Some of the circlize functions are very hard to understand and very particular about how arguments and data are passed to them. If it looks like I'm doing something strange with the data, it probably is because I am but that was the only way I found it to work. I am very open to suggestions for optimization but just know that some of these things are not as straight forward. 

To get familiar with circlize here's the manual: https://jokergoo.github.io/circlize_book/book/ It has a lot of examples so it is very helpful but some stuff is very confusing, feel free to bug me about things in my code that look weird and I will try to explain how I came to that if it isn't clear. 

There are examples. I wouldn't get too caught up in what the examples are showing with the data, they are just to illustrate how to use the functions. 

Here's the rendered notebook of examples: https://cansavvy.github.io/openpbta-notebook-concept/chromosomal-instability/01b-visualization-cnv-sv.nb.html

- [x] This analysis has been added to continuous integration. (It is 

#### Documentation Checklist
I added items on each of these things to try to make it clear the circos plots parts are added to the `chromosomal-instability` analysis. Let me know if I need to add more. 


analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,338 @@
...
+    bed_df <- bed_df %>%
+      dplyr::mutate(chr = paste0("chr", chr))
+  }
+
+  # If we aren't adding on a track to a current plot...
+  if (!add_track) {
+    # Initialize the plot
+    circlize::circos.clear()
+    circlize::circos.par(start.degree = 90)
+    circlize::circos.initializeWithIdeogram(species = "hg38")
```
jashapiro: Do we always want Ideograms? I feel like this should be optional, if we can do it.

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,338 @@
...
+    circlize::circos.genomicTrackPlotRegion(bed_df,
+      track.height = track_height,
+      # Establish the ylim
+      ylim = c(y_min, y_max),
+      panel.fun = function(region, value, ...) {
+        circlize::circos.genomicPoints(region, value,
+          col = palette_col
+        )
+      }
+    )
```
jashapiro: I'm fairly confident you tested this, but according to the docs: 
> If there is only one numeric column, the length of col can be either one or number of rows of region

So you should be able to use the **unmodified** bed_df and 
```
circlize::circos.genomicPoints(region, value[[1]],
          col = value[[2]]
)
```

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,338 @@
...
+    # Add a rectangle track
+    circlize::circos.genomicTrackPlotRegion(bed_df,
+      track.height = track_height,
+      # Establish the ylim
+      ylim = c(y_min, y_max),
+      panel.fun = function(region, value, ...) {
+        i <- circlize::getI(...)
+        circlize::circos.genomicRect(region, value[[1]],
+          ytop = value[[1]] + 0.4,
+          ybottom = value[[1]] - 0.4,
```
jashapiro: Should we make the rect size an option for this function? So +-0.4 would be derived from an argument, rather than hard coded here.

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,338 @@
...
+  if (!add_track) {
+    # Initialize the plot
+    circlize::circos.clear()
+    circlize::circos.par(start.degree = 90)
+    circlize::circos.initializeWithIdeogram(species = "hg38")
+  }
+
+  circlize::circos.genomicLink(bed_df_1,
+    bed_df_2,
+    col = circlize::rand_color(nrow(bed_df_1),
```
jashapiro: I don't like the random colors here. Allow a color column here too, perhaps?

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: I'll try again, but I do remember trying this and it wouldn't work.

analyses/chromosomal-instability/01b-visualization-cnv-sv.Rmd
```diff
@@ -0,0 +1,277 @@
...
+cnv_df <- data.table::fread(
+  file.path(data_dir, "pbta-cnv-cnvkit.seg.gz"),
+  data.table = FALSE
+) %>%
+  # Let's make a variable that specifies whether the data is a deletion or amplication.
+  dplyr::mutate(
+    direction = dplyr::case_when(
+      seg.mean < 0 ~ "del",
+      seg.mean > 0 ~ "amp",
+      seg.mean == 0 ~ "neutral"
```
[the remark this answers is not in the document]
cansavvy: Will do.

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: Per our discussion in person, I will add an option that allows a user to drop the cytobands.

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,338 @@
...
+  return(bed_df)
+}
+
+circos_map_plot <- function(df,
+                            add_track = FALSE,
+                            samples_col = "samples",
+                            sample_names = NULL,
+                            chr_col = "chrom",
+                            start_col = "start",
+                            end_col = "end",
```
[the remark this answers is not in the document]
cansavvy: All the data file seem to have different column names so this is my way to try to make it flexible. How desirable/advantageous is it to have shorter function calls?

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: Yes, I can do that.

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: Yes, this was one of those things, along with many other parameters that should probably be its own option, but I wasn't sure which of the option ideas I had to prioritize.

analyses/chromosomal-instability/util/circos-plots.R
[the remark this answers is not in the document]
jashapiro: It can stay just as flexible. But personally this is a case where I would find myself renaming data frames to match the default column names so the function call itself would be shorter. But definitely don't remove the option!

analyses/chromosomal-instability/util/circos-plots.R
jashapiro: Correct, it is the cytobands I cared about. just didn't know which option was what.

analyses/chromosomal-instability/01b-visualization-cnv-sv.Rmd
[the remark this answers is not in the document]
cansavvy: So for copy num, I don't see any negative numbers so if I see a 0 or 1 that means a deletion? 2 would be a neutral? Above 2 is amplification?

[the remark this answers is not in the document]
jaclyn-taroni: I would use the same logic as the `focal-cn-file-preparation` module, which takes into account tumor ploidy, see:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/86bd634a928db3f460328210e38c7dc1797d1304/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L87

cansavvy:
@josh, should I keep an option to do the colors automagically and then add the ability to pass color columns? Or should I get rid of automagic bit and make it so you have to pass a color column?

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: So this works with `genomicPoints`, but not `genomicLines` which is a pain.

jashapiro:
> @jashapiro , should I keep an option to do the colors automagically and then add the ability to pass color columns? Or should I get rid of automagic bit and make it so you have to pass a color column?

If you can make the column of colors work for all cases, I would drop the automatic. But maybe move some of that code into your example notebook, to show people ways to make the color column.

analyses/chromosomal-instability/util/circos-plots.R
cansavvy: I've left the lines in, because I think there may be uses for them, but I don't highlight them in the example notebook.

cansavvy:
@jashapiro , feel free to take another gander at this when you are ready. Let me know if your suggestions have been adequately addressed.

jaclyn-taroni:
I know there were some questions about the translocation code - I wanted to point out some specific code from a link on the issue (#397) that you could check out if you have not already @cansavvy. Here's the file from the UMCCR pipeline for dealing with Manta output (notably it accepts VCF files, not the output of AnnotSV so wouldn't work off the shelf): https://github.com/pdiakumis/rock/blob/ea6e72adf0d57a0c02c4ac8502c704d10f9559c7/R/sv_manta.R

Specifically, I'd look at this function: https://github.com/pdiakumis/rock/blob/ea6e72adf0d57a0c02c4ac8502c704d10f9559c7/R/sv_manta.R#L98

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,313 @@
...
+    color_col = color_col
+  )
+
+  # Establish ymin and ymax before further manipulations
+  y_min <- min(bed_df$y_val, na.rm = TRUE)
+  y_max <- max(bed_df$y_val, na.rm = TRUE)
+
+  # If a single color is specified:
+  if (single_color != "none" & color_col == "none") {
+    # If no color key is specified, just make the color black.
```
jashapiro:
```suggestion
    # If a single color is specified, replace the color column with that value
```

analyses/chromosomal-instability/util/circos-plots.R
```diff
@@ -0,0 +1,313 @@
...
+    bed_df <- bed_df %>%
+      dplyr::mutate(color_col = single_color[1])
```
jashapiro: Maybe you should warn if length(single_color)> 1

analyses/chromosomal-instability/01b-visualization-cnv-sv.Rmd
```diff
@@ -0,0 +1,348 @@
...
+  dplyr::inner_join(dplyr::filter(transloc_df, pair_num == 1),
+    by = "match_id",
+    suffix = c("1", "2")
+  )
+```
+
+Let's get some sample IDs to play with in these examples. 
+
+```{r}
+samples_for_examples <- sample(cnv_df$ID, 5)
```
jashapiro: I love this variable name!

PR #485, opened 2020-01-29
WIP: Resolving "gene-level" copy number alterations with broader alterations

jaclyn-taroni:

The `focal-cn-file-preparation` module was focused on gene-level changes, without the context of broader changes. So it may be the case that adjacent genes in the same cytoband or arm would all be called as a loss where it would be more appropriate to call the chromosome arm as a loss. This work-in-progress is my attempt to resolve that issue.

View the rendered notebook here: https://jaclyn-taroni.github.io/openpbta-notebook-concept/04-consensus-cytoband-and-chrom-arm.nb.html

#### What was your approach?

Note that we have to keep in mind that the neutral changes are filtered out of the annotated file (e.g., output of `03-prepare-cn-file.R`) and only things that overlap with exons are included in that 

Following up on an in-person suggestion from @jashapiro - I create a data frame that maps gene symbols to cytobands using the `org.Hs.eg.db` package and count how many gene symbols map to a given cytoband. If more than half of the gene symbols in a cytoband have the same status, then we call the cytoband as that status (e.g., if >50% of the genes in a cytoband are called a loss, we call the cytoband a loss). 

I'm also creating what I think of as a long "broadest regions" data.frame, where for a given WGS biospecimen we only keep track of the broadest region that has a copy number alteration and then filter out anything on a more narrow scale (arm > cytoband > gene symbol). I think the filtering could use more work, because there appear to be cases where the cytoband status and arm status don't agree.

#186 

Cytoband proportion steps

I am hoping to discuss the approach in general. I don't have a full picture of how often the cytoband status and arm status disagree and _why_, but wanted to have others take a look.

I'm not adding this to CI right now because I'm using a local copy of the v14 GISTIC file https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/453#issuecomment-579340618

#### Documentation Checklist


jaclyn-taroni:
Closing this in favor of #644.

Issue #247, opened 2019-11-08
Proposed Analysis: Add additional subtypes to medulloblastoma tumors


jharenza:
#### Scientific goals

_What are the scientific goals of the analysis?_
Further classify Medulloblastoma SHH subtype.

#### Proposed methods

_What methods do you plan to use to accomplish the scientific goals?_
Assess _TP53_ mutation status for samples classified as SHH subtype to further classify into SHH,TP53-mutated and SHH,TP53-wildtype. (See WHO book chapter below)

**SHH,TP53-mutated**
-SHH subtype with _TP53_ mutation

**SHH,TP53-wildtype**
-SHH subtype, no _TP53_ mutation

#### Required input data

_What input data will you use for this analysis?_
SNVs for _TP53_ status, copy number, SVs, fusions, RNA expression, histologies file

#### Proposed timeline
_What is the timeline for the analysis?_
1-2 days

#### Relevant literature

_If there is relevant scientific literature, put links to those items here._
[Medulloblastoma WHO 2016 book chapter.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/3831401/Medulloblastoma.WHO.2016.book.chapter.pdf)

jaclyn-taroni:
@mkoptyra and I will start on this next week.

jaclyn-taroni:
@mkoptyra - @cansavvy is going to pick up this ticket so you can focus on chordoma subtyping (#250)

jaclyn-taroni:
Working on #249 this weekend brought this question up for me - is it _any_ _TP53_ mutation (how the code is currently written) or does this mean “_TP53_ mutations that are likely to be functional?” How do folks usually approach this?

jaclyn-taroni:
Taking a look at the chapter - looks like the _TP53_ are often in exons 4 through 8 and are germline mutations about half the time, which we would not capture here.

jharenza:
@jaclyn-taroni nice catch on this - I think that perhaps we should just leave MB as the 4 subtypes, given we do not have germline data in this cohort and I think we would miss some if we attempted it. @mkoptyra, do you agree?

cgreene:
@jharenza : would the folks writing the germline paper be willing to share the presence/absence of mutations in key genes for subtyping, or would they view that as something that could not be shared?

jharenza:
[...]

PR #495, opened 2020-02-02
Deprecate `molecular-subtyping-SHH-tp53`

jaclyn-taroni:

#247 was closed 5 days ago because we don't have all the necessary information to further divide SHH MB samples into SHH, _TP53_-mutated and  SHH, _TP53_-wildtype. Here I'm adding documentation to the `molecular-subtyping-SHH-tp53` README and to the `analyses` README that notes this module is deprecated and that the results should not be used for subtyping.

#### Documentation Checklist


jaclyn-taroni:
Good idea @cbethell, I'll make that change!

PR #256, opened 2019-11-11
make template more amenable to completion

cgreene:
#### Purpose/implementation

The issue templates are not being used. This clarifies that the headers refer to questions that should be answered.

#### Issue

The cases where issues and pull requests are not following templates.

#### Directions for reviewers

Do you think this will enhance completion? Do you notice any places where the wording should be revised for grammar or clarity?

#### Results

N/A

#### Docker and continuous integration

N/A

PR #498, opened 2020-02-02
Remove redundant question in update-an-analysis issue template

jaclyn-taroni:

In #256, we made the HTML comments their own headers. As a result, there's some redundancy in the `update an analysis` issue template that's tripped me up several times recently as I've filled out these tickets.


jaclyn-taroni:
Documentation only change - merging before CI finishes!

PR #492, opened 2020-01-31
Updates to `chromosomal-instability` 

cansavvy:
#### What scientific question is your analysis addressing?
The items on #487 :

- CNV consensus seg file needs to be used.
- The excluded regions BED file needs to be used
      - For the denominator for calculating densities
      - For marking NA on heatmap
- Flexibility parameters need to be added for calling breakpoints as the same between SV and CNV data. See [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/461#discussion_r369325462)

#### What was your approach?
- **Excluded regions BED**:
Regions are considered `NA` based on the percentage of the bin that is covered by an `cnv_excluded_regions.bed`. The percent cutoff for each bin can be set by `perc_cutoff` argument for `break_density` function calculation.
- **Flexibility parameters**:
CNV and SV data are now combined by an intersection the flexibility of which is controlled by `--gap 5` parameter. This gap parameter is passed directly to GenomicRanges' `mergeByOverlaps` function's `maxgap` parameter. E.g. for a gap parameter of 5, a CNV or SV breakpoint within 5 bp of each other for the same sample will be considered the same.

- Somewhat related but not really, I also added a `drop-sex` argument to the 00-setup script. 

#### What GitHub issue does your pull request address?
#487

#### Which areas should receive a particularly close look?
- The break density function is where the percent overlap is calculated. Can someone make sure that is representing what I think it's representing? 
- 00-setup-breakpoint-data.R's changes for the intersection between CNV and SV also should have a careful look. 
- Are the added options: `--gap`, `perc_overlap` and `--drop-sex` clear and seem like the logic that surrounds them is sensible? 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes. Let me know what changes you suggest for the plots. 

#### What types of results are included (e.g., table, figure)?
The plots are updated to be like this: 
cnv_breaks_heatmap

#### What is your summary of the results?
Here's the rendered html: https://cansavvy.github.io/openpbta-notebook-concept/chromosomal-instability/01-plot-chromosomal-instability.nb.html

#### Reproducibility Checklist

These items were done previously.
- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

Issue #487, opened 2020-01-30
Updated analysis: Chromosomal Instability Tweaks


cansavvy:
#### What analysis module should be updated and why?
`chromosomal-instability` has some open issues that need to be addressed in a particular order. Originally discussed on #394 This issue and #488 are the remaining issues from the #394 ticket. 

#### Why should the module be updated?
Now that the consensus CNV file and its associated BED files are ready and annotated (#467 and #480 and etc)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
*In order, but the first and second items can probably be in the same PR:*
- [x] CNV consensus seg file needs to be used. 
- [x] The excluded regions BED file needs to be used for the denominator for calculating densities. 
   - [x] These regions need to be labeled distinctly on the plot as being `NA`.
- [x] Flexibility parameters need to be added for calling breakpoints as the same between SV and CNV data. See [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/461#discussion_r369325462)

#### What input data should be used? Which data were used in the version being updated?
v13 for now, but `TODO`s should be placed for when the consensus CNV file is added to the data release in v14. 

#### When do you expect the revised analysis will be completed?
I don't *think* these changes should require too much overhaul, should be able to file a PR today. 

#### Who will complete the updated analysis?
@cansavvy

PR #518, opened 2020-02-05
Chr instability: PR 1 of 3: Calculate binned breakpoint counts

cansavvy:
The next 3 PRs are the redone versions of the material that was included in #492. 

This first PR includes the first of three notebooks which contain the material that was originally all in the previous 01 notebook.

I'd recommend taking a look at the README edits first because they describe the overall flow. If you don't like or don't think it's clear there, then let me know before we get into the details. 

This first notebook calculates the data that was formatted up in the 00 script and makes it prepared to be plotted in the next two notebooks. 

The items on #487 :
- CNV consensus seg file needs to be used.
- The excluded regions BED file needs to be used
      - For the denominator for calculating densities
      - For marking NA on heatmap
- Flexibility parameters need to be added for calling breakpoints as the same between SV and CNV data. See [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/461#discussion_r369325462)

#### What was your approach?

**Reorganization** 
- `00-setup-breakpoint-data.R` CNV and SV data are transformed into single breakpoint data as well as their intersection (This has not changed) 

The previous 01 notebook has been split into three notebooks: 
- `01-localization-of-breakpoints.Rmd` uses the data in `breaks_lists.RDS` to co-localize and map breakpoints by bins across the genome.
- `02a-plot-chr-instability-heatmaps.Rmd` uses `_binned_breakpoint_counts.tsv"` datasets to create three heatmaps for the intersection, CNV, and SV data respectively.
- `02b-plot-chr-instability-by-histology.Rmd` uses `intersection_of_breaks_densities.tsv` and `histology_breakpoint_binned_counts.RDS` to plot breakpoint densities by `short_histology` group. 

#### Added options/parameters:
- `return_vector`: this was added to by pass some of the lapply madness that was being used to get the data as we need it for the heatmap. Essentially we don't need the data in a GenomicRanges format for the heatmap, so I've added an option to only return the simple vector instead. 
- **Excluded regions BED**:
Regions are considered `NA` based on the percentage of the bin that is covered by an `cnv_excluded_regions.bed`. The percent cutoff for each bin can be set by `perc_cutoff` argument for `break_density` function calculation.
- **Flexibility parameters**:
CNV and SV data are now combined by an intersection the flexibility of which is controlled by `--gap 5` parameter. This gap parameter is passed directly to GenomicRanges' `mergeByOverlaps` function's `maxgap` parameter. E.g. for a gap parameter of 5, a CNV or SV breakpoint within 5 bp of each other for the same sample will be considered the same.
- Somewhat related but not really, I also added a `drop-sex` argument to the 00-setup script. 

#487

- How do you feel about the reorganization? 
- Given the changes to this module, should the circos notebook be broken into its own module? It doesn't relate to the other notebooks as much now. 
- The break density function is where the percent overlap is calculated. Can someone make sure that is representing what I think it's representing? 
- 00-setup-breakpoint-data.R's changes for the intersection between CNV and SV also should have a careful look. 
- Are the added options: `--gap`, `perc_overlap` and `--drop-sex` clear and seem like the logic that surrounds them is sensible? 

The plots will be in the next two notebooks which I will add in the next two PRs. 

Here's the rendered html: https://cansavvy.github.io/openpbta-notebook-concept/chromosomal-instability/01-localization-of-breakpoints.nb.html

These items were done previously.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date (Note it includes documentation for the upcoming notebooks as well. 


analyses/chromosomal-instability/00-setup-breakpoint-data.R
```diff
@@ -215,14 +245,48 @@ sv_breaks <- data.frame(
...
-  dplyr::distinct(samples, chrom, coord, .keep_all = TRUE)
+# Remake this in case some samples' data got filtered out in the process
+common_samples <- dplyr::intersect(
+  unique(cnv_breaks$samples),
+  unique(sv_breaks$samples)
+)
+
+######################### Create intersection of breaks ########################
+# Make an intersection of breaks data.frame.
+intersection_of_breaks <- lapply(common_samples, function(sample_id) {
```
jashapiro: I think I figured out why I don't love this. its because the function you define here uses sv_breaks and cnv_breaks, neither of which is an argument to the function itself. I'd rather see something like a function `intersect_breaks(sample_id, sv, cnv)` followed by `lapply(common_samples, intersect_breaks, sv = sv_breaks, cnv = cnv_breaks)`

cansavvy: Made this its own specific argument outside of the lapply [in this commit. ](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/518/commits/31c81f816218d54ad9c1c1af3d77d42050c0bf68)

jashapiro:
Okay, I figured out why I am confused. My mental model and the code are opposite. See photo. This does not make one right or wrong, just different.
IMG_0758

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,279 @@
...
+# Bring along the sample IDs
+names(sample_densities) <- common_samples
+```
+
+Turn each dataset into its own data.frame and write it to its own TSV file. 
+This will be used by `02-a-plot-chr-instability-heatmaps.Rmd` for heatmaps. 
+
+```{r}
+# Turn intersection data into its own data.frame
+intersection_df <- purrr::map(sample_densities, "intersection_of_breaks") %>%
```
jashapiro: I found this confusing, until I found this note on the purrr cheat sheet.
Screen Shot 2020-02-05 at 14 11 09

jashapiro:
> Okay, I figured out why I am confused. My mental model and the code are opposite. 

But they can be made the same with (`purrr::transpose()`)[https://purrr.tidyverse.org/reference/transpose.html]!! Hooray for new functions

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -185,8 +200,55 @@ break_density <- function(breaks_df = NULL,
...
+    # Get the sum of the length of all excluded regions for each bin. 
+    excluded_length_per_bin <- tapply(na_regions@ranges@width, # Get length of each sequence
+                                      na_overlaps@from, # Index of which bin it overlaps
+                                      sum) # Add up per bin
+    
+    # Get the total bin length for each bin that has excluded regions
+    bin_length <- bins[unique(na_overlaps@from)]@ranges@width
+    
+    # Calculate the percent overlap, adding up the na regions within a bin
+    pct_overlap <- excluded_length_per_bin / bin_length
```
jashapiro: Is there a way to be sure that these two are in the same order? It seems like they should be, but is it guaranteed?

analyses/chromosomal-instability/util/chr-break-calculate.R
```diff
@@ -185,8 +200,55 @@ break_density <- function(breaks_df = NULL,
...
+
+    # Get the bin indices that correspond to less than the cutoff
+    bin_indices <- na_overlaps@from[which(pct_overlap > perc_cutoff)]
```
jashapiro: If `unique()` above on line 227 is doing something, then these pct_overlaps won't necessarily be the same indexes, right? Or is the `unique` not needed there?

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,279 @@
...
+It contains both SV data and CNV data for its intersections. 
+We'll go by SV data here. 
+
+```{r}
+# For simplicity's sake, change the name of the intersection column we want to use
+breaks_list$intersection_of_breaks <- breaks_list$intersection_of_breaks %>%
+  dplyr::rename(
+    chrom = sv_ranges.seqnames,
+    coord = sv_ranges.start
+  )
```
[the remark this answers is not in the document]
cansavvy: Yes. But also I'm not happy with my "solution" so if you have a better one, I'd like to use it. When I implemented the gap option, this means the coordinates that are intersecting here are not *exactly* identical between CNV and SV data. Which means we have two sets of coordinates and here I've arbitarily chose to use the SV coordinates to represent the intersection data. But I could have just as easily use the CNV coordinates. But I don't want to get rid of these column names when I originally make the data because I don't want this information to get lost.

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,279 @@
...
+```
+
+Calculate sample breakpoint densities across the binned genome.
+
+```{r}
+# Get a big list of break densities for each sample.
+sample_densities <- lapply(common_samples, function(sample_id) {
+  all_break_density(breaks_list = breaks_list,
```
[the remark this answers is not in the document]
cansavvy: Okay.

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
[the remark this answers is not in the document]
cansavvy: Given your most recent great find of `purrr::transpose`, this step doesn't happen anymore anyway.

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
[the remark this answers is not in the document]
jashapiro: That makes sense. I guess if there were a reason that the positions were not the same, you could calculate the average of the start and end positions, but that seems like it is probably overkill.

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Per our in person discussions and tests, I adjusted this and adjusted these steps to be less rickety. See [this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/518/commits/beba0bc80429a467a07244e3ffbf4a42a0f7239e)

analyses/chromosomal-instability/util/chr-break-calculate.R
cansavvy: Yes. I think we found this was not exactly right. But is now fixed with most recent commit.

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,275 @@
...
+
+```{r}
+common_samples <- unique(breaks_list$intersection_of_breaks$samples)
+```
+
+Set up metadata
+
+```{r}
+# Read in the metadata
+metadata <- readr::read_tsv(file.path(data_dir, "pbta-histologies.tsv")) %>%
```
jashapiro: Move to files section

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,275 @@
...
+  dplyr::filter(Kids_First_Biospecimen_ID %in% common_samples)
+```
+
+Set up chromosome size data that will be used for creating genome bins.
+It just so happens that this BED file: `WGS.hg38.strelka2.unpadded.bed` is actually 
+just a list of the chromosome sizes so we are using that for now. 
+
+```{r}
+# Set up Chr sizes
+chr_sizes_bed <- readr::read_tsv(file.path(data_dir, "WGS.hg38.strelka2.unpadded.bed"),
```
jashapiro: To the files section, Batman!

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,275 @@
...
+too high of a percentage of uncallable regions. 
+
+```{r}
+uncallable_bed <- readr::read_tsv(
+  file.path(
+    "..",
+    "copy_number_consensus_call",
+    "ref",
+    "cnv_excluded_regions.bed"
+  ),
```
jashapiro: Another file? Where should this go?

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,275 @@
...
+    )
+}
+```
+
+### Read in data 
+
+Import data that was set up in breakpoint format in `00-setup-breakpoint-data.R`.
+
+```{r}
+breaks_list <- readr::read_rds(file.path(output_dir, "breaks_lists.RDS"))
```
jashapiro: Move to files section.

analyses/chromosomal-instability/01-localization-of-breakpoints.Rmd
```diff
@@ -0,0 +1,275 @@
...
+names(group_densities) <- tumor_types
+```
+
+Save list of `GenomicRanges` objects to an RDS file.
+
+```{r}
+# Save to an RDS file
+readr::write_rds(
+  group_densities,
+  file.path(output_dir, "histology_breakpoint_binned_counts.RDS")
```
jashapiro: Maybe this should go up there too. But not as critical.

Issue #257, opened 2019-11-11
Proposed Analysis: Run brain TCGA SNV data through PBTA SNV caller pipeline


cansavvy:
#### What are the scientific goals of the analysis?

Related to #3 Tumor Mutation Burden Analysis. 

In order to make the pediatric brain tumor SNV comparison to TCGA adult data more interpretable, we should run those data through the same mutation callers and parameters as was used for the PBTA data.

#### What methods do you plan to use to accomplish the scientific goals?

This would include running the brain-related TCGA data through Strelka2, Lancet, and Mutect2 each with whatever parameters are used for PBTA data. Upon recieving the output from these callers for the TCGA data, I would make SNV consensus file for TCGA data in the [same way as I did for the PBTA data](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers).

I think VarDict can be excluded from this analysis, since it has been oversensitive in its calls with the PBTA data. 

#### What input data are required for this analysis?

These brain related TCGA data:
TCGA-LGG
TCGA-GBM
TCGA-PCPG

These data can be obtained from https://portal.gdc.cancer.gov/repository but require controlled access. 

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

This is something that I am unsure of, but I expect it would be about the same amount of time it took to run the PBTA dataset through the mutation callers. 

#### Who will complete the analysis (please add a GitHub handle here if relevant)?

@jharenza who do you suggest? 

#### What relevant scientific literature relates to this analysis?

GBM : https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3910500/
LGG : https://www.nejm.org/doi/full/10.1056/NEJMoa1402121
PCPG: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5643159/pdf/nihms849815.pdf

jharenza:
@yuankunzhu can you share here the figures you recently generated comparing GDC pipelines with our pipelines? We did a pilot comparison between TMB from their pipeline and ours and do not see much of a difference in TMB, so in an effort to save money and time on computation, we likely will not be performing this analysis.

cgreene:
@jharenza : the processing has changed quite a bit since the initial dataset (adding new callers, etc). It might be wise to process at least some random subset of the TCGA tumors with at least some subset of callers and then compare over regions that have adequate coverage for that caller with our samples.

[...]

Issue #512, opened 2020-02-04
TCGA Lancet data has only "calls" with t_alt_count = 0 or NA


cansavvy:
#### What data file(s) does this issue pertain to?
The TCGA Lancet data : `pbta-tcga-snv-lancet.vep.maf.gz` of version 14

#### What release are you using?

v14

#### Put your question or report your issue here.
I was confused because Lancet's data for TCGA was not agreeing at all with Mutect or Strelka. 
I looked into the VAF distributions and saw Lancet was all zeroes, because `t_alt_count` is only zeroes and NAs. `t_alt_count`'s  of 0 shouldn't be calls. Did something happen with a filtering step? It looks like it may have been filtered by `n_alt_count` > 0 ? 
```
lancet <- data.table::fread("../../data/pbta-tcga-snv-lancet.vep.maf.gz", data.table = FALSE)
summary(lancet$t_alt_count)

  Min. 1st Qu.  Median    Mean 3rd Qu.    Max.    NA's 
      0       0       0       0       0       0     622 
```
As a positive control, and by contrast, Mutect and Strelka show no 0's for `t_alt_count`
```
mutect <- data.table::fread("../../data/pbta-tcga-snv-mutect2.vep.maf.gz", data.table = FALSE)
summary(mutect$t_alt_count)
Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
   2.00    4.00    8.00   19.09   21.00 3528.00 
```

jharenza:
Hi @cansavvy - thanks for noticing this - was this also an issue in V13? I am assuming so, since this file should not have changed. I am cc-ing @tkoganti, @migbro, and @yuankunzhu for this.

jaclyn-taroni:
I can confirm that it was also an issue for v13:
```r
> lancet <- data.table::fread("data/release-v13-20200116/pbta-tcga-snv-lancet.vep.maf.gz", data.table = FALSE)
> summary(lancet$t_alt_count)
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max.    NA's 
      0       0       0       0       0       0     622 
```

jharenza:
Quick spot check on individual MAFs - seems they are all like this, so not a merge/filter issue (no filtering for these). I asked those above to check out the run, as it appears something went wrong there.

jharenza:
[...]

PR #521, opened 2020-02-06
TCGA Consensus Run

cansavvy:
If we run the TCGA data through the same variant callers and consensus methods we used for PBTA data, do we get a comparison that more accurately meets our expectations? (aka that adults have a higher tumor mutation burden). 

This PR is draft because the results are still quite raw and I don't totally trust them yet. But I want to get this posted in case someone notices something with the consensus pipeline for TCGA that I have missed. 

#### What was your approach?
- I ran the TCGA files from v14 through the same pipeline we used for the PBTA data. You will note that there were a few metadata cleaning steps that needed to be adjusted for TCGA so there is a new option (`--tcga`) that needs to be used for the 03-calculate-tmb.R script.

#257 

- Is there anything where the TCGA data is not being handled properly by the snv-caller pipeline? 
- I was assuming that the same BED files should be used for TMB calculations
- Is there anything that needs adjusting for the TCGA data that I have missed? 

There are results (see notebooks below), but I don't trust them yet. They need some fine tooth combing and investigating. 

Here are the updated rendered notebooks: 
https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers_plots-tcga.nb.html

 https://cansavvy.github.io/openpbta-notebook-concept/tmb-compare-tcga/compare-tmb-update.nb.html

For reference, this is what this comparison notebook looked like before using the updated data: 
https://cansavvy.github.io/openpbta-notebook-concept/tmb-compare-tcga/compare-tmb.nb.html

The comparison doesn't look like what we'd expect, need to go over all methods to determine whether there is anything in the analysis that is amiss. 

#### Documentation Checklist


jashapiro:
Looking at the agreement among callers, the TCGA Lancet results still seem strange. Here are the agreements among callers from the PBTA samples for exons: 


download-4
And here are the TCGA samples
download-3


Aside from the lack of VarDict, the big difference is that the Lancet calls are **much** more likely to be Lancet-only in the TCGA set (even after adding in Lancet + VarDict in the PBTA), and we have a much smaller proportion of of agreement among all three (or four) callers.

The nucleotide biases in the Lancet samples also seem very different between the two data sets. In PBTA all callers have similar biases, but in the TCGA data, the Lancet data is very skewed toward `A>C` and `T>G`. (Mutect looks a bit strange in TCGA too, but not nearly as strange).
download-2


I'm wondering if it is possible that the Lancet data was processed with a different genome build? Or some other difference in the way lancet was run on the two sets that would result in odd patterns like this?

I am tagging in @jharenza, @tkoganti, @migbro, and @yuankunzhu for this, as they were all mentioned in #512

jharenza:
@jashapiro `GRCh38.d1.vd1.fa` was the genome version used for TCGA [here](https://gdc.cancer.gov/about-data/data-harmonization-and-generation/gdc-reference-files). It [looks like](https://portal.gdc.cancer.gov/files/69d03ec5-2e0d-45ea-a568-4740815007c0) this is patch 0, while PBTA is patch 12. At the time of the initial run, I recall it was decided to not convert from BAM->FQ->CRAM because that process was going to push back getting the results for several days and it wouldn't make it into the last data release (but I guess that is moot now!). 

@tkoganti also just pointed out that the BED files used were the same as were used for the PBTA WXS, but we should have used the WXS BED files for whichever kit(s) were used in TCGA. @cansavvy do I recall from another ticket that you had found these BED files accessible or kit information somewhere?

We will try to locate the kit information and WXS BED files and put this in as an end-to-end workflow to-do for the next release.

jashapiro:
I don't feel like patch changes should make enough of a difference to explain these discrepancies, but I don't really have other ideas why lancet should behave so differently with this set when it was fairly concordant in the PBTA set. We will investigate some more at this end.

cansavvy:
> @cansavvy do I recall from another ticket that you had found these BED files accessible or kit information somewhere?

Because I understood the BED files you use to be variant caller specific, I used the same BED files for the TCGA data as I did for the PBTA data. Previously I had used the MC3 data and their associated WXS BED files, but I am not sure if it makes sense to use it here. The target BED regions file can be found on this page: https://gdc.cancer.gov/about-data/publications/mc3-2017
It's the gencode.v19.basic.exome.bed that I used. 

> We will try to locate the kit information and WXS BED files and put this in as an end-to-end workflow to-do for the next release.

If you can share whatever BED files you use in the release, then I can calculate TMB based on them.

jharenza:
Thanks @cansavvy!

So the normal workflow should be to use in the variant calling the same BED intervals as were used to create the WXS or targeted capture panel. Since we only had WXS from one lab and they only used one specific library prep kit for this, it was easy to just use that one WXS BED (``StrexomeLite_hg38_liftover_100bp_padded.bed`) we had in the release. This changed when we processed the targeted panel data, as the BED had fewer regions, and that is the `StrexomeLite_Targets_CrossMap_hg38_filtered_chr_prefixed.bed`.

I somehow missed the fact that we used the first BED above when processing TCGA data. We should have used the BED you just pointed out. That being said, you can process with what is in the release now, while we redo this processing using the correct BED. I noticed that gencode v19 is in hg19, so we will have to liftover those regions to hg38, then add the 100bp padding, as we did before, then process and release those BEDs as well. Does that help?

cansavvy:
Going to close this PR because I think it served its purpose of sharing the revamped TCGA data and  facilitating the conversation we had around it.

PR #528, opened 2020-02-07
Account for tp53_nf1_score when choosing biospecimens to subset for CI

### LINKED ISSUE #527: CI: Enrich for samples with TP53 and NF1 mutations that are in the stranded dataset
We need to account for the requirements to successfully run `tp53_nf1_score` when generating files in CI. Specifically - we can skip the poly-A and test only the stranded dataset (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/523/commits/da8db99b50236d74ed1a5ee9571d152b383f3248) but we have to make sure to do the following:

* Include positive examples in the SNV files (e.g., biospecimen IDs with  _TP53_ and _NF1_ that are likely to be functional/pass the filtering here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/a5dd6c122750fa145d8a60629f79ed856ee94c36/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R) However - this is not enough (this didn't fix it: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/515/commits/18f335e8296321272ae6eabeef73f7407012f686).
* We also need select those biospecimens that have RNA-seq data in the stranded dataset - I might approach this by including a notebook in `create-subset-files` that looks into this and then hardcoding those values.

I'm going to comment that module out of CI until we make changes to `create-subset-files` so it doesn't block things. The alternative would be to include error handling that skips the ROC steps, but I think that puts us in a situation where we're not testing everything.

jaclyn-taroni:

Here I am adding a short notebook that examines and randomly selects positive examples of _TP53_ and _NF1_ mutations that pass the criteria in [`analyses/tp53_nf1_score/00-tp53-nf1-alterations.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/a5dd6c122750fa145d8a60629f79ed856ee94c36/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R) when those biospecimen IDs come from samples that are represented in the stranded RNA-seq dataset. You can view that notebook here: https://jaclyn-taroni.github.io/openpbta-notebook-concept/enrich-positive-examples.nb.html

I then hardcode these biospecimen IDs (both the DNA-seq and stranded RNA-seq IDs) into `analyses/create-subset-files/01-get_biospecimen_identifiers.R` and add them to the list of biospecimen IDs that are used to generate the subset files for CI -- specifically, the stranded RNA-seq files and any file that begins with the prefix `pbta-snv`.

I tested the `tp53_nf1_score` code locally and the stranded section should pass CI when the new files are generated and uploaded to S3.

Closes #527

`create-subset-files` does not get run in CI.

#### Documentation Checklist


jaclyn-taroni:
@cansavvy can we get this running on AWS this morning please? Thank you!

jaclyn-taroni:
The `tp53_nf1_score` step passed CI 🎊

analyses/create-subset-files/01-get_biospecimen_identifiers.R
```diff
@@ -181,6 +181,13 @@ num_nonmatched <- ceiling(0.1 * num_matched_participants)
 # set the seed
 set.seed(opt$seed)
 
+#### Samples we need to include to run tp53_nf1_score --------------------------
+
+tp53_dnaseq <- c("BS_3E5C1PN1", "BS_KHSYAB3J", "BS_ZR75EKKX")
```
cansavvy: May wanna put a comment that explains how these came from the 00-enrich-positive-examples notebook.

PR #490, opened 2020-01-30
Ependymoma subtyping

tkoganti:
## Purpose/implementation
ependymoma subtyping 

#### What scientific question is your analysis addressing?
Create file with  fusion results, gene expression, GISTIC, breaks density and gsva scores got ependymoma samples according to the ticket here (https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/245)

#### What was your approach?
1) Take only ependymoma samples from `pbta-histologies.tsv`.
2) Based on primary site, classify them as supratentorial/ infratentorial (Some as "NA" since they s ay `Ventricles` as primary site)
3) Match corresponding RNA and DNA sample BSID's based on common `sample_id` column in `pbta-histologies.tsv`
4) Based on the files mentioned above in scientific question, fill up table for ependymoma samples based on specific fusion, gene expression, broad CNA etc. 
 

#### Which areas should receive a particularly close look?
1) All scripts should run from OpenPBTA folder. Any feedback on that? 
2) Column headers and file formats for final files okay? 

### Results
OpenPBTA/OpenPBTA-analysis/analyses/molecular-subtyping-EPN/results/EPN_all_data.tsv

#### What types of results are included (e.g., table, figure)?
tables 

#### What is your summary of the results?
The information in final EPN_all_data.tsv will help subgroup samples into different  subtypes

PR #532, opened 2020-02-10
Chr instability: PR 3 of 3: Histology plots

cansavvy:
This includes redone versions of the material that was included in #492. 

This last PR has the last of three notebooks which contain the material that was originally all in the previous 01 notebook.

How does chromosomal instability relate to histology group? 

#### What was your approach?
This takes the binned counts from the 01-localization notebook and makes histology plots for them. 

#487

Most of this material has been reviewed previously, but it now is in its own notebook. 
How do you feel about the readability of it? Are there other aesthetic changes that need to be made to the plots? 

Here's the rendered html: https://cansavvy.github.io/openpbta-notebook-concept/chromosomal-instability/02b-plot-chr-instability-by-histology.nb.html

These items were done previously.

#### Documentation Checklist
These items were done previously.
- [x] This analysis module has a `README` and it is up to date (Note it includes documentation for the upcoming notebooks as well. 


analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,195 @@
...
+
+```{r}
+breaks_cdf <- total_break_densities_df  %>%
+  # Only plot histologies groups with more than `min_samples` number of samples
+  dplyr::group_by(short_histology, add = TRUE) %>%
+  dplyr::filter(dplyr::n() > params$min_samples) %>%
+  # Calculate histology group mean
+  dplyr::mutate(
+    hist_mean = mean(breaks_density),
+    hist_rank = rank(breaks_density) / dplyr::n()
```
jashapiro: Why is this divided by the group size? Oh... it is not the rank, but a relative rank, got it.

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,195 @@
...
+  ggplot2::ylab("Chromosomal Breaks per Mb") +
+  # Transform to log10 make non-log y-axis labels
+  ggplot2::scale_y_continuous(trans = "log10", breaks = c(0, 1, 10, 30)) +
+  ggplot2::theme(legend.position = "none") +
+  # Get rid of x-axis ticks
+  ggplot2::theme(
+    axis.text.x = ggplot2::element_blank(),
+    axis.ticks.x = ggplot2::element_blank(),
+    strip.text = ggplot2::element_text(size = 10)
+  )
```
jashapiro: I did some editing here for a nicer plot, IMO. Main changes are strip.placement, strip.position, strip.background and axis.text.x.

```r
  ggplot2::ggplot(ggplot2::aes(
    x = hist_rank, 
    y = breaks_density,
    color = short_histology
  )) +
  ggplot2::geom_point() +
  # Add summary line for mean
  ggplot2::geom_segment(x = 0, xend = 1, color = "grey", 
                        ggplot2::aes( y = hist_mean, yend = hist_mean)) +
  # Separate by histology
  ggplot2::facet_wrap(~short_histology, nrow = 1, strip.position = "bottom") +
  ggplot2::theme_classic() +
  ggplot2::xlab("") +
  ggplot2::ylab("Chromosomal Breaks per Mb") +
  # Transform to log10 make non-log y-axis labels
  ggplot2::scale_y_continuous(trans = "log10", breaks = c(0, 1, 3, 10, 30)) +
  ggplot2::scale_x_continuous(limits = c(-0.1, 1.1), breaks = c()) +
  ggplot2::theme(legend.position = "none") +
  ggplot2::theme(
    axis.text.x = ggplot2::element_blank(),
    axis.ticks.x = ggplot2::element_blank(),
    strip.placement = "outside",
    strip.text = ggplot2::element_text(size = 10, angle = 90, hjust = 1), 
    strip.background = ggplot2::element_rect(fill = NA, color = NA)
  )
```

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,195 @@
...
+  ggplot2::geom_point() +
+  # Add summary line for mean
+  ggplot2::geom_hline(ggplot2::aes(yintercept = hist_mean)) +
+  # Separate by histology
+  ggplot2::facet_wrap(~short_histology, nrow = 1) +
+  ggplot2::theme_classic() +
+  ggplot2::xlab("") +
+  ggplot2::ylab("Chromosomal Breaks per Mb") +
+  # Transform to log10 make non-log y-axis labels
+  ggplot2::scale_y_continuous(trans = "log10", breaks = c(0, 1, 10, 30)) +
```
jashapiro: The breaks aren't showing up, presumably because "breaks_density" is tiny. Lets just plot total breaks/sample? label as Breaks per genome?

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,195 @@
...
+```{r}
+# Declare name of zip file
+zip_file <- paste0(hist_plots_dir, ".zip")
+
+# Remove any current zip_file of this name so we can overwrite it
+if (file.exists(zip_file)) {
+  file.remove(zip_file)
+}
+# Zip up the plots
+zip(zip_file, hist_plots_dir)
```
jashapiro: When you make the zip file this way, they get put in plots/tumor-type within the folder, which is not really what we want.

Instead, you can try something like: 
```r
gzfile = paste0(hist_plots_dir, ".tar.gz")
plot_files = list.files(hist_plots_dir, full.names = TRUE, pattern = ".png")
tar(gzfile, plot_files, compression = "gzip")
```

jashapiro: Found the flag...
```suggestion
zip(zip_file, hist_plots_dir, extras = "-j")
```

cansavvy: For future reference, this didn't work^.

cansavvy:
Incorporated your changes, @jashapiro 👍 Let me know what you think.

cansavvy:
Now that we fixed the PNG issues and added the set -e thing, was there anything else that needs to be addressed, @jashapiro ?

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,198 @@
...
+  )) +
+  ggplot2::geom_point() +
+  # Add summary line for mean
+  ggplot2::geom_segment(x = 0, xend = 1, color = "grey", 
+                        ggplot2::aes( y = hist_mean, yend = hist_mean)) +
+  # Separate by histology
+  ggplot2::facet_wrap(~short_histology, nrow = 1, strip.position = "bottom") +
+  ggplot2::theme_classic() +
+  ggplot2::xlab("") +
+  ggplot2::ylab("Chromosomal Breaks per Mb") +
```
jashapiro:
```suggestion
  ggplot2::ylab("Chromosomal Breaks per Genome") +
```

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,198 @@
...
+  # Transform to log10 make non-log y-axis labels
+  ggplot2::scale_y_continuous(trans = "log10", breaks = c(0, 1, 3, 10, 30)) +
```
jashapiro: This transformation hides 0 values.
Perhaps try:
```suggestion
  ggplot2::scale_y_continuous(trans = "log1p", breaks = c(0, 1, 3, 10, 30)) +
```

cansavvy:
Alrighty, @jashapiro , let me know if we think everything is addressed now.

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,165 @@
...
+each bin.
+
+```{r}
+purrr::imap(hist_break_local, function(.x, name = .y) {
+  # Make the combo plot
+  multipanel_break_plot(
+    granges_list = .x,
+    plot_name = name,
+    y_val = "total_counts",
+    y_lab = "Total Breaks per Mb",
```
jashapiro: Lets make the label match the data.
```suggestion
    y_lab = "Median Breaks per Mb",
```

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
```diff
@@ -0,0 +1,165 @@
...
+
+### Plot the breakpoints for each tumor-type
+
+Here we will plot median number of break points for the tumor-type group per 
+each bin.
+
+```{r}
+purrr::imap(hist_break_local, function(.x, name = .y) {
+  # Make the combo plot
+  multipanel_break_plot(
```
jashapiro: When I run this locally, I get a big list of NULL outputs, which isn't a problem, but isn't so pretty. My quick suggestion would be have `multipanel_break_plot` return something (the plot object?), and then have the call to `purrr::imap()` saved to something new.

Or, change the block to hide output with results=FALSE, though for some reason this is giving trouble

analyses/chromosomal-instability/util/chr-break-plot.R
```diff
@@ -115,7 +115,71 @@ multipanel_break_plot <- function(granges_list,
...
 
-  # Print out the plot while we are here
-  full_plot
+breaks_cdf_plot <- function(density_file, metadata_df, cdf_plot_file) {
+  # Given a genome wide breaks density file path, plot the CDF distribution 
+  # for it by histology.
+  #
+  # Args:
+  #   density_file: a chromosomal breaks density file path where each sample is
+  #                 a row with these columns: samples, experimental_strategy,
```
jashapiro: Is experimental_strategy required here? Seems like only samples and breaks_count is used in this function.

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
cansavvy: Does this mean you want me to use median number of counts instead?

cansavvy: It is still total counts right now. Unless I'm misunderstanding things. If it were median breaks it would have been `y_val = "median_counts"` above that.

analyses/chromosomal-instability/util/chr-break-plot.R
cansavvy: Oh you are right. My comment is wrong.

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
jashapiro: Sorry, I should have been less prescriptive: The trouble is this disagrees with the description on line 131.

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
cansavvy: I added a `results=FALSE` so you don't have to see this hideous output. :)

analyses/chromosomal-instability/02b-plot-chr-instability-by-histology.Rmd
cansavvy: Gotcha. The description on line 131 is what's wrong. Fixing now.

analyses/chromosomal-instability/00-setup-breakpoint-data.R
```diff
@@ -356,7 +358,7 @@ breaks_density_list <- lapply(breaks_list, function(breaks_df) {
       samples, experimental_strategy, genome_size
     ) %>%
     # Count number of mutations for that sample
-    dplyr::summarize(breaks_count = dplyr::n()) %>%
+    dplyr::summarize(breaks_count = sum(!is.na(chrom))) %>%
```
jashapiro: I just realized, while thinking about https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/490#issuecomment-585542405, that this may make some samples that should be NA into zeros. If the sample was not in the consensus seg file, it should be NA for CNV breaks and SV breaks  I *think* such a sample here would end up with a zero here.

cansavvy: Those do end up with zeroes. If NA is preferred I can make those changes. I’ll just have to convert them back to zeroes for the CDF plots, unless we think those samples should be dropped from the CDF plots?

jaclyn-taroni: Yes, drop them if they are missing. If you don't drop them, it says to your audience that there are _n_ samples that have 0 breakpoints when we don't have evidence either way.

cansavvy: So is it correct to say that all WGS samples have been ran through both CNV detection pipelines, so if they don't show up in consensus CNV file or the SV file then they should be `NA`s for break density?

cansavvy: If I'm understanding this correctly, then there will be no such thing as a `0`, only NAs where the minimum `break_counts` would 1.

jaclyn-taroni: @cansavvy take a look at the documentation for https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/copy_number_consensus_call and also see https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/copy_number_consensus_call/results/uncalled_samples.tsv. I would also check the IDs in the  original files to be sure.

jashapiro: The docs linked to should explain, but yes, there are both zeros and NAs. Some samples were deemed “uncallable” and should be NA. Others were called but had no CNVs and should be 0.

analyses/chromosomal-instability/00-setup-breakpoint-data.R
```diff
@@ -355,10 +356,18 @@ breaks_density_list <- lapply(breaks_list, function(breaks_df) {
...
-    dplyr::mutate(breaks_density = breaks_count / (genome_size / 1000000))
+    # Count number of mutations for that sample but find out if it is NA
+    dplyr::summarize(is_na = any(is.na(chrom)), 
+                     breaks_count = dplyr::n()) %>%
+    # Calculate breaks density, but put NA for breaks_count if the sample was not 
+    # in the SV or CNV data originally
+    dplyr::mutate(breaks_count = dplyr::case_when(
+      !is_na ~ as.numeric(breaks_count), 
+      is_na ~ as.numeric(NA)
+      ),
```
jashapiro: Unfortunately, this isn't actually doing what we really want it to do. 
The trouble is that anything with a true zero count will get an NA in chrom column, just like something that is actually missing. 

I think the thing to do is to add a column to `metadata` called `surveyed` which would indicate if the sample is in  `unique(c(cnv_samples, sv_samples))`. Then if `surveyed` is TRUE, set the `is_na` `break_counts` to 0, and to NA otherwise.

something like:
```r
dplyr::mutate(
    breaks_count = dplyr::case_when(
      !is_na ~ as.numeric(breaks_count), 
      is_na & surveyed ~ as.numeric(0),
      TRUE ~ as.numeric(NA)
      ), 
```

jashapiro:
Okay, I think this is good to go. :shipit:

PR #537, opened 2020-02-14
Improve focal-cn-file-preparation documentation, filter copy neutral out of sex chromosome files

jaclyn-taroni:

I wanted to improve the `focal-cn-file-preparation` documentation to explicitly note that we filter copy neutral regions out of the files to save on memory when running the annotation step and on disk space. While I was looking at that, I noticed we did not do that for sex chromosomes so I've also made that change.

Context is https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/490#issuecomment-585991614

Is this documentation an improvement?

#### Documentation Checklist


analyses/focal-cn-file-preparation/03-prepare-cn-file.R
```diff
@@ -289,7 +289,7 @@ readr::write_tsv(autosome_annotated_cn,
 if (xy_flag) {
   # Filter to just the X and Y chromosomes
```
cbethell: We can also add here that we filter out the neutral copy number statuses, and the brief explanation for this decision.

jaclyn-taroni:
Thanks @cbethell !

Issue #529, opened 2020-02-10
Add MATLAB Runtime and the relevant MATLAB programs to project Docker container


jaclyn-taroni:
We currently run GISTIC upstream of the project. It would be great to include the GISTIC steps in the project repository, as that would give us a better opportunity to document and encapsulate the entire CNA _analysis file_ pipeline and give us more flexibility for how GISTIC is run (e.g., on individual histologies with larger sample sizes, different parameters for a broad change) if desired. Notably, we may also want to use something like MutSigCV (#517) and that would have similar constraints (more on that below). 

To include the GISTIC or MutSigCV steps within the project, we will need to install these on the project Docker container. Both of these programs appear to have executable versions that can be run if the MATLAB Compiler Runtime libraries are installed.

From the GISTIC `README.txt`:

> The executable is the gene pattern module 'gp_gistic2_from_seg'. This
module depends on the libraries in the MATLAB Compiler Runtime (MCR)
version 8.3 (R2014a).

And from the [MutSig documentation](https://software.broadinstitute.org/cancer/cga/mutsig_run):

> If you do not have a license for Matlab, you can run the compiled version of MutSigCV using the free Matlab MCR:
>
> `run_MutSigCV.sh <path_to_MCR> mutations.maf coverage.txt covariates.txt output.txt`

At least GISTIC comes with an MCR Installer, but the size of that installation is non-negligible (~0.5 GB), so we should plan to install it _once_ and symlink it to the individual tools' folders.

Here is an example Dockerfile for MCR: https://hub.docker.com/r/usuresearch/matlab-runtime/dockerfile (H/T @cgreene)

jaclyn-taroni:
I am going to look into this today.

jaclyn-taroni:
@jharenza - relevant to this issue but also more generally - what is the markers file used here: https://github.com/d3b-center/OpenPBTA-workflows/blob/cb87a2b725d0e41d34a88436492830802c40f7f0/bash/run-gistic.sh#L16

jaclyn-taroni:
It seems like it might be optional?

jharenza:
[...]

PR #540, opened 2020-02-14
Docker image 'housekeeping' to prepare for additional MCR versions

jaclyn-taroni:

We're interested in adding other versions of MATLAB Compiler Runtime required for other methods to the project Docker image (#545, #546). Here I'm changing a folder name to potentially make things a bit cleaner/clearer down the road. I'm also removing some compressed files that were added to layers during installation.

Related to #529


jaclyn-taroni:
> But how did you know R2014a was v83?

They tell you in the `Release (MATLAB Runtime Version#)` column here: https://www.mathworks.com/products/compiler/matlab-runtime.html

PR #534, opened 2020-02-12
Clustering of HGAT samples with `BRAF V600E` mutations

cbethell:
#### Purpose/implementation Section

The purpose of this PR is to add a visualization to highlight the HGAT samples without histone mutations that have BRAF V600E mutations in a t-SNE/UMAP clustering of all of the data by broad histology. 

#### What scientific question is your analysis addressing?

If the HGAT samples with BRAF V600E mutations cluster with the LGAT, then we think we can re-categorize them.

#### What was your approach?

I used the functions (some of which I adapted) from the `analyses/transcriptomic-dimension-reduction` module to perform the t-SNE and UMAP clustering, produce a list of plots (saved in the `scratch` directory), and produce two final multipanel plots (saved in this module's `plots` directory).

#### What GitHub issue does your pull request address?

This PR addresses issue #499.

#### Which areas should receive a particularly close look?

The generated plots. I filtered the data.frame to include only broad histologies relevant to HGAT and LGAT samples (including all histologies overwhelmed the plot with colors and there aren't an adequate amount of shapes to represent them). Would you have any alternative suggestions to the decision I made here?

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

Yes, this is ready for review (although I may be updating comments and possibly implement any refactoring I may think of in going through the PR). 

#### What types of results are included (e.g., table, figure)?

Two multipanel plots are produced and stored in `plots` as follows:
- `hgg_polya.pdf`
- `hgg_stranded.pdf`

Each multipanel plot shows the visualization of results from the t-SNE and UMAP clustering.

#### What is your summary of the results?

The rendered notebook can be found [here](https://cbethell.github.io/open-pbta-output/09-HGG-with-braf-clustering.nb.html).

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #542, opened 2020-02-17
Rerun `transcriptomic-dimension-reduction` module

cbethell:
- rerun module with v14 data files 
- add `generate-multipanel-plot-functions.R` script in `util`
- implement use of custom function in `generate-multipanel-plot.R` 
- propogate changes to `03-multipanel-plots.sh`

The purpose of this PR is to rerun the `transcriptomic-dimension-reduction` module using the v14 data release files. 

This PR also adds a function script that can be used in the `generate-multipanel-plot.R` script of this module, in the `09-HGG-with-braf-clustering.Rmd` notebook in the `analyses/molecular-subtyping-HGG` module (open PR #534), and in any other future instances where the generation of a multipanel plot is needed.

None, this PR is just updating the module and refactoring a script.

This PR does not address one issue in particular, it instead stems from the [comment made on open PR #534](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/534#discussion_r379169175).

The`generate-multipanel-plot-functions.R` script in this module's `util` directory and the `generate-multipanel-plot.R` script in this module's `scripts` directory should both receive a close look as these contain the new changes.

No additional results have been added to this module.

This PR is an update to the module so this all was done previously.

#### Documentation Checklist


analyses/transcriptomic-dimension-reduction/scripts/generate-multipanel-plot.R
```diff
@@ -40,6 +46,11 @@ if (!dir.exists(plot_directory)) {
   dir.create(plot_directory, recursive = TRUE)
 }
 
+#### Source custom function ---------------------------------------------------
```
jaclyn-taroni: Why not hard code the script that needs to be sourced? See another example from this module.

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/46cf6ccb119312ccae6122ac94c51710df01f6da/analyses/transcriptomic-dimension-reduction/scripts/run-dimension-reduction.R#L56

Do you see any reason why the input would change?

cbethell: Good question. I do not anticipate the input changing so I'll go ahead and hard code this.

cbethell:
In the last commit, I was able to retain the values that were "missing" from the `molecular_subtype` variable in the aligned output TSV files, but was unable to determine the source of variation in the dimension reduction scores. 

I confirmed that the values in the expression files (from v13 and v14 releases) do match, and the seed remained the same (hence, my confusion).

jaclyn-taroni:
I think the only method with notable changes is t-SNE. However, I've been poking around locally using the project Docker image and running the same thing multiple times does not result in different values, which suggests to me that the `set.seed` is being respected/is in the correct spot/etc. The first time I ran it locally (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/542/commits/9e58fe2aa613900cf16503f70d638b7e4790b516) I _did_ get different values than what was in this branch (minor changes in PCA, differences in t-SNE, UMAP was the same) so I am wondering if it comes down to some difference in environment.

Issue #229, opened 2019-11-05
Proposed Analysis: Comparative RNA-Seq analysis


GeoffLyle:
#### Scientific goals

- Create correlation matrices for polyA samples and ribodeplete (stranded) samples using gene expression data.
- Generate gene outlier thresholds for ribodeplete samples.
- List outlier genes for each ribodeplete sample.
- Discover trends in outlier genes by tumor subtype.
- Investigate possible targeted therapeutics based on outlier expression profile.

#### Proposed methods

- The correlations will be calculated using pairwise Spearman correlation of the RNA-Seq gene expression profiles.
- The gene expression profile data will be filtered using a developed method described in Vaske et al.
Note: This process is the precursor to creating a TumorMap (https://tumormap.ucsc.edu/). While programmatic creation of a TumorMap is not currently open access, we can add instructions for users on how to manually generate a TumorMap on the website.
- Outlier thresholds will be calculated according to the Tukey method.
- Discovery of outlier genes and possible targeted therapeutics will be based on the Treehouse CARE method. https://github.com/UCSC-Treehouse/CARE
  - The CARE method will be used to find outliers using the entire population as a cohort as well as subsamples from the population to disentangle sample specific outliers from tumor subtype outliers.
- Application of novel transfer learning method trained on CCLE response data to Open-PBTA samples will predict response to therapeutics.

#### Required input data

`pbta-gene-expression-kallisto.polya.rds`
`pbta-gene-expression-kallisto.stranded.rds`

Is it possible to get RSEM TPM values for this data?
That would make it easier for us to adapt existing algorithms to work with the Open-PBTA data.

#### Proposed timeline

8 weeks

#### Relevant literature

Method published in [Vaske et al. Jama Open Network. 2019.] (https://jamanetwork.com/journals/jamanetworkopen/article-abstract/2753519)

jharenza:
hi @GeoffLyle! Thanks for proposing this analysis! We currently have RSEM FPKM values available in our data release: `pbta-gene-expression-rsem-fpkm.polya.rds` and `pbta-gene-expression-rsem-fpkm.stranded.rds`. Since FPKM and TPM from RSEM are highly correlated, will this do, or does your analysis require TPM? If the latter, I can file an issue to add the TPM data during our next sprint, which starts Nov. 13 and the data could be expected sometime around the end of next week.

GeoffLyle:
[...]

PR #544, opened 2020-02-19
(Comparative RNA-Seq Analysis) Apply sample filters (QC + tumor status); convert scratch files to .feather

e-t-k:

Gene expression outliers are only meaningful if the samples which comprise the background distribution have valid expression data; inclusion of any invalid samples affects the outlier results for all, including valid samples.

#### What was your approach?

When generating normalized dataset, drop samples that did not pass MEND QC (ie, have fewer than 10 million Mapped Exonic Non-Duplicate reads); also remove non-Tumor samples per @jaclyn-taroni's [helpful guidelines](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/229#issuecomment-575919379).

Other changes:

- Modify the gene filters output file in step 1 to retain which filter (Expression or Variance) was responsible for dropping any dropped gene.
Use .feather instead of .rds for scratch files since this is in fact python.
- Consolidate path handling in step 01 into the `main()`  function to match step 2.

#229 

- ~~I'm filing this as a draft pull request solely in the hopes that this will enable Allow Edits from Maintainers (see https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/513#discussion_r375573075); I don't need feedback on splitting into multiple pulls.~~ [Edit: This doesn't seem to have helped. --ETK]

- The new methods in step 01, get_qc_status and filter_samples.

- [Step 01 line 29](https://github.com/UCSC-Treehouse/OpenPBTA-analysis/blob/filter-samples-and-cleanup-stuff/analyses/comparative-RNASeq-analysis/01-correlation-matrix.py#L29) has a lot going on in it. I can break it up if you prefer - thoughts?

- Step 01 is beginning to accumulate a lot of arguments - would it be better to hardcode the `pbta-mend-qc-manifest.tsv` , `pbta-mend-qc-results.tar.gz`, `pbta-histologies.tsv ` paths under the assumption they will stay the same for all datasets?

n/a; no resulting figures/tables

A data file listing outlier genes for all samples (as well as QC status) is provided at `results/rsem-tpm-stranded-gene_expression_outliers.tsv.gz`; this file has been generating using the updated (sample filtered) code.

None yet. Next steps are *Discover trends in outlier genes by tumor subtype*  and *Investigate possible targeted therapeutics based on outlier expression profile*, both to be addressed.

#### Documentation Checklist


e-t-k:
And once again I seem to be unable to "Allow Edits from Maintainers"; apologies. For my next pull I'll try forking it under my own identity rather than Treehouse and see if that clears it up. Stay tuned.

jashapiro:
> Step 01 is beginning to accumulate a lot of arguments - would it be better to hardcode the pbta-mend-qc-manifest.tsv , pbta-mend-qc-results.tar.gz, pbta-histologies.tsv paths under the assumption they will stay the same for all datasets?

I think the best move here, now that you have three steps running set up and running, is to wrap them in a shell script with all of the file paths set there, and then call the shell script in CI with a single `-run:` step. That can also provide a nice usage example for each of the scripts that is easier to find.

There are a few examples of this scattered around the repo, but this is [one recent example](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-EPN/run-molecular-subtyping-EPN.sh) I was working with.

analyses/comparative-RNASeq-analysis/01-correlation-matrix.py
```diff
@@ -11,34 +12,84 @@
...
+    raw_samples = utils.read_rds(input_matrix)
+
+    if nofilter:
+        print_v("Skipped the QC and tumor status filters; using all samples.")
+        return raw_samples
+
+    # Filter to only samples with a MEND QC status of PASS
+    qc_status = get_qc_status(mend_results_path, mend_manifest_path)
+    qc_pass_ids= [k for k, v in qc_status.items() if v == "PASS"]
+    qc_pass_samples = raw_samples[set(raw_samples.columns).intersection(qc_pass_ids)]
```
jashapiro: While it should be the case that every sample has a QC call, is that something we ever check? I'd assume you would not want to include a sample without a matched QC, but it might be worth calling out separately in the log?

analyses/comparative-RNASeq-analysis/01-correlation-matrix.py
```diff
@@ -11,34 +12,84 @@
...
-def prepare_expression(input_matrix,
-                       scratch_dir,
+def get_qc_status(mend_results_path, mend_manifest_path):
+    """Takes: mend results tgz containing bam_umend_qc.tsv files, and
+    manifest containing mapping from filename to sample id.
+    returns dictionary from sample id to mend status: PASS or FAIL"""
+    mend_tgz = tarfile.open(mend_results_path)
+    qc_files = (i for i in mend_tgz.getmembers() if i.name.endswith("bam_umend_qc.tsv"))
+    # Dictionary of filename (UUID.bam_umend_qc.tsv) to PASS or FAIL string
+    filename_map = { i.name : mend_tgz.extractfile(i).readlines()[1][-5:-1].decode("utf-8") for i in qc_files }
```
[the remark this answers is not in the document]
e-t-k: Added a check for `PASS`/`FAIL`, throwing a ValueError if something else is found.

I'll add the shell script in the next pull.

analyses/comparative-RNASeq-analysis/01-correlation-matrix.py
e-t-k: Added a check for samples that didn't have matched QC. Since this indicates missing QC rather than existing QC that has somehow gone wrong (as your previous comment does), I'm comfortable moving forward without those samples rather than crashing the whole analysis.

e-t-k:
I've also updated 01 to maintain column (samples) order from original .Rds -> outlier results, because it sometimes changed between runs due to filtering on sets which is inconvenient for diff testing. I don't have any further changes planned for this pull unless the ci fails.

analyses/comparative-RNASeq-analysis/01-correlation-matrix.py
jashapiro: Yup, this seems like correct approach.

jashapiro:
Looks great!

Issue #547, opened 2020-02-19
Proposed Analysis: comparison of GISTIC results (specific histology versus entire cohort)


cbethell:
#### What are the scientific goals of the analysis?

To determine whether or not the GISTIC results are in agreement when we compare those of a specific histology to those of the entire PBTA cohort.

For context, we found that there are some histologies in the PBTA cohort that contain more samples than others (ie. LGAT). This means that histologies with higher n samples may be driving the results of our analyses. This comparison analysis will help us decide whether or not we should handle our downstream analyses in a histology specific manner.

#### What methods do you plan to use to accomplish the scientific goals?

In `analyses/run-gistic/results` there are four zip files. One contains the GISTIC results for the entire cohort, and the other three contain the GISTIC results for three individual histologies (LGAT, HGAT, and medulloblastoma).

The plan for this analysis is to: 
1. Visualize the agreement/disagreement in the `scores.gistic` files for each of the individual histologies compared to that for the entire cohort by adapting the code in [`analyses/cnv-chrom-plot/gistic_plot.Rmd`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/56f82b9f23d369a5e208c7564ce8fc22fa0b64dd/analyses/cnv-chrom-plot/gistic_plot.Rmd#L132). 
2. Use a Venn diagram to visualize the agreement/disagreement in the lists of amplified and deleted genes for each of the individual histologies compared to that for the entire cohort.
3. Use appropriate visualizations to observe the degree of agreement between each of the GISTIC result files that the individual histologies have in common with the entire cohort.

This will also be represented in a R notebook.

#### What input data are required for this analysis?

The input data required for this analysis include:
- `analyses/run-gistic/results/pbta-cnv-consensus-gistic.zip`
- `analyses/run-gistic/results/pbta-cnv-consensus-hgat-gistic.zip`
- `analyses/run-gistic/results/pbta-cnv-consensus-lgat-gistic.zip`
- `analyses/run-gistic/results/pbta-cnv-consensus-medulloblastoma-gistic.zip`

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

~2 days (rough estimate)

#### Who will complete the analysis (please add a GitHub handle here if relevant)?

@cbethell

jaclyn-taroni:
Closed via the linked pull requests – subsequent analyses are tracked in #560

Issue #387, opened 2019-12-31
Updated analysis: RNA expression of copy number losses


jaclyn-taroni:
#### What analysis module should be updated and why?

[`focal-cn-file-preparation`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/focal-cn-file-preparation), specifically the `02-rna-expression-validation.R`. 

#### Why should the module be updated?

As noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/367#issue-356029454, this step gets OOM-killed for me locally and we now include collapsed RNA-seq matrices in the data download that this script could use.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

At the moment, it is not clear to me what step(s) requires a lot of RAM or takes a long time to run. This file is the first place I would look: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/focal-cn-file-preparation/util/rna-expression-functions.R

#### What input data should be used? Which data were used in the version being updated?

Previously, RSEM FPKM files were used. I propose that we use `pbta-gene-expression-rsem-fpkm-collapsed.polya.rds` and `pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds`. 

We also are currently using the ControlFreeC file produced by that module (`analyses/focal-cn-file-preparation/results/controlfreec_annotated_cn_autosomes.tsv.gz`), but we will eventually want to move to using some kind of consensus file (#128).

#### When do you expect the revised analysis will be completed?

This may take 1-3 days.

#### Who will complete the updated analysis?

Not sure.

jaclyn-taroni:
The `focal-cn-file-preparation` module is being revamped (and sped up!) over on #452. It's worth noting that those revisions may have addressed the _major_ bottleneck for that analysis module. I think profiling and improving the RNA-seq expression levels of losses code is a good step to follow #452. As such, I am going to mark this `in progress` and assign @cbethell. As noted above, we also to look at the expression levels for copy number losses that are in the consensus calls.

cbethell:
In addition to the plots in PR #493, density plots were generated using the consensus SEG autosomes files, annotated in the `focal-cn-file-preparation` module. These plots use z-scored expression values (polyA and stranded expression files are handled separately) to look at the density of copy number calls, specifically looking to validate the losses. 

[...]

PR #549, opened 2020-02-20
GISTIC comparison of cohort results vs individual histology results

cbethell:

The purpose of this PR is to identify, if any, disagreement between GISTIC results for the entire cohort versus the three individual histologies.

This PR addresses the question of whether or not we should be handling analyses that incorporate the GISTIC results separately based on individual histology.

#### What was your approach?

My approach is to visualize the disagreements/agreements between the GISTIC result files for the entire cohort and compare this to each of the individual histologies' result files.
A slightly more detailed idea of my approach is found in [the original comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/547#issue-567707513) on reference issue #547.

This PR addresses issue #547.

The plots should receive a particularly close look. 
- Are these plots adequately representing the data in an informative way?
- Are there suggestions for additional plot types that may be useful if implemented?

Yes.

The plots can be seen in the [rendered notebook](https://cbethell.github.io/open-pbta-output/gistic-cohort-vs-histology-comparison.nb.html) of the R markdown file as it is currently formatted in this PR.

Plots showing the agreements/disagreements between the `scores.gistic` files are included in the `plots` directory of this module.

This PR suggests that the results in the `scores.gistic` files and the `amp/del_genes.conf_90.txt` files differ between the entire cohort and the three individual histologies used for comparison.

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile. (No new dependencies required).

#### Documentation Checklist


analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
```diff
@@ -0,0 +1,342 @@
...
+                         plot_title = "Medulloblastoma scores versus entire cohort",
+                         output_directory = plots_dir,
+                         output_filename = "medulloblastoma_gistic_scores_multipanel_plot.png")
+```
+
+## Compare `amp_genes.conf_90.txt` and `del_genes.conf_90.txt` files
+
+### Custom function
+```{r}
+plot_amp_del_genes <-
```
jaclyn-taroni: I think what is being compared in the Venn diagram is entire strings? Imagine a situation where the wide peaks between the histology-specific run and the entire cohort GISTIC run have a lot of overlap but are not exactly the same. It could be the case that 70% of the genes from the wide peaks from either run would overlap but show up as a zero in the the Venn diagram. We would want the 70% of genes in both to be reflected in the Venn diagram. So what we're after here is looking at the overlap between vectors of gene symbols, where the vector will contain the unique set of gene symbols that are in _any_ of the significant wide peaks. Does that make sense?

analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
```diff
@@ -0,0 +1,342 @@
...
+    #   cohort_genes_file: the file path to the amp/del genes file for
+    #                           the entire cohort
+    #   histology_genes_file: the file path to the amp/del genes file
+    #                              for an individual histology
+    #
+    # Returns:
+    #  This function displays a Venn Diagram that represents the data that
+    #  overlaps/does not overlap between the two given files
+    
+    # Read in files
```
jaclyn-taroni: You are treating the two files in the exact same way, which is a good indication to me that this can be wrapped up in a function. I can imagine a situation where you might want to examine what genes are in the vectors that underlie the Venn diagram, so you can separate things out into two function: 1) read in the file and process the vector and 2) create the Venn diagram.

With the first function, you'll only need to process the entire cohort results _once_.

analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
```diff
@@ -0,0 +1,342 @@
...
+      dplyr::select(-c("V1", "V2", "V3", "V4")) %>%
+      dplyr::select(summary_info, dplyr::everything()) %>%
+      tidyr::gather(V5, V6,-summary_info) %>%
+      dplyr::group_by(summary_info) %>%
+      dplyr::mutate(genes = paste0(sort(unique(V6)), collapse = ", ")) %>%
+      dplyr::select(-V5,-V6) %>%
+      dplyr::distinct() %>%
+      dplyr::mutate(summary = paste0(summary_info, genes)) %>%
+      dplyr::filter(!(summary == "NA, NA, NA, NA")) %>%
+      dplyr::arrange()
```
jaclyn-taroni: For the amplification and deletion files, GISTIC produces "ragged columns" per [their documentation](https://www.genepattern.org/modules/docs/GISTIC_2.0). It is often helpful to use lists any time you find yourself in a situation where columns or rows are of different sizes and you have a whole bunch of trailing blanks.

Here's what I came up with for the first function mentioned above, getting the vector of genes of interest for the Venn diagram:

```r
get_genes_vector <- function(genes_file) {
  genes_ragged_df <- data.table::fread(genes_file,
                                       data.table = FALSE)
  genes_list <- as.list(genes_ragged_df)
  genes_list <- genes_list %>%
    # This removes any element from the list that is all NA -- most likely a
    # result of reading in a ragged array
    purrr::discard(~ all(is.na(.))) %>%
    # This removes the element of the list that is essentially the "header" for
    # the file
    purrr::discard(~ any(str_detect(., "cytoband|q value"))) %>% 
    # Remove any broad peaks with q-value > 0.05
    purrr::discard(~ .[3] > 0.05) %>%
    # Remove everything before and including the wide peak boundaries for each
    # remaining element of the list
    purrr::modify(~ .[-c(1:(str_which(., "chr")))]) %>%
    # Remove blanks -- result of ragged data.frame
    purrr::modify(~ .[. != ""])
  # This will give us the vector of all the genes that were included
  genes_vector <- unique(unname(unlist(genes_list)))
}
```

analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
cbethell: Yes, that does make sense.

analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
cbethell: Noted, I'll give this function a try and commit the changes.

jaclyn-taroni:
This looks ready to be documented to me @cbethell! We might want to consider how we could restructure all of the copy number modules such they are all under the same directory. For example I'm going to create a new one called `expression-cna` to try out some ideas about #387 rather than add to `focal-cn-file-preparation`. The restructuring is way beyond this scope of this particular pull request, but is something that's on my mind.

cbethell:
> This looks ready to be documented to me @cbethell! We might want to consider how we could restructure all of the copy number modules such they are all under the same directory. For example I'm going to create a new one called `expression-cna` to try out some ideas about #387 rather than add to `focal-cn-file-preparation`. The restructuring is way beyond this scope of this particular pull request, but is something that's on my mind.

@jaclyn-taroni sounds good to me! I started a README for this module but will likely update that and the `analyses/README.md` file in a separate PR and mark this one ready for review. I will also give the restructuring of the copy number modules some thought. 

Would you also give the files in `run-gistic/results` another look (if you haven't already) to ensure that we covered all the possible file comparisons in this PR? I attempted to compare the `all_lesions.conf_90.txt` files as well but it appears that this would be redundant so I believe we covered all but a second opinion won't hurt.

jaclyn-taroni:
@cbethell I would say the results here are sufficient to give me pause about using the GISTIC results from the entire cohort without including `all_lesions.conf_90.txt`. The data in `all_lesions.conf_90.txt` also seems like it would be more difficult to represent.

analyses/gistic-cohort-vs-histology-comparison/gistic-cohort-vs-histology-comparison.Rmd
```diff
@@ -0,0 +1,348 @@
...
+```{r}
+# Source `generate-multipanel-plot-functions.R` script
+source(file.path("..",
+                 "..",
+                 "analyses",
+                 "transcriptomic-dimension-reduction",
+                 "util",
+                 "generate-multipanel-plot-functions.R"))
+```
+
```
jaclyn-taroni: Can you stick all your functions under this section?

PR #557, opened 2020-02-25
Lancet exploration notebook series: Part 1 TCGA-PBTA comparison variations

cansavvy:

This first notebook of three was an exploration into how different callers were resulting in different PBTA- TCGA comparisons. 

There are three questions and plots to accompany them : 
1) Is the read depth different between PBTA and TCGA
2) Do TMB comparisons results change if we calculate TMB with each caller by itself? 
3) How much do the TCGA and PBTA overlap in their target WXS regions? 

#### What was your approach?

Respectively: 
1) Plot the read densities for TCGA and PBTA
2) Recreate the TMB comparison CDF plot for each caller
3) Create Venndiagrams of the overlaps of the target regions

#548 

Two headlining notes before you get into the nitty gritty: 
1) This PR is huge. Let me know if you want me to split it up. 
2) Note that this analysis is gonna be retired almost as soon as we get it in (and its already outdated with the BED file changes that are still in progress #564 and #566 )

There are some parts of this analysis that could be DRY'ed up, but it also will be retired as soon as we get it confirmed. Let me know what parts of this code clearly need work and which we will leave as is.

Here's the rendered notebook: 
https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/explore-tcga-pbta.nb.html

The results of this notebook were what was added to this PDF report:

[TCGAvsPBTAconsensus.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4251116/TCGAvsPBTAconsensus.pdf)

#### Documentation Checklist
This analysis is not permanent and has not been added to the main READMEs


.circleci/config.yml
```diff
@@ -167,6 +167,9 @@ jobs:
           name: Telomerase activity
           command: ./scripts/run_in_ci.sh bash analyses/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh
 
+      - run:
+          name: PBTA vs TCGA explore 
+          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/snv-callers/lancet-wxs-test/explore-tcga-pbta.Rmd', clean = TRUE)"
```
cansavvy:
```suggestion
          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd', clean = TRUE)"
```

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+```
+
+### Special functions 
+
+Function for setting up PBTA data from the database. 
+This database was originally created by the [run consensus bash script](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/snv-callers/run_caller_consensus_analysis-pbta.sh)
+
+```{r}
+set_up_pbta <- function(data.name) {
+  # Given the name of dataset in the PBTA database, set up the TCGA data.
```
jashapiro: PBTA or TCGA here?  It seems like the function might be bets to be db agnostic, and just take a name and a database file; then it can collapse this and the later one, with just one argument added.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+    as.data.frame() %>%
+    # Tack on the metadata columns we want
+    dplyr::inner_join(
+      pbta_metadata %>%
+        dplyr::select(
+          Tumor_Sample_Barcode = Kids_First_Biospecimen_ID,
+          experimental_strategy, short_histology
+        )
+    ) 
+  return(df)
```
jashapiro: Don't forget to clean up!
```suggestion
  DBI::dbDisconnect(con)
  return(df)
```

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+                        name_1,
+                        name_2,
+                        plot_name) {
+  # Given two GenomicRanges objects make a VennDiagram of their overlap
+  
+  # Find intersection
+  overlaps <- GenomicRanges::intersect(bed_granges_1, bed_granges_2)
+
+  # Percent of TCGA Target region covered by overlap
+  sum(overlaps@ranges@width) / sum(bed_granges_2@ranges@width)
```
jashapiro: Do we need to collapse (reduce?) these before summing? I see that you do it later, but I  feel like the more bombproof way do to it is in this function.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+), .id = "dataset") %>%
+  dplyr::mutate(dataset = as.factor(dataset))
+```
+
+We'll plot the tumor sequencing depth for each caller as a series of density plots. 
+
+```{r}
+lancet %>%
+  ggplot2::ggplot(ggplot2::aes(x = log10(t_depth), color = dataset)) +
+  ggplot2::geom_density()
```
jashapiro: The bandwidth seems too small for these plots, and inconsistent, using `geom_density(bw = 1)` might be a good idea, though you will want to play with bandwidth to get a reasonable result.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+  cdf_plot <- tmb_df %>%
+    as.data.frame() %>%
+    dplyr::mutate(short_histology = tools::toTitleCase(short_histology)) %>%
+    # Only plot histologies groups with more than `min_samples` number of samples
+    dplyr::group_by(short_histology, add = TRUE) %>%
+    # Only keep groups with this amount of samples
+    dplyr::filter(dplyr::n() > 5) %>%
+    # Calculate histology group mean
+    dplyr::mutate(
+      hist_mean = mean(tmb),
```
jashapiro: median here too?

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+tmb_cdf_plot <- function(tmb_df, plot_title) {
+  # Given a data.frame of TMB data, plot it as a CDF plot and save it as a png. 
+  #
+  # Args:
+  #   tmb_df: a chromosomal breaks density file path where each sample is
+  #                 a row with `samples` and `breaks_count` columns. 
+  #   plot_title: to be used for the ggplot2::ggtitle and what it will be saved 
+  #                as a png as.
+  
+  cdf_plot <- tmb_df %>%
```
jashapiro: I wouldn't complain if this function was pulled out to a utility script and sourced as needed. It seems a good general use thing for people to fine easily.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+    bed_wxs = file.path(data_dir, "intersect_cds_lancet_WXS.bed")
+  ),
+  tcga = calculate_tmb(tcga_consensus,
+    bed_wgs = file.path(data_dir, "intersect_cds_lancet_strelka_mutect_WGS.bed"),
+    bed_wxs = file.path(ref_dir, "intersect_cds_gencode_liftover_WXS.bed")
+  )
+), .id = "datasets")
+
+# Make a CDF plot
+tmb_cdf_plot(consensus_tmb, plot_title = "Consensus")
```
jashapiro: This plot doesn't look so bad! Is this different from the main tcga-pbta comparison somehow?

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+tcga_bed <- readr::read_tsv("https://api.gdc.cancer.gov/data/7f0d3ab9-8bef-4e3b-928a-6090caae885b", col_names = c("chr", "start", "end")) %>%
+  dplyr::filter(!is.na(chr)) %>%
+  dplyr::mutate(chr = paste0("chr", chr)) %>%
+  readr::write_tsv(file.path(ref_dir, "gencode.v19.basic.exome.tsv"),
+    col_names = FALSE
+  )
+```
+
+After using [UCSC BED liftover](https://genome.ucsc.edu/cgi-bin/hgLiftOver) to convert 
+from the Target BED regions for TCGA data from hg19 to hg38 (What the PBTA data is using).
```
jashapiro: This will need updating when we produce this file directly (see #597)

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,591 @@
...
+Find overlap between the TCGA (non-liftover) and PBTA data. 
+
+```{r}
+bed_overlap(pbta_granges,
+  tcga_granges,
+  plot_name = "PBTA vs TCGA WXS target BED",
+  name_1 = "PBTA",
+  name_2 = "TCGA"
+)
+```
```
jashapiro: I don't think this is necessary. We don't expect unlifted data to work, so I would simply eliminate these comparisons completely.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: I kind of think we should have this archived as we looked at originally since this doesn't go toward any plots we will use.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: Yes. The BED files here were wrong #566 was not implemented yet and #568 is still at large. So the PBTA TMBs got larger because the BED files got smaller.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: I think this analysis is getting retired immediately since it was exploratory.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: I thought of doing this but the differences in metadata made it so I'd have to have weird/complicated arguments.

cansavvy: Unless you have a suggestion for getting around that.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: But when I make this plot into a function I can make it median instead.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: This is something I have on my TO-DO list but I think it would be better if it were in the module that has the "real" plot (tmb-compare-tcga) and not with a notebook that will be retired. I will file an issue in my own repo so we have this written down.

cansavvy: Tis documented : https://github.com/cansavvy/openpbta-notebook-concept/issues/6

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
cansavvy: It wasn't as weird as I thought it would be.

analyses/snv-callers/lancet-wxs-tests/explore-tcga-pbta.Rmd
```diff
@@ -0,0 +1,641 @@
...
+    dplyr::select(cols_to_keep) %>%
+    # Turn into data.frame
+    as.data.frame() 
+  
+  # This step is only needed for TCGA data
+  if (is_tcga) {
+    df <- df %>% 
+      # Shorten the Tumor_Sample_Barcode so it matches
+      dplyr::mutate(Tumor_Sample_Barcode = substr(Tumor_Sample_Barcode, 0, 12))
+  }
```
jashapiro: I think you can just do this always, and save an argument. Taking the substr of a 12 character barcode will just return the barcode.

PR #562, opened 2020-02-26
TCGA Consensus and Comparison Revised (1 of 2)

cansavvy:
If we run the TCGA data through the same variant callers and consensus methods we used for PBTA data, do we get a comparison that more accurately meets our expectations? (aka that adults have a higher tumor mutation burden). 

The next PR in this series will carry over the new dataset to the `tmb-compare-tcga` analysis.

#### What was your approach?
- Both TMB calculations for both PBTA and TCGA have been switched to only use Mutect2 and Strelka2. I had to re-run these on AWS. 
- I ran the TCGA files from v14 through the same pipeline we used for the PBTA data. You will note that there were a few metadata cleaning steps that needed to be adjusted for TCGA so there is a new option (`--tcga`) that needs to be used for the 03-calculate-tmb.R script.
- You may note that I removed the word "Consensus" from the results files. This is because these TMB results are no longer using the official consensus maf file. Is there another file notation that would make this more clear? 

#257 

- The BED files for PBTA do need to be fixed see #564 , but that will be in a different PR. 
- All mentions of WGS related files in the TCGA are placeholders because there is not actually any WGS data in TCGA. I haven't made adjustments to make it so WGS options can be dropped completely. Perhaps this is a thing for the next PR? 
- Is there anything where the TCGA data is not being handled properly by the snv-caller pipeline? 

### Results

This analysis needs to be run in AWS, and because of issues noted for the bed files for PBTA #564 I will end up re-running this in AWS. So do not be too concerned with the exact results, but more what we need to know at this stage is if the changes in the tmb calculations otherwise seem right. 

#### Documentation Checklist


analyses/README.md
```diff
@@ -39,7 +39,7 @@ Note that _nearly all_ modules use the harmonized clinical data file (`pbta-hist
...
+| [`snv-callers`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers) | `pbta-snv-lancet.vep.maf.gz` <br> `pbta-snv-mutect2.vep.maf.gz` <br> `pbta-snv-strelka2.vep.maf.gz` <br> `pbta-snv-vardict.vep.maf.gz` <br> `tcga-snv-lancet.vep.maf.gz` <br> `tcga-snv-mutect2.vep.maf.gz` <br> `tcga-snv-strelka2.vep.maf.gz` | Generates consensus SNV and indel calls for PBTA and TCGA data; calculates tumor mutation burden using the consensus calls | `results/consensus/pbta-snv-consensus-mutation.maf.tsv.gz` <br> `results/consensus/pbta-snv-consensus-mutation-tmb.tsv` <br> `results/consensus/pbta-snv-consensus-mutation-tmb-coding.tsv`(included in data download; too large for tracking via GitHub) <br> `results/consensus/tcga-snv-consensus-mutation.maf.tsv.gz` <br> `results/consensus/tcga-snv-consensus-mutation-tmb.tsv` <br> `results/consensus/tcga-snv-consensus-mutation-tmb-coding.tsv`
```
cansavvy: This section also needs to be edited to remove the "consensus" part of the file name for the tmb files if we agree with my notation change.

analyses/snv-callers/.gitignore
```diff
@@ -1,3 +1,4 @@
 # ignore folders with big results files
 results
-ref_files
+ref_files/*
+!ref_files/gencode.v19.basic.exome.hg38liftover.bed
```
cansavvy: For now, this needs to be added in somewhere, but I think it will be in a future data release.

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +140,109 @@ if (!dir.exists(opt$output)) {
   dir.create(opt$output, recursive = TRUE)
 }
 
-# Declare output file paths
-tmb_coding_file <- file.path(opt$output, "pbta-snv-consensus-mutation-tmb-coding.tsv")
-tmb_all_file <- file.path(opt$output, "pbta-snv-consensus-mutation-tmb-all.tsv")
+# Get data name
+data_name <- ifelse(opt$tcga, "tcga", "pbta")
+
+# Declare output file based on data_name
```
cansavvy: I took out the word consensus here. Let me know if that is better or if there is another file name that would be more clear.

analyses/snv-callers/README.md
```diff
@@ -30,15 +30,22 @@ To run the evaluations and comparisons of all the SNV callers, call the bash scr
 ```
 bash run_caller_analysis.sh
 ```
+
+For the TCGA data, it has its own script to run the same methods:
```
jashapiro:
```suggestion
The TCGA data is processed by its own script to run the same methods:
```

analyses/snv-callers/README.md
```diff
@@ -30,15 +30,22 @@ To run the evaluations and comparisons of all the SNV callers, call the bash scr
...
+```
+bash run_caller_analysis-tcga.sh
+```
+
 This bash script will return:
 
 - Comparison plots in a notebook: [`compare_snv_callers_plots.nb.html`](https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers_plots.nb.html).
 - A zip file containing:
-  - `pbta-snv-consensus-mutation.maf.tsv` - is  [MAF-like file](#consensus-mutation-call) that contains the snvs that were called by all three of these callers for a given sample are saved to this file.
+  - `pbta/tcga-snv-consensus-mutation.maf.tsv` - is  [MAF-like file](#consensus-mutation-call) that contains the snvs that were called by all three of these callers for a given sample are saved to this file.
```
jashapiro:
```suggestion
  - `[pbta|tcga]-snv-consensus-mutation.maf.tsv` - [MAF-like files](#consensus-mutation-call) that contain the snvs that were called by all three callers (mutect2, strelka2, and lancet) for a given sample.
```

jashapiro:
```suggestion
  - `[pbta|tcga]-snv-consensus-mutation.maf.tsv` - [MAF-like files](#consensus-mutation-call) that contain the snvs that were called by all callers described below for a given sample.
```

analyses/snv-callers/README.md
```diff
@@ -55,14 +62,15 @@ This is following the [code used in
...
+The consensus mutations called are those that are shared among all of Strelka2, Mutect2, and Lancet.
 Mutations were considered to be the same if they were identical in the following field: `Chromosome`, `Start_Position`, `Reference_Allele`,  `Allele`, and `Tumor_Sample_Barcode`.
 As Strelka2 does not call multinucleotide variants (MNV), but instead calls each component SNV as a separate mutation, MNV calls from Mutect2 and Lancet were separated into consecutive SNVs before comparison with Strelka2.
 
 ### Tumor Mutation Burden Calculation
 
 For each experimental strategy and TMB calculation, the intersection of the genomic regions effectively being surveyed are used.
-These genomic regions are used for first filtering mutations to these regions and then for using the size in bp of the genomic regions surveyed as the TMB denominator. 
+Only Strelka2 and Mutect2 agreement was used for TMB calculations. This is because of some [complications identified with Lancet's performance on WXS data](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers/lancet-wxs-tests).
```
jashapiro:
```suggestion
Only Strelka2 and Mutect2 agreement were used for TMB calculations. 
This is because of some [complications identified with Lancet's performance on WXS data](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers/lancet-wxs-tests).
```

analyses/snv-callers/README.md
```diff
@@ -55,14 +62,15 @@ This is following the [code used in
...
+These genomic regions are used for first filtering mutations to these regions and then for using the size in bp of the genomic regions surveyed as the TMB denominator.
```
jashapiro:
```suggestion
These genomic regions are used for first filtering mutations to these regions and then for calculating the size of the genomic regions surveyed as the TMB denominator.
```

analyses/snv-callers/README.md
```diff
@@ -82,12 +90,12 @@ This file is included in the data download.
 SNVs outside of these coding sequences are filtered out before being summed and used for TMB calculations like such:
```
jashapiro:
```suggestion
SNVs outside of these coding sequences are filtered out before being summed and used for TMB calculations as follows:
```

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
```diff
@@ -2,7 +2,7 @@
 # C. Savonen
 # CCDL for ALSF 2019
 
-# Purpose:Run an consensus analysis of SNV callers
+# Purpose:Run an consensus analysis for PBTA of SNV callers
```
jashapiro:
```suggestion
# Purpose: Run a consensus analysis for PBTA of SNV callers
```

jashapiro:
```suggestion
# Purpose: Run a consensus analysis for PBTA of SNV callers
```

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +141,109 @@ if (!dir.exists(opt$output)) {
...
+strelka_mutect_mnv <- strelka %>%
+  dplyr::inner_join(split_mutect_df,
+    by = join_cols,
+    copy = TRUE
+  ) %>%
+  as.data.frame()
+
+if (opt$tcga) {
+  strelka_mutect_mnv <- strelka_mutect_mnv %>%
+    dplyr::mutate(Tumor_Sample_Barcode = substr(Tumor_Sample_Barcode, 0, 12))
```
jashapiro: What is going on here? Can you add a comment?

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +141,109 @@ if (!dir.exists(opt$output)) {
...
+}
 
+# Split up MNVs
```
jashapiro: Is this comment correct? Aren't you merging in the mnvs with the snvs here?

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +141,109 @@ if (!dir.exists(opt$output)) {
...
+      experimental_strategy = "WXS", # This field doesn't exist for this data, but all is WXS
+      short_histology = Primary_diagnosis
+    ) # This field is named differently
 
-# Make sure that we have metadata for all these samples.
-if (!all(unique(maf_df$Tumor_Sample_Barcode) %in% metadata$Tumor_Sample_Barcode)) {
-  stop("There are samples in this MAF file that are not in the metadata.")
+  # Manifest files only have first 12 letters of the barcode so we gotta chop the end off
+  strelka_mutect_maf_df <- strelka_mutect_maf_df %>%
+    dplyr::mutate(Tumor_Sample_Barcode = substr(Tumor_Sample_Barcode, 0, 12))
```
jashapiro: Okay, you explain this here, but this is the second time you are doing it. Do you need it in the previous case, or is this one sufficient, as `strelka_mutect_mnv` is part of `strelka_mutect_maf_df`, if I understand line 201 correctly?

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +141,109 @@ if (!dir.exists(opt$output)) {
...
-    ))
+                      dplyr::select(
+                        "Tumor_Sample_Barcode",
+                        "experimental_strategy",
+                        "short_histology"
+                      ))
 
-############################# Calculate TMB ####################################
+# .x is messing up the maf_to_granges function
+colnames(strelka_mutect_maf_df) <- gsub("\\.x$", "", colnames(strelka_mutect_maf_df))
```
jashapiro: This should only be added if there are repeated column names that are not used for the join, right? I'm confused why they are showing up at all.

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +141,109 @@ if (!dir.exists(opt$output)) {
...
+      short_histology = Primary_diagnosis
+    ) # This field is named differently
 
-# Make sure that we have metadata for all these samples.
-if (!all(unique(maf_df$Tumor_Sample_Barcode) %in% metadata$Tumor_Sample_Barcode)) {
-  stop("There are samples in this MAF file that are not in the metadata.")
+  # Manifest files only have first 12 letters of the barcode so we gotta chop the end off
+  strelka_mutect_maf_df <- strelka_mutect_maf_df %>%
+    dplyr::mutate(Tumor_Sample_Barcode = substr(Tumor_Sample_Barcode, 0, 12))
+} else {
```
jashapiro:
```suggestion
} else { # pbta data 
```

analyses/snv-callers/scripts/03-calculate_tmb.R
cansavvy: In the TCGA MAF files, the Tumor_Sample_Barcode has the biospecimen information but only the first 12 characters are used for matching to the the metadata

analyses/snv-callers/scripts/03-calculate_tmb.R
cansavvy: This comment must have snuck in in the shuffle.

analyses/snv-callers/scripts/03-calculate_tmb.R
cansavvy: This is actually unnecessary. The manifest doesn't need this done, only the MAF file does. I double checked that and now am deleting this step 👍

analyses/snv-callers/scripts/03-calculate_tmb.R
cansavvy: IDK, This is something we had previously too: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b1b73fe321a97fa82d85c86d20bd85635aabba25/analyses/snv-callers/scripts/03-calculate_tmb.R#L278 It shows up after the union_all step.

cansavvy: I think we want union instead of union_all http://www.datasciencemadesimple.com/union-union_all-function-r-using-dplyr-union-data-frames/

cansavvy: I think I've made the changes we need. I switched to `union` and added in some `selects`  See if what I did seems reasonable/more right to you.

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -143,46 +151,119 @@ if (!dir.exists(opt$output)) {
...
+  dplyr::inner_join(mutect %>% 
+                      dplyr::select(join_cols), 
+                    by = join_cols, 
+                    copy = TRUE) %>%
+  as.data.frame()
+
+# Get Multi-nucleotide calls from mutect as SNVs
+split_mutect_df <- split_mnv(mutect) %>% 
+  dplyr::select(join_cols) %>% 
+  dplyr::ungroup() %>%
```
cansavvy: I'm a little concerned about this add. It kept wanting to add a "temp_id" column after doing split_mnv.

jashapiro: I just renovated the split_mnv function to ungroup there rather than here.
```suggestion
```

analyses/snv-callers/scripts/03-calculate_tmb.R
jashapiro: I don't think it should matter if it is `union` or `union_all`, as there should not be any shared sites. Pretty sure I used union_all for speed, but I'm sure it doesn't matter much. The by= is there so we keep all of the columns that might not have equal values, one of which is the send position, which is probably what is screwing up the maf_to_granges.

analyses/snv-callers/README.md
```diff
@@ -30,15 +30,22 @@ To run the evaluations and comparisons of all the SNV callers, call the bash scr
...
+```
+bash run_caller_analysis-tcga.sh
+```
+
 This bash script will return:
 
 - Comparison plots in a notebook: [`compare_snv_callers_plots.nb.html`](https://cansavvy.github.io/openpbta-notebook-concept/snv-callers/compare_snv_callers_plots.nb.html).
 - A zip file containing:
-  - `pbta-snv-consensus-mutation.maf.tsv` - is  [MAF-like file](#consensus-mutation-call) that contains the snvs that were called by all three of these callers for a given sample are saved to this file.
+- `[pbta|tcga]-snv-consensus-mutation.maf.tsv` - [MAF-like files](#consensus-mutation-call) that contain the snvs that were called by all callers described below for a given sample.
```
jashapiro:
```suggestion
- `(pbta|tcga)-snv-consensus-mutation.maf.tsv` - [MAF-like files](#consensus-mutation-call) that contain the snvs that were called by all callers described below for a given sample.
```

cansavvy:
> Those changes don't seem to be related to the rest of the changes in this PR, so perhaps those files should be reverted? Or will they be updated later?

Yeah, that will be in a next PR after I have the real data run from AWS. I didn't realize they were changed, I might have ran the data through as a check at some point. I'll revert remove the TCGA notebook for now and have a next PR with the "real" one.

jashapiro:
>  I'll revert remove the TCGA notebook for now and have a next PR with the "real" one.

While you are at it, probably worth removing the .pngs as well.

Issue #574, opened 2020-02-28
Fix v15 breaking changes 


jaclyn-taroni:
To quote the release notes being added in #569, we're changing the names of well-enough-used columns in the clinical file:

> - Change `disease_type_old` to `pathology_diagnosis` and `disease_type_new` to `integrated_diagnosis` per request in [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/543#issuecomment-589248029).

I know this change to `pbta-histologies.tsv` will break a number of things. The purpose of this issue is to track what will need to be changed as a result. Not only will the column names need to be updated, but we will also need to rerun any notebooks, change documentation, etc.

### Anticipated issues

Here I'll list what I know needs to change in modules that *are not deprecated*.

* Some of the modeling steps of `gene-set-enrichment-analysis` use `disease_type_new`: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L123 
Luckily the ` gsva_anova_tukey` function is already flexible! https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/gene-set-enrichment-analysis/util/hallmark_models.R#L35

* The first step of `interaction-plots` uses the `disease_type_new` column to generate lists of samples:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/interaction-plots/scripts/01-disease-specimen-lists.R#L97 
  Documentation associated with that option will also need to change.

* We filter out ATRT and MB samples in `molecular-subtyping-embryonal` using `disease_type_old` 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd#L128
and check `disease_type_new` as well:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd#L136
Also in this module, we use both the `disease_type` columns in the subtyping and generating final tables quite a bit starting around
[...]

PR #576, opened 2020-02-29
Find any unanticipated v15 breaking changes

jaclyn-taroni:

We know the changes to `pbta-histologies.tsv` is going to cause some modules to fail CI. Those modules are tracked in #574. This is a draft pull request where I've commented those out to see if we can identify any additional changes we'll need to make as part of the v15 release process.


jaclyn-taroni:
Okay this pull request has come to the end of its usefulness. I've included a new section of the issue about v15 breaking changes #574 and added new tickets #577, #578, #579.

PR #582, opened 2020-03-02
Update Interaction plots for v15

jashapiro:

Updating analysis for v15

#### What was your approach?

Replaced `disease-type-new` with `integrated-diagnosis` and reran on full data
Changes in subset used for CI were also required due to changes in the included subset.

#574

CI config needs to be reverted after passing, and will not pass at that point

#### Documentation Checklist


jashapiro:
Passed CI with only this module running.

analyses/interaction-plots/scripts/01-disease-specimen-lists.R
```diff
@@ -85,16 +85,15 @@ out_file <- file.path(root_dir, opts$outfile)
 
 #### Read files
 
-meta_df <- readr::read_tsv(meta_file, col_types = readr::cols())
+meta_df <- readr::read_tsv(meta_file, col_types = readr::cols(), guess_max = 10000)
```
cbethell: What is the reason behind adding the `guess_max` argument to the function here?

jashapiro: It prevents a warning when readr guesses the wrong data type for some columns (columns that are blank for the first 1000 samples, in this case)

jashapiro:
Merging over CI failure (for other modules)

PR #587, opened 2020-03-02
Does Shatterseek work with v15 changes? 

cansavvy:
I'm just running this on circleCI to see if Shatterseek needs changes. According to [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/574#issuecomment-593516925), it may actually be fine. 

I think there's a way to have done this locally, but this seemed easier for now.


cansavvy:
Turns out Shatterseek is fine! I'm gonna close this now.

PR #589, opened 2020-03-02
Deprecate Selection Strategy Comparison

jashapiro:

The analysis in `selection-strategy-comparison` is no longer needed, as the decision to split up the selection strategy calls are well established.

#### What was your approach?

Comment out analysis from CI, and mark the analysis as deprecated in the `analysis/README.md` table

#574


jashapiro:
Merging without test; no code additions, only commenting out of CI

Issue #578, opened 2020-02-29
Updated analysis: fusion-summary with v15 fusion files and downstream


jaclyn-taroni:
#### What analysis module should be updated and why?

`fusion-summary` creates files of fusions of interest from the `pbta-fusion-putative-oncogenic.tsv` file, which is being updated in v15 (#569). The `molecular-subtyping-embryonal` and `molecular-subtyping-EPN` are downstream of the `fusion-summary` module.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

The `01-fusion-summary.Rmd` notebook in`fusion-summary` needs to be updated. It is insufficient to rerun it because new processing criteria for the fusions was used and now it appears, at least in the case of the embryonal tumor fusions of interest, there are new instances of fusions that previously need to be filled in as all `0`:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/fusion-summary/01-fusion-summary.Rmd#L159

So these chunks

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/fusion-summary/01-fusion-summary.Rmd#L153

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/286ff25022930024bb9812e3cfad5410a2cf49c8/analyses/fusion-summary/01-fusion-summary.Rmd#L168

Need to be looked at carefully to see what fusions of interest need to be filled in with all `0`. Notably, the TSV results of module then needs to go on deck for v16 release, so let's plan to get a `planned data release` issue filed at the same time a PR for this update gets filed!

Then the downstream subtyping modules (`molecular-subtyping-embryonal` and `molecular-subtyping-EPN`) then need to be rerun and checked for any change in interpretation.

#### What input data should be used? Which data were used in the version being updated?

`pbta-fusion-putative-oncogenic.tsv`

#### When do you expect the revised analysis will be completed?

I think this should be completed around the same time as #574. ~1 day barring any unforeseen complications.

#### Who will complete the updated analysis?

Likely someone from the CCDL.

jaclyn-taroni:
We also have to keep an eye on the logic for CI in `01-fusion-summary.Rmd`. We may need to take an approach where we enrich for certain fusions in the subset files for CI or include the full `pbta-fusion-putative-oncogenic.tsv`.

[...]

PR #590, opened 2020-03-02
V15 Fix molecular-subtype-chordoma analysis

cansavvy:

There are not enough chordoma samples in the subset CircleCI files to test so I switched to the approach that was used in the other subtyping modules. 

Looks like the `rename` steps were not working because of library load and it was not calling the `dplyr::rename` but another `rename`. So to fix I added a series of `dplyr::`, reran the notebook with v15 and pushed the results. 

#574 

These items were all taken care of previously. 

#### Documentation Checklist


cansavvy:
@jaclyn-taroni I have two fix ideas for the lack of chordoma samples in subset for this notebook. 

1) Force the subset to include chordoma samples by adding another step in `create-subset-files` (This is the more painful route since we'd have to re-run the subsetting steps). 
2) Make this filter step in the notebook only occur when it is NOT being run in CircleCI:  
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0d9a85ea585f25bade2a62c5e54539270021d4dd/analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd#L147

At this point in time, we are thinking 2 would be okay, but I wanna clear it by you before I implement it.

jaclyn-taroni:
There is another option - have the first step of this module be it’s own subsetting where all resulting files are committed to the repository. That is what we do for other subtyping modules.

cansavvy:
> There is another option - have the first step of this module be it’s own subsetting where all resulting files are committed to the repository. That is what we do for other subtyping modules.

I was just noticing this for the other subtyping modules. I will try to follow their suit then.

analyses/molecular-subtyping-chordoma/00-subset-files-for-chordoma.R
```diff
@@ -0,0 +1,79 @@
...
+
+# Set path to results and plots directory
+results_dir <-
+  file.path(root_dir, "analyses", "molecular-subtyping-chordoma", "chordoma-subset")
+
+if (!dir.exists(results_dir)) {
+  dir.create(results_dir)
+}
+
+scratch_dir <- file.path(root_dir, "scratch")
```
jashapiro: Suggestion: create a subdirectory in `scratch` for this analysis to avoid collisions.

analyses/molecular-subtyping-chordoma/00-subset-files-for-chordoma.R
```diff
@@ -0,0 +1,79 @@
...
+select_metadata <- metadata %>%
+  dplyr::select(sample_id,
+                Kids_First_Participant_ID,
+                Kids_First_Biospecimen_ID)
+
+#### Filter metadata -----------------------------------------------------------
+chordoma_df <- metadata %>%
+  dplyr::filter(short_histology == "Chordoma",
+                sample_type == "Tumor",
+                composition == "Solid Tissue")
```
jashapiro: We don't need all of the metadata, only the chordoma, right?
```suggestion
#### Filter metadata -----------------------------------------------------------
chordoma_df <- metadata %>%
  dplyr::filter(short_histology == "Chordoma",
                sample_type == "Tumor",
                composition == "Solid Tissue")
                
# Select wanted columns in metadata for merging and assign to a new object
select_metadata <- chordoma_df %>%
  dplyr::select(sample_id,
                Kids_First_Participant_ID,
                Kids_First_Biospecimen_ID)
```

analyses/molecular-subtyping-chordoma/00-subset-files-for-chordoma.R
```diff
@@ -0,0 +1,79 @@
...
+cn_metadata <-  data.table::fread(file.path("data",
+                                            "consensus_seg_annotated_cn_autosomes.tsv.gz")) %>%
+  dplyr::left_join(select_metadata,
+                   by = c("biospecimen_id" = "Kids_First_Biospecimen_ID")) %>%
+  dplyr::select(gene_symbol,
+                sample_id,
+                Kids_First_Participant_ID,
+                biospecimen_id,
+                status) %>%
+  dplyr::filter(sample_id %in% chordoma_df$sample_id) %>%
```
jashapiro: If we use the `select_metadata` created after filtering to chordoma_df, this line is not needed
```suggestion
```

.circleci/config.yml
```diff
@@ -45,8 +45,8 @@ jobs:
           command:  OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-ATRT/run-molecular-subtyping-ATRT.sh
 
       - run:
-          name: Molecular subtyping Chordoma
-          command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd', clean = TRUE)"
+           name: Molecular subtyping Chordoma
+           command: OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-chordoma/run-molecular-subtyping-chordoma.sh
```
jashapiro: Spacing...
```suggestion
          name: Molecular subtyping Chordoma
          command: OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-chordoma/run-molecular-subtyping-chordoma.sh
```

analyses/molecular-subtyping-chordoma/README.md
```diff
@@ -1,10 +1,14 @@
...
+This module looks at _SMARCB1_ focal copy status and expression levels.
 
-This module consists of a single notebook that looks at _SMARCB1_ focal copy status and expression levels.
 It can be run via the command line with the following:
 
 ```
-Rscript -e "rmarkdown::render('01-Subtype-chordoma.Rmd', clean = TRUE)"
+bash run-molecular-subtyping-chordoma.sh
 ```
+This bash script the same regardless of where it is called and will first subset the data to `Chordoma` samples
```
cbethell: Looks like you're missing a verb here, correct?

analyses/molecular-subtyping-chordoma/00-subset-files-for-chordoma.R
cansavvy: This is leftover from ATRT, I don't use the scratch at all. Gonna delete this line.

analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd
```diff
@@ -18,32 +18,42 @@ library(ggplot2)
...
-focal_cn_df <- 
-  read_tsv(file.path("..", "focal-cn-file-preparation", "results",
-                     "consensus_seg_annotated_cn_autosomes.tsv.gz"))
+# Read in non-subsetted metadata
+histologies_df <- readr::read_tsv(file.path("..", "..", "data", "pbta-histologies.tsv"))
+
+# Subset metadata
+subset_metadata <- histologies_df %>%
+  dplyr::filter(short_histology == "Chordoma") %>%
+  select(
```
cbethell: You're calling the `filter` function above using `dplyr::` but not doing the same for `select` here. Is this because `filter` was behaving similarly to `rename` as you noted in your [original comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/590#issue-382545384)? If so, then disregard this comment. If not, removing the `dplyr::` altogether (except for in the case of the `rename` function) or adding the `dplyr::` to the rest of the functions would be nice for consistency purposes.

jashapiro:
Most fixes are done, but `fusion-summary` is not failing because there are no ependymoma fusions in the data set. The notebook currently skips non-ATRT embryonal samples (https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/94fbb85cdaff3bcc4fff0828958f60f6322726bc/analyses/fusion-summary/01-fusion-summary.Rmd#L154), so we could add the same switch to the ependymoma samples: (https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/94fbb85cdaff3bcc4fff0828958f60f6322726bc/analyses/fusion-summary/01-fusion-summary.Rmd#L169)

I have verified that this runs fine with the full dataset, but this would leave no testing of the output functions.

Thoughts @jaclyn-taroni ?

jaclyn-taroni:
`fusion-summary` is commented out in https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/a6e2923f04b846ea310c55e58780aec0ee97e0ad https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/569 because it's going to require a bigger change (because the fusion files are part of what is changing in v15) which is tracked with #578. I intended for #578 to wait until after v15 so we don't hold things up.

jashapiro:
Meant to leave that comment in the main thread...

analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd
```diff
@@ -141,16 +133,15 @@ Look at SMARCB1 expression values only in chordoma
...
-smarcb1_expression <- smarcb1_expression[, which(colnames(expression_data) %in% chordoma_samples) ]
-smarcb1_expression
+smarcb1_expression <- subset_expression_data[which(rownames(subset_expression_data) == "SMARCB1"), ]
 ```
 
 ```{r}
-# remove large expression matrix that's no longer needed
-rm(expression_data)
+# now only the columns correspond to chordoma samples
+smarcb1_expression <- smarcb1_expression[, which(colnames(subset_expression_data) %in% subset_metadata$Kids_First_Biospecimen_ID) ]
```
cbethell: This line looks weird (format-wise), did you run this through a styler?

cansavvy: I did but it doesn't do anything with really long lines. I'd rather use dplyr::filter here anyway but was trying to keep the original code here.

jashapiro:
Merging before expected CI failure.

analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd
cansavvy: `filter` also has a base:: function that can mess with stuff sometimes.

Issue #571, opened 2020-02-28
[Transition plan] Publication-ready figures that don't have publication-ready content


jaclyn-taroni:
The CCDL data science team (@cbethell @cansavvy @jashapiro and @jaclyn-taroni) has other commitments that require us to wind down the phase of the project where we are _actively analyzing data_ and move to maintenance and review. We plan to use our next sprint (March 2-13) to make this transition.

Specifically, our goal is to get publication-ready figures added to the manuscript repository  that may not have the publication-ready content (e.g., the underlying data or gene lists are subject to change) and to do so in a manner where it is easy to update figures as the underlying data changes.

To this end, we will:

1. Write code to produce publication-ready multi-panel plots using a unified color palette and plot theme (#510)
2. Add a shell script that is designed to run _all steps_ required to get from the data download to the publication-ready figures either to the root of the repository or to `scripts`
3. Add a markdown file that documents the process to generate the figures including requirements (e.g., recommended RAM)
4. File a series of pull requests to the manuscript repository ([`AlexsLemonade/OpenPBTA-manuscript`](https://github.com/AlexsLemonade/OpenPBTA-manuscript/)) that use links to the figure PNGs in the `master` branch of this repository so that any changes committed back to this repo are automatic propagated to the manuscript repository.

### Figures we will prepare

* Sample distribution (from [`sample-distribution-analysis`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/sample-distribution-analysis)) and GitHub contributions (#572); see: https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/9b65cd00cef2b59bad7059f0f96b483b7354b8e4/content/images/figure_sketches/fig1-proposed-sketch.png
* Tumor mutation burden (from [`snv-callers`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers)) and mutational signatures (bubble plot from [`mutational-signatures`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/mutational-signatures)): https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/9b65cd00cef2b59bad7059f0f96b483b7354b8e4/content/images/figure_sketches/fig2-proposed-sketch.png
[...]

Issue #510, opened 2020-02-04
Updated analysis: Unified color palette for plots


cansavvy:
#### What analysis module should be updated and why?
All modules that have plots with colors. 

We should probably prioritize plots that will be in the main document? But we probably want the unified color palette to also extend to non-main figures. 

#### What changes need to be made? Please provide enough detail for another participant to make the update.

We should have a unified color palette. This helps interpretability and aesthetics. 
`simplecolors` R package has some helpful tools and nice vignette: 
https://cran.r-project.org/web/packages/simplecolors/vignettes/intro.html

For `ggplot2` plots, colors can be designated using [scale_fill_manual and scale_color manual](https://ggplot2.tidyverse.org/reference/scale_manual.html). 

**Which colors do we generally want to default to?**
- For histology: There are many places where data are colored by `short_histology`, so having the colors for each group in particular would help readers follow along better. Can use an numeric approach to try to get ~36 colors as different as possible. I started implementing this 
- Could use [`colorblindr`'s palette](https://github.com/clauswilke/colorblindr) selection for guidance on some variable color choices. 
Here's an example of what I mean, but I haven't yet tested these colors: 
```
# Get values that can be used to make colors equi distant hues away for the 
# number of histology groups we have
col_val <- seq(from = 0, to = 1, 
               length.out = length(unique(df$short_histology)))

# Translate into colors
col_key <- hsv(h = col_val, s = col_val, v = 1)

# Make this named based on histology
names(col_key) <- unique(df$short_histology)

# Make the same order as the data.frame
col_key <-  as.character(dplyr::recode(df$short_histology, !!!col_key))

# Make the names 
names(col_key) <- as.character(rownames(df))
```

- For heatmaps or other continuous numeric variable data, we should choose a general color palette. Some instances want color functions, so we can use `colorRamp` for these instances, but we should decide what hex codes/colors should be used (I'm not suggesting necessarily the ones I have below). 
```
col_fun <- circlize::colorRamp2(
  c(0, .25, .5, 1, 3),
  c("#edf8fb", "#b2e2e2", "#66c2a4", "#2ca25f", "#006d2c")
)

```
### Modules with plots that will need to be color palette unified: 
I've tagged myself on the modules I will be responsible for updating the palette for, others can add themselves for other modules. 

| Module | Person who will update the plots | Plots in this module to be updated? |
|--------|----------------------------------|----------------------|
[...]

Issue #231, opened 2019-11-05
Include or establish a 'blacklist' of likely false positive genes?


jaclyn-taroni:
Some genes might show up as highly mutated, but just happen to be very large (e.g., _TTN_) and are unlikely to be informative. Does a list of genes of this nature already exist out there somewhere in the literature?

If so, including that list somewhere could 1) help contributors with less specific domain expertise filter and interpret their results and 2) allow analysts to use the same list for consistency.

jharenza:
There is [this paper](https://www.ncbi.nlm.nih.gov/pubmed/25466818) and [correction](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5706417/) describing FLAGS - Table S4 in the supplement may be a good one to use, and MAFTOOLS actually spits out a warning about these when in your dataset to make users aware. The caution to omitting FLAGS is that some of these could be real (Eg: I saw ROS1, proto-oncogene in this list and while I am not sure if it is flagged because a certain site is a FLAG, other mutated sites/expression/fusions could be important). Maybe I would suggest using this list and if any of the genes are found in the oncogene/TSG list from @kgaonkar6's fusion analysis, then we don't remove. I had instead, in the past, selected specific genes of interest to use (also not ideal because of bias and would miss novelty, but was needed for landscape type work to validate the tumor genomics).

jaclyn-taroni:
Context for this comment is https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/595#issuecomment-594773187.

The FLAGS list can be obtained via this link: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5706417/bin/12920_2017_309_MOESM3_ESM.txt

Here is the (abbreviated) list used in maftools: https://github.com/PoisonAlien/maftools/blob/acef26fa99e2619c4051d7792025e7a16c13b43b/R/summarizeMaf.R#L38

PR #595, opened 2020-03-03
Create Gene by Disease bar plot

jashapiro:

In addition to cooccurrence data, it would be useful to see a plot that showed how often some of the more frequently mutated genes were mutated in each disease type. This analysis adds a version of that plot. 

#### What was your approach?

As I intend to incorporate this plot into the cooccurrence plot for all samples, I made this plot in the same plotting script that plots the cooccurrences, but the plots are not joined yet; that will come in a separate PR.

None yet, but it is on the way to #571

`plots/gene_disease_top50.png`

This analysis should be updated with #510 

Yes; comments on the form of the `gene_disease_top50.png` figure are welcomed!

Figures and tables of disease/gene mutation frequency

#### Documentation Checklist


jaclyn-taroni:
Documenting a discussion we had that is outside the scope of this pull request. I happened to notice _TTN_ showing up in these plots which prompted a discussion about #231. Because [`cooccurrence` accepts a list of genes](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b5a33838d1e9bd7e7913a89201ec26125c16c94c/analyses/interaction-plots/scripts/cooccur_functions.R#L68) ([see also](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/9ee9f948ec386df9bf14e4e02aa2d798a01b3e4d/analyses/interaction-plots/scripts/02-process_mutations.R#L305)), it will be straightforward to implement a step where we filter based on FLAGS. I'm going to add more information to #231.

analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -133,7 +133,7 @@ option_list <- list(
   make_option(
     opt_str = "--vaf",
     type = "numeric",
-    default = 0.2,
+    default = 0.05,
```
jaclyn-taroni: Is there a particular reason you are changing this default value?

analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -276,27 +277,53 @@ maf_filtered <- maf_df %>%
...
+}
+# count mutated samples by disease types
+gene_disease_counts <- gene_sample_counts %>%
+  dplyr::filter(gene %in% top_count_genes) %>%
+  dplyr::left_join(sample_meta, 
+                   by = c("sample" = "Kids_First_Biospecimen_ID")) %>%
+  dplyr::group_by(gene, disease = integrated_diagnosis) %>%
+  dplyr::summarize(mutant_samples = dplyr::n(),
+                   total_muts = sum(mutations),
+                   muts_per_sample = mean(mutations)) %>%
```
jaclyn-taroni: I have not read far enough to know what this is used for, but making this ever so slightly more descriptive could be useful
```suggestion
                   avg_muts_per_sample = mean(mutations)) %>%
```

analyses/interaction-plots/scripts/03-plot_interactions.R
```diff
@@ -74,12 +90,22 @@ cooccur_df <-
 
 labels <- unique(c(cooccur_df$label1, cooccur_df$label2))
 
+# check the order of the labels to be decreasing by mut count
```
jaclyn-taroni: The string manipulation through here is because you have `<gene symbol> (<count>)` correct?

analyses/interaction-plots/scripts/03-plot_interactions.R
```diff
@@ -132,3 +158,56 @@ cooccur_plot <- ggplot(
...
+disease_plot <- ggplot(
+  disease_df,
+  aes(x = gene, y = mutant_samples, fill = disease_factor)) + 
+  geom_col() +
+  labs(
+    x = "",
+    y = "Samples with mutations",
+    fill = "Diagnosis"
+  ) + 
+  scale_fill_manual(values = okabe_ito) +
```
jaclyn-taroni: I think you can do:
```suggestion
  colorblindr::scale_fill_OkabeIto() + 
```
and remove lines 187-188

analyses/interaction-plots/scripts/03-plot_interactions.R
jashapiro: That is correct. It is a bit of a hack, if I'm being honest, but it prevented having to rewrite previous code.

analyses/interaction-plots/scripts/03-plot_interactions.R
jashapiro: Indeed I can. I just today saw that `colorblindr` is already installed; I was avoiding it to prevent having to `install::github()`, but that is moot!

analyses/interaction-plots/scripts/02-process_mutations.R
jashapiro: I think it isn't used now, but it is available for use later.

analyses/interaction-plots/scripts/02-process_mutations.R
jashapiro: Not really, I guess I just thought it was too strict before.

Issue #560, opened 2020-02-26
Proposed Analysis: GISTIC vs. focal-cn-file-preparation comparison


cbethell:
#### What are the scientific goals of the analysis?

The scientific goal of the analysis is to gather evidence needed to validate whether or not the way we handle the copy number data in this project is reasonable and the best possible way in which we could handle it. 

To do this, we compare our calls to GISTIC's calls. More specifically, we want to compare the GISTIC gene level status calls (and cytoband status calls later) with the focal CN calls we prepare in the `focal-cn-file-preparation` module. We know that GISTIC is a tool that is widely used, and although it relies on recurrence and may not be appropriate for every histology we have in our cohort, we want to see if our `focal-cn-file-preparation` analysis gives us the same answer that GISTIC would. Once, we've collected this evidence, we will get an expert's opinion on the interpretation of the evidence.

#### What methods do you plan to use to accomplish the scientific goals?

I plan to execute this analysis in multiple steps/notebooks:
- 1. The first notebook will tidy the GISTIC data files (`all_lesions.conf_90.txt`, `amp_genes.conf_90.txt`, and `del_genes.conf_90.txt`). This is done to format the GISTIC data in a way that is comparable with the gene level calls in the `focal-cn-file-preparation/results` files. The output will therefore be a table that can be consumed in the second notebook for comparison to our focal CN files.
This notebook will also produce a separate table with GISTIC's cytoband data for comparison to our cytoband status calls once they are generated (#497). 
The table will have the following columns:

      | cytoband | Kids_First_Biospecimen_ID | status |
      |----------|---------------------------|--------|

- 2. The second notebook will take the tidy GISTIC data and count the number of samples for an individual histology (the focus will be LGAT for now) that are found in a particular amplification/deletion peak, for the corresponding genes (found in the same amplification/deletion peak according to the data from the `amp_genes.conf_90.txt` and`del_genes.conf_90.txt` files).
The output of the notebook will look similar to this sketch outlined by @jaclyn-taroni:
![gistic-comparison-sketch](https://user-images.githubusercontent.com/43576623/75369745-01fc4180-5892-11ea-9a45-39d88d10552c.jpg)

These steps were broadly attempted in open PR #559. 
This PR will now be adapted to implement step 1 of the plan above and a separate PR will be filed to address step 2.

#### What input data are required for this analysis?

The input data required for this includes:
[...]

PR #559, opened 2020-02-25
Comparison of GISTIC gene level calls - Data Prep

cbethell:
### Purpose/implementation Section

The purpose of this PR is to prepare GISTIC's results file for comparison of GISTIC's gene level calls with the calls we have prepared in `analyses/focal-cn-file-preparation/results`.

#### What scientific question is your analysis addressing?

This is addressing the question of whether or not our calls agree with GISTIC's calls, to further determine if our handling of the copy number data is reasonable in this project.

#### What was your approach?

My approach was to take the `all_lesions.conf_90.txt`, `amp_genes.conf_90.txt`, and `del_genes.conf_90.txt` GISTIC result files for an individual histology (LGAT) and prepare them for consumption in the upcoming second notebook of this analysis and described in the original comment of the reference issue #560:

> > The first notebook will tidy the GISTIC data files (`all_lesions.conf_90.txt`, `amp_genes.conf_90.txt`, and `del_genes.conf_90.txt`). This is done to format the GISTIC data in a way that is comparable with the gene level calls in the focal-cn-file-preparation/results files. The output will therefore be a table that can be consumed in the second notebook for comparison to our focal CN files.
> 
> Either a sketch or example markdown table work

@jaclyn-taroni does the following update to the original comment now seem suffice?
>
> * 1. The first notebook will tidy the GISTIC data files (`all_lesions.conf_90.txt`, `amp_genes.conf_90.txt`, and `del_genes.conf_90.txt`). This is done to format the GISTIC data in a way that is comparable with the gene level calls in the `focal-cn-file-preparation/results` files. The output will therefore be a table that can be consumed in the second notebook for comparison to our focal CN files. The table will tentatively have the following columns:
>      
>      
>      
>      | gene_symbol | Kids_First_Biospecimen_ID | status | detection_peak |
>      | ------------- | -------------------------- | -------| ---------------- |
>      |  _TBXAS1_  | BS_xxxxxx | gain | Amplification Peak 3 |
>      
>      
>      
>      where `gene_symbol` values are retrieved from the
>      `amp_genes.conf_90.txt`/`del_genes.conf_90.txt` files, `Kids_First_Biospecimen_ID` and
>      `status` values are retrieved from the `all_lesions.conf_90.txt` file, and `detection_peak`
>      values are matched between the `amp`/`del` files and the `all_lesions` file.
>   

#### What GitHub issue does your pull request address?

This PR addresses issue #560.

#### Which areas should receive a particularly close look?

[...]

PR #600, opened 2020-03-04
Comparison of GISTIC gene level calls - Data Wrangling/Tables

cbethell:

The purpose of this PR is to compare GISTIC's gene level calls with the calls we have prepared in `analyses/focal-cn-file-preparation/results`.

This is addressing the question of whether or not our calls agree with GISTIC's calls, to further determine if our handling of the copy number data is reasonable in this project.

#### What was your approach?

My approach was to take the files prepared in PR #559 and tally the focal/GISTIC calls per gene for each call category (loss, gain, amplification, neutral), noting that we filter the neutral calls out of our focal CN files and the GISTIC files do not contain amplification calls.

This PR addresses issue #560.

- The `tally_calls` function should receive a close look and can likely use some refactoring.
- I also displayed the tables in the notebook in order of highest loss count. Should this be displayed differently?
- Should this analysis now go beyond using just the data for the LGAT histology and entire cohort?

Yes, the code and tables are ready for review.

Tables of results are included in the results directory as follows:
- cohort_gene_level_focal_cn_counts.tsv
- cohort_gene_level_gistic_cn_counts.tsv
- lgat_gene_level_focal_cn_counts.tsv
- lgat_gene_level_gistic_cn_counts.tsv

where the files ending in `_focal_cn_counts.tsv` are counts of our focal CN calls with GISTIC's detection peak information and gene_symbol, and file ending in `_gistic_cn_counts.tsv` are counts of GISTIC's calls with wide peak information and gene symbol.

The [rendered notebook](https://cbethell.github.io/open-pbta-output/03-GISTIC-gene-level-cn-status-comparison.nb.html) displays the data saved in these files.

When looking at the entire cohort's and LGAT's GISTIC calls compared to our calls, our calls appear to mostly agree except in the amplification/neutral fields (where we have amplification calls only in the focal data and neutral calls only in the GISTIC). 
The GISTIC counts files are also larger in size than the focal counts files due to the presence more genes with calls in the GISTIC files (the set of genes in the focal results tables is a subset of the genes in the GISTIC results tables).

#### Documentation Checklist


jaclyn-taroni:
I am noticing that the highest a tally goes is 4 and we have over 200 LGAT samples. Based on what is in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/560#issue-571553936, I would expect the gain and loss columns to be counting the number of times that status occurs in LGAT.

cbethell:
@jaclyn-taroni RE your comment below:
> I am noticing that the highest a tally goes is 4 and we have over 200 LGAT samples. Based on what is in [#560 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/560#issue-571553936), I would expect the gain and loss columns to be counting the number of times that status occurs in LGAT.

My first impression was that this must be fine because we're looking at it per gene, but your comment made me take another look at the data and now I think that my first impression was not quite right. I'll take another look at the code to see where things went wrong.

jaclyn-taroni:
Yea, taking a quick look at the `all_lesions.conf_90.txt` file for LGAT only - I would expect all genes in  `Amplification Peak 1` to have a count of 13 gains.

analyses/compare-gistic/results/lgat_gene_level_focal_cn_counts.tsv
```diff
@@ -0,0 +1,119 @@
+gene_symbol	detection_peak	loss	gain	amplification
+ACP6	Deletion Peak 1	243	243	247
```
jaclyn-taroni: I think we have 242 LGAT samples, so this is missing some kind of `distinct` step I bet.

analyses/compare-gistic/results/lgat_gene_level_gistic_cn_counts.tsv
```diff
@@ -0,0 +1,4581 @@
+gene_symbol	detection_peak	loss	gain	neutral
+A3GALT2	Deletion Peak 1	8	8	476
```
jaclyn-taroni: For each of these rows, there are the same number of gains and losses.

cbethell:
The updated rendered notebook, where I implemented the `distinct` function where needed and removed the step that was removing the sample IDs before we tally them can be viewed [here](https://cbethell.github.io/open-pbta-output/03-GISTIC-gene-level-cn-status-comparison.nb.html).

The numbers appear to make more sense (correspond more with what's in the `all_lesions.conf_90.txt` files), but there are still rows in which the number of gains and losses are the same as in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/600#discussion_r387986299).

Do you have any further thoughts on this @jaclyn-taroni?

cbethell:
I realized that I should have probably used `inner_join` rather than `left_join` in the step merging calls. The updated rendered notebook reflecting this change can be seen [here](https://cbethell.github.io/open-pbta-output/03-GISTIC-gene-level-cn-status-comparison.nb.html#run_tally_calls_function_on_entire_cohort_gistic_calls). GISTIC's counts appear to make more sense now.

cbethell:
In [commit 2c87e7c](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/600/commits/2c87e7cbfefb04ff5c1cf8054afd4390d611b11c), I propagated the changes made to this module in WIP PR #614 (the changes made are further noted in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/614#issuecomment-595954412)).

Here's the [rendered `03-GISTIC-gene-level-cn-status-comparison.nb.Rmd` notebook](https://cbethell.github.io/open-pbta-output/03-GISTIC-gene-level-cn-status-comparison.nb.html), reflecting the results of these changes.

jaclyn-taroni:
I've made changes here such that preparing the gene-level GISTIC status separates two concepts from one another (and makes two data frames):

1. The gene to peak mapping or assignment, which also contains information like q-value, wide peak boundaries, etc.
2. The gene status in an individual sample (e.g., gain, loss, neutral)

The reason we were double and triple counting some genes is that peaks can be overlapping. In the LGAT case at least, we are able avoid double and triple counting by filtering based on `residual q value`, which is (from GISTIC docs):

>  The q-value of the peak region after removing (“peeling off”)
amplifications or deletions that overlap other more significant peak regions in the
same chromosome.

I then join these two together in the `03` notebook which now tallies each status at the gene-level and doesn't do much else. I left some ideas about some plots we could consider making in here though it is not used yet (I thought it would be easier to comment out and come back to it).

In general, it seems there are less gene-level non-neutral calls in the GISTIC data. My intuition is that this is expected, but I'd have to dig into _why_ a bit more. I figured that this is a decent size as it is and digging in more could wait. 

@cansavvy can you take another look?

jaclyn-taroni:
@cansavvy I forgot to mention some of the ideas that you reviewed earlier on the now-closed draft pull request #614 are over here. You have not yet review _this pull request_.

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  # robust
+  if ((method == "exon") & any(c(is.null(biospecimens_included), 
+                                 is.null(genes_included)))) {
+    stop("The 'exon' method is intended to be used with the 
+         biospecimens_included and genes_included arguments")
+  }
+  
+  # If a vector of biospecimen IDs is provided, filter only to those biospecimen
+  # ids
+  if (!is.null(biospecimens_included)) {
```
cansavvy: Is there an instance where one would use `biospecimens_included` but would *not* be using `method = "exon"`? The documentation makes me think no. If so, it might be better to separate these error handlers by `method`. It would make it easier to follow at least. However if there is an intention to allow biospecimen filtering for `method = "gistic"` then I would change the documentation to make that more clear and that it is just required for `method = "exon"`.

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+    # method that we used.
+    filter(str_detect(gene, "\\[", negate = TRUE)) %>%
+    # Count the number of times a copy number alteration occurs in an individual
+    # gene
+    group_by(gene, status) %>%
+    tally() %>%
+    spread(status, n) %>%
+    # Any instances where a copy number alteration is NA are equivalent to a 
+    # count of zero
+    replace_na(list(loss = 0,
```
cansavvy: It is unclear to me what this function is doing with `amplification` statuses?

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+    # gene
+    group_by(gene, status) %>%
+    tally() %>%
+    spread(status, n) %>%
+    # Any instances where a copy number alteration is NA are equivalent to a 
+    # count of zero
+    replace_na(list(loss = 0, 
+                    gain = 0))
+  
+  # In the GISTIC case, missing values in the neutral column are the same as
```
cansavvy: If `na_replace` steps are something that is always different between `gistic` and `exon` methods it may be easier to follow and more intentional if this is made into a parallel if statement like lines 118 to 123.

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  # zero counts
+  if (method == "gistic") {
+    gene_tally_df <- gene_tally_df %>%
+      replace_na(list(neutral = 0))
+  }
+  
+  # Because neutral calls are not included in the annotated consensus file,
+  # we can assume the number of neutral calls is the number of samples included 
+  # minus any alterations (gain + loss)
+  if (method == "exon") {
```
cansavvy: Ah. Maybe just move the `replace_na` step from line 115 to being within this if statement.

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  
+  gene_tally_df %>%
+    # Add detection peak information to this data.frame. Peaks can be 
+    # overlapping in GISTIC and contain the same gene symbols, so we need to 
+    # summarize the detection_peak column
+    inner_join(gene_mapping_df %>%
+                 select(gene,
+                        detection_peak),
+               by = "gene") %>%
+    group_by(gene, gain, loss, neutral) %>%
```
cansavvy: Still unclear what is happening to `amplification`? Are we intentionally ignoring it?

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
```diff
@@ -0,0 +1,382 @@
...
+  lgat_exon_tally <- get_gene_tally(gene_status_df = consensus_focal_cn_calls,
+                                    gene_mapping_df = lgat_gene_mapping,
+                                    method = "exon",
+                                    biospecimens_included = lgat_ids_in_consensus,
+                                    genes_included = lgat_gistic_tally$gene) %>%
+    write_tsv(lgat_exon_file)
+}
+```
+
+<!--
```
cansavvy: Is this going to be removed as well?

analyses/compare-gistic/util/GISTIC-comparison-functions.R
```diff
@@ -215,62 +202,69 @@ plot_genes_venn_diagram_wrapper <- function(cohort_genes_file,
...
-  # the IDs of the samples in the corresponding peaks along with
-  # with their CN status call.
+  # read in and tidy this data into a two data.frames that are written to file.
   #
   # Args:
   #   all_lesions_file: file path to GISTIC's `all_lesion.conf_90.txt` file
   #   amp_genes_file: file path to GISTIC's `amp_genes.conf_90.txt` file
   #   del_genes_file: file path to GISTIC's `del_genes.conf_90.txt` file
-  #   output_filename: string to save the output file as
+  #   gene_mapping_filepath: TSV file path for the first data.frame that
```
cansavvy:
```suggestion
  #   gene_mapping_filepath: file path for output TSV which has these columns: 
```
If this isn't what you mean correct me.

cansavvy: My wording isn't great, but I just want to clarify that these are files that don't currently exist but will be made by this function. ?

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
jaclyn-taroni: Good point! I think in development, because we are filtering the `focal-cn-file-preparation` files to genes in the GISTIC files, I didn't see any instances of `amplification` but I believe the right way to deal with this is to recode all amplifications as gains in the `focal-cn-file-preparation` specifically when we are tallying things for comparison.

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
cansavvy: Yeah, I saw that, but was wondering if that was something we could do now?

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
cansavvy: Ah. If the step is the same for both `gistic` and `exon` then that's fine. When I was running through some of these steps it was unclear to me whether that step needed to be applied to both or just that it didn't hurt anything to be applied to both. But I suppose in either situation this is fine.

jaclyn-taroni:
> There are quite a few commented-out sections in the 03 notebook. What is the plan with those?

The plan is to leave them in because then if we come back to it, it's easier to find what I was thinking at the time.

cansavvy:
> The plan is to leave them in because then if we come back to it, it's easier to find what I was thinking at the time.

Can we add a comment or two to explain then why they are commented out but not added to the code? Or what explicitly you are thinking in regards to those commented out code chunks?

analyses/compare-gistic/03-GISTIC-gene-level-tally.Rmd
jaclyn-taroni: Good catch – it didn't matter for the LGAT data that I used in development, but it _did_ matter for the entire cohort.

jaclyn-taroni:
Ready for another look @cansavvy! 👀  Regarding
> It would probably be good if someone that's not me reviews them if that hasn't already happened.

I looked at those changes on Sunday when I made the first set of changes to this PR.

PR #602, opened 2020-03-04
CN Status Heatmap (PR 1 of 2)

cansavvy:
#### What scientific question is your analysis addressing?
We wanted a summary visualization of copy number status.

This first PR only has the main data handling functions and docs updates, the next PR will show the results and the functions in action. 

#### What was your approach?
Create a summary heatmap of copy number status from the consensus CNV call data. 
This is done by binning the genome and calculating the segment's coverage of the
CNV consensus segments. 
A bin is declared a particular copy number status if that status's base pair 
coverage is a certain threshold percentage larger than the other statuses' 
coverage. 

#### What GitHub issue does your pull request address?
Issue: #594 

#### Which areas should receive a particularly close look?
The context of the next PR may mean this PR's code will be tweaked later, that's okay. Mainly I'm looking for if this looks reasonable to begin with. 

- Can you take a close look at the chr coordinate manipulations and calculations and try to help me make sure that these data are what I think they are? 
- Should a bin's copy number status be called only comparing gain and loss or should neutral be compared too? (Currently I have only gain and loss being used to make the call, but I have in previous commits, used neutral segments). 
- Are there comments or issues on how these functions handle the data? 
- Suggestions for efficiency? 

### Results
Will be added in the next PR. 

#### Reproducibility Checklist
- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist
**These items already existed but have been updated with the added notebook. 
- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

Issue #594, opened 2020-03-03
Proposed Analysis: CN status heatmap


cansavvy:
#### What are the scientific goals of the analysis?
Create a visualization that can show the CN copy number status across samples.
This would be much like the heatmaps in [chromosomal-instability](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd) except this heatmap would be samples x copy number status across the genome (three colors - amp, neutral, del). 

#### What methods do you plan to use to accomplish the scientific goals?
I plan to mainly use the code I developed in [chromosomal-instability](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd) but adapt it for copy number status.

#### What input data are required for this analysis?
`pbta-cnv-consensus.seg` from the data release. 

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
A day?

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@cansavvy 

#### What relevant scientific literature relates to this analysis?
I don't know of any.

cansavvy:
What I need to figure out more specifically is what the unit of columns for this heatmap will be. Do we want: 
- each pixel is a cytoband? 
- each pixel is a gene? 
- each pixel is a particular bin of width of the genome (which requires me to make a summary call for each bin) - this is probably the least easily interpretable idea.

I think cytoband makes the most sense, so I'll go with that unless someone weighs in differently.

cansavvy:
The pixel is a gene form of this heatmap did not turn out to be useful. After consulting @jaclyn-taroni and @jashapiro , we decided to . move to a window approach where binned portions of the genome are called as a loss, gain or amplification based on the number of bp that are called as each. Will also require a certain (percent of bases?) threshold to avoid calling a whole window as gained/loss if only a few bases were involved in that call.

jaclyn-taroni:
@cansavvy can we close this? I believe any updates you want to make are tracked elsewhere. If that is the case, can you post where they are tracked and close this issue please.

cansavvy:
What's left to be done for the CNV status plot is tracked here: https://github.com/cansavvy/openpbta-notebook-concept/issues/7

PR #603, opened 2020-03-04
CN Status Heatmap (PR 2 of 2)

cansavvy:
We wanted a summary visualization of copy number status.

This second PR has the main notebook where the functions from the previous PR are implemented. 

#### What was your approach?
Create a summary heatmap of copy number status from the consensus CNV call data. 
This is done by binning the genome and calculating the segment's coverage of the
CNV consensus segments. 
A bin is declared a particular copy number status if that status's base pair 
coverage is a certain threshold percentage larger than the other statuses' 
coverage. 

Issue: #594 

- Plot coordinates look wrong. Unsure why. 
- Do the cuttoff's seem reasonable? 
- Keep in mind colors will need to be updated with (#510) but are there other aesthetic issues with the plot?  
- Any gotchas? 

### Results
See the rendered notebook here: 
https://cansavvy.github.io/openpbta-notebook-concept/cnv-chrom-plot/cn_status_heatmap.nb.html

#### Documentation Checklist
**These items already existed but have been updated in #602


cansavvy:
> I did however notice that the bottom CN status annotation bar on the final heatmap does not appear to reflect all of the calls (it appears to reflect only losses and gains although the legend also includes `neutral`, `unstable` and `uncallable`).

I'm not seeing this? When I'm looking at the plot it looks like all of the labels are in the legend? Sometimes the Markdown document cuts off part of it in the preview, but if you look at the pdf, it looks fine.

cbethell:
> I'm not seeing this? When I'm looking at the plot it looks like all of the labels are in the legend? Sometimes the Markdown document cuts off part of it in the preview, but if you look at the pdf, it looks fine.

The legend looks fine, I am referring to the bottom annotation red and blue bar. That bottom bar reflects the chromosomal majority CN status, is that correct?

analyses/cnv-chrom-plot/util/bin-coverage.R
```diff
@@ -139,11 +139,11 @@ call_bin_status <- function(sample_id,
     # the frac_delta_threshold
     dplyr::mutate(
       status =  dplyr::case_when(
-        frac_uncallable > uncallable_threshold ~ "uncallable"
-        frac_gain > threshold ~ "gain",
-        frac_loss > threshold ~ "loss",
-        frac_neutral > threshold ~ "neutral",
-        TRUE ~ "unstable"
+        frac_uncallable > frac_uncallable_val ~ "uncallable",
```
cansavvy, line 142: Is this what we wanted here? Or should we keep what's in master?

jashapiro:
> > I'm not seeing this? When I'm looking at the plot it looks like all of the labels are in the legend? Sometimes the Markdown document cuts off part of it in the preview, but if you look at the pdf, it looks fine.
> 
> The legend looks fine, I am referring to the bottom annotation red and blue bar. That bottom bar reflects the chromosomal majority CN status, is that correct?

I think what you are looking at is just the visual cue for where each chromosome starts and ends. Perhaps the color should be changed to avoid confusion with calls though.

analyses/cnv-chrom-plot/util/bin-coverage.R
jashapiro: It looks like this is correct, as it now matches the function arguments. However, the docs for the argument should also be changed to match.

jashapiro: Unless you were looking at the order of the final two options, but I _think_ that what you have is correct, because you got strange results the other way? I don't fully remember.

cansavvy: Yes. I think we had settled on this. https://alexslemonade.slack.com/archives/CNH4FND1C/p1586786889017900

Will make sure docs are updated though.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -0,0 +1,452 @@
...
+### Set up consensus copy number data
+
+```{r}
+# Read in the segment copy number data
+seg_data <- data.table::fread(file.path(
+  input_dir,
+  "pbta-cnv-consensus.seg.gz"
+),
+data.table = FALSE
+)
```
jashapiro, line 124: Indentation here is strange.
```suggestion
seg_data <- data.table::fread(
  file.path(
    input_dir,
    "pbta-cnv-consensus.seg.gz"
  ),
  data.table = FALSE
)
```

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -0,0 +1,452 @@
...
+```{r}
+seg_data <- seg_data %>%
+  # Join the histology column to this data
+  dplyr::inner_join(dplyr::select(
+    metadata,
+    "Kids_First_Biospecimen_ID",
+    "short_histology",
+    "tumor_ploidy"
+  ),
+  by = c("ID" = "Kids_First_Biospecimen_ID")
```
jashapiro, line 138: Same thing... try not to open 2 sets of parens on the same line that don't then close together, as it makes for non-semantic indentation.
```suggestion
seg_data <- seg_data %>%
  # Join the histology column to this data
  dplyr::inner_join(
    dplyr::select(
      metadata,
      "Kids_First_Biospecimen_ID",
      "short_histology",
      "tumor_ploidy"
    ),
    by = c("ID" = "Kids_First_Biospecimen_ID")
```

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -0,0 +1,452 @@
...
+  bin_calls_mat,
+  name = "CN status",
+  col = color_key,
+  cluster_columns = FALSE,
+  cluster_rows = FALSE,
+  show_column_names = FALSE,
+  show_row_names = FALSE,
+  bottom_annotation = chr_annot,
+  right_annotation = hist_annot,
+  heatmap_legend_param = list(nrow = 1)
```
jashapiro, line 427: The heatmap itself is drawn is rasterized, which is mostly fine, but may be a bit lower resolution than ideal. Can we see what happens if we bump it up, both to file size and image quality?  You might even go higher than this, depending on results. It is a bit hard to interpret the docs https://jokergoo.github.io/ComplexHeatmap-reference/book/a-single-heatmap.html#heatmap-as-raster-image, but if the general setting is equivalent to 72dpi, we might want to go as far as 4 here.
```suggestion
  heatmap_legend_param = list(nrow = 1),
  raster_quality = 2
```

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
cansavvy: I ran `styler` on it. It doesn't always make perfect choices.

jashapiro: Bad `styler`, no biscuit!

Issue #519, opened 2020-02-05
Updated analysis: Immune deconvolution


jainpayal022:
#### What changes need to be made? Please provide enough detail for another participant to make the update.
For the correlation plot (CIBERSORT)- update the colors so that text on the deep blue circles are visible
For both the hearmaps, separate the brain tumor from non-brain tumor rows (histiocytic and lymphoma) so as to see how brain tumors immune landscape is different. 

#### What input data should be used? Which data were used in the version being updated?
same as before

#### When do you expect the revised analysis will be completed?
Unsure

#### Who will complete the updated analysis?
Komal

jharenza:
@komalsrathi can you work on these minor updates when you have a chance? Thanks!

komalsrathi:
@jharenza sure

komalsrathi:
@jainpayal022 @jharenza I am slightly wary of this approach as the sample sizes of the non-brain tumors are pretty small - 1 for lymphomas and 5 for histiocytic tumors. I have still submitted a pull request and working on some minor changes that Jaclyn suggested.

jharenza:
@komalsrathi good catch - I think perhaps then we should remove lymphomas (and perhaps add a minimum N for histologies plotted in such figures). The histiocytic tumors are a good positive control (could possibly go in a supplement).

jharenza:
closed via #536

PR #611, opened 2020-03-06
Immune deconv

komalsrathi:

Immune Deconvolution

Update with v15 data (removed redundancy in histology file)

Yes

#### Documentation Checklist


jaclyn-taroni:
Hi @komalsrathi - this looks good to me and I am happy to approve and merge. I wanted to ask first if you wanted to use this PR as an opportunity to address your concerns about separating out the non-brain cancers from https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/519#issuecomment-586352879.

komalsrathi:
@jaclyn-taroni I was just going to comment on that. I have not heard back from Payal - who initially opened the issue. I thought the original heatmaps were pretty reasonable (per the methods). I think it only makes sense to separate the brain tumors and non-brain ones if any interesting inference can be drawn from it. If we are keeping the original heatmaps (which I would prefer), then I don't think lymphomas should be kept out because even though it is an N of 1, we are just visualizing the average scores per histology per cell type and that 1 lymphoma sample shows nicely that it is enriched in B-cell types. @jharenza thoughts?

jaclyn-taroni:
Tagging @jainpayal022 for discussion

jharenza:
@komalsrathi and @jaclyn-taroni - I think our rationale for separating these was to determine if we could see any signal in CNS tumors, since the non-CNS tumors were so enriched that they blew out any signal. I do think histiocytic and lymphomas are really nice controls and worth keeping, even if lymphoma has an N of 1. Separating these out, we now see some signal in germ cell, mesenchymal, LGAT, metastatic, some lower level of others. I think how we go forward depends on what we would like to say about the tumors - if we put them back together, this is probably suitable for a supplement, because we can't tease out the immune composition within the brain tumors. It is possible we could expand the CNS figure to look at subtypes of certain tumors we have subtypes for and see something - maybe worth exploring?

komalsrathi:
@jharenza I am currently using `broad_histology`, are the subtypes in `pathology_diagnosis`?  Let me know, I will regenerate those figures and upload here for review.

jharenza:
they are in `molecular_subtype` - we just have MB and high-grade gliomas/DMGs subtyped thus far

komalsrathi:
@jharenza With molecular subtypes: I am using `broad_histology` where no subtype info, and `broad_histlogy-molecular_subtype` where subtype info present:

[heatmap_CIBERSORT(abs.).pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4307620/heatmap_CIBERSORT.abs.pdf)
[heatmap_xCell.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4307621/heatmap_xCell.pdf)
[corrplot_xCell_vs_CIBERSORT(abs.).pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4307622/corrplot_xCell_vs_CIBERSORT.abs.pdf)

jaclyn-taroni:
@jharenza - what would you like to do here? Do you want to use 
> `broad_histology` where no subtype info, and` broad_histlogy-molecular_subtype` where subtype info present
If so, this code needs to be updated and reviewed before merging.

jharenza:
I think this looks more informative with the subtypes, but also would make more sense for:
1. Medulloblastomas to be annotated as such, rather than `Embryonal tumor-subtype`
2. Embryonal tumor without subtypes to be annotated as the `ATRT`s they are.

I think these tweaks will be helpful for better expert and audience review.

jaclyn-taroni:
@jharenza if you think that _most figures_ should be organized that way, I strongly encourage you to file a discussion ticket about that because it's going to be easier to start to address while we are working on #571.

jharenza:
@jaclyn-taroni sure - I think it depends. For oncoprints, for ex, we could add subtype as a track. Let me create the discussion ticket!

jharenza:
Discussion ticket here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/627

komalsrathi:
@jharenza @jaclyn-taroni I have made these changes locally (reg. MB and ATRTs). Because the discussion is still open on the best way to annotate histologies, do you want me to wait until a decision is made and then push the changes?

jaclyn-taroni:
@komalsrathi it comes down to your preference. Would you rather get this in as it stands and open another pull request later if things change or would you preference to wait and make that change in this PR?

komalsrathi:
> @komalsrathi it comes down to your preference. Would you rather get this in as it stands and open another pull request later if things change or would your preference be to wait and make that change in this PR?

@jaclyn-taroni The first option sounds good to me.

analyses/immune-deconv/02-summary-plots.R
```diff
@@ -29,6 +29,16 @@ deconvout <- opt$input
...
+# if short histology is Medulloblastoma or ATRT, use them as broad histology
+deconv.res$broad_histology <- ifelse(deconv.res$short_histology %in% c("ATRT", "Medulloblastoma"),
+                                     deconv.res$short_histology, deconv.res$broad_histology)
+
+# add molecular subtype info
+deconv.res$broad_histology <- ifelse(is.na(deconv.res$molecular_subtype), 
+                                     deconv.res$broad_histology, 
+                                     paste0(deconv.res$broad_histology, '-', deconv.res$molecular_subtype))
+
+
```
jaclyn-taroni: My preference would be to create a new variable/column (perhaps called something like `disease_labels`) that uses the logic you have here but doesn't replace `broad_histology`. The new column could then be used downstream. I think this is less confusing than someone reading `broad_histology` in your heatmap code and having the end result in the plot using "custom" labels. You will have to revisit this anyway (related to https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/627), so I think that change can wait.

komalsrathi: Hmm, makes sense. Thank you!

komalsrathi:
@jaclyn-taroni Just checking in - was any decision made regarding the histology labels?

jaclyn-taroni:
Hi @komalsrathi, no I think we are waiting for more folks from D3b with domain expertise to weigh in on #627 (+ trying to wrap up remaining subtyping tickets).

PR #615, opened 2020-03-06
Ignore open_pbta_envs.txt

jaclyn-taroni:

`open_pbta_envs.txt` is generated as part of `scripts/run_in_ci.sh`.

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/cdcd37701a2ff6bb1eaccb1c51af54a141bcfdd1/scripts/run_in_ci.sh#L34

I'm adding this file to `.gitignore` at the root of the repository because sometimes I test things locally and I don't think we want to track this file.

Are there any downsides to this that I am missing?


jaclyn-taroni:
The CI passed last time and this is not a code change, merging without the check.

Issue #497, opened 2020-02-02
Updated analysis: generate cytoband copy number status file for consumption


jaclyn-taroni:
Related to the open draft pull request #485 and #186 

#### What analysis module should be updated and why?

The annotated output of `focal-cn-file-preparation` (e.g., the contents of [`results`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/c2ec68909df4480580de91cdff3b2a0018208343/analyses/focal-cn-file-preparation/results)) contains some information about cytobands. Specifically, the ranges in a SEG file are mapped to gene symbols when there's overlap with exons. We then use the Ensembl gene IDs to map to cytobands via the `org.Hs.eg.db` package. In #485, I looked at assigning cytoband status based on the proportion of gene symbols annotated to that cytoband. The results sometimes disagree with the GISTIC arm status (GISTIC cutoff is 0.98 of arm for an event). We can approach this more directly and that will perhaps lead to more agreement between cytoband status and arm status.

The output of this analysis will feed into https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/488. Here's a photo from a whiteboarding session - the ❌  indicate things that should be removed and replaced by this ticket.

![Image from iOS (24)](https://user-images.githubusercontent.com/19534205/73615481-2be57f80-45d6-11ea-9b62-da6f0b85f537.jpg)

#### What changes need to be made? Please provide enough detail for another participant to make the update.

We can use [`bedtools intersect`](https://bedtools.readthedocs.io/en/latest/content/tools/intersect.html) to look at the overlap between a data frame with copy number status (e.g., `pbta-cnv-controlfreec.tsv.gz` or SEG file processed in `focal-cn-file-preparation`) with cytoband data from UCSC (`http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz`). 

[...]

PR #617, opened 2020-03-09
Add cytoband to copy number files using bedtools intersect

cbethell:

The purpose of this PR is to generate cytoband copy number status consensus files for consumption by downstream analyses.

As noted in the original comment on PR #497, the current annotated output of focal-cn-file-preparation (e.g., the contents of results) contains some information about cytobands. However, the cytoband status sometimes disagree with the GISTIC arm status (GISTIC cutoff is 0.98 of arm for an event). Using the approach in this PR, we can hopefully this issue more directly and hopefully find more agreement between cytoband status and arm status.

#### What was your approach?

My approach was to prepare the cytoband file retrieved from the UCSC database and the `consensus_seg_with_status.tsv` file prepared in `02-add-ploidy-consensus.Rmd` to be in format required by bedtools functions. I also separated the `consensus_seg_with_status.tsv` file into gains and losses and saved these as individual bed files.

I then used bedtools coverage to retrieve the coverage ratio for each of this files using the UCSC cytoband bed file.

In `03-add-cytoband-status-consensus.Rmd`, I merge the data produced using bedtools coverage and denote the dominant status for each chromosome arm using this data. I then add and compare GISTIC's arm status data. **Should this be broken up into 2 separate notebooks?**

This PR addresses issue #497.

- The implementation of the bedtools coverage function and the file this produces should receive a particularly close look.
- The logic determining the `dominant_status` for our consensus calls should also receive some close attention.

Currently, this PR produces a number of intermediate `bed` files stored in the project's scratch directory, but one final results file as follows:
- `results/consensus_seg_gistic_cytoband_status.tsv`

This results file contains the chromosome arm and the dominant status calls for our consensus data and for GISTIC's data.

**Should I be saving this file before the addition of the GISTIC data?**

The chromosome arm calls appear to agree for a total of 35 out of 48 instances (this total excludes the `_alt` chromosomes that I left in the final file for the purpose of being thorough -- these chromosomes are labeled `uncallable`).

#### Documentation Checklist


analyses/focal-cn-file-preparation/00-prepare-bed-files.R
```diff
@@ -0,0 +1,80 @@
...
+  data.table::fread(
+    "http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz"
+  )
+
+# Select variables needed in the UCSC cytoband data -- must be in the
+# required bedtools format: chr, start, end 
+ucsc_cytoband_bed <- ucsc_cytoband %>%
+  dplyr::select(chr = V1, start = V2, end = V3, cytoband = V4) %>%
+  dplyr::mutate(cytoband = paste0(gsub("chr", "", chr), cytoband),
+                chr = gsub("_.*","", chr)) %>%
```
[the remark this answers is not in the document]
cbethell: This step, in cases like: `chr10_GL383545v1_alt`, is removing everything after the first `_` character inclusive. This is done to make it comparable with the `consensus_with_status.tsv` file.

analyses/focal-cn-file-preparation/00-prepare-bed-files.R
```diff
@@ -0,0 +1,80 @@
...
+ucsc_cytoband <-
+  data.table::fread(
+    "http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz"
+  )
+
+# Select variables needed in the UCSC cytoband data -- must be in the
+# required bedtools format: chr, start, end 
+ucsc_cytoband_bed <- ucsc_cytoband %>%
+  dplyr::select(chr = V1, start = V2, end = V3, cytoband = V4) %>%
+  dplyr::mutate(cytoband = paste0(gsub("chr", "", chr), cytoband),
```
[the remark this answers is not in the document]
cbethell: This is an outdated step and not necessary here, will remove 👍

analyses/focal-cn-file-preparation/00-prepare-bed-files.R
```diff
@@ -0,0 +1,80 @@
...
+# cytoband calls.
+ucsc_cytoband <-
+  data.table::fread(
+    "http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz"
+  )
+
+# Select variables needed in the UCSC cytoband data -- must be in the
+# required bedtools format: chr, start, end 
+ucsc_cytoband_bed <- ucsc_cytoband %>%
+  dplyr::select(chr = V1, start = V2, end = V3, cytoband = V4) %>%
```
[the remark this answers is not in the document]
cbethell: Gotcha, will make this change.

analyses/focal-cn-file-preparation/00-prepare-bed-files.R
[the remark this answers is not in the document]
cansavvy: AHH I see. Okay cool. I think in most cases in this project we've dropped `_alt` chromosomes, but I could be wrong.

cbethell:
@cansavvy re:

> **IF** it doesn't throw any downstream analyses off (this is an *if* of unknown size) then you could just alter [this dplyr::select line to move the biospecimen id to the end](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b9c61c35a2f81316aaae2cfdee2afcfe09c496b6/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L123). And then you can use the `consensus_seg_with_status.tsv` file with bedtools exactly how it is. However, I do not know how much that will throw things off, so you should look into that before using this idea. 

I did give this a thought before filing the PR but was worried about downstream analyses, but will look into it and let you know what I find.

analyses/focal-cn-file-preparation/00-prepare-bed-files.R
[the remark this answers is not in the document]
cansavvy: > @cansavvy re:
> 
> > **IF** it doesn't throw any downstream analyses off (this is an _if_ of unknown size) then you could just alter [this dplyr::select line to move the biospecimen id to the end](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b9c61c35a2f81316aaae2cfdee2afcfe09c496b6/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L123). And then you can use the `consensus_seg_with_status.tsv` file with bedtools exactly how it is. However, I do not know how much that will throw things off, so you should look into that before using this idea.
> 
> I did give this a thought before filing the PR but was worried about downstream analyses, but will look into it and let you know what I find.

An alternative that is less risky, then is to save a BED ready version of that file in `scratch` directory at the end of that 02 notebook.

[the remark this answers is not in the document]
cbethell: Ah okay, will also look into this before making the change.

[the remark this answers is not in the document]
cansavvy: By BED ready version, that only means moving that biospecimen column out of the first 3 column spots.

cansavvy:
More specifically @cbethell, given your comments and findings here, here's what I'd suggest: 

Step 1) Change the 02-add-ploidy-consensus.Rmd to save a "`bedtools` ready" of the `cytoband_with_status.tsv` by saving a version of the file in `scratch` where you move the `Kids_First_Biospecimen_ID` column to the end of the file in using this kind of [dplyr::select line](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b9c61c35a2f81316aaae2cfdee2afcfe09c496b6/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L123)

Step 2) Have your already `run-prepare-cn-file.sh` bash script `wget` the UCSC cytoband file with the url you have in this script. 

Step 3) Continue on with the [intersect steps you have laid out in your bash script](https://github.com/cbethell/OpenPBTA-analysis/blob/e745ad935a3f39fbe9006afec20cc17d691ba243/analyses/focal-cn-file-preparation/run-prepare-cn.sh#L39). Though we should discuss if -wa option is what you want here. [See this doc](https://bedtools.readthedocs.io/en/latest/content/tools/intersect.html)

jaclyn-taroni:
`-wa` was suggested by @jashapiro and is in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/497, but is probably worth discussion / re-evaluating so everyone is on the same page.

jashapiro:
I'm not clear on what information `intersect_with_cytoband.tsv` is supposed to capture? (Also, since it is still a `.bed` file it should probably be named that) As it stands (`bedtools intersect -wa -f 0.1`), it is capturing which cytobands have at least 10% of their length present the consensus status, but it is not retaining any information about *which* status (and seems to be super redundant, if the current file is correct)

cansavvy:
> I'm not clear on what information `intersect_with_cytoband.tsv` is supposed to capture? (Also, since it is still a `.bed` file it should probably be named that) As it stands (`bedtools intersect -wa -f 0.1`), it is capturing which cytobands have at least 10% of their length present the consensus status, but it is not retaining any information about _which_ status (and seems to be super redundant, if the current file is correct)

Gotcha. Okay. I think my confusion may have been over what our end file results goal here was. We want UCSC's cytoband reports and do not care about losing cytobands as they are reported by our data.

cbethell:
Upon further discussion with @jashapiro @cansavvy @jaclyn-taroni, the following steps will be taken:

1. Use `bedtools subtract -A -f` to filter out cytobands from the UCSC cytoband file with too much uncalled data.
2. Save separate `bed` files for the losses and gains reported in `consensus_seg_with_status.tsv`.
3. Use `bedtools intersect -wa -f 0.75` between the filtered callable cytobands file (provided to the `-a` flag) and each of the consensus seg bed files (one for losses and one for gains, provided to the `-b` flag).
4. Then read these files into R and annotate these files with genes.

jaclyn-taroni:
> 4. Then read these files into R and annotate these files with genes.

I don't think you need to annotate these files with genes, rather you want to annotate them with status: `callable`, `loss`, `gain`

cbethell:
> I don't think you need to annotate these files with genes, rather you want to annotate them with status: `callable`, `loss`, `gain`

Okay, so the final output we want here is a table with fields similar to the following:
| chr | start | end | cytoband | status |
| --- | ----- | ---- | --------- | ------ |
| chr1 | 0  | 2300000 | p36.33 | loss |
| chr4 | 106700000  | 113200000 | q25 | callable |

Where `callable` could be a possible `neutral` call?

cansavvy:
> > I don't think you need to annotate these files with genes, rather you want to annotate them with status: `callable`, `loss`, `gain`
> 
> Okay, so the final output we want here is a table with fields similar to the following:
> 
> chr	start	end	cytoband	status
> chr1	0	2300000	p36.33	loss
> chr4	106700000	113200000	q25	callable
> Where `callable` could be a possible `neutral` call?

I think callable would be any range that is within `callable` regions but not called as `gain` or `loss`.

jashapiro:
I think you might want every cytoband in there, so I would have 4 possibilities for status: `loss`, `neutral`, `gain`, and `uncalled`


| chr | start | end | cytoband | status |
| --- | ----- | ---- | --------- | ------ |
| chr1 | 0  | 2300000 | p36.33 | loss |
| chr4 | 106700000  | 113200000 | q25 | callable |

jaclyn-taroni:
> I think you might want every cytoband in there, so I would have 4 possibilities for status: `loss`, `neutral`, `gain`, and `uncalled`

This seems like it would be the most flexible for downstream analyses, but I believe it may require you to read an additional file in (the original cytoband file) to capture the `uncalled`.

jashapiro:
An alternative worth exploring might be to use `bedtools coverage` to capture the actual fraction of a cytoband that is covered by loss, gain, or uncalled.

https://bedtools.readthedocs.io/en/latest/content/tools/coverage.html

This could allow you to make a table like the following by combining three separate coverage calls 

| chr  | start | end  | cytoband | loss_fraction | gain_fraction | uncalled_fraction | status  |
|------|-------|------|----------|---------------|---------------|-------------------|---------|
| chr1 | 1000  | 2000 | 1p1      | 0.8           | 0.02          | 0.1               | loss    |
| chr1 | 2000  | 3000 | 1p2      | 0.0           | 0.01          | 0.02              | neutral |

cbethell:
> An alternative worth exploring might be to use `bedtools coverage` to capture the actual fraction of a cytoband that is covered by loss, gain, or uncalled.
> 
> https://bedtools.readthedocs.io/en/latest/content/tools/coverage.html
> 
> This could allow you to make a table like the following by combining three separate coverage calls
> 
> chr	start	end	cytoband	loss_fraction	gain_fraction	uncalled_fraction	status
> chr1	1000	2000	1p1	0.8	0.02	0.1	loss
> chr1	2000	3000	1p2	0.0	0.01	0.02	neutral
> This would give flexibility on choosing cutoffs later without rerunning the whole thing.

Nice, I'll look into implementing this tool. 

My understanding here is that `bedtools subtract` would still be used to retain the callable cytoband regions, then `bedtools coverage` would be used (in place of `bedtools intersect`) for each consensus seg with status file (losses, gains, and uncalled regions). Correct?

jashapiro:
> My understanding here is that `bedtools subtract` would still be used to retain the callable cytoband regions, then `bedtools coverage` would be used (in place of `bedtools intersect`) for each consensus seg with status file (losses, gains, and uncalled regions). Correct?

I think you could skip the `subtract` and just go to `coverage` for gains, losses, and uncallable. Then you would read in the three coverage results files to merge them together and make "status" calls.

cbethell:
> I think you could skip the `subtract` and just go to `coverage` for gains, losses, and uncallable. Then you would read in the three coverage results files to merge them together and make "status" calls.

@jashapiro I tried to implement this in the [last commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/617/commits/1c4eea5498a2b772c8c41e34b67b00ca75559a93), but was unsuccessful thus far as the script seems to get stuck at the first `bedtools coverage` implementation in the shell script but does not throw an error. Do you have an idea of why this may be?

jashapiro:
@cbethell I am not sure what would be happening there, but I have a couple of quick thoughts. 

One is that you should not need the `-f` flag for the coverage calculation.
The other is that you may be able to speed it up by making sure the bed files are sorted and then adding `--sorted` throughout. I suggest doing this with an `arrange()` statement in your R script before printing, which I will add as a suggestion in just a sec.

analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd
```diff
@@ -146,3 +146,31 @@ add_status_df %>%
...
+bed_status_df <- add_status_df %>%
+  select(chrom, loc.start, loc.end, everything())
+
+losses_bed_status_df <- add_status_df %>%
+  select(chrom, loc.start, loc.end, everything()) %>%
+  filter(status == "loss")
+
+gains_bed_status_df <- add_status_df %>%
+  select(chrom, loc.start, loc.end, everything()) %>%
+  filter(status == "gain")
```
jashapiro: This makes sure the bed tables are sorted before output.

```suggestion
bed_status_df <- add_status_df %>%
  select(chrom, loc.start, loc.end, everything()) %>%
  arrange(chrom, loc.start, loc.end)

losses_bed_status_df <- bed_status_df %>%
  filter(status == "loss")

gains_bed_status_df <- bed_status_df %>%
  filter(status == "gain")
```

analyses/focal-cn-file-preparation/run-prepare-cn.sh
```diff
@@ -18,74 +18,105 @@ cd "$script_directory" || exit
...
+results_dir=../../analyses/focal-cn-file-preparation/results
 histologies_file=${data_dir}/pbta-histologies.tsv
 gtf_file=${data_dir}/gencode.v27.primary_assembly.annotation.gtf.gz
 goi_file=../../analyses/oncoprint-landscape/driver-lists/brain-goi-list-long.txt
 independent_specimens_file=${data_dir}/independent-specimens.wgswxs.primary.tsv
+ucsc_bed_file=${results_dir}/ucsc_cytoband.bed
+consensus_bed_file=${scratch_dir}/consensus_seg_with_status.tsv
+loss_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_losses.tsv
+gain_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_gains.tsv
+callable_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_callable.tsv
```
jashapiro: I think these file names where changed elsewhere, so I am noting that here.
```suggestion
consensus_bed_file=${scratch_dir}/consensus_seg_with_status.bed
loss_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_losses.bed
gain_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_gains.bed
callable_intersect_with_cytoband_file=${scratch_dir}/intersect_with_cytoband_callable.bed
```

analyses/focal-cn-file-preparation/run-prepare-cn.sh
```diff
@@ -18,74 +18,105 @@ cd "$script_directory" || exit
...
+# cytoband data and the `scratch/consensus_with_status.tsv` file prepared in
+# `02-add-ploidy-consensus.Rmd`
 
-libraryStrategies=("polya" "stranded")
-chromosomesType=("autosomes" "x_and_y")
-for strategy in ${libraryStrategies[@]}; do
+bedtools coverage \
+    -a ${scratch_dir}/ucsc_cytoband.bed \
+    -b ${scratch_dir}/consensus_seg_with_status_losses.bed \
+    -f 0.75 \
```
jashapiro: If you have sorted the `-b` file, then you can speed this up with `--sorted` (I don't think we need `-f` here).

The same thing applies to the two `bedtools coverage` calls below this one.
```suggestion
    --sorted \
```

cbethell:
The notebook added in the [last commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/617/commits/337a2644645db8e3c23b1f62ec11d67e27c247fa) can be seen [here](https://cbethell.github.io/open-pbta-output/03-add-cytoband-status-consensus.nb.html). 

In this notebook so far,
- I merged the data prepared by `bedtools coverage` in the shell script (also done in this PR) and created a `dominant_status` column using the coverage values. 
- I then read in the original cytoband file to define `uncallable` values in the `dominant_status` field. However, these rows seem to be `_alt` chromosomes (I believe we are dropping these but I left them in for verification that we should ignore these instances).
- I also assigned the `uncallable` value to rows where the callable coverage ratio  == 0. This step should probably receive a close look as I am not sure that this was the right decision made here.
- My next steps will be adding a chromosome arm field and adding the GISTIC broad arm data for comparison.

cbethell:
> I think the way this is currently implemented it is not capturing the per-sample coverage that we really need, so I suggested a way to do that by breaking things into lots of separate files and then some bash-fu to do all the separate parts.
> 
> The bash scripting for `bedtools` is getting extensive, so I would also recommend pulling that out into its own script, separate from the `run-prepare-cn.sh` script, and then calling that script from within the main runner script. That should make debugging just that section easier.

Thanks @jashapiro! I'll separate the bash scripting for `bedtools` into its own script, attempt to implement your code suggestions below and let you know if I come across any issues or questions.

analyses/focal-cn-file-preparation/run-bedtools.sh
```diff
@@ -0,0 +1,64 @@
...
+#
+# Usage: bash run-bedtools.sh
+
+scratch_dir=../../scratch
+callable_intersect_with_cytoband_file=${scratch_dir}/cytoband_status/intersect_with_cytoband_callable.bed
+loss_intersect_with_cytoband_file=${scratch_dir}/cytoband_status/intersect_with_cytoband_losses.bed
+gain_intersect_with_cytoband_file=${scratch_dir}/cytoband_status/intersect_with_cytoband_gains.bed
+
+# Download and save UCSC cytoband file as bed file
+wget -O ${scratch_dir}/ucsc_cytoband.bed http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz
```
jashapiro:
```suggestion
wget -O - http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz \
  | gunzip -c > ${scratch_dir}/ucsc_cytoband.bed 
```

cbethell:
In the [last commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/617/commits/69a3ee5bea23d7b094dc87e2c4a66824ca966ef7), I added a notebook named `05-define-most-focal-units.Rmd` to start to dig into defining the most focal units for recurrent CNVs using the cytoband status calls determined in this PR. (I also updated the `README` to reflect the changes in the PR in its current state.)

_Note: This notebook is beyond the scope of this PR and was added to demonstrate my thought process on how we would want to tackle the part of the analysis that focuses on finding the most focal units._

The reason I added it to this PR is because it determines the final file that comes out of this PR.

In other words, should this PR save a file of our CN calls with `chromsome_arm`, `cytoband`, `dominant_status`, and `Kids_First_Biospecimen_ID`, for consumption in a subsequent notebook that will take this information and keep focal CN cytoband calls where possible, and denote broad chromosome arm calls were they exist? (This file was saved and added in the last commit)

If not, what should the final file of this PR look like with the next step being defining the most focal units in mind?

jaclyn-taroni:
> In other words, should this PR save a file of our CN calls with `chromosome_arm`, `cytoband`, `dominant_status`, and `Kids_First_Biospecimen_ID`, for consumption in a subsequent notebook that will take this information and keep focal CN cytoband calls where possible, and denote broad chromosome arm calls were they exist? (This file was saved and added in the last commit)

In this description, it is not clear to me how we would distinguish an arm loss from a cytoband loss. My kneejerk reaction is that you would want to generate arm status separately using a similar approach to how the cytoband calls were generated (you may want to supply a different value to `-f` for example). That being said, I have not been keeping up with this as closely as @jashapiro and may have said something contradictory.

jashapiro:
> > In other words, should this PR save a file of our CN calls with `chromosome_arm`, `cytoband`, `dominant_status`, and `Kids_First_Biospecimen_ID`, for consumption in a subsequent notebook that will take this information and keep focal CN cytoband calls where possible, and denote broad chromosome arm calls were they exist? (This file was saved and added in the last commit)
> 
> In this description, it is not clear to me how we would distinguish an arm loss from a cytoband loss. My kneejerk reaction is that you would want to generate arm status separately using a similar approach to how the cytoband calls were generated (you may want to supply a different value to `-f` for example). That being said, I have not been keeping up with this as closely as @jashapiro and may have said something contradictory.

I think if we have all of the cytoband calls, we can readily combine them to identify arm losses with some simple `dplyr::group_by()` and `mutate()`. We are not using `-f` for the the coverage calculations, as we get more raw data out (% of each cytoband covered) that we can make (and adjust) calls from within the R script which is what makes the "dominant" call.  That said, I think the arm calls I think should wait for the next PR. I'd like to see this one get in with the table of cytoband stats/calls.

jaclyn-taroni:
Okay sounds good 👍

cbethell:
> I'd like to see this one get in with the table of cytoband stats/calls.

This is what I wanted to get to the bottom of, the final output table that we would like to see from this PR.
The addition of the `05` nb was more of a how "this is how we may want to use the output of this PR in the next step, so is this how I should save the output". **This will be removed in the upcoming commit.**

So to make sure that we are on the same page here, the final table from this PR should include the following fields:

|chr | cytoband | coverage_ratio_callable | coverage_ratio_gains | coverage_ratio_losses | dominant_status |
| -- | ---------- | ----------------- | --------------- | ----------------- | ---------------|

Is that correct?

jashapiro:
> So to make sure that we are on the same page here, the final table from this PR should include the following fields:
> 
> chr	cytoband	coverage_ratio_callable	coverage_ratio_gains	coverage_ratio_losses	dominant_status
> Is that correct?

That seems right to me. You also had a column for **chromosome arm** in there in the last version I looked at, and I would leave that in.

I might change the `coverage_ratio_*` labels to `callable_fraction`, `gain_fraction` and `loss_fraction` too, just to make them a bit simpler.

I do think I would save the gistic comparison for a separate PR.

cbethell:
> That seems right to me. You also had a column for chromosome arm in there in the last version I > > looked at, and I would leave that in.
>
> I might change the coverage_ratio_* labels to callable_fraction, gain_fraction and loss_fraction too, > just to make them a bit simpler.
>
> I do think I would save the gistic comparison for a separate PR.

Okay sounds good, these changes will be in the upcoming commit.

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,160 @@
...
+## Wrangle and merge consensus cytoband status data
+
+### Filter each of the cytoband status data.frames
+
+```{r}
+all_callable_cytoband_status_df <-
+  all_callable_cytoband_status_df %>%
+  select(
+    chr = X1,
+    cytoband = X4,
```
jashapiro: I am adding in `band_length` here so that we can output this to properly calculate the fractions of each arm that are gains, losses, etc. in the next step. You won't need to keep it for any of the other data frames, as it will be the same, so the join will retain the info. 
```suggestion
    cytoband = X4,
    band_length = X8,
```

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,160 @@
...
+### Add chromosome arm column
+
+```{r}
+# Add a column that tells us the position of the p or q and then use this to
+# split the cytoband column
+final_df <- final_df %>%
+  mutate(cytoband_with_arm = paste0(gsub("chr", "", chr), cytoband),
+         chromosome_arm = gsub("(p|q).*", "\\1", cytoband_with_arm)) %>%
+  select(-cytoband_with_arm)
+```
```
jashapiro: This step should be moved to after the UCSC data is merged back in, otherwise those uncallable cytobands might not have arm assignments.

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,160 @@
...
+
+### Read in and join original UCSC cytoband data
+
+This step is to define any additional uncallable cytoband regions.
+
+```{r message = FALSE}
+ucsc_cytoband_df <- data.table::fread("http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz", data.table = FALSE)
+
+ucsc_cytoband_df <- ucsc_cytoband_df %>%
+  select(chr = V1, cytoband = V4)
```
jashapiro: band_length = 
A couple of suggestions here: one to add the band_length as above, the other to remove non-canonical chromosomes (and mitochondria) which we don't call for CNVs anyway.

```suggestion
  mutate(band_length = V3 - V2) %>%
  select(chr = V1, cytoband = V4, band_length) %>%
  filter(chr %in% paste0("chr", 1:23)) # remove nonstandard chroms.
```

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,160 @@
...
+
+# Join the cytoband data from the original UCSC cytoband file with the data
+# bedtools coverage files wrangled in the steps above
+final_df <- final_df %>%
+  right_join(ucsc_cytoband_df, by = c("chr", "cytoband")) %>%
```
jashapiro: Adding `band_length` to prevent duplication of band_length columns
```suggestion
  right_join(ucsc_cytoband_df, by = c("chr", "cytoband", "band_length")) %>%
```

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,160 @@
...
+  select(Kids_First_Biospecimen_ID, chr, cytoband, dominant_status, everything())
```
jashapiro:
```suggestion
  select(Kids_First_Biospecimen_ID, chr, cytoband, dominant_status, band_length, everything())
```

analyses/focal-cn-file-preparation/README.md
```diff
@@ -28,9 +30,14 @@ RUN_ORIGINAL=1 bash analyses/focal-cn-file-preparation/run-prepare-cn.sh
...
 However, there are instances in the consensus SEG file where `copy.num` is `NA` which are removed.
-See the notebook for more information.
+See the notebook for more information. This notebook also prepares lists of copy number bed files by sample for use in the implementation of bedtools coverage in `run-bedtools.snakemake`.
 
-* `03-prepare-cn-file.R` - This script performs the ranges to annotation mapping using the GENCODE v27 GTF included via the data download step; it takes the ControlFreeC file or a SEG (e.g., CNVkit, consensus SEG) file prepared with `01-add-ploidy-cnvkit.Rmd` and  `02-add-ploidy-cnvkit.Rmd` as input.
+* `03-add-cytoband-status-consensus.Rmd` - This notebook reads in the bedtools coverage output files and defines the dominant copy number status for each chromosome arm. The output of this notebook is a table with the following columns:
+
+  | `Kids_First_Biospecimen_ID` | chr | cytoband | dominant_status | callable_fraction | gain_fraction | loss_fraction | chromosome_arm |
+  |----------------|--------|-------------|--------|---------|-------------|---------|---------------|
```
jashapiro: You will want to update this to reflect any added columns with changes above.

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
```diff
@@ -0,0 +1,171 @@
...
+
+ucsc_cytoband_df <- ucsc_cytoband_df %>%
+  mutate(band_length = V3 - V2) %>%
+  select(chr = V1, cytoband = V4, band_length) %>%
+  filter(chr %in% paste0("chr", 1:23)) # remove nonstandard chroms.
+
+# Join the cytoband data from the original UCSC cytoband file with the data
+# bedtools coverage files wrangled in the steps above
+final_df <- final_df %>%
+  right_join(ucsc_cytoband_df, by = c("chr", "cytoband", "band_length"))
```
jashapiro: Realizing this whole step may be redundant, since the `bedtools coverage` is including all of the ucsc cytobands already.

cbethell:
> Looks good! I am approving this, but I have two additional comments...
> 
> One is whether we need to add in the ucsc cytobands at all, as the `bedtools coverage` results really should have all the ones we care about, and it looks like the merge is resulting in no added rows now.
> 
> I also did a bit of spot checking, and there does seem to be something a bit funny going on with sample BS_CBMAWSAR. For some reason, that one is not getting "uncallable" calls as I would expect (if one sample is uncallable in a region, all of them should be).
> 
> For example, here:
> 
> ```
> BS_JTBM5TSE	chr1	p11.2	uncallable	1300000	0.2165369	0	0.2165369	1p
> BS_DE26D072	chr1	p11.2	uncallable	1300000	0.2165369	0	0.2165369	1p
> BS_CBMAWSAR	chr1	p11.2	neutral	1300000	1	0	0	1p
> BS_ZS1QRMXS	chr1	p11.2	uncallable	1300000	0.2165369	0	0	1p
> BS_5R0HHQ1Y	chr1	p11.2	uncallable	1300000	0.2165369	0	0	1p
> ```
> 
> So something is funny there, which may warrant investigation. Most likely it isn't showing up in one of the original tables for some reason? Or it shouldn't be included at all?

After some investigation, I found that the BS_CBMAWSAR sample, with the addition of three other samples( namely, BS_H50HR85Y, BS_FF73TT6D, and BS_80078QDG), all have a callable ratio of 1 for this region compared to the remainder of the samples with a callable ratio of 0.2165369. However, these samples appear to agree with all other samples of the cohort in other cytoband regions.
Note: This information was determined using the `intersect_with_cytoband_callable.bed` file

Would you still recommend removing these samples?

analyses/focal-cn-file-preparation/03-add-cytoband-status-consensus.Rmd
cbethell: Agreed, I will remove this section.

jashapiro:
I don't think anything different should be done with these

> Would you still recommend removing these samples?

No, I would leave them. I have some guesses why that is happening, but I don't think it should be a major issue in general. My guess is that it is where there is a very large segment that crosses an uncallable region. Shouldn't happen that often, but should be something we are aware of, I guess. 

We will probably want to add logic at some point that says if a band is mostly uncallable, then it should be uncallable for all samples. But other bands for that sample should stay as they are.

analyses/focal-cn-file-preparation/run-bedtools.snakemake
```diff
@@ -0,0 +1,75 @@
...
+
+rule get_cytobands:
+    output:
+        scratch_dir + "/ucsc_cytoband.bed"
+    params:
+        url = "http://hgdownload.cse.ucsc.edu/goldenpath/hg38/database/cytoBand.txt.gz"
+    shell:
+        "wget -O - {params.url}"
+        " | gunzip -c"
+        " | grep '^chr[0-9]\+\s' " # filter to only canonical autosomes
```
jashapiro:
```suggestion
        " | grep '^chr[0-9XY]\+\s' " # filter to only canonical autosomes
```

cbethell: I'm on it 👍

PR #612, opened 2020-03-06
Functionalize CDF TMB Plot

cansavvy:
#### What scientific question is your analysis addressing?
No new scientific question, just moving the CDF TMB plot to its own file so it is more accessible. 

Note that GitHub can't track that the same exact code moved from the notebook to the R script but it is the same exact code. 

#### What was your approach?
Put the CDF plot it its own file so it can be sourced anywhere, Also added more documentation. 

#### What GitHub issue does your pull request address?
Related to [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/557#discussion_r387949966) and [this issue](https://github.com/cansavvy/openpbta-notebook-concept/issues/6).

### Results
No new results. 

#### Reproducibility Checklist
No changes to this.

PR #622, opened 2020-03-10
Unified color palette set up

cansavvy:
#### What scientific question is your analysis addressing?
In order to help the interpretability and aesthetics of our figures, we are moving toward a unified color palette. 

#### What was your approach?
Main focus here is implementing a method of making our color palettes in a place where they can be easily grabbed and implemented in any analysis module. This is what the new base folder `figures` will contain (as well as the figure running scripts that are in progress #613 and future PRs) There is a README that is added that explains the rationale for each of the color palettes.

You will see changes made to `chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd` that isn't final, because I needed to test this in a real scenario. I would recommend ignoring the chromosomal-instability items and looking at the color-palette.R script and README first.

#### What GitHub issue does your pull request address?
#510 

#### Which areas should receive a particularly close look?
- Do we need TSV files for each of the palettes listed in the README table? Right now the RDS method for the other lists of hex codes is probably fine for R plots but if we add other non-R plots in the future we need a file for that too. 
- What's the easiest way for people to grab these? 
- I included 9 Hex code lists for gradients, but that might be overkill? Any feedback on this? 
- Are there any color palettes we need beyond the colors included here (see the README table) that should be included and are not? 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
No. Not yet. 

#### What types of results are included (e.g., table, figure)?
No particular results, but some documentation, a script and some hex codes. 
Here are those new color palettes in action: https://cansavvy.github.io/openpbta-notebook-concept/chromosomal-instability/02a-plot-chr-instability-heatmaps.nb.html

#### Reproducibility Checklist
This has not been added to CI but potentially should be added at some point. 
- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This color palette folder has a `README` and it is up to date.
- [ ] The analytical code is documented and contains comments.

Issue #601, opened 2020-03-04
Planned Data Release: V16


jharenza:
#### What data file(s) does this issue pertain to?
pbta-histologies.tsv
pbta-tcga-manifest.tsv           
pbta-tcga-snv-lancet.vep.maf.gz  
pbta-tcga-snv-mutect2.vep.maf.gz  
pbta-tcga-snv-strelka2.vep.maf.gz

#### What release are you using?

V15

#### Put your question or report your issue here.
- [ ] Updated TCGA MAFs per #568 
- [ ] Update TCGA manifest file to include BED file used per sample
- [ ] Add new TCGA BED files, padded and unpadded
- [ ] Remove old TCGA BED file `gencode.v19.basic.exome.hg38liftover.100bp_padded.bed`
- [ ] Update histologies file to include additional information per tickets #607, #610, #637 ~potentially #509 and #608 #245~ (@jaclyn-taroni edit per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/601#issuecomment-603343136)

Placeholder for V16 release to include D3b team: @baileyckelly @chris-s-friedman @yuankunzhu @allisonheath

jaclyn-taroni:
Because we made changes to the `pbta-fusion-putative-oncogenic.tsv` with v15, we've had to update the `fusion-summary` files that are included in a release (`fusion_summary_embryonal_foi.tsv` and `fusion_summary_ependymoma_foi.tsv`): https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/516201f91f174c495f6d7964f4a03d6c7c6fb227/analyses/fusion-summary/results

allisonheath:
For the v16 release, @baileyckelly and @chris-s-friedman are moving the histology file generation over to a new database workflow to make the manual changes more trackable. We are working on replicating v15 first, expected to be done by the end of this week and then will start working on the new issues noted in the ticket early next. Will update if this changes.

allisonheath:
And can we go ahead and add https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/624 as well?

jaclyn-taroni:
@tkoganti gave me a heads up that the updated TCGA files are available pre-v16 if folks would like to take a look.

Here's the link to the s3 bucket and folder `https://s3.amazonaws.com/kf-openaccess-us-east-1-prd-pbta/data/TCGA_mar-12-2020/`

And the file list:

```
README.md
[...]

PR #625, opened 2020-03-10
chromosomal-instability cdf function updates

cansavvy:
No new scientific question, just a re-vamp of some plots that can use a recently functionalize CDF plotter that has been merged (#612). 

#### What was your approach?
I used the function I recently separated into its own script and now made this 02b-plot-chr-instability-by-histology.Rmd CDF plots use this function instead of having their own separate CDF function in the `chr-break-plot.R` script. 

[This issue](https://github.com/cansavvy/openpbta-notebook-concept/issues/6)

- Is there anything missing with this update? 
- Outdated comments? 
- Note that the color palette needs to be updated but that will be in a different PR.

Yes. 

### Results
Results are not actually different just how we got the plots is. 

***No changes were made to the below items. 

#### Documentation Checklist


jaclyn-taroni:
These CDF plots are not the same. 

* SV breaks is empty: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=3e9d2ec#diff-3e9d2ec9cb00ce64531c1ad8f191458f
* ATRT, for example, is missing from the intersection plot even though I suspect that it meets the minimum samples threshold: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=f0bbbb2#diff-f0bbbb2f537c67ca9388a0ed5e13aaca

I thought it might be an issue with the underlying density files needing to be rerun, but I got:

```
$ bash run_breakpoint_analysis.sh
run_breakpoint_analysis.sh: line 19: scratch/cnv_surveyed_wgs.bed: No such file or directory
```

This is following the directions in the README - I think it's a paths issue which I will address in an upcoming commit. If, however, I need to run something _upstream_ before I can generated `cnv_surveyed_wgs.bed`, we should get that added to the documentation as well. Also wondering if it's a good time to address some of the TODOs in the shell script (may be beyond the scope of this pull request).

jaclyn-taroni:
I ran into trouble debugging this due to the instructions in the README, which is a very easy fix that I will address, and the design pattern here:  https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/796fcf2eb32e970fb2995cd593746a40265cfb98/analyses/chromosomal-instability/00-setup-breakpoint-data.R#L104

I am going to alter this module to remove the pattern where an Rscript accepts the path relative to the root directory and instead accepts a "generic" path that should be relative to the current working directory when you run it.

I will then make it such that the bash script has a working directory of where it is saved:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/796fcf2eb32e970fb2995cd593746a40265cfb98/analyses/interaction-plots/01-create-interaction-plots.sh#L14

That ensures that the shell script will run correctly no matter where it is called from.

jaclyn-taroni:
Rerunning all steps did not "fix" the plots. I am going to leave it there and let @cansavvy take another look.

cansavvy:
- [SV breaks plot is fixed](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=3e9d2ec#diff-3e9d2ec9cb00ce64531c1ad8f191458f)
- [ATRT plot is here and looks fine](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=a51bf31#diff-a51bf313b3e96e31d2ed4867dac8633b)
- The TODOs here are either not applicable at this time [because we don't have WXS data](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/7e000418b106e34040325d2fbd642f63080966c8/analyses/chromosomal-instability/run_breakpoint_analysis.sh#L23), or cannot really be addressed until [the unsurveyed cnv regions of the genome gets added to the data release ](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/7e000418b106e34040325d2fbd642f63080966c8/analyses/chromosomal-instability/run_breakpoint_analysis.sh#L16) . Speaking of which, is that to happen in #601 v16?
This all being said, I edited the comments around those steps in the bash script so it is more clear why they are still TODOs that have not been acted upon. 

To be addressed but not in this particular PR: 
- Color palette needs to be applied (But need #622 to be merged first). 
- Documentation around relative file paths needs to be made more clear (This is tracked here: #632)

Is there anything else that is still at large for this PR, @jaclyn-taroni ? I believe your comments have been addressed (or are being tracked elsewhere) and this runs fine for me and CircleCI right now.

jaclyn-taroni:
Two quick comments:

> * ATRT, for example, is missing from the intersection plot even though I suspect that it meets the minimum samples threshold: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=f0bbbb2#diff-f0bbbb2f537c67ca9388a0ed5e13aaca

The plot linked there is `analyses/chromosomal-instability/plots/intersection_breaks_cdf_plot.png` not `analyses/chromosomal-instability/plots/tumor-type/ATRT_breaks.png` which you link to above. It still looks like ATRT is not in `intersection_breaks_cdf_plot.png` in the most recent update to me. I wanted to verify that this is expected.

> Documentation around relative file paths needs to be made more clear (This is tracked here: #632)

I removed this pattern in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/commits/323db4b110d0660c34b619f45897bd2c033a358f, but perhaps I missed something.

cansavvy:
> The plot linked there is `analyses/chromosomal-instability/plots/intersection_breaks_cdf_plot.png` not `analyses/chromosomal-instability/plots/tumor-type/ATRT_breaks.png` which you link to above. 

I may have accidentally linked to the wrong plot. 

>It still looks like ATRT is not in `intersection_breaks_cdf_plot.png` in the most recent update to me. I wanted to verify that this is expected.

ATRT is there its just not the first one listed because it is order by median and previously the median was not being calculated with `na.rm` so this has changed the order of the histologies in the plots: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=f0bbbb2#diff-f0bbbb2f537c67ca9388a0ed5e13aaca

cansavvy:
> ATRT is there its just not the first one listed because it is order by median and previously the median was not being calculated with `na.rm` so this has changed the order of the histologies in the plots: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625/files?short_path=f0bbbb2#diff-f0bbbb2f537c67ca9388a0ed5e13aaca

Actually, I think the change is the reverse of what I said here, the `na.rm = TRUE` was there before but is not now. I'm going to add back `na.rm = TRUE` because I think that is more what we want.

jaclyn-taroni:
👍 I thought it was alphabetical in both so that is my confusion

cansavvy:
> 👍 I thought it was alphabetical in both so that is my confusion

It's good we got to the bottom of it. I didn't realize the `na.rm = TRUE` got lost in the shuffle.

jaclyn-taroni:
The system is working!

cansavvy:
Okay, @jaclyn-taroni , feel free to take another look to see if everything has been addressed.

Issue #629, opened 2020-03-13
Figure Panel: Sample distribution across histologies plot


cbethell:
As a part of [figure 1 of the manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/04.results.md#the-open-pediatric-brain-tumor-atlas). The Github contributions figure mentioned in ticket #572 is not yet available, I will therefore be working on the following subset of this panel: 
<img width="1003" alt="Screen Shot 2020-03-13 at 12 37 11 PM" src="https://user-images.githubusercontent.com/43576623/76641075-64f30700-6527-11ea-9570-9806fa9e036a.png">

- Two options for representing the sample distribution will be provided: a barplot and a treemap.
- Then a separate script for putting these plots in a figure panel will be created (setting this script up to accept the other figures that will later be a part of this panel as seen in the [sketch](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/04.results.md#the-open-pediatric-brain-tumor-atlas).

jaclyn-taroni:
Closed via #633 - if any changes should be made, I think a new ticket is appropriate (notes to self: update a figure issue template could be helpful).

PR #613, opened 2020-03-06
Figure 2 mutational-landscape run and some set up for figure scripts

cansavvy:
### Purpose/implementation Section

** Note that this includes changes that occurred in #612, ignore those changes GitHub will ignore them once that PR is merged. 

#### What scientific question is your analysis addressing?
This is adding a script for the mutational landscape figure and associated documentation for re-running figures as described on #571. 

#### What was your approach?
There is a scripts/figure-scripts folder where I added my `fig2-mutational-landscape.R` script. 
This script re-runs all steps leading up to the mutational-landscape and tmb-comparison multipanel figure described in this [manuscript section](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/04.results.md#landscape-of-mutational-processes). 

**Note that after this plot is merged I will file a PR in the manuscript repo to link this plot in the manuscript. 

In the instructions I put that manubot should be refreshed in someway? I don't know if this is accurate. We should replace those instructions with something more specific and accurate. 

#### What GitHub issue does your pull request address?
#571 and https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/84

### Directions for reviewers. Tell potential reviewers what kind of feedback you are soliciting.

- How do we feel about the file and script structure I set up? 
- I took a whack at making the associated README and made a table to start, but it will undoubtedly need some editing. 
- Any aesthetic changes that need to be made to the plot? (Yes the data is still wrong, but the original issue #571 assumes this)

### Results
fig2-mutational-landscapes

#### Reproducibility Checklist
This should still be added to CircleCI in someway, but I am unsure if we want to bog down the same set of tests? What should happen to test this script? 

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist
- [x] It has its own README in the figures folder
- [x] The analytical code is documented and contains comments.

PR #633, opened 2020-03-16
Figure 1 sample distribution plot

cbethell:

The purpose of this PR is to add a script to make the sample distribution across cancer types plot publication ready.

This PR pulls out the necessary results and steps required to rerun figure 1 plots when necessary.

#### What was your approach?

As in @cansavvy's PR #613, 
I created a script that takes the necessary results files from the `sample-distribution-analysis` module and reruns the sample distribution bar plot and saved this script in a directory named `figures/scripts`.

The figure1 plot is saved in a directory named `figures/pngs`.

This PR addresses issue #629.

- Should there be any changes in the format of the script?
- Although there is only one fourth of the figure1 panel represented, is there anything that should change in this final plot thus far?
- Should the color scheme of the sample distribution plot be more distinct?

fig1-openpbta-distribution

In the [figure 1 sketch in the manuscript repository](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/04.results.md#the-open-pediatric-brain-tumor-atlas), the B and D
aspects of the panel are not represented in a plot or table. 

I am thinking that these should be represented in tables where we have a value/feature column and a description column. Should I go ahead and implement this idea or are there any other suggestions for how these parts of the panel should be represented?

Also, in this PR is a sample distribution barplot and a sample distribution treemap as follows:
[distribution_across_cancer_types.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4339218/distribution_across_cancer_types.pdf)

[distribution_across_cancer_types_treemap.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4339220/distribution_across_cancer_types_treemap.pdf)

Which plot would we prefer to be represented in the final figure 1 panel?
I believe the barplot is easier to follow. 

Yes, it is mature enough for review (although it is not yet complete -- the three other aspects of the figure panel need to be generated and added).

Plots as follows:
- `analyses/sample-distribution-analysis/plots/distribution_across_cancer_types.pdf`
image
- `analyses/sample-distribution-analysis/plots/distribution_across_cancer_types_treemap.pdf`
image
- `figures/pngs/fig1-openpbta-distribution.png`
image

#### Documentation Checklist


cbethell:
Here is the updated figure 1 plot with a treemap (generated using the [treemapify package](https://github.com/wilkox/treemapify)) representing the sample distribution across histologies and cancer types (accounting for the inclusion of 3 additional plots/panels sometime in the future):

fig1-openpbta-distribution

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+
+# Reorder the columns to be displayed in descending order by count on the plot
+disease_expression$integrated_diagnosis <- with(disease_expression,
+                                                reorder(integrated_diagnosis, -count))
+
+#### Re-run the individual plots ----------------------------------------------
+
+# Create a bar plot of sample distribution across cancer types
+gg_types_bar <- disease_expression %>%
+  ggplot2::ggplot(ggplot2::aes(x = integrated_diagnosis, y = count, fill = count)) +
```
jashapiro: I don't tend to like having redundant info in a figure, so having the fill also represent the count is probably not needed here.  However, we may well want to use the colors from the final color schemes https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/510 to color the bars by histology.  Either way, we probably don't want to have the color legend at all, as it does not provide any real information, so you will want to add to theme `legend.position = 'none'`

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+    max_size = max(size),
+    color = gradient_n_pal(
+      sequential_hcl(
+        6,
+        h = 360 * index[1]/n,
+        c = c(45, 20),
+        l = c(30, 80),
+        power = .5)
+    )(size/max_size)
+  )
```
jashapiro: We will presumably want to use #510 for this as well, possibly adjusted in lightness?

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+                               grow = T,
+                               reflow = T,
+                               alpha = 0.6,
+                               colour = "#FAFAFA",
+                               min.size = 0) +
+    geom_treemap_subgroup2_text(place = "centre",
+                                grow = T,
+                                alpha = 0.8,
+                                colour = "#FAFAFA",
+                                min.size = 0) +
```
jashapiro: Again here, I'm not sure where each set of labels are defined, but the smallest labels seem redundant and add visual noise

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+                  ncol = 2,
+                  labels = c("B", "D"),
+                  widths = c(1.6, 2),
+                  font.label = list(size = 20)
+),
+nrow = 2,
+heights = c(2, 2)
+)
+
+# Save to PNG
```
jashapiro: PDF?

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+ggplot2::ggsave(file.path(output_dir, "fig1-openpbta-distribution.png"),
+                width = 16.5, height = 15,
```
jashapiro: Here too, maybe a smaller final plot: I'd suggest no bigger than 8x10 (or 10x8) so it can print at "full" size on a single sheet.

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+## this plot as NULL
+github_contributions_plot <- NULL
+
+## TODO: Re-run or load in plots of the project features and assays -- for now
+## we will define these plots as NULL
+project_assays_plot <- NULL
+project_features_plot <- NULL
+
+#### Assemble multipanel plot -------------------------------------------------
+ggpubr::ggarrange(ggpubr::ggarrange(treemap,
```
jashapiro: I'm going to suggest that you try using `patchwork` here? I think it is perhaps a bit simpler in syntax?

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+  ggplot2::labs(x = "Cancer Types", y = "Count",
+                title = "Sample Distribution Across Cancer Types") +
+  ggplot2::scale_y_continuous(breaks = seq(0, 500, by = 100)) +
+  ggplot2::theme(axis.text.x = ggplot2::element_text(
+    angle = 75,
+    hjust = 1,
+    size = 8
+  ),
+  panel.grid = ggplot2::element_blank()) +
+  ggplot2::geom_text(nudge_y = 5.5, size = 1,
```
jashapiro: I'm a little worried this text is too small.

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,142 @@
...
+  readr::read_tsv(file.path(sample_distribution_dir, "results", "plots_df.tsv"))
+
+# Reorder the columns to be displayed in descending order by count on the plot
+disease_expression$integrated_diagnosis <- with(disease_expression,
+                                                reorder(integrated_diagnosis, -count))
+
+#### Re-run the individual plots ----------------------------------------------
+
+# Create a bar plot of sample distribution across cancer types
+gg_types_bar <- disease_expression %>%
```
jashapiro: Is this output anywhere?

analyses/sample-distribution-analysis/02-multilayer-plots.R
```diff
@@ -60,7 +66,54 @@ final_df <- histologies_df %>%
...
+                    place = "centre",
+                    grow = F,
+                    reflow=T) +
+  geom_treemap_subgroup_text(place = "top",
+                             grow = T,
+                             reflow = T,
+                             alpha = 0.6,
+                             colour = "#FAFAFA",
+                             min.size = 0) +
+  geom_treemap_subgroup2_text(place = "centre",
```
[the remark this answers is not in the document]
cbethell: The large labels here are broad histology and short histology, the smaller labels are the `integrated_diagnosis` values. Would we only want to keep broad histology and integrated diagnosis values, or should I keep broad and short histology values and look into making the representation of the integrated diagnosis text less redundant?

figures/scripts/fig1-sample-distribution.R
cbethell: No, I can take it out now I believe. It was originally in the panel but I believe we decided to go with a treemap instead of the barplot.

cbethell:
Here is the figure after making the changes mentioned above: [fig1-openpbta-distribution.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4354649/fig1-openpbta-distribution.pdf)

cbethell:
@jashapiro, what are your thoughts on the figure in the [above comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/633#issuecomment-601165392), now that changes have been made based on your previous review comments?

figures/scripts/fig1-sample-distribution.R
jashapiro: I'm wrong on this... manubot doesn't like pdf, so back to png.

jashapiro:
That looks better. I don't know if I understand what the individual blocks are though. I would have thought they were samples, but then why are there different sizes?

It seems like the broad histology may be a bit too redundant  with the other info, except for HGAT, so maybe we should just use the `integrated_diagnosis` for labels (keep broad fro grouping). I am partly reacting to the overlapping text in HGAT and ATRT, but more to the duplication of information..

cbethell:
> That looks better. I don't know if I understand what the individual blocks are though. I would have thought they were samples, but then why are there different sizes?
> 
> It seems like the broad histology may be a bit too redundant with the other info, except for HGAT, so maybe we should just use the `integrated_diagnosis` for labels (keep broad fro grouping). I am partly reacting to the overlapping text in HGAT and ATRT, but more to the duplication of information..

@jashapiro the `short_histology` and `integrated_diagnosis` fields are currently being used to represent data in that figure. 

The size of the individual blocks represent the number of samples in each specific histology and within that represents the number of samples in each `integrated_diagnosis` field within each histology. Does that make sense?

As for the overlapping of text, I can spend more time with the package documentation to get rid of this (it seems to be a default feature of the function that I have not yet figured out how to manually change).

Otherwise, does the figure appear to be what we would want in the figure 1 panel? (As opposed to the other plots, namely the treemap and the barplot, in this PR's [original comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/633#issue-389361947))

jashapiro:
> The size of the individual blocks represent the number of samples in each specific histology and within that represents the number of samples in each `integrated_diagnosis` field within each histology. Does that make sense?

Not really! I'm looking, for example, at the upper right of LGAT... some of the blocks there are larger than the others, but I can't understand why that would be. And I'm not sure why the blocks in LGAT are larger than the ones in DNET, for example. It feels like there may be some way the sample numbers are being doubly weighted in the figure, which might throw off the proportions. We don't really have more than half LGAT (do we?), but that is how the figure looks.

cbethell:
> Not really! I'm looking, for example, at the upper right of LGAT... some of the blocks there are larger than the others, but I can't understand why that would be. And I'm not sure why the blocks in LGAT are larger than the ones in DNET, for example. It feels like there may be some way the sample numbers are being doubly weighted in the figure, which might throw off the proportions. We don't really have more than half LGAT (do we?), but that is how the figure looks.

@jashapiro ah okay, I understand what you are getting at now. No we do not have more than half LGAT (LGAT does however represent about 25% of the cohort). I will look at the data.frame being used again and more into the function documentation to see how I can fix this. 

Will tag you with an update once I've done that.

jaclyn-taroni:
If there's a possibility of double-counting happening, my gut reaction is to check that normal WGS samples are not being counted = double counting tumor-normal pairs.

cbethell:
> If there's a possibility of double-counting happening, my gut reaction is to check that normal WGS samples are not being counted = double counting tumor-normal pairs.

I'll check for this as well, but I believe I found the issue to be that I was using the data.frame that was formatted for use with the treemap function that was previously implemented(that data.frame was redundant as required by the function).

jashapiro:
Looked at the [current figure](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e50a86ca2b71f895d61cfd00d16b40d78958ae38/analyses/sample-distribution-analysis/plots/distribution_across_cancer_types_treemap.pdf) , and the proportions look correct, but now blocks aren't grouping by broad_histology? For example, the HGAT tumor types are scattered about...

cbethell:
> Looked at the [current figure](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e50a86ca2b71f895d61cfd00d16b40d78958ae38/analyses/sample-distribution-analysis/plots/distribution_across_cancer_types_treemap.pdf) , and the proportions look correct, but now blocks aren't grouping by broad_histology? For example, the HGAT tumor types are scattered about...

You are correct. If we want to group by `broad_histology`, we would have to include the `broad_histology` values on the treemap. That plot looks a lot noisier (as we've previously determined). Here's that plot: 
[fig1-openpbta-distribution.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4376087/fig1-openpbta-distribution.pdf)

However, I was able to group by `short_histology` and this plot looks a bit cleaner than the plot that includes `broad_histology` (which is expected since it shows less information). 
[fig1-openpbta-distribution.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4376103/fig1-openpbta-distribution.pdf)

Should I maybe increase the `alpha` value for the `broad_histology` plot?

jashapiro:
Okay! That second plot looks good, but should we be using broad_histology **instead** of short_histology there? Are they the same info, just coded differently?

cbethell:
> Okay! That second plot looks good, but should we be using broad_histology **instead** of short_histology there? Are they the same info, just coded differently?

They are not the same info, but they are very similar. However, for example, the broad histology "embryonal tumor" consists of the short histologies "medulloblastoma", "ATRT", etc. While the broad histology "Low-grade astrocytotic tumor" only consists of the short histology "LGAT", in this case this would be representing the same info with different coding.

Here is the plot using `broad_histology` instead of `short_histology`: 
[fig1-openpbta-distribution.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4376490/fig1-openpbta-distribution.pdf)

jashapiro:
I don't really know which of those is better for this plot; I think I would lean toward broad_histology, but I have not particularly strong feelings. 

Text size and overlaps I think is a still challenge. I have a sense that  the max size for the integrated subtypes is too large, as I would want it to be smaller than the broad_histology in general. It may also reduce overlaps if broad is in the lower left and integrated is in the upper right? That is just a guess... There may have to be some fiddling to get it to look best. 

I guess the next question is coloring etc. We probably want to add in the standardized color palette, but maybe that should be a separate PR? This plot is probably a good one to use to refine the color palette as defined in #510 and #622; in particular we may be able to give types with the same broad_histology more similar colors to add some semantic value to those choices, as well as to make this plot more visually appealing.

cbethell:
> I don't really know which of those is better for this plot; I think I would lean toward broad_histology, but I have not particularly strong feelings.
> 
> Text size and overlaps I think is a still challenge. I have a sense that the max size for the integrated subtypes is too large, as I would want it to be smaller than the broad_histology in general. It may also reduce overlaps if broad is in the lower left and integrated is in the upper right? That is just a guess... There may have to be some fiddling to get it to look best.
> 
> I guess the next question is coloring etc. We probably want to add in the standardized color palette, but maybe that should be a separate PR? This plot is probably a good one to use to refine the color palette as defined in #510 and #622; in particular we may be able to give types with the same broad_histology more similar colors to add some semantic value to those choices, as well as to make this plot more visually appealing.

Okay I'll fiddle with it some more to try to get rid of the overlapping. 
I also already implemented the color palette for histology from #510 in these plots.

analyses/sample-distribution-analysis/02-multilayer-plots.R
```diff
@@ -60,7 +66,63 @@ final_df <- histologies_df %>%
...
+  ) +
+  geom_treemap_subgroup_text(
+    place = "bottomright",
+    grow = T,
+    reflow = T,
+    alpha = 0.6,
+    colour = "#FAFAFA",
+    min.size = 0
+  ) +
+  theme(legend.position = "none")
```
jashapiro: Ah, I see. You do have the colors specified here, but `ggplot` uses them as factors and then assigns its own colors. To get it to use the imported colors we need to add `scale_fill_identity()`
```suggestion
  theme(legend.position = "none") +
  scale_fill_identity()
```

cbethell:
Okay I added the `scale_fill_identity()` argument, decreased the labels' size and adjusted the positions to address the overlapping issue.

What are your thoughts now on the updated plot: 
[fig1-openpbta-distribution.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4376850/fig1-openpbta-distribution.pdf)

jashapiro:
Looks good! I would guess there could still be tweaks on the text, but I don't know what is best there. 

As to colors and #510 , I want to tag in @cansavvy and @dvenprasad; I think this plot can be used to adjust color order for `histology_color_palette.tsv`. (In a separate PR!) The current colors are also defined by short_histology, which ends up with some odd groupings compared to broad_histology, mostly due to the dispersion of `Other` categories.

figures/scripts/fig1-sample-distribution.R
```diff
@@ -0,0 +1,130 @@
...
+project_features_plot <- NULL
+
+#### Assemble multipanel plot -------------------------------------------------
+
+# Combine plots with patchwork
+# Layout of the four plots will be two over the other two
+# (2 columns and 2 rows)
+combined_plot <- treemap + project_features_plot +
+  project_assays_plot + github_contributions_plot +
+  plot_layout(ncol = 2, nrow = 2) &
```
jashapiro: Add letters to the subplots.
```suggestion
  plot_layout(ncol = 2, nrow = 2) +
  plot_annotation(tag_levels = 'A') &
```

PR #639, opened 2020-03-19
Add publication quality Cooccurrence plot to figure generation scripts

jashapiro:

The goal of this PR is to add another publication quality figure (not necessarily with publication quality results) to the figures/ directory and figure generation scripts. 

This partially fulfills goals of #571 

#### What was your approach?

I did some minor refactoring of the figure generation scripts, including a name change.

For the cooccurrence plot, there are some further minor changes, notably excluding the to 50 commonly mutated genes as defined by the FLAGS data set. 

#571, in part

Are the changes to the main figure generation script reasonable/acceptable?

Is the choice of 50 FLAGs genes to exclude reasonable, or should there be a different list of genes excluded?

Colors probably still need to be updated with the common color pallete in #622 

Yes

#### Documentation Checklist


figures/generate-figures.sh
```diff
@@ -3,26 +3,39 @@
...
 
-## Interaction plots 
+######################
+## Interaction plots
+
+# Run the main figure generation script
+bash ${analyses_dir}/interaction-plots/01-create-interaction-plots.sh
+
+# Copy the main figure to final directory
+cp ${analyses_dir}/interaction-plots/plots/combined_top50.pdf pdfs/mutation_cooccurrence_figure.pdf
```
jaclyn-taroni: For manubot purposes, I think we may want a PNG copy as well but not entirely sure. From the [Usage documentation about figures](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/USAGE.md#figures):

> For images created by the manuscript authors that are hosted elsewhere on GitHub, we recommend using a [versioned](https://help.github.com/en/github/managing-files-in-a-repository/getting-permanent-links-to-files) GitHub URL to embed figures, thereby preserving exact image provenance. When embedding SVG images hosted on GitHub, it's necessary to append ?sanitize=true to the raw.githubusercontent.com URL.

We can move forward with the PDF and find out.

analyses/interaction-plots/01-create-interaction-plots.sh
```diff
@@ -49,6 +49,14 @@ if [ "$ALL" -gt "0" ]; then
 fi
 
 
+# Get FLAG file and add header
```
jaclyn-taroni: The README for this module should get updated to include a citation for FLAGS, etc.

figures/generate-figures.sh
jashapiro: I think I will just change back to PNG. Sad to lose resolution, but if manubot isn't going to support it, we will just have to live with it.

Issue #486, opened 2020-01-29
Discussion: possible improvements for copy number processing / analyses 


jaclyn-taroni:
We at the CCDL have been thinking quite a bit about the copy number data this sprint (#479 #485 #480 #482 #463 #467 #476). Here I want to document some potential improvements that have come up during various discussions. I don't know if anything here will rise to the level of "must-have" and it may also be more appropriate to split individual notes/points into their own, more fleshed out tickets.

* From a discussion with @jashapiro (please tell me if I've got this wrong 🙂 ) - anything that goes through [`copy_number_consensus_call`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/copy_number_consensus_call) with a copy number of 2 in CNVkit is essentially considered a neutral change because CNVkit doesn't take into account ploidy ([A Note on Ploidy](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#a-note-on-ploidy)). So it's possible in some cases we are underreporting losses (or gains if there are haploid samples, but we don't think this is the case). Here's the distribution of `tumor_ploidy` from v13 of `pbta-histologies.tsv` for reference:
```r
> histologies_df %>% 
+   filter(experimental_strategy == "WGS", 
+          sample_type == "Tumor") %>% 
+   group_by(tumor_ploidy) %>% 
+   tally()
# A tibble: 3 x 2
  tumor_ploidy     n
         <dbl> <int>
1            2   690
2            3   134
3            4   116
```
* We broadly call copy number changes, i.e., we'll only label an event as a loss without considering whether the loss is homozygous or hemizygous or how many copies are lost (see [A Note on Ploidy](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#a-note-on-ploidy) again) in the [`focal-cn-file-preparation`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/focal-cn-file-preparation) steps for SEG files (CNVkit, consensus SEG). In the consensus case at least, we can potentially use the `seg.mean` values to guide us with any kind of thresholding - see this section of the notebook for processing the consensus SEG file: https://alexslemonade.github.io/OpenPBTA-analysis/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.nb.html#does_segmean_agree_with_status

[...]

PR #644, opened 2020-03-23
Define most focal units of recurrent CNVs

cbethell:

The purpose of this PR is to define the _most focal_ units of recurrent CNVs. 

Related to PR #485, this PR is working to compile a table/list of the _most focal_ units of recurrent CNVs, _most focal_ meaning: 

- If a chromosome arm is not clearly defined as a gain or loss (and is callable) we look to define the cytoband level status
- If a cytoband is not clearly defined as a gain or loss (and is callable) we then look to define the gene level status

This PR is working to compile this information into one table.

#### What was your approach?

Using the file generated in PR #617:
- The chromosome arm statuses are determined by counting the amount of instances of each call on each chromosome arm and defining the arm status based on the status call that is responsible for more than 90% of the total arm calls
- If a chromosome arm is not clearly defined as a gain or loss (and is more than 50% callable), look to define the cytoband level status
- If a cytoband is not clearly defined as a gain or loss (and is more than 50% callable), look to define the gene level status
- Compile this information into one table

Related: #486

- The `running_in_ci` flag was placed around the gene level steps because the testing subset files throw an error that a certain status is not found (as the subset files do not have each status at least once at the gene level after filtering out arms and cytobands that are non-neutral).
**Is there possibly a more elegant solution here?**

- The threshold used to determine the most dominant status for each level should also receive a closer look.

Note that in this PR I also replace an `inner_join` in the `03-add-cytoband-status.Rmd` notebook, that was unintentially added in [this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/617/commits/e9bf7a78dc8cec5d81fcef990fbea5fed4d155bc) of PR #617 with `left_join` (as was originally implemented and approved) -- @jashapiro you may want to take a look at this and the resulting file to make sure its still on par with what you approved.

A final table of results can be found at `results/consensus_seg_most_focal_cn_status.tsv.gz`

**The rendered html output of the notebook in this PR as it is currently can be seen [here](https://cbethell.github.io/open-pbta-output/05-define-most-focal-cn-units.nb.html).**

#### Documentation Checklist


analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+  select(-c("uncallable", "unstable")) %>%
+  replace_na(list(
+    gain = 0,
+    loss = 0,
+    neutral = 0
+  )) %>%
+  group_by(chromosome_arm) %>%
+  # The logic here is to define the arm status based on the majority of calls
+  # in the chromosome arm -- if the number of calls for a specific status is
+  # responsible for more than 50% of the total calls in that arm, then it is
```
cansavvy: Technically, this is calling things when it is 49% or higher. e.g. if something is. 49.1% it will be called. I'd like to know how close these percentages are when something is called.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+  # used to define the arm status
+  mutate(
+    arm_dominant = case_when(
+      gain > (0.49 * (gain + loss + neutral)) ~ "arm_gain",
+      loss > (0.49 * (gain + loss + neutral)) ~ "arm_loss",
```
cansavvy: I don't know if this happens at all, but technically you could have two calls could have numbers between .49 and .50 and both would be `TRUE`, in which case, does `case_when` go in order? Also does this ever happen? I haven't looked at the data enough to know.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+  spread(status, n) %>%
+  replace_na(list(
+    gain = 0,
+    loss = 0,
+    amplification = 0
+  )) %>%
+  group_by(cytoband) %>%
+  # The logic here is to define the arm status based on the majority of calls
+  # in the cytoband -- if the number of calls for a specific status is
+  # responsible for more than 50% of the total calls in that cytoband, then it is
```
cansavvy: Same comments here as above.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+consensus_seg_autosomes_df_develop <- consensus_seg_autosomes_df %>%
+  select(biospecimen_id, gene_symbol, cytoband, status) %>%
+  mutate(chromosome_arm = gsub("(p|q).*", "\\1", cytoband)) %>%
+  filter(chromosome_arm %in% neutral_arm_status_df$chromosome_arm)
+
+# Now count the cytoband level calls (for each status call) and define
+# the cytoband as that status if more than 50% of the total counts are
+# for that particular status
+cytoband_status_count <- consensus_seg_autosomes_df_develop %>%
+  count(cytoband, status) %>%
```
cansavvy: Looks like this section is the same as lines 67 - 89,  may want to make it a function? But we should hammer down the logic for majority calls.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+
+if (!(running_in_ci)) {
+# Now count the gene level calls (for each status call) and define
+# the gene as that status if more than 50% of the total counts are
+# for that particular status
+gene_status_count <- consensus_seg_autosomes_df_develop %>%
+  count(gene_symbol, status) %>%
+  # Spread the data -- each row represents a unique chromosome arm
+  spread(status, n) %>%
+  replace_na(list(
```
cansavvy: Now since this thing is happening a third time, I think a function may be in order.

analyses/focal-cn-file-preparation/run-prepare-cn.sh
```diff
@@ -44,58 +47,61 @@ Rscript --vanilla 04-prepare-cn-file.R \
...
-#           --independent_specimens_file $independent_specimens_file \
-#           --metadata $histologies_file \
-#           --goi_list $goi_file \
-#           --filename_lead ${filename}_${chromosome_type}_${strategy}
-#       done
-#     done
-#   done
-# 
-# fi
+# Define most focal units of recurrent CNVs
```
cansavvy: Why was this commented out previously?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+
+# Display table
+gene_status_count
+}
+```
+
+## Combine arm, cytoband, and gene level status data
+
+```{r}
+if (!(running_in_ci)) {
```
cansavvy: Can you add a comment about why this can't be run in CI?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Can call it something like `status_majority_caller` or something else.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,213 @@
...
+  html_notebook:
+    toc: TRUE
+    toc_float: TRUE
+author: Chante Bethell for ALSF CCDL
+date: 2020
+params:
+  is_ci: FALSE
+---
+
+This notebook defines the most focal recurrent copy number units by separating out entire chromsome arm losses and gains.
```
cansavvy: Can you add your nice explanation that you have on your PR, here? It really is nice and sets up the reader well for what this notebook is doing.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: The idea here was really to call an arm a loss/gain/neutral if that status is responsible for more than half of the calls in the arm.

_However_, the 0.49 value was included when I noticed that a lot of neutral dominant arm statuses and cytoband loss statuses (no gains) were in the results table.

That being said, I should probably raise the threshold back up to 0.5 (or higher) as I now notice instances where rows that should be neutral (same amount of gain and loss calls) are not called as neutral (with this logic).

I can wait to do this however, after I've calculated percentages for each call and we've determined an appropriate logic to make these calls.

Does this kind of answer your question?

analyses/focal-cn-file-preparation/run-prepare-cn.sh
cbethell: For development reasons, it is still relevant for the production of the files that were added to this module previous to this PR.

cbethell:
> To ask these questions you may want to calculate percentages first, and then in a separate step where you make the call.
> You could draft up some quick and dirty plots that shows the percent of each status and separate on the x-axis what the final call was. Ideally if we are calling things right, we should see that there are hardly ever "ties" but there's a clear differentiation between gains and losses.

I will implement these suggestions in the upcoming commits and post an updated comment here once I've summarized the findings.

cbethell:
@cansavvy here is the [rendered notebook](https://cbethell.github.io/open-pbta-output/05-define-most-focal-cn-units.nb.html) of my first go at implementing the suggestions you made above.

You can let me know if that's what you were thinking the plots should look like or if I should tweak them (or if I should tweak anything else at this point).

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Well if you are looking for more than half, I think it makes sense to keep the logic consistent with that, so > 0.5 , partially so people don't get confused, but partially so that if there is an instance where the 0.01 difference matters, you are still all set.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+
+  if (region_variable == "chromosome_arm") {
+    status_count <- status_df %>%
+      count(!!region_sym, dominant_status) %>%
+      # Spread the data -- each row represents a unique chromosome arm
+      spread(dominant_status, n) %>%
+      replace_na(list(
+        gain = 0,
+        loss = 0,
+        neutral = 0
```
cansavvy: I think this section can be the same regardless if you add: 
```suggestion
        neutral = 0, 
        amplification = 0
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+  # Return:
+  #   status_count: data.frame with percentage values for each unique status in
+  #                 each unique region (arm/cytoband/gene) and the dominant
+  #                 status for that region
+
+  region_sym <- rlang::sym(region_variable)
+
+  if (region_variable == "chromosome_arm") {
+    status_count <- status_df %>%
+      count(!!region_sym, dominant_status) %>%
```
cansavvy: Am I understanding this that you want this to use `dominant_status` for `region_variable = chromosome_arm`, but should use `status` if `region_variable = cytoband` or `gene_symbol`?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+      )) %>%
+      group_by(!!region_sym) %>%
+      # The logic here is to define the arm status based on the majority of calls
+      # in the region -- if the number of calls for a specific status is
+      # responsible for more than 50% of the total calls in that region, then it is
+      # used to define the region's status
+      mutate(
+        percent_gain = gain / (gain + loss + amplification),
+        percent_loss = loss / (gain + loss + amplification),
+        percent_amplification = amplification / (gain + loss + amplification),
```
cansavvy: Looks like the only difference here is that its reporting `amplification` instead of `neutral`?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+            percent_gain > 1.5 * percent_loss ~ "gain",
+          percent_loss > 0.5 &
+            percent_loss > 1.5 * percent_gain ~ "loss",
+          percent_neutral > 0.5 &
+            percent_neutral > 1.5 * percent_loss &
+            percent_neutral > 1.5 * percent_gain ~ "neutral",
+          TRUE ~ "none"
+        )
+      )
+  } else {
```
cansavvy: I would advise making the else more specific, because if for some reason region_variable isn't `chromosome_arm`, `cytoband` or `gene_symbol` this would still run anyway and we don't necessarily want that?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: I think we can still shorten up this section.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+      # in the region -- if the number of calls for a specific status is
+      # responsible for more than 50% of the total calls in that region, then it is
+      # used to define the region's status
+      mutate(
+        percent_gain = gain / (gain + loss + neutral),
+        percent_loss = loss / (gain + loss + neutral),
+        percent_neutral = neutral / (gain + loss + neutral),
+        dominant_status = case_when(
+          percent_gain > 0.5 &
+            percent_gain > 1.5 * percent_loss ~ "gain",
```
cansavvy: Can you explain these logic statements? I'm not saying they aren't what we want, but I wanna understand the changes you made here and what you are intending.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Let me know if you wanna pair code on this. I have some ideas on this, but might be easiest to pair up on it.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+      )) %>%
+      group_by(!!region_sym) %>%
+      # The logic here is to define the arm status based on the majority of calls
+      # in the region -- if the number of calls for a specific status is
+      # responsible for more than 50% of the total calls in that region, then it is
+      # used to define the region's status
+      mutate(
+        percent_gain = gain / (gain + loss + neutral),
+        percent_loss = loss / (gain + loss + neutral),
+        percent_neutral = neutral / (gain + loss + neutral),
```
cansavvy: I think we should also have a step where we exclude making calls on regions that have a certain percentage uncallable.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+
+# Display table
+cytoband_status_count
+```
+
+### Now we want to use the annotated CN data to define the gene level gains/losses 
+
+```{r}
+# Filter the annotated CN data to include only these cytobands
+consensus_seg_autosomes_df_develop <- consensus_seg_autosomes_df %>%
```
cansavvy: It looks like you are reusing the name `consensus_seg_autosomes_df_develop`a few times. This is a tad hard to follow. Can you make them different things? We should probably see if we can streamline this some, or add this formatting step to the main function.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+  status_sym <- rlang::sym(status_variable)
+
+  status_plot <- ggplot2::ggplot(
+    status_count_df,
+    ggplot2::aes(
+      x = dominant_status,
+      y = !!status_sym
+    )
+  ) +
+    ggplot2::geom_point() +
```
cansavvy: Can you make this a jitter or a sina plot so the points get a separated from each other.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+arm_loss_status_plot
+
+arm_neutral_status_plot <- plot_dominant_status_calls(
+  arm_status_count,
+  "percent_neutral"
+)
+arm_neutral_status_plot
+```
+
+### Plot cytoband status calls
```
cansavvy: You could shorten up these sections by making these facet wrap plots.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,357 @@
...
+```{r}
+# The logic variable `running_in_ci` is needed here because the CI testing
+# files do not contain any of the genes in the `gene_status_count` data.frame
+# generated above (when `running_in_ci` == FALSE)
+if (!(running_in_ci)) {
+  final_df <- consensus_seg_autosomes_df %>%
+    mutate(chromosome_arm = gsub("(p|q).*", "\\1", cytoband)) %>%
+    left_join(arm_status_count, by = "chromosome_arm") %>%
+    left_join(cytoband_status_count, by = "cytoband") %>%
+    left_join(gene_status_count, by = "gene_symbol")
```
cansavvy: Can you use a `suffix` argument here and have them labeled with their origins? right now it looks like they have the default `.x` and `.y`s.

cansavvy: We may also want to have a final column that summarizes which status we should be listening to. Aka if `arm_status` is declared, report that, then `cytoband`, then `gene_symbol`. But, we may want some more information about the downstream applications for this for knowing on how this final table should be formatted.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: Yes, that is correct. The data.frame that `cytoband` and `gene_symbol` data are retrieved from has the column name `status` and not `dominant_status`.

However, now thinking about how this can be refactored, I can just rename the data.frame column name to be the same across data.frames.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: I am sure that I probably went about implementing this logic is the least efficient way (it's probably not doing what I intended), but the idea here is that we want the status percentage to be responsible for more than 50% of calls in that region _and_ to be greater than 1.5 times the alternate status call.

This was an attempt to account for a loss/gain dominant status call being made when there is just slim difference between calls.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: Sure we pair on this. Would you prefer that I address your other comments and push those changes first? Or woyld you rather pair on it sooner?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: If you go with the renaming route, I would keep that renaming just to the function, but a preferred solution might be to just specify another tidyeval so that you don't lose this column name information.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Go ahead and address the comments I have and then maybe we should schedule a time to try to rework this main function.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: Ahh okay, I'll give the second solution mentioned here a try first.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Ah. Okay. Well as a more general statement, I think whatever the logic is that we have for deciding majority status we should probably these methods pretty upfront laid out in documentation (probably at the top of the notebook and READMEs as well as comment). Not only what the methods are for the majority status call decisions but also what was our reasoning behind that.

cbethell: Gotcha, will do.

cansavvy: That being said, I think we are getting closer, but there are two things that we may want to think about when calling majority status (and these are related). 
1) Not calling a region if too much of it is uncallable
2) Not calling a region if the absolute number of calls are too small (related to 1) e.g. 3 calls vs 2 calls would technically be greater than 50% but we probably don't have enough data to make that call. 

But lastly, I'm wondering whether we need a > 0.5 cutoff in addition to a 1.5x cutoff or if raising the threshold of the percent cutoff might establish the same thing?

cbethell: > But lastly, I'm wondering whether we need a > 0.5 cutoff in addition to a 1.5x cutoff or if raising the threshold of the percent cutoff might establish the same thing?

Good question, I'll test this out and include my findings in the notebook (in addition to the points you made above).

cansavvy: No need do an entire side analysis but, my point is just that whatever decision logic we come up with we should make sure we are clear about what it is and why we think this gives us a reasonable answer.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,415 @@
...
+  #   status_count_df: data.frame with percentage values for each unique status
+  #                    in each unique region (arm/cytoband/gene) and the
+  #                    dominant status for that region
+  #   region_variable: string of the region (which will also be a column name)
+  #                    that the data.frame holds percentage values for
+  # Return:
+  #   status_plot: plot representing the dominant status call on the x-axis and
+  #                the percentage values on the y-axis
+  
+  status_count_df %>%
```
cansavvy: I know I sent you this code, but can you add comments at these steps?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,415 @@
...
+  "status", 
+  percent_threshold = percent_threshold
+)
+
+# Display table
+arm_status_count %>%
+  group_by(dominant_status)
+```
+
+These are the chromosome arms that have not been defined as gain or loss -- we want to define their cytoband/gene level status
```
cansavvy:
```suggestion
These are the chromosome arms that have not been defined as gain or loss -- we want to define their cytoband/gene-level status
```
Figure we should be consistent with the `-` (This was probably me who left this)

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,415 @@
...
+# Run `plot_dominant_status_calls` function for the chromosome arm calls
+plot_dominant_status_calls(
+  arm_status_count,
+  "chromosome_arm"
+)
+```
+
+### Plot cytoband status calls
+
+```{r fig.width = 10}
```
cansavvy: Don't need this `fig.width` anymore.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,415 @@
...
+        ", ",
+        dominant_status.cytoband,
+        ", ",
+        dominant_status
+      ),
+      # Here we want to define the most focal call based on the arm, cytoband,
+      # and gene status information -- if a loss/gain is defined at the arm
+      # level then the focal call will be "arm_loss" or "arm_gain" respectively,
+      # and so on.
+      focal_call = case_when(focal_call == "loss, NA, NA" ~ "arm_loss",
```
cansavvy: I think we need to rework this last section here. In part because of something doesn't fall underneath the categories you lay out here then it automatically gets labeled as "Other" which is not a super helpful label.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,415 @@
...
+                             focal_call == "neutral, uncallable, NA" ~ "uncallable",
+                             focal_call == "neutral, loss, NA" ~ "cytoband_loss",
+                             focal_call == "neutral, NA, NA" ~ "arm_neutral",
+                             focal_call == "neutral, unstable, neutral" ~ "gene_neutral",
+                             focal_call == "neutral, unstable, loss" ~ "gene_loss",
+                             TRUE ~ "Other")
+    )
+} else {
+  final_df <- consensus_seg_autosomes_df %>%
```
cansavvy: So in CI it won't do the recoding?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: I added this because these particular plots have the two additional labels on the x-axis so without the `fig.width` argument the labels start to overlap and become hard to read.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: Oops, I'll add the recoding here.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: Oh Sorry, I misread this chunk. It still has a plot. Nevermind!

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cansavvy: I think we might refactor this section once we determine what `final_df` needs to look like, so you can hold off for now.

cbethell: Okay sounds good

cansavvy:
@cbethell , can we reformat the `final_df` table to follow the format of @jaclyn-taroni 's set up from this branch: 

https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/resolve-cytoband-arm/analyses/focal-cn-file-preparation/results/consensus_autosomes_largest_region_tidy.tsv.gz

It would look like this and neutrals and uncallables would be excluded from the final list: 
Kids_First_Biospecimen_ID | region | status
-- | -- | --
BS_007JTNB8 | 10q | loss
BS_00TRPEQX | 17p | gain
...| ... | ..

cbethell:
@cansavvy, I have reformatted the final table as you mentioned in the [above comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/644#issuecomment-606066189) and also added a `region_type` column variable to the final table to make filtering by region easier. 

The updated notebook can be seen [here](https://cbethell.github.io/open-pbta-output/05-define-most-focal-cn-units.nb.html).

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,515 @@
...
+    mutate(chromosome_arm = gsub("(p|q).*", "\\1", cytoband)) %>%
+    inner_join(arm_status_count,
+               by = "chromosome_arm") %>%
+    left_join(cytoband_status_count,
+              by = "cytoband",
+              suffix = c(".arm", ".cytoband")) %>%
+    left_join(gene_status_count,
+              by = "gene_symbol",
+              suffix = c(".arm", ".gene")) %>%
+    mutate(
```
cansavvy: If I'm understanding this correctly, I think you can remove lines 373 - 393 since we are not using `focal_call` variable anymore.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,515 @@
...
+                             focal_call == "neutral, NA, NA" ~ "arm_neutral",
+                             focal_call == "neutral, unstable, neutral" ~ "gene_neutral",
+                             focal_call == "neutral, unstable, loss" ~ "gene_loss",
+                             TRUE ~ "Other")
+    )
+} else {
+  combined_status_count_df <- consensus_seg_autosomes_df %>%
+    mutate(chromosome_arm = gsub("(p|q).*", "\\1", cytoband)) %>%
+    inner_join(arm_status_count, by = "chromosome_arm") %>%
+    mutate(focal_call = dominant_status)
```
cansavvy: You'll be able to get rid of this line too.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,515 @@
...
+}
+
+# Display combined status counts table
```
cansavvy: I think 400- 404 can also be removed? Or at least reworked.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,515 @@
...
+  final_arm_status_df <- combined_status_count_df %>%
+    # Filter to only non-neutral chromosome arms
+    filter(dominant_status != "neutral") %>%
+    select(
+      Kids_First_Biospecimen_ID = biospecimen_id,
+      region = chromosome_arm,
+      status = dominant_status
+    ) %>%
+    distinct() %>%
+    mutate(region_type = "chromosome_arm")
```
cansavvy: You can get rid of these mutate steps, if at your `bind_rows` step at 479, you use `.id` argument.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,515 @@
...
+```
+
+## Combine arm, cytoband, and gene-level status data
+
+```{r}
+# The logic variable `running_in_ci` is needed here because the CI testing
+# files do not contain any of the genes in the `gene_status_count` data.frame
+# generated above (when `running_in_ci` == FALSE)
+if (!(running_in_ci)) {
+  combined_status_count_df <- consensus_seg_autosomes_df %>%
```
cansavvy: Can you explain why you are joining this to `consensus_seg_autosomes_df`? I think we get to what we need in the final table without it unless I am missing something.

cbethell: This was for the purpose of retaining sample ID values.

cansavvy: Retaining? That means the sample ID values are there already right?

cbethell: Sorry, retrieving*. No they are not already there. This may be something that needs to be fixed in the function.

cansavvy: Yes. I'm thinking so. We need this to be a by-sample analysis. Aka if in given sample it's called at the arm > cytoband > gene level, then we go with the broadest level call. Sorry I did not catch this before. Should not be too bad to fix, just need to make sure that `status_majority_caller` carries over the biospecimen IDs.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,445 @@
...
+      )
+  }
+
+  return(status_percent_df)
+}
+```
+
+```{r}
+plot_dominant_status_calls <- function(status_count_df) {
+  # TODO: Update this documentation to reflect the select `starts_with` stategy.
```
cansavvy, line 162: Don't forget to update this documentation! @cbethell

analyses/focal-cn-file-preparation/README.md
```diff
@@ -45,6 +45,15 @@ See the notebook for more information. This notebook also prepares lists of copy
...
+  
+* `05-define-most-focal-cn-units.Rmd` - This notebook defines the _most focal_ recurrent copy number units by removing focal changes that are within entire chromosome arm losses and gains.
+_Most focal_ here meaning if a chromosome arm is not clearly defined as a gain or loss (and is callable) we look to define the cytoband level status.
+Similarly, if a cytoband is not clearly defined as a gain or loss (and is callable) we then look to define the gene level status.
+To make these calls, the following decisions around cutoffs were made:
+	- The percentage of calls a particular status needs to be above to be called the majority status is currently 	90%.
+	This decision was made to ensure that the status is not only the majority status but is also 	significantly called more than the other status values in the region.
+	- The percentage of a region (arm, cytoband, or gene) that should be callable is more than 50%.
+	This decision was made because it seems reasonable to expect a region to be more than 50% callable for a 	dominant status call to be made.
```
jashapiro, line 56: These two cutoffs seem quite reasonable to me (as a starting point, at least).

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,445 @@
...
+  # find the dominant status of the cytoband/gene region by calculating
+  # the percentages of the region that each call represents.
+  #
+  # Args:
+  #   status_df: data.frame with cytoband/gene-level copy number status data
+  #   region_variable: string of the column name/region to calculate copy number
+  #                    status percentages for
+  #   status_column_name: string of the column name that holds the relevant copy
+  #                       number status data
+  #   id_column_name: string of the column name that holds the sample id values
```
jashapiro, line 72: Just removing a bit of wordiness. 
```suggestion
  #   region_variable: the column name/region to calculate copy number
  #                    status percentages for
  #   status_column_name: the column name that holds the relevant copy
  #                       number status data
  #   id_column_name: the column name that holds the sample id values
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,445 @@
...
+```{r}
+status_majority_caller <- function(status_df,
+                                   region_variable,
+                                   status_column_name,
+                                   id_column_name,
+                                   percent_threshold_val,
+                                   uncallable_threshold_val = uncallable_threshold) {
+  # Given a data.frame with cytoband/gene-level copy number status data,
+  # find the dominant status of the cytoband/gene region by calculating
+  # the percentages of the region that each call represents.
```
jashapiro, line 64: I find this description confusing. Can you put this in simpler terms?

cansavvy:
I like the weighted mean approach, right now we don't have any sort of accountability for differences in length, so I think that sounds like a good way to do it. 

I am a bit confused by the meaning of this paragraph: 
> Right now, it appears that calculations are being done based on the gene-level calls, with counts translated up to arms, but we have data for the cytobands which seems like a much more natural place to look for these high-level calls.

Do you think you could point out in the code what you are referring to specifically? 

> One disadvantage compared to your current approach is that it loses the "amplification" calls...

Should the amplifications calls be recoded to `gain`s and this is resolved? Or is this something else that needs to be done?

jashapiro:
>  Yes. I think you are right. I imagine it might be more exact to move to a percent length based assessment? Or is that not necessary and using weighted means is sufficient?

Weighted mean of call fraction and length will give the correct fractions for the total.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+
+We want to include cytoband and gene-level calls for chromosome arms that have not been defined as a gain or loss to make the cytoband-level majority calls.
+
+```{r}
+# Now count the cytoband level calls (for each status call) and define
+# the cytoband as that status if more than 90% of the total counts are
+# for that particular status
+cytoband_status_count <- consensus_seg_cytoband_status_df %>%
+  # Filter the annotated CN data to include only neutral chromosome arms
+  filter(chromosome_arm %in% neutral_arms) %>%
```
jashapiro, line 224: I realized that this may actually miss some important calls, like when a deletion occurs within an otherwise duplicated region. 

So think the approach here might have to be a bit more nuanced than just filtering on arms. I think we want to join to the arm df, then filter on the merged data, with an extra caveat...

Below is a suggestion, but note that I am not necessarily using the correct column names, so this will not work as is.

```suggestion
cytoband_status_count <- consensus_seg_cytoband_status_df %>%
  left_join(arm_status,  by = c("Kids_First_Biospecimen_ID", "chromosome_arm")
  # Filter the annotated CN data to include only neutral chromosome arms and disagreements
  filter(chromosome_arm %in% c("neutral", "uncallable", "unstable") |
         (dominant_cytoband_status != dominant_arm_status & # bands that disagree with arm, but are not neutral (or uncallable)
         !(dominant_cytoband_status %in% c("neutral", "uncallable", "unstable"))) %>%
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+  select(
+    Kids_First_Biospecimen_ID,
+    cytoband,
+    loss_fraction,
+    gain_fraction,
+    callable_fraction,
+    dominant_status = dominant_cytoband_status
+  ) %>%
+  mutate(
+    dominant_status = case_when(
```
jashapiro, line 234:
```suggestion
    dominant_cytoband_status = case_when(
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+
+### Determine chromosome arm status
+
+```{r}
+# Use our custom function to make the arm status calls
+arm_status_count <- status_majority_caller(
+  consensus_seg_cytoband_status_df,
+  "chromosome_arm",
+  status_threshold
+)
```
jashapiro, line 198: I think the clarity of what status has been called is useful here as above.
```suggestion
) %>% 
rename(dominant_arm_status = dominant_status)
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+  filter(chromosome_arm %in% neutral_arms) %>%
+  select(
+    Kids_First_Biospecimen_ID,
+    cytoband,
+    loss_fraction,
+    gain_fraction,
+    callable_fraction,
+    dominant_status = dominant_cytoband_status
+  ) %>%
+  mutate(
```
jashapiro, line 233: With the thought that we may want to keep places where the cytoband status disagrees with the arm status, this `mutate` step should be moved before the `filter` step (against my usual practice!)

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+```
+
+### Determine gene-level status
+
+```{r}
+# These are the cytobands that have not been defined as gain or loss --
+# we want to define their gene-level status
+neutral_cytobands <- cytoband_status_count %>%
+  filter(dominant_status %in% c("unstable", "neutral")) %>%
+  dplyr::pull("cytoband")
```
jashapiro, line 255: Similarly to above, I think we want to do this via a `left_join` and comparisons within each row. The reason being we need to preserve the relationship between samples and cytobands, not just keep all cytobands with at least one neutral call.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+```{r}
+if (!(running_in_ci)) {
+  # Now count the gene-level calls (for each status call) and define
+  # the gene as that status if more than 50% of the total counts are
+  # for that particular status
+  gene_status_count <- status_majority_caller(
+    combined_neutral_status_cytoband_df,
+    "gene_symbol",
+    status_threshold
+  )
```
jashapiro, line 278: I don't understand what this is meant to do. Each sample should only have one call for each gene, right?

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+#   # Run `plot_dominant_status_calls` function for the gene-level calls
+#   plot_dominant_status_calls(gene_status_count)
+# }
+```
+
+## Combine arm, cytoband, and gene-level status data
+
+```{r}
+# For each of the datasets we're joining, we'll only keep these columns:
+cols_to_keep <- c("Kids_First_Biospecimen_ID", "dominant_status")
```
jashapiro, line 320: If we do leave the upper dfs with `dominant_arm_status` and `dominant_cytoband_status`, this may need to change. I might have a small preference for leaving the more specific column names and changing only down here, but that is a small preference, so either way you feel comfortable.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+    select(
+      Kids_First_Biospecimen_ID,
+      status = dominant_status,
+      region,
+      region_type
+    )
+}
+
+# Print out preview
+final_df
```
jashapiro, line 373: I think this final output will include neutral calls (and possibly unstable and uncallable) for cytobands and arms, but not for genes (neutral calls were not included in `consensus_seg_annotated_cn_autosomes`). We probably don't want to include neutral calls here either, though I am not sure how we should treat unstable or uncallable regions.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
jashapiro: Note that when we join these genes to cytobands, some of the cytoband columns in `consensus_seg_cytoband_status_df` have more than one cytoband as genes can cross boundaries. We will need to add logic to fill those in as possible.

We will also have to be careful here to be sure the logic flows correctly, as we don't want to make mistakes in filtering at this point because of any filtering we might have done at the cytoband level! That is, if we filtered a cytoband because it was the same call as the arm that contains it, we still want to filter out the genes within that cytoband that have the same call!

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+# Use our custom function to make the arm status calls
+arm_status_count <- status_majority_caller(
+  consensus_seg_cytoband_status_df,
+  "chromosome_arm",
+  status_threshold
+)
+
+# Display table
+arm_status_count %>%
+  group_by(dominant_status)
```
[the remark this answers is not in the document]
cbethell: I was using it for easier viewing for the different calls in the rendered html format, but removed it in the last commit because it's not needed.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,394 @@
...
+```
+
+These are the chromosome arms that have not been defined as gain or loss -- we want to define their cytoband/gene-level status
+
+```{r}
+# Let's get a vector of the neutral arms
+neutral_arms <- arm_status_count %>%
+  filter(dominant_status == "neutral") %>%
```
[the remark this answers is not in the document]
cbethell: Gotcha, got rid of this section in the last commit.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: You are correct, it was redundant and I believe I resolved this in the last commit.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: In the last commit, I filtered out the neutral calls but left the unstable and uncallable calls until we decide how we should handle them.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: You are correct, it was redundant and I believe I resolved this in the last commit.

cbethell:
@jashapiro, I believe I addressed your most recent review comments [here in this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/644/commits/249884183b7416d6ddd9114c18908de5a0035259), with the exception of:
> We will also need to decide what to do with the "amplification" calls at the gene level; I tend to think we should demote them all to "gain" for consistency, but maybe have a separate column that denotes what might have been called an amplification? Ideally, we would tie things back to the seg.mean call, but I think that is probably beyond the scope here.
>
>I am also not quite sure what to do about the fact that the gene level calls were done with respect to ploidy, but the cytoband calls were not. We may have to revisit the cytoband calls to add this in.

Please let me know if I misinterpreted any of your comments or suggestions.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,272 @@
...
+      "biospecimen_id" = "Kids_First_Biospecimen_ID",
+      "cytoband_no_ranges" = "cytoband"
+    )
+  ) %>%
+  # Filter the annotated CN data to include only neutral cytobands and disagreements
+  filter(
+    dominant_cytoband_status %in% c("neutral", "uncallable", "unstable") |
+      (status != dominant_cytoband_status &
+         # bands that disagree with arm, but are not neutral (or uncallable)
+         !(status %in% c("neutral", "uncallable", "unstable")
```
jashapiro, line 192: There are no neutral calls at the gene level, so I think we can skip this part of the filter.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,272 @@
...
+    consensus_seg_cytoband_status_df,
+    by = c(
+      "biospecimen_id" = "Kids_First_Biospecimen_ID",
+      "cytoband_no_ranges" = "cytoband"
+    )
+  ) %>%
+  # Filter the annotated CN data to include only neutral cytobands and disagreements
+  filter(
+    dominant_cytoband_status %in% c("neutral", "uncallable", "unstable") |
+      (status != dominant_cytoband_status &
```
jashapiro, line 190: This will catch all of the places where the gene level call is "amplification" because we don't make that call at the cytoband level. I think the right thing to do is to change all "amplification" calls to "gain" for the purposes of this script and its output.  

I also have some thought that we should also be looking at arm-level calls at this point. I can imagine a case where an arm is a **gain** and a cytoband within it is **neutral** and then a gene within that band is a **gain**. We would not be including the cytoband in the final set (on the assumption it is a false negative), but by the current logic we would include the individual gene, but I don't think we should. The logic here may get a bit messy, but I think start is to be sure to join in the arm level calls as well, so we have both `dominant_cytoband_status` and `dominant_arm_status` to look at before making decisions.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,272 @@
...
+         # bands that disagree with arm, but are not neutral (or uncallable)
+         !(status %in% c("neutral", "uncallable", "unstable")
+      ))
+    ) %>%
+  select(
+    Kids_First_Biospecimen_ID = biospecimen_id,
+    gene_symbol,
+    loss_fraction,
+    gain_fraction,
+    callable_fraction,
```
jashapiro, line 200: These are cytoband level info, so we probably don't want these in this gene-level output.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,272 @@
...
+    status
+  )
```
jashapiro, line 202: I think the `distinct()` we need because of the band separation into multiple rows can go right here.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: In the most recent commit, I recoded the "amplification" calls to be "gain" calls as you suggested here.

I also _believe_ that I implemented the logic of including individual gene calls using both the `dominant_cytoband_status` and `dominant_arm_status` data.

Again, please let me know if I misinterpreted this comment in the way I re-formatted the code.

cbethell:
@jashapiro, this is ready for another look 👍

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,293 @@
...
+    by = c("biospecimen_id" = "Kids_First_Biospecimen_ID",
+           "cytoband")
+  ) %>%
+  # Filter the annotated CN data to include only neutral arms and cytobands,
+  # and disagreements
+  filter(
+    dominant_arm_status %in% c("neutral", "uncallable", "unstable") |
+      dominant_cytoband_status %in% c("neutral", "uncallable", "unstable") |
+      status != dominant_arm_status |
+      status != dominant_cytoband_status
```
jashapiro, line 216: This is the critical bit, and the logic is tough! I think we need some `&`s in here and parenthesis... I'm going to try to figure it out, with annotation.  I would love another set of eyes on this (as many sets as possible?)

```suggestion
  filter(
    # Case 1) Gene call disagrees with both arm and cytoband (This captures most 
    # we want to keep, including all of when arm and cytoband are neutral, uncallable 
    # or unstable, since we have no neutral gene calls in this df):
    (status != dominant_arm_status & status != dominant_cytoband_status)
    # Case 2) Gene call disagrees with a non-neutral cytoband call. 
    # Keep no matter what the arm status
    | (status != dominant_cytoband_status 
       & dominant_cytoband_status %in% c("gain", "loss"))
    # I think that captures everything we want. Cases we don't want include:
    # gene & arm agree, but cytoband is neutral
    # gene & cytoband agree
    # all 3 agree
```

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,293 @@
...
+      callable_fraction = weighted.mean(callable_fraction, band_length)
+    ) %>%
+    # Define dominant status based the weighted means meeting a status
+    # threshold
+    mutate(
+      dominant_arm_status = case_when(
+        callable_fraction < (1 - uncallable_threshold) ~ "uncallable",
+        loss_fraction > status_threshold ~ "loss",
+        gain_fraction > status_threshold ~ "gain",
+        loss_fraction + gain_fraction > status_threshold ~ "unstable",
```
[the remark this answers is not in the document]
cansavvy: I wonder if in an example situation where one status is >> than the other whether unstable is the right call. For example, if `loss_fraction_arm` is 0.89  and `gain_fraction_arm`  is 0.01 whether this should be called as `unstable` but I guess we could be asking these kinds of questions all day and maybe they don't actually happen often if at all?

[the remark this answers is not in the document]
jashapiro: Yes and also yes. I think this is a place where refinement could happen, but I don't know how much of a difference it will make in final conclusions. If we were to refine it, we would probablyu need to add another threshold for unstable, but I don't think we would want to call a region with 80% gain and 10%  loss as "neutral". In some sense, if all this does is catch a few things that fall just below threshold, that might be a good thing. Note that the consequences for the final table are the same: neutral and unstable are treated essentially the same way when looking at smaller regions/genes.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
cbethell: Ah, this logic makes much more sense to me, thank you for the added comments here!

jashapiro: Those comments were not just for you! They were for me as I was writing it. Not even me tomorrow; me today!

cbethell:
I implemented the latest of @jashapiro's suggestions in the [last commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/644/commits/f58c4d1b87f5d79211266f0581f73effe1474ada), including the logic mentioned in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/644#discussion_r406787187).

I ran it with the logic in that comment and produced a new output file for folks to take a look at when (or if) thinking about how this logic may or may not need to be adjusted.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
jaclyn-taroni: Checking my understanding here - we _do_ have neutrals for cytoband and arm. @jashapiro can you tell me a little bit more about
```
# gene & arm agree, but cytoband is neutral
```
as well.

jashapiro: Correct, we have neutrals for cytoband and arm. But if we call the whole arm as loss (for example), but the cytoband is called neutral, the assumption I was making is that the cytoband is probably a false negative, and we would already have the arm that "covers" all of the genes in this region, so an individual gene that is lost in that cytoband would not be worth of calling out as "focal".

jaclyn-taroni: Thanks @jashapiro ! This logic seems correct to me.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -0,0 +1,306 @@
...
+  gene = select(filtered_consensus_gene_status,
+                cols_to_keep,
+                region = gene_symbol),
+  .id = "region_type"
+) %>%
+  # Reorder columns more sensibly
+  select(Kids_First_Biospecimen_ID,
+         status = dominant_status,
+         region,
+         region_type)
```
jashapiro, line 282: This is not critical, but it might be worth sorting by ID, just to keep things in a somewhat reasonable order.
If we want to do even better, we could merge back in start and end positions for each region? Doing so might save a lot of redundant work in later scripts.

```suggestion
         region_type) %>%
    arrange(Kids_First_Biospecimen_ID)
```

cbethell: Gotcha, added this change in the last commit.

cbethell:
> My only suggestion is really a forward-looking one that we might want to retain region start and end positions for each region in the final table to make later analysis/plots a bit easier. I do not think that change should be in this PR, but might be noted as a potential future enhancement.

In other words, in a future PR, you may imagine getting the chromosomal start and end regions from the consensus seg file?

jashapiro:
> @jashapiro In other words, in a future PR, you may imagine getting the chromosomal start and end regions from the consensus seg file?

I imagine we would get them from the original annotation files that define arms, bands, and genes, but yes. If that information would be broadly used, it might make sense to include it here, rather than requiring those annotations being repeatedly added back. But I don't think it shoudl be done now.

Once we figure out why CI is failing, this should go in as it is.

cbethell:
> I imagine we would get them from the original annotation files that define arms, bands, and genes, but yes. If that information would be broadly used, it might make sense to include it here, rather than requiring those annotations being repeatedly added back. But I don't think it shoudl be done now.
> 
> Once we figure out why CI is failing, this should go in as it is.

Ahh okay, understood 👍 I will work on getting that ready in a separate PR.

jashapiro:
> Ahh okay, understood 👍 I will work on getting that ready in a separate PR.

I would work on other things first as I would view this as very low priority at this point. I really just wanted to document the thought.

PR #655, opened 2020-03-27
Unified color palette: mutational signatures

cansavvy:

Updating the color palette for the bubble matrix plots.

#### What was your approach?
- Updated the color palette in the `mutational-signatures` and the `figures/scripts/fig2-mutational-landscape.R`
- Made it so the`bubble_matrix_plot` function will accept a color palette. 

#622 

How do we feel about the new color palette? 
Note that Figure 2 needs to be updated, but it requires me to re-run the whole thing and I'm gonna hold off until we get the new BED files.

Yes.

Same results, the color palette is different. 

I didn't make any changes to this. 

#### Documentation Checklist
This is the same. 


analyses/mutational-signatures/util/mut_sig_functions.R
```diff
@@ -110,38 +115,48 @@ bubble_matrix_plot <- function(sig_num_df, label = "none") {
...
-      med_num = median(mut_per_mb)
+      med_num = median(mut_per_mb, na.rm = TRUE)
     )
   
   # Turn 0's into NAs so they are not actually plotted
   grouped_sig_num$prop_tumors[which(grouped_sig_num$prop_tumors == 0)] <-  NA
-
+  
+  # If color palette is not specified use color brewer's YlGnBu
+  if (is.na(color_palette[1])) {
```
jaclyn-taroni:
```suggestion
  if (any(is.na(color_palette))) {
```
might accomplish what you want and be more robust?

cansavvy:
> The code looks good - I would not necessarily expect to see changes in `analyses/mutational-signatures/results/nature_signatures_results.tsv` and `analyses/mutational-signatures/results/cosmic_signatures_results.tsv` and I can't really look at the diff easily. Can you comment on that?

Took a look into this. Data is the same, but the `sample_color` column got added in. I suppose we don't really want that in the results, so I will add in a quick `select` line to get rid of that.

Issue #645, opened 2020-03-24
Proposed Analysis: Include pathology feedback in repository


jaclyn-taroni:
Related to #609 

#### What are the goals of the analysis?

In https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/609#issuecomment-602821376, we have some feedback from CHOP pathologists Maria Rita Santi and Angela Viaene on `integrated_diagnosis`.  Many of the `Notes` contain:

> Updated via OpenPBTA subtyping

Which is to say are the results of molecular subtyping modules included in the repository.  From that same comment: 

> 1. `PT_7E3V3JFX` specimens were consistent with the original EPN dx, so pathology would call this a rare EPN, H3 K28 mutated tumor, rather than DMG.
> 2. `PT_AQWDQW27` specimen was consistent with meningioma, even though it has a hallmark EPN fusion, so pathology would also call this a rare meningioma with a _YAP1_ fusion.
> 3. Because 1 is a rare tumor (maybe first seen), the logic of searching for all H3 K28 mutations in [HGG subtyping](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/249) would convert this sample by default - how to handle this? 
> 4. Pathology confirmed this HGG BRAF V600E mutant tumor, `BS_H1XPVS9A` [https://cbethell.github.io/open-pbta-output/09-HGG-with-braf-clustering.nb.html#identify_sample_that_clusters_with_lgat), to be a LGAT (PXA). I updated `molecular_subtype` here based on what it would look like, but this should come through via the LGAT [subtyping](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/631) ticket. How should we add this info?

I think a short-term solution to address all of these notes in this repository is to create a new module `molecular-subtyping-pathology` that is essentially downstream of all other molecular subtyping modules that:

1. Compiles all the subtyping calls from all modules
2. Addresses these notes in a notebook that documents why changes need to be made

_Note: now might be a good time to consider grouping all of the molecular subtyping modules a directory (`analyses/molecular-subtyping`) where that directory has a README that explains the need for `molecular-subtyping-pathology` - alternatively put the downstream steps in `analyses/molecular-subtyping` itself. Might also consider a refactor to include common helper functions but that's a future problem._

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

~1 week - it will be multistep if we address the italicized note above

[...]

PR #642, opened 2020-03-20
add lgat subtyping script

kgaonkar6:
### Purpose/implementation Section
Subset metadata ,fusion and consensus maf from v15 data release to biospecimen where short_histology == LGAT and subtype samples depending on presence of BRAF fusion/V600E mutation or wildtype/not classified.

#### What scientific question is your analysis addressing?
How are lgat sample subtyped depending on the presence/absence of braf fusion/v600e mutations?

#### What was your approach?
I looked for short_histology == LGAT ; FusionName=="KIAA1549--BRAF" and BRAF==1 or BRAF_V600E==Yes respectively in the following files:
`pbta-histologies.tsv`
`pbta-fusion-putative-oncogenic.tsv`
`pbta-fusion-recurrently-fused-genes-bysample.tsv`
`lgat-subset/LGAT_snv_subset.tsv`

#### What GitHub issue does your pull request address?
#631

#### Which areas should receive a particularly close look?
Does this satisfy all conditions? 

#### Is there anything that you want to discuss further?
I added "LGG,To be classified" in addition to molecular_subtype column values in #631 because I think it helps capture the condition that some data (in most case it's WGS) is missing and so cannot be clearly classified? 
Like for example if a patient has no BRAF fusion in RNAseq samples and WGS is not available

 
#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
table

#### What is your summary of the results?
lgat samples are annotated with a molecular subtype as per #631 in:
analyses/molecular-subtyping-LGAT/results/lgat_subtyping.tsv

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

Issue #631, opened 2020-03-14
Proposed Analysis: Molecular subtyping for LGAT 


jaclyn-taroni:
#### What are the scientific goals of the analysis?

Breaking things out from https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/627#issuecomment-598789232 - **Molecular subtyping of LGAT** into LGG BRAF fusion, LGG BRAF V600E, or LGG BRAF WT

#### Analysis steps

1. Subset the consensus mutation file mentioned below to biospecimens where `short_histology == LGAT` and commit them to the repo. This may be necessary for any subsequent steps to pass CI, otherwise the number of samples that meet this criterion may be low (less of a concern with LGAT, though). This subsetting step will not go into CI. Check out [`molecular-subtyping-ATRT`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/7e000418b106e34040325d2fbd642f63080966c8/analyses/molecular-subtyping-ATRT) for an example. We'll want to set things up such that if someone runs things locally, the subset files will be generated using the version of the data symlinked in `data/`.

2. Identify LGAT biospecimens with _BRAF_ fusions using `pbta-fusion-recurrently-fused-genes-bysample.tsv`. _KIAA1549--BRAF_ fusions are of particular interest, but to get specific fusions we would have to use `pbta-fusion-putative-oncogenic.tsv`.

3. Identify LGAT biospecimens with BRAF V600E mutations in `pbta-snv-consensus-mutation.maf.tsv.gz`.

4. Create a final table that contains:

| `Kids_First_Participant_ID` | `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | `molecular_subtype` |
|----------------------------|-------------|------------------------------------|-----------------------|---------------|
| PT_XXXXXX | 7316-XXXX | BS_XXXXXXXX | BS_XXXXXXXX | LGG, BRAF wildtype | 
| PT_XXXXXX | 7316-XXXX | BS_XXXXXXXX | BS_XXXXXXXX | LGG, BRAF V600E |
| PT_XXXXXX | 7316-XXXX | BS_XXXXXXXX | BS_XXXXXXXX | LGG, BRAF fusion | 
| PT_XXXXXX | 7316-XXXX | BS_XXXXXXXX | BS_XXXXXXXX | LGG, BRAF V600E/fusion |   

[...]

PR #661, opened 2020-03-28
#645 Part 3: Required upstream changes to LGAT subtyping

jaclyn-taroni:

In addressing #645, I found that there were a lot of duplicated identifiers in the LGAT final table. The LGAT final table followed the general format where RNA-seq and DNA-seq samples were joined via `Kids_First_Participant_ID` and `sample_id` to get these columns:

| `Kids_First_Participant_ID` | `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | 
|---|---|---|---|

`lgat_subtyping.tsv` had 503 rows (excluding the header), but there are only 503 LGAT biospecimens. Because of the joining based on `Kids_First_Participant_ID` and `sample_id`, I would expect that the final table should have fewer rows than there are LGAT biospecimens. 

#### What was your approach?

As discussed in this thread https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/642#discussion_r396116933, both `pbta-fusion-recurrently-fused-genes-bysample.tsv` and  `pbta-fusion-putative-oncogenic.tsv` were used to identify LGAT samples with _BRAF_ fusions because the recurrently fused genes file only looked at independent RNA-seq specimens https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/df7a3a4e0334232c90baa5951c6d696ace1359d9/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L72

However, only _KIAA1549--BRAF_ fusions were looked at in `pbta-fusion-putative-oncogenic.tsv`. I believe this does not account for the possibility that an RNA-seq specimen was not selected in `pbta-fusion-recurrently-fused-genes-bysample.tsv`, but had a _BRAF_ fusion other than _KIAA1549--BRAF_. I also thought the use of two fusion files may have attributed to the identifier duplication issue. 

So, I've revised the `molecular-subtyping-LGAT` module to look for any _BRAF_ fusion in the `pbta-fusion-putative-oncogenic.tsv` file and bypassed `pbta-fusion-recurrently-fused-genes-bysample.tsv` step (I forgot about the independent specimen step when I wrote up #631 😞 ). I'm also saving the table used to make the subtyping calls as a TSV file so there is a bit more transparency during review and if we were to revisit this later.

Here's the logic for the subtyping – I tried to match what was original there and what I think is consistent with the intention behind #631 (which I filed on behalf of @jharenza) but definitely could use another opinion:

* If there is no fusion data and no mutation data available, a sample is labeled `LGG, To be classified`
* If no _BRAF_ fusion is detected and there is no mutation data available, a sample is labeled `LGG, To be classified`
* If there is no fusion data available or _BRAF_ fusion detected and a BRAF V600E mutation is present, a sample is labeled `LGG, BRAF V600E`
* If there is no fusion data available or _BRAF_ fusion detected and a BRAF V600E mutation is absent, a sample is labeled `LGG, BRAF wildtype`
* If there is a _BRAF_ fusion detected but no V600E mutation or no mutation data available, a sample is labeled `LGG, BRAF fusion`
* If there is a _BRAF_ fusion detected and a BRAF V600E mutation is present, a sample is labeled `LGG, BRAF fusion/V600E`

Related to #645 

Subtyping logic!

Scientific decision around the _BRAF_ fusions. I'm going to highlight a single example that changes as part of this pull request.

Here is a row in the file that is currently in master:
```
PT_20ZM6THA	7316-3629	BS_9X9E86DY	NA	LGG, BRAF wildtype
```

And what's being added (`BS_9X9E86DY` is a DNA-seq biospecimen identifier):
```
PT_20ZM6THA	7316-3629	BS_9X9E86DY	BS_E94QRJDW	LGG, BRAF fusion
```

And if we look at the table that underlies the subtyping, we see that this sample has no BRAF V600E mutation but does have two fusions (_BRAF--RNF130_, _RNF130--BRAF_):

```
BS_9X9E86DY	No	PT_20ZM6THA	7316-3629	BS_E94QRJDW	BRAF--RNF130, RNF130--BRAF
```

The fact that both detected fusions have the same partners but in different positions seems like it might be artefactual / something that we would filter out but I really do not know enough about fusions to say. I also considered that maybe that's something that gets filtered out in `analyses/fusion_filtering/06-recurrent-fusions-per-histology.R` but I don't think so? All this to say, it would be good to have a discussion.

#### Documentation Checklist


jaclyn-taroni:
@kgaonkar6 I think this one is going require us putting our heads together a bit!

kgaonkar6:
That's a good question reciprocal fusion partners,  I see from the callers annotation that this specific fusion in discussion doesn't come up as a artifact. Also in my opinion I think because they are not prevalent in multiple samples in unrelated histologies and genes are in different chromosomes and have enough junction supporting reads RNF130--BRAF has (10 in arriba annotated as translocation and 25 in starfusion) this fusion seems to be true and it's also found in previous literature https://cancer.sanger.ac.uk/cosmic/fusion/overview?fid=199318&gid=296121 and breakpoints are splice-sites. The reciprocal BRAF--RNF130 is a splice-site fusion, has 1 read support (which passes our condition ) in arriba and is a medlum confident call annotated as translocation as per arriba. 

I also  vaguely remember @jharenza mentioning that these reciprocal fusions might infact be considered as true calls because of underlying translocation/SV events as well, please correct me if I'm wrong.

kgaonkar6:
Also good catch on the NAs for DNA samples and multiple BRAF fusion rows! Thanks!

analyses/molecular-subtyping-LGAT/01-make-lgat-final-table.Rmd
```diff
@@ -1,91 +1,139 @@
...
+    (FusionName %in% c(NA, "None"))  &
+      BRAF_V600E == "No" ~ "LGG, BRAF wildtype",
+    # If there is a BRAF fusion detected but no BRAF V600E / the absence of 
+    # a DNA-seq sample, call is BRAF fusion
+    !(FusionName %in% c(NA, "None")) & 
+      (BRAF_V600E %in% c("No", NA)) ~ "LGG, BRAF fusion",
+    # If there is a BRAF fusion detected and a BRAF V600E mutation, note both
+    !(FusionName %in% c(NA, "None")) & 
+      BRAF_V600E == "Yes" ~ "LGG, BRAF fusion/V600E"
+  )) %>%
```
kgaonkar6:
```suggestion
final_table <- all_lgat_df %>%
  mutate(molecular_subtype = case_when(
    # When we don't have RNA-seq and mutation data, this is to be classified
    is.na(FusionName) & is.na(BRAF_V600E) ~ "LGG, To be classified",
    # When there is no BRAF fusion detected and there is no DNA-seq data
    # to be classified
    FusionName == "None" & is.na(BRAF_V600E) ~ "LGG, To be classified",
    # If there is no fusion data or BRAF fusion detection, but there is a 
    # BRAF V600E mutation, call BRAF V600E
    (FusionName %in% c(NA, "None")) & 
      BRAF_V600E == "Yes" ~ "LGG, BRAF V600E",
    # If there's no fusion data or BRAF fusion detected or V600E, wildtype call 
    (FusionName %in% c(NA, "None"))  &
      BRAF_V600E == "No" ~ "LGG, BRAF wildtype",
    # If there is a BRAF fusion detected but no BRAF V600E / the absence of 
    # a DNA-seq sample, call is BRAF fusion
    !(FusionName %in% c(NA, "None")) & 
      (BRAF_V600E %in% c("No", NA)) ~ "LGG, BRAF fusion",
    # If there is a BRAF fusion detected and a BRAF V600E mutation, note both
    !(FusionName %in% c(NA, "None")) & 
      BRAF_V600E == "Yes" ~ "LGG, BRAF fusion/V600E"
  )) %>%
```

it's an and logic here right?

jaclyn-taroni:
@kgaonkar6 

>  Should we reopen the previous issue to update this logic?

I think it's sufficient to add it to the README of this module. 

Based on what you've described about the reciprocal fusions, I think it's fine for us to use them in the manner that we are (e.g., presence/absence of **any** _BRAF_ fusion) for subtyping.

jaclyn-taroni:
I'm sorry, I fired off that comment before I finished it - do you agree with leaving the handling of fusions as is @kgaonkar6 ?

kgaonkar6:
I believe that the reciprocal BRAF fusion could be true calls however reading through https://www.frontiersin.org/articles/10.3389/fonc.2015.00054/full mentioned in the original ticket reference,  it looks like all KIAA1549-BRAF fusion need to have constitutive kinase activity and overexpression of BRAF  should we include these conditions for all other BRAF fusions as well?
- kinase domain retention == "Yes" 
- Fusion_Type=="in-frame" ( so that if kinase domain is retained its coded in the fusion transcript?)
- overexpression of BRAF == "Yes"

The bit in the paper that talk about BRAF fusions:
"The constitutive kinase activity of KIAA1549:BRAF fusion oncoprotein is due to the loss of the BRAF N-terminal auto-inhibitory domain which usually regulates BRAF activity (15, 28). Lin et al. demonstrated that the KIAA1549:BRAF fusion is transcribed from the KIAA1549 gene promoter and thus is expressed at higher levels than wild-type BRAF, so the fusion oncoprotein is not only constitutively active but also over expressed giving two mechanisms of aberrant activity (27)."

kgaonkar6:
I used https://github.com/d3b-center/annoFuse/blob/domain_annotation/R/getPfamDomain.R to get the domain retention status and https://github.com/d3b-center/annoFuse/blob/domain_annotation/R/plotSummary.R to plot the figures below:
the breakpoint in BRAF--RNF130 is within the kinase domain so the kinase domain maybe be partially retained image 
However in RNF130--BRAF the whole kinase  domain is retained 
image


You'll also see the information from arriba visualization plots however we don't have a text file from arriba with the domain retention information. (StarFusion has a column with the domain information though)
[556b5e35-1662-47e7-93ef-c34bb02dfe48.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4405824/556b5e35-1662-47e7-93ef-c34bb02dfe48.pdf)

jaclyn-taroni:
@kgaonkar6 - if the logic for this generally looks good, my inclination is to add a fusion cleaning step that takes into account the kinase domain retention prior to the notebook that puts together the final table and get this merged. We can either add the "fusion cleaning" this to `fusion-summary` _or_ in a notebook before this one in this module and try to have the output of that be compatible with `pbta-fusion-putative-oncogenic.tsv` such that we have the same column names used in this notebook. Can you file a new issue (likely using the update an analysis template) outlining what you've found? I think a focused issue will be easier to find and a good way to both solicit opinions from others and document scientific decision making.

kgaonkar6:
Thanks! I'll add a new issue explaining what we see here. I like the idea of adding it as a notebook before this one which has the same columns used from `pbta-fusion-putative-oncogenic.tsv`.

Issue #641, opened 2020-03-19
Figure Panel: Oncoprint Landscape


cbethell:
As depicted in the sketch of [the landscape of genomic alterations figure in the manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/04.results.md#landscape-of-genomic-alterations), the oncoprint figure found in `analyses/oncoprint-landscape/plots` should be updated and made publication ready. (Related to: #571)

Updating/making this figure publication ready will consist of:
-  Incorporating the list of genes in the `gene_disease_top50.tsv` generated and saved in the `results` directory of `analyses/interaction-plots` to filter to genes with mutations of interest interest that should be represented on the oncoprint
- Incorporating a list of CNA genes that should be represented on the oncoprint (after PR #617 is merged, lists of the "most focal unit" generated (WIP PR #485) and recurrent copy number alterations determined (#488)
- Updating the script to use the most recent consensus files where applicable (consensus CN files)
- Updating the script to use the color palettes generated in PR #510
- Creating a separate script for putting the oncoprint in a figure panel

I plan to work on this sometime this sprint.

jaclyn-taroni:
I'm going to call this closed by #666 and subsumed by #706.

PR #666, opened 2020-03-31
Generate publication ready oncoprint landscape figure

cbethell:

The purpose of this PR is to add a script to make the oncoprint landscape figure publication ready as depicted in the sketch [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/images/figure_sketches/oncoprint-proposed-sketch.png).

This PR pulls out the necessary results and steps required to rerun the oncoprint figure panel plots when necessary.

#### What was your approach?

- I took the bulk of the main functionality steps in `01-plot-oncoprint.R` and compiled them into a function script found at `util/oncoplot-functions.R`
- I then sourced this script in `01-plot-oncoprint.R` and in `figures/fig3-oncoprint-landscape.R`
- The `figures/fig3-oncoprint-landscape.R` script reruns the primary-plus (primary and secondary specimens) and primary-only versions of the oncoprint that take in the genes of interest list.

- I also adjusted the `run-oncoprint.sh` script from the `oncoprint-landscape` directory to take in the consensus seg autosomes file instead of the controlfreec autosomes file and removed the steps relevant to the gene lists that were previously being implemented.

This PR addresses issue #641.

- Does this appear to be the most efficient manner for re-running the oncoprint landscape plots?
- I was unsuccessful in my attempt to use `patchwork` to assemble the final figure due to what I believe is just the functionality of the `oncoplot` function.
It does not allow me to save the individual plots as an object in the R environment, therefore preventing me from using the `patchwork::plot_layout()`function.
**Are there any suggestions for an alternative here?**
- I also added the short histology names and hex codes to the `oncoplot_palette.R` script so that they can be sourced along with the remaining hex codes that are relevant to the oncoprint.
However, I am not sure that this was the best decision so this may need a close look as well.

**Note: The generation of the oncoplot step causes the R session to crash on my local machine.
Therefore, the current results in this PR were produced using the testing subset data.**

Note also that:
- the color palettes found in `figures/palettes` still need to be incorporated
- the copy number calls will be adjusted upon completion of PR #644 and this may affect the figure generation script
- no plots are included in this PR yet (due to the session crashing locally)

Yes, although a method to assembly the individual plots into a final figure is still left to be determined.

- The four oncoprint plots in `analyses/oncoprint-landscape/plots` have been updated using the consensus seg autosomes file, the gene list from the `interaction-plots/results` directory, and the additional hex codes in the updated `oncoplot-palette.R`

- There are also two plots saved in `figures/png` named `fig3a-oncoprint-landscape.png` and `fig3b-oncoprint-landscape.png`.
As mentioned above, these represent the primary-plus and primary-only oncoprints respectively, both filtered to represent only the genes of interest.
**These plots are here temporarily until a solution for assembling them into one final figure is determined.**

**Note also: In an attempt to find a solution for assembling these plots into one final figure, I implemented the `coOncoplot` function in the fig3 script and saved the output at `figures/pngs/fig3-oncoprint-landscape.png`.
This format is not ideal as it places the oncoprints side by side but it is a possible solution.**

I have not yet added the figure script to CI as I have not successfully ran it locally.

#### Documentation Checklist


figures/scripts/fig3-oncoprint-landscape.R
```diff
@@ -0,0 +1,240 @@
...
+      "analyses",
+      "interaction-plots",
+      "results",
+      "gene_disease_top50.tsv"
+    )
+  )
+
+#### Custom function ----------------------------------------------------------
+
+prepare_and_plot_oncoprint <- function(maf_df,
```
cansavvy, line 131: How is this section different or the same from the `01-plot-oncoprint.R`?

cansavvy: Can you put a link to  where this code originally comes from? Also is there a way we can minimize the duplication of the same code? (I get that we want to render the final png in one place so the answer to this may be no. )

cansavvy: Another idea to be thinking about is if you could make this function into its own `util` file in oncoprint-landscape folder, and then source it. Use it in `oncoprint-landscape` and here. (I did a similar thing for the CDF plotting function if you want to look for an example of that. This makes it so you only have to change this function once in the future and it will be changed in both the final figure and the figures in the `oncoprint-landscape`.

analyses/oncoprint-landscape/run-oncoprint.sh
```diff
@@ -15,26 +15,28 @@ cd "$script_directory" || exit
 
 #### Download the driver lists from the PPTC repository
 
-mkdir -p driver-lists
-wget -N --directory-prefix=driver-lists -O driver-lists/brain-goi-list-long.txt https://github.com/marislab/create-pptc-pdx-oncoprints/raw/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list-old.txt
-wget -N --directory-prefix=driver-lists -O driver-lists/brain-goi-list-short.txt https://github.com/marislab/create-pptc-pdx-oncoprints/raw/2c9ed2a2331bcef3003d6aa25a130485d76a3535/data/brain-goi-list.txt
+# mkdir -p driver-lists
```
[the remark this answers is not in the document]
cbethell: I commented it out for development reasons mainly. We are using the gene list from the `interaction-plots` directory in this PR instead of these driver gene lists so this is technically not needed here unless we want to leave the option in to implement them instead.

figures/generate-figures.sh
```diff
@@ -48,11 +50,33 @@ bash ${analyses_dir}/interaction-plots/01-create-interaction-plots.sh
 # Copy the main figure to final directory
 cp ${analyses_dir}/interaction-plots/plots/combined_top50.png pngs/mutation_cooccurrence_figure.png
 
+######################
+## Oncoprint plot(s)
 
+# Primary only oncoprint
```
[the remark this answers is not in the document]
cbethell: This distinguished in the `independent-samples` module [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/79a698904bb5f2e73c9eeef2918f48c80d6323e7/analyses/independent-samples/README.md#summary).

figures/scripts/fig3-oncoprint-landscape.R
```diff
@@ -0,0 +1,240 @@
...
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+
+# Path to the data obtained via `bash download-data.sh`.
+data_dir <- file.path(root_dir, "data")
+
+# Declare output directory
+output_dir <- file.path(root_dir, "figures", "pngs")
+
+# Source the color palette for plots
+source(
```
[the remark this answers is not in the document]
cbethell: That's a good idea. However, these colors would also need to be further tested and approved as I do not believe they were before.

figures/scripts/fig3-oncoprint-landscape.R
cbethell: > How is this section different or the same from the 01-plot-oncoprint.R?

This section in particular is the same. After making this script I did wonder if there was a way for me to take the two output plots needed from `01-plot-oncoprint.R` and read them into a script to be assembled into a figure panel but have not determined a way to do this as yet.

That being said, your suggestion:
> Another idea to be thinking about is if you could make this function into its own util file in oncoprint-landscape folder, and then source it. Use it in oncoprint-landscape and here. (I did a similar thing for the CDF plotting function if you want to look for an example of that. This makes it so you only have to change this function once in the future and it will be changed in both the final figure and the figures in the oncoprint-landscape.

seems that it will likely be the best solution in this situation.

figures/scripts/fig3-oncoprint-landscape.R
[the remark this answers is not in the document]
cansavvy: Can you put a short description of this color palette file in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/510?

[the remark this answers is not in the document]
cansavvy: I am not 100% sure what we will want to do with it, but I think it needs to be in the README at least, or needs to be incorporated into another color palette strategy at most. But if we have it recorded there we can see what else pops up as we are working on these figures.

analyses/oncoprint-landscape/util/oncoplot-functions.R
```diff
@@ -0,0 +1,103 @@
...
+        "Hem_Deletion",
+        "amplification",
+        "gain",
+        "loss"
+      )
+    )
+  
+  # If this function is being run in the figures generation script, return the
+  # maf object (instead of the final plot)
+  if (running_in_figures_script) {
```
cansavvy, line 75: Can you explain this distinction? Why are you plotting it in one situation and not in the other? Is there a way we can revise `analyses/oncoprint-landscape/01-plot-oncoprint.R` so we have less of a branch off? Secondly, I would have expected this function to always plot based on the name `prepare_and_plot_oncoprint`.

cbethell: The issue here is that the `coOncoplot` function (that I used to lay out the two specific oncoprints side by side in the figure generation script expects two maf objects, while the `oncoplot` function used to produce each individual oncoprint in the analyses/oncoprint-landscape/plots module only expects one maf object.

Should I maybe add the coOncoplot step to this function for `running_in_figures_script` = `TRUE` and add the function argument `maf_df_2` = `NULL` for the purpose of consistency?

analyses/oncoprint-landscape/util/oncoplot-functions.R
```diff
@@ -0,0 +1,103 @@
...
+  #             represented on the oncoprint. This is NULL by default.
+  #   color_palette: vector of colors (hex codes) corresponding to the data
+  #                  that will be represented on the oncoprint
+  #   running_in_figures_script: TRUE or FALSE statement indicating whether
+  #                              or not this function is being run in the
+  #                              figures generation script
+  
+  if (!is.null(fusion_df)) {
+    # Bind rows of maf and fusion data frames
+    maf_df <- dplyr::bind_rows(maf_df, fusion_df)
```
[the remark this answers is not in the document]
cbethell: Yes it does create NAs in the data frame. `fusion_df` has 5 columns, 4 of which overlap with `maf_df` which has 135 columns.

I'll also note that `fusion_df` is significantly smaller than `maf_df` so I am not sure what you initially meant here by _a lot_ of NAs but when running this using the subset file, the `fusion_df` data accounts for 32 of the 12147 observations in the data frame created using `bind_rows`. In other words, there are 32 rows that contain NAs.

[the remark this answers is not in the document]
cansavvy: It might be best to be a bit more intentional about the columns being bound so that way if something fails, we get a warning kicked back. In other words, what columns do you need downstream? Can we do a `select` of those columns and then bind the `fusion_df` and `maf_df` of just those columns? This way we can also double check what information is being passed so we can construct this in a way to know if we are missing information at some point.

analyses/oncoprint-landscape/util/oncoplot-functions.R
cansavvy: Do you end up using `coOncoplot` in both scenarios? If yes, then, let's find a way to incorporate it into the `prepare_and_plot_oncoprint ` function, if no, then we might want to remove it from this particular function.

cbethell: No, should I should remove the `oncoplot` step from this function and put it in the `01-plot-oncoprint.R` script in that case?

cansavvy: Upon digging into the code a bit I would say, yes. Put the `oncoprint` step into `01-plot-oncoprint.R` and remove it from `prepare_and_plot_oncoprint`. Then you will should probably: 

1) Get rid of the `running_in_figures_script` argument. 
2) Rename `prepare_and_plot_oncoprint` it something that indicates it won't do the plotting, but is preparing it for the plotting. `prepare_for_oncoprint` or something like that.

figures/scripts/fig3-oncoprint-landscape.R
```diff
@@ -0,0 +1,177 @@
...
+  height = 30,
+  units = "cm",
+  res = 300
+)
+# Given a maf file, plot an oncoprint of the variants in the
+# dataset and save as a png file.
+coOncoplot(
+  primary_plus,
+  primary_only,
+  clinicalFeatures1 = c(
```
cansavvy, line 153: Looks like `clinicalFeatures1` and `clinicalFeatures2` are being fed identical vectors? If so, I'd take this vector out separately and name it something.

figures/scripts/fig3-oncoprint-landscape.R
```diff
@@ -0,0 +1,177 @@
...
+
+#### Directories and Files -----------------------------------------------------
+
+# Detect the ".git" folder -- this will in the project root directory.
+# Use this as the root directory to ensure proper execution, no matter where
+# it is called from.
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+
+# Path to the data obtained via `00-map-to-sample_id.R`
+scratch_oncoprint_dir <- file.path(root_dir, "scratch", "oncoprint_files")
```
cansavvy, line 31: Just checking, does this directory get all its necessary files from lines 53 - 76 of `generate_figures.sh`?

cbethell: That is correct.

cbethell:
@cansavvy, I believe I implemented all of the suggestions you made above with the exception of:
>It might be best to be a bit more intentional about the columns being bound so that way if something fails, we get a warning kicked back. In other words, what columns do you need downstream? Can we do a select of those columns and then bind the fusion_df and maf_df of just those columns? This way we can also double check what information is being passed so we can construct this in a way to know if we are missing information at some point.

The reason being that when I attempted to select the variables that I believed were necessary and that were common to both data.frames, I encountered an error. So I need to do a little more digging into this to figure out what column variables are needed and which ones aren't.

In the meantime, this is ready for another look to ensure that I interpreted your other review comments correctly.

analyses/oncoprint-landscape/util/oncoplot-functions.R
```diff
@@ -0,0 +1,77 @@
...
+                               fusion_df = NULL) {
+  # Given maf, cnv, and fusion data.frames prepared in `00-map-to-sample_id.R`,
+  # along with the relevant metadata prepared in the `interaction-plots`
+  # directory, return a maf object. Note that the fusion data.frame argument is
+  # optional here and is NULL by default.
+  #
+  # Args:
+  #   maf_df: data.frame with data from a MAF file
+  #   cnv_df: data.frame with copy number variant data
+  #   metadata: data.frame with the relavant metadata
```
cansavvy, line 23:
```suggestion
  #   metadata: data.frame with the relevant metadata
```

analyses/oncoprint-landscape/util/oncoplot-palette.R
```diff
@@ -29,4 +29,42 @@ color_palette <- c("Missense_Mutation" = "#35978f",
...
+                   "Meningioma" =	"#3d0099",
+                   "MPNST" = "#c200f2",
+                   "Neurofibroma" =	"#917399",
+                   "none" =	"#f1f1f1",
+                   "Oligodendroglioma" = "#f279da",
+                   "Other" = "#cc0052",
+                   "Pineoblastoma" = "#994d6b",
+                   "Schwannoma" =	"#4d2636",
+                   "Teratoma" =	"#ffbfd9"
+)
```
jaclyn-taroni, line 70: If the palette in `figures/palettes/histology_color_palette.tsv` were to be updated, you would have to also make changes to this code. Instead, I would find a way to use `figures/palettes/histology_color_palette.tsv` and I'd even consider moving the OncoPrint palette to a TSV files in `figures/palettes`.

jaclyn-taroni: Looks like this has come up before! 😅 https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/510#issuecomment-607523137 Probably means we should do it!

jaclyn-taroni:
@cbethell regarding

> I was unsuccessful in my attempt to use patchwork to assemble the final figure due to what I believe is just the functionality of the oncoplot function.
It does not allow me to save the individual plots as an object in the R environment, therefore preventing me from using the patchwork::plot_layout()function.
Are there any suggestions for an alternative here?

Is there anything on #675 you can borrow?

cbethell:
> @cbethell regarding
> 
> > I was unsuccessful in my attempt to use patchwork to assemble the final figure due to what I believe is just the functionality of the oncoplot function.
> > It does not allow me to save the individual plots as an object in the R environment, therefore preventing me from using the patchwork::plot_layout()function.
> > Are there any suggestions for an alternative here?
> 
> Is there anything on #675 you can borrow?

I can give it a try, it looks like using the methods there to save and combine the plots into a multipanel figure should work!

cbethell:
@jaclyn-taroni, in this PR's latest commits, I implemented the use of the `figures/palettes/histology_color_palette.tsv` and `figures/palettes/divergent_color_palette.tsv` files to create an `figures/palettes/oncoprint_color_palette.tsv` file for use in the `oncoprint-landscape` module and in the `figures` directory. I reran the plots and added a description of this palette to `figures/README.md`. 

If the production/usage of this palette seems suffice, I can file a new PR to add the production of this palette to the `figures/scripts/color_palettes.R` script. _(Note: This removed the need for the hard coded `analyses/oncoprint-landscape/util/oncoplot-palette.R` script._

I also added the execution of the `analyses/focal-cn-file-preparation/run-prepare-cn.sh` to `figures/generate-figures.sh` upon the realization that this will be needed to produce the focal CN file that we want incorporated into the oncoprints (although this techinically will not produce the file we want until PR #686 is merged, hence the `consensus_seg_recurrent_focal_cn_units.tsv` file being added in this PR).

Additionally, as discussed, I was able to copy the final oncoprint figure from `analyses/oncoprint-landscape/plots` to `figures/pngs`, removing the need for an individual figure 3 generation script.

I also want to note that I added a rough estimate of the RAM requirements to the figure summary table of `figures/README.md`. I am not sure how far of this may be so if you have any suggestions for the best way to figure that out and make sure that estimate is accurate, please let me know!

With _ALL_ of that being said, this PR is ready for another look.

figures/README.md
```diff
@@ -60,6 +61,7 @@ Biospecimens without a `short_histology` designation are coded as `none` and ass
...
...d.it/150x40/2166ac/FFFFFF?text=2166ac) <br>Fusion:![4393c3](https://placehold.it/150x40/4393c3/FFFFFF?text=4393c3) <br>Multi_Hit:![92c5de](https://placehold.it/150x40/92c5de/FFFFFF?text=92c5de) <br>Multi_Hit_Fusion:![d1e5f0](https://placehold.it/150x40/d1e5f0/FFFFFF?text=d1e5f0) <br>Missense_Mutation:![f7f7f7](https://placehold.it/150x40/f7f7f7/FFFFFF?text=f7f7f7) <br>Nonsense_Mutation:![fddbc7](https://placehold.it/150x40/fddbc7/FFFFFF?text=fddbc7) <br>In_Frame_Del:![f4a582](https://placehold.it/150x40/f4a582/FFFFFF?text=f4a582) <br>In_Frame_Ins:![d6604d](https://placehold.it/150x40/d6604d/FFFFFF?text=d6604d) <br>Frame_Shift_Del:![b2182b](https://placehold.it/150x40/b2182b/FFFFFF?text=b2182b) <br>Frame_Shift_Ins:![67001f](https://placehold.it/150x40/67001f/FFFFFF?text=67001f) <br>Splice_Site:![f1f1f1](https://placehold.it/150x40/f1f1f1/FFFFFF?text=f1f1f1) | A named vector of hex codes assigned to each `short_histology` and to each `CNV`, `SNV` and `Fusion` category | For plotting an oncoprint figure, this vector incorporates the `histology_color_palette.tsv` and `divergent_color_palette.tsv` to provide hex codes for `short_histology` values as well as `CNVs`, `SNVs`, and `Fusions`
```
jaclyn-taroni, line 64: This does not look like it got updated with the most recent changes to the palette.

cbethell: You are correct, will do that in the upcoming commit.

jaclyn-taroni:
In addition to https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/686#issuecomment-632671064, I'd like to see the README for the OncoPrint module updated in a future PR. It is out of date. Since you are no longer snagging the brain cancer genes of interest in your shell script, you should document where you got them from in the README. Can you file an issue summarizing the changes that would come in future PRs @cbethell? (Alternatively, update #641.)

jaclyn-taroni:
I will merge once CI passes.

jaclyn-taroni:
This has, in fact, passed. I don't know why this still says pending.

Issue #643, opened 2020-03-23
Figure: Transcriptomic overview


jaclyn-taroni:
One of the figures mentioned in #571 is a "transcriptomic overview" with the following panels:

![Image from iOS (32)](https://user-images.githubusercontent.com/19534205/77316399-be092a80-6cdf-11ea-84f2-c6ff0ea5c8a8.jpg)

The GSVA and immune deconvolution steps both require `collapse-rnaseq`. I will use `broad_histology` for now and add TODOs about how we might deal with additional information (#627). Because we might want to change the labeling, I am planning to use the data from `gene-set-enrichment-analysis` and `immune-deconv` and to create a new script(s) to generate these panels.

jaclyn-taroni:
I'm going to start with `stranded` only for now.

Issue #541, opened 2020-02-17
v14 PBTA Histologies


komalsrathi:
#### What data file(s) does this issue pertain to?

pbta-histologies.tsv

#### What release are you using?

v14

#### Put your question or report your issue here.

broad_histology column has `Embryonal Tumor` and `Embryonal tumor` which needs to be standardized.

jharenza:
closed via #569

Issue #630, opened 2020-03-14
Use snakemake for pub-ready figure generation


jaclyn-taroni:
_Originally suggested by @jashapiro._

With #613, we're using a large bash script to regenerate figures. Some kind of workflow management system would be _better_. [`snakemake` is already on the project Docker container](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/7e000418b106e34040325d2fbd642f63080966c8/Dockerfile#L190).  The CCDL does not have bandwidth to implement this as part of our initial effort to get publication-ready figures together (#571) but wanted to document this potential improvement.

jashapiro:
Since I did do a bit of proof of concept work on this, I should link it here for anybody who might work on this in the future. 

https://github.com/jashapiro/OpenPBTA-analysis/blob/jashapiro/snakemake-results/Snakefile

One thing to note in that implementation is that it uses scripts as inputs in an attempt to capture both changes in data files AND analysis/figure generation code. But it might not catch all changes, esepcially in scripts called by the defined scripts.

sjspielman:
@jaclyn-taroni are we still interested in doing this? If so, it would be good to do in conjunction with/instead of #1261

jaclyn-taroni:
Are we interested? Sure, but I think we'd need to estimate the amount of effort before I comment on whether or not we _should_ do it.

sjspielman:
My gut tells me the effort would be a little too high, _but_ this depends on how much is in what @jashapiro had previously [written up](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/630#issuecomment-772583999). This link is now long gone it seems and I can't find the branch [here](https://github.com/jashapiro/OpenPBTA-analysis/branches/all). @jashapiro, still have this locally?

jashapiro:
In some previous cleaning, I must have removed that branch on github, but I did have it locally, and it is now [back up](https://github.com/jashapiro/OpenPBTA-analysis/blob/jashapiro/snakemake-results/Snakefile). Note that that file was setting up to do _all_ the analysis: not just the figures, but also the analysis that generated the inputs to those figures. Doing just the figures should be a bit easier/

[...]

PR #669, opened 2020-04-02
#643 Part 1: Upstream changes required for transcriptomic overview figure

jaclyn-taroni:

Generating the transcriptomic overview in #643 requires several upstream modules to be run:

* `transcriptomic-dimension-reduction`
* `collapse-rnaseq`
* `gene-set-enrichment-analysis`
* `immune-deconv`

I found that this application required a few changes to upstream modules, so I decided to file a pull request before I completed #643. I've held off on documentation around this figure for now.

#### What was your approach?

See directions for reviews for more information.

Begins to address #643

I would like to direct your attention to the following:

* The changes to `figures/generate-figures.sh` include the upstream changes required before putting together the figure outlined in #643. This probably needs the _closest_ look.
* I altered `analyses/gene-set-enrichment-analysis/01-conduct-gsea-analysis.R` and the shell script for that module `analyses/gene-set-enrichment-analysis/run-gsea.sh` such that the assumptions about where input and output files are located are relaxed.
* Because we don't include CIBERSORT in this repository and I don't have access to it, I am using xCell and MCP counter for `immune-deconv`. This uncovered an issue with how we were handle this case in `analyses/immune-deconv/01-immune-deconv.R`. 

Any changes in the TSV files related to `transcriptomic-dimension-reduction` appear to be due to #541 

I am ignoring the RData output of `analyses/immune-deconv/01-immune-deconv.R` that doesn't use CIBERSORT/is specifically for generating figures - what do folks think of this approach?

For the altered modules.

#### Documentation Checklist


sjspielman:
Some initial questions/comments before I dive in more tomorrow:

+  Inconsistent use of `--vanilla` flag when running R scripts that generate figures. Do we care? I'm not sure if it matters, but in case it does, we should do it for all or none.

+ I can see `figures/generate-figures.sh` becoming rather large and potentially unwieldy - especially since a lot of the code isn't explicitly generating a figure but rather prepping data, as is the case in this PR. Is there a good reason *not* to break this down into other files within `figures/`? Smaller shells scripts for cohesive groups of figures which employ the same data, and then call all those scripts from `generate-figures.sh`?

jaclyn-taroni:
We have some ideas about how to improve how we generate figures, namely #630. We’ve chosen to go unwieldy as a first pass.

jaclyn-taroni:
> Inconsistent use of --vanilla flag when running R scripts that generate figures. Do we care? I'm not sure if it matters, but in case it does, we should do it for all or none.

I don't think we care so much, but I added it in.

jaclyn-taroni:
@sjspielman I think this is ready for another look (provided the immune deconvolution step passes CI). I made some changes such that the second method for `analyses/immune-deconv/01-immune-deconv.R` can be NULL, so when we generate figures it will use xCell.

analyses/immune-deconv/01-immune-deconv.R
```diff
@@ -67,37 +69,36 @@ clin <- read.delim(clin.file, stringsAsFactors = F)
...
     inner_join(clin.sub, by = c("sample" = "Kids_First_Biospecimen_ID"))
-  
+
   return(res)
 }
 
 # Deconvolute using xCell and the second chosen method
-# these two methods have the max number of immune cell types
-deconv.method <- c("xcell", deconv.method)
+deconv.method <- unique(c("xcell", deconv.method))
```
sjspielman, line 95: Good call here, makes sense!!!

Issue #653, opened 2020-03-26
Updated analysis: add CNV alterations to evaluate tp53, nf1 classifier  


kgaonkar6:
#### What analysis module should be updated and why?
Tp53 and nf1 classifier by Greg Way was used to get TP53 and NF1 inactivation scores 
https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/165
As you might know we only used SNVs to get tp53/nf1 altered status for evaluating the classifier results, should we add CNV calls as well to identify tp53/nf1 deletion inactivation status of samples as per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/165#issuecomment-572823015 ?

#### What changes need to be made? Please provide enough detail for another participant to 
extend https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R to include CNV deletion (amplification as well?) calls to analyses/tp53_nf1_score/results/TP53_NF1_snv_alteration.tsv which will then be used to evaluate the classifier results https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/02-evaluate-classifier.py.

#### What input data should be used? Which data were used in the version being updated?
data/consensus_seg_annotated_cn_autosomes.tsv.gz maybe ?

#### When do you expect the revised analysis will be completed?
1 week

#### Who will complete the updated analysis?
@kgaonkar6

kgaonkar6:
Hi @jaclyn-taroni ! To gather the copy number alterations I’m using consensus_seg_annotated_cn_autosomes.tsv.gz  since it looks like is being used for certain subtyping analysis and is also annotated with gene_symbol. When I look at the status column there's loss, gain and amp 
`cnvConsesus$status %>% unique()`
`[1] “loss”          “gain”          “amplification”`
I used the all "loss" data in the evaluation step and it seems it's not a good match to evaluate looking at the ROC curve for stranded data here: 
![stranded_TP53](https://user-images.githubusercontent.com/34580719/78585076-3c4ced00-7807-11ea-9dff-03d2131e572f.png)
![stranded_NF1](https://user-images.githubusercontent.com/34580719/78585534-0825fc00-7808-11ea-8adf-04284a63c1b4.png)

[...]

Issue #665, opened 2020-03-31
Updated analysis: Cleaning BRAF fusions for LGAT subtyping 


kgaonkar6:
#### What analysis module should be updated and why?
**molecular-subtyping-LGAT**
In the LGAT subset we see known KIAA1549--BRAF fusion and some other BRAF fusions as you can see in FusionName column here:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-LGAT/results/lgat_cleaned_all_table.tsv

So @jaclyn-taroni  and I want to discuss here if **any** BRAF fusions found in LGAT sample should be included in molecular_subtype== "LGG,BRAF fusion" or would the fused BRAF need to satisfy other conditions like kinase domain retention and overexpression as mentioned in for the oncogenic KIAA1549--BRAF fusion
https://www.frontiersin.org/articles/10.3389/fonc.2015.00054/full 
"The constitutive kinase activity of KIAA1549:BRAF fusion oncoprotein is due to the loss of the BRAF N-terminal auto-inhibitory domain which usually regulates BRAF activity (15, 28). Lin et al. demonstrated that the KIAA1549:BRAF fusion is transcribed from the KIAA1549 gene promoter and thus is expressed at higher levels than wild-type BRAF, so the fusion oncoprotein is not only constitutively active but also over expressed giving two mechanisms of aberrant activity (27)."

A example of other BRAF fusion can be seen in this sample: has no BRAF V600E mutation but does have two fusions (BRAF--RNF130_, RNF130--BRAF):
```
BS_9X9E86DY	No	PT_20ZM6THA	7316-3629	BS_E94QRJDW	BRAF--RNF130, RNF130--BRAF
```

Here I've plotted the breakpoints for BRAF gene in each fusion to show the differences in kinase domain retention status per fusion
RNF130--BRAF is a cosmic fusion https://cancer.sanger.ac.uk/cosmic/fusion/overview?fid=199318&gid=296121. In-frame fusion and also retains the kinase domain because the breakpoint is before the kinase domain
![image](https://user-images.githubusercontent.com/34580719/78035805-95b0aa00-7337-11ea-8d49-264e4eb92022.png)
BRAF--RNF130 is the reciprocal frame-shift fusion that doesn't fully retain the kinase domain as you can see the breakpoint is within the kinase domain
[...]

PR #673, opened 2020-04-07
Tp53/NF1 classifier: Evaluate classifier using CNV alterations

kgaonkar6:
Update TP53/NF1 alterations that are used in the classifier evaluation step for analysis: [tp53_nf1_score](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tp53_nf1_score)

We used SNVs to get tp53/nf1 altered status for evaluating the classifier results, should we add CNV calls as well to identify tp53/nf1 deletion inactivation status of samples as per #165 (comment) ?

#### What was your approach?
Add TP53/NF1 loss status from data/consensus_seg_annotated_cn_autosomes.tsv.gz to results/TP53_NF1_snv_alteration.tsv 

#653 

TP53/NF1 loss status selection from data/consensus_seg_annotated_cn_autosomes.tsv.gz
[here](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/87360e83ec6d4030e6fa9015d32a4fa1589a48d1/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R#L83)
And "No_Tp53_NF1_alt" status assignment [here](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/87360e83ec6d4030e6fa9015d32a4fa1589a48d1/analyses/tp53_nf1_score/00-tp53-nf1-alterations.R#L125)

- I've assigned "No_TP53_NF1_alt" to TP53 amplification/gain status since we are looking for TP53/NF1 loss conditions. 
- Bs_ids which have neutral TP53/NF1 status are not in the consensus_seg_annotated_cn_autosomes.tsv.gz these are also annotated as "No_TP53_NF1_alt.

Yes

figures

Results might be updated, it seems the addition of all "losses" is not a good match to evaluate looking at the ROC curve for stranded data here: 
stranded_TP53
stranded_NF1

#### Documentation Checklist


jaclyn-taroni:
Hi @kgaonkar6 - over on #653, we talked about the "permissiveness" of the annotations (thinking of https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/653#issuecomment-610518874, specifically). Some changes got merged into [`focal-cn-file-preparation`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/focal-cn-file-preparation) such that there is now a file that accounts for broad CNA and is a table of the "most focal" unit (arm CNA vs. cytoband CNA vs. gene CNA). That file is: `analyses/focal-cn-file-preparation/results/consensus_seg_most_focal_cn_status.tsv.gz`. Naively, I expect that the broader calls might be a little more robust and we might want to try looking at the cytoband level for _TP53_ losses.

More generally, though - what are we hoping to gain or understand from the classifier analysis? (Note: I filed #165 because we were contacted via email about doing that analysis.) This could end up being a bit of a rabbit hole and I am trying to weigh this against tasks like #665.

kgaonkar6:
I was just trying to close loose ends with this PR :D . I'm not sure how it weighs in preferences to the BRAF fusion analysis . 

The  aim of running the classifier was to identify if there are samples which that have high tp53 inactivation score ( thus predicted to have inactivated tp53 profile) where maybe TP53 is not actually mutated or deleted or we don't have the dna data to look at that.

jaclyn-taroni:
> I was just trying to close loose ends with this PR :D .

Yea, I gotcha. Worth taking a step back, though!

> The aim of running the classifier was to identify if there are samples which that have high tp53 inactivation score ( thus predicted to have inactivated tp53 profile) where maybe TP53 is not actually mutated or deleted or we don't have the dna data to look at that.

So I think the handling of ploidy across all the copy number calls is somewhat inconsistent, which concerns me when it comes to these updates. 

Would it be more interesting, instead, to start by looking at negative (e.g., non-mutated) cases of _TP53_ that have the highest inactivation scores?

kgaonkar6:
> Worth taking a step back, though!

Right let me know if you think this can be de-prioritized as well .

> Would it be more interesting, instead, to start by looking at negative (e.g., non-mutated) cases of _TP53_ that have the highest inactivation scores?

yeah I think that would be interesting !

cgreene:
"Would it be more interesting, instead, to start by looking at negative (e.g., non-mutated) cases of TP53 that have the highest inactivation scores?"

We have had really good luck with these types of approaches revealing interesting biology in the past! Strong agree on exploring the top set of these. What are the confidences like? Moving classifiers out of the setting in which they are trained can often lead to probabilities that are a bit compressed towards the middle.

kgaonkar6:
Oh yeah I did work on the confidence intervals for ROC, I'll add that code to this PR as well.
ROC for the tp53_scores
ci_roc_tp53

ROC for nf1_scores
ci_roc_NF1

jaclyn-taroni:
@cgreene when you say 

> What are the confidences like? Moving classifiers out of the setting in which they are trained can often lead to probabilities that are a bit compressed towards the middle.

Are you looking for the confidence intervals for ROC or are you referring to something else?

cgreene:
The distribution of probabilities from the logistic regression classifier. I'm guessing that instead of running from 0-1 they'll be more narrowly distributed around the middle.

kgaonkar6:
I'm no expert in classifiers but from the code it looks like there is only a logit transformed score https://github.com/kgaonkar6/OpenPBTA-analysis/blob/a203b23478b9b3bb0ed1f19b78b2b04342f0bf6c/analyses/tp53_nf1_score/utils.py#L47 

Looks like p value can be calculated as such : 1/(1+exp(L)) ?

https://data-se.netlify.app/2017/01/24/convert_logit2prob/
Screen Shot 2020-05-22 at 9 58 27 AM

cgreene:
@kgaonkar6 isn't that what this line already does? https://github.com/kgaonkar6/OpenPBTA-analysis/blob/a203b23478b9b3bb0ed1f19b78b2b04342f0bf6c/analyses/tp53_nf1_score/utils.py#L49

I think you can just plot the output of the classifier.

kgaonkar6:
oh yeah that is what the scores are, ok I'll plot those!

kgaonkar6:
Boxplots for scores does look like mostly distributed around the 0.5 mark:

polya_score_dist
stranded_score_dist

cgreene:
@kgaonkar6 : can you plot separately by genotyping-derived status too? TP53 only, unless you want to do NF1. It doesn't look like NF1 is working in this setting from the results above.

kgaonkar6:
The tp53_status as x axis in the plots below are derived from the genomic data (1= has tp53 alteration and 0= doesn't have tp53 alterations)

polya_score_dist
TP53 status | freq
--- | ---
1   | 29
0   | 29

stranded_score_dist
TP53 status |freq
--- | ---
0   | 685
1   | 285

cgreene:
Ok! Yes - this makes a lot of sense. There is a difference by TP53 status and it goes in the anticipated direction - so the classifier seems to be porting successfully to this setting. On the other hand, the _probabilities_ coming out of the classifier look like they're likely to be a bit compressed.

You could make a calibration curve ( description if helpful: https://scikit-learn.org/stable/auto_examples/calibration/plot_calibration_curve.html ) if we really want to dig into understanding if the classifier is under- or over-confident in this setting. Based on what I see here, my guess is that it's under-confident in the poly-A setting and maybe overconfident in the stranded setting.

I don't think that's really necessary though, if we consider the results that come out as a ranked list instead of probabilities. If you took the top "false positives" (those with 0 status but with very high confidence scores), it could help to investigate those to see if they fall into certain tumor types, have certain mutations, or have other evidence of TP53 loss. In other settings we've found this type of analysis quite fruitful.

kgaonkar6:
> If you took the top "false positives" (those with 0 status but with very high confidence scores), it could help to investigate those to see if they fall into certain tumor types, have certain mutations, or have other evidence of TP53 loss. In other settings we've found this type of analysis quite fruitful.

Sounds good! I can start by looking into top 100 "false positives" from stranded dataset.

jaclyn-taroni:
@kgaonkar6 do you want to close this pull request and work on the ranked list stuff in a new one? Not sure if those boxplots included CNV alterations. (I'm after how relevant what your next steps will be are to #653.)

kgaonkar6:
I did include the CNV alterations (TP53 "loss" from consensus_seg_annotated_cn_autosomes.tsv.gz ) tp53_status counts in the boxplots above, would you suggest only using SNVs for now ?

If we are only interested in using SNV alterations then I'll close this PR like you said, since the next steps ie. looking at top false positives should probably be a different PR.

jaclyn-taroni:
> I did include the CNV alterations (TP53 "loss" from consensus_seg_annotated_cn_autosomes.tsv.gz ) tp53_status counts in the boxplots above, would you suggest only using SNVs for now ?

Nope, not necessarily -  I would be interested in the difference in scores between _TP53_ losses and _TP53_ coding mutations. In that case, I will review this - including the poor performance for ROC seems appropriate. I'd like to see those boxplots make it into the repo, too, but that can be a separate PR.

kgaonkar6:
Thanks Jaclyn! I'll work on the next PR .

Issue #648, opened 2020-03-26
Proposed Analysis: Survival plots HGG subtypes


kgaonkar6:
#### What are the scientific goals of the analysis?
As per @jharenza suggestions:
Identify the difference in survival 
- between molecular subtypes. 
- between samples with fusions (recurrent/known fusion which might be included in molecular subtypes like in EPN with RELA and YAP fusions but in high grade glioma we find REV3L-FYN and recurrent MET fusions these can be explored as well) . 
- between presence and absence of fusions in subgroups (I'm just thinking aloud should this also include presence and absence of SV changes? )
I've also just realized a lot of this was discussed as part of #18 :D 

#### What input data are required for this analysis?
method:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/survival-analysis/util/survival_models.R by @cansavvy 
file selection:
initial tumor samples (if multiple choose randomly) for Overall survival plots 
files: 
pbta-histologies.tsv for molecular-subtype information 
fusion-summary files for known/recurrent fusions
pbta-fusion-putative-oncogene.tsv for fusion presence and absence

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
1 week
Yes it is a multi-step analysis
1 ) Subset and format clinical data to get molecular subgroups, fusion file for presence-absence
2 ) Run survival code for each subgroup 
 
#### Who will complete the analysis (please add a GitHub handle here if relevant)?
I can work on this using @cansavvy notebook for survival 

#### What relevant scientific literature relates to this analysis?
[DIPG subgroups](https://user-images.githubusercontent.com/34580719/77652488-aaaac900-6f44-11ea-8385-b4d8b0e7202f.png)
[ more glioma survival plots](https://www.cell.com/cms/10.1016/j.ccell.2017.08.017/attachment/3551e2c1-9f69-450c-bda7-040543dcfc9b/mmc1)

jaclyn-taroni:
@kgaonkar6 I think we should pick one of these areas that is the most straightforward to implement so we can kick the tires of `survival-analysis`. HGG subtypes might be a good starting point because, based on your comment, it seems like we can compare our results to previous literature. What do you think?

kgaonkar6:
Sounds good to me! 
Yeah I think KM plots for HGG/DIPG subtypes could be compared to plots from https://www.cell.com/cancer-cell/fulltext/S1535-6108(17)30362-8

[...]

PR #674, opened 2020-04-09
Survival curves for HGG subtypes

kgaonkar6:
Plot survival curves for HGG subtypes

Are there differences in survival for HGG subtypes? 
How do they compare to published data?

#### What was your approach?
**method:**
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/survival-analysis/util/survival_models.R by @cansavvy
**Sample selection:**
broad_histology was == "Diffuse astrocytic and oligodendroglial tumor"
experimental_strategy != "RNA-seq" (because subtyping in HGG is from DNA aberations and samples where only RNA-Seq is available could not be subtyped)
broad_composition != "cell-line"
Kids_First_Biospecimen_ID==independent_primary$Kids_First_Biospecimen_ID

Survival plots comparing all subtypes as well as each subtype compared to wildtype were saved.

#648 

I'm expecting the survival plot to be per patient so I checked if each patient has a single molecular subtype. Should I add a logic in the code to remove these patient samples or use the independent_primary list like I did sound good?
I have plotted all subtypes here should there be a logic to only plot subtypes that have enough *N*?

This is from Chris Jones' meta analysis of glioma paper
Screen Shot 2020-04-08 at 8 08 05 PM
Screen Shot 2020-04-08 at 8 07 58 PM

Since we have a good number of samples in DMG, H3 K28 and HGG, H3 wildtype I think it does match the trends seen the glioma paper. 
> H3.3K27M were spread throughout the midline and pons, where they account for 63.0% DIPG and 59.7% non-brainstem midline tumors. In all locations (including ten cases reported to present in the cortex), these mutations conferred a significantly shorter time to death from disease (overall median 11 months, 2 year overall survival 4.7%)

I'm not so sure if we can compare other HGG subtypes survival scores since it looks like very few N are present. I believe in the 1000 glioma paper it's HGG+LGG so if we also do that for survival analysis we might have more Ns in BRAF V600E subset.

Yes

Survival plots comparing all subtypes:
plots/survival_curve_HGG_DMG.pdf
Survival plots comparing each subtype to wildtype:
plots/Kap_meier_Diffuse_astrocytic_and_oligodendroglial_tumor*pdf

median and CI for survival by comparing all subtypes:
plots/survival_curve_HGG_DMG.tsv
median and CI for survival by comparing each subtype to wildtype:
plots/Kap_meier_Diffuse_astrocytic_and_oligodendroglial_tumor*tsv 

DMG, K28M mutants have median survival of 1 year

#### Documentation Checklist


analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+## Import the metadata
+
+```{r Read in metadata}
+# choose Diffuse astrocytic and oligodendroglial tumor samples
+# remove RNA-Seq because subtyping in HGG is from DNA abberations and samples where only RNA-Seq is available cannot be subtyped
+# remove cell-lines from this analysis as well
+metadata <- readr::read_tsv(metadata_file,
+                            col_types = readr::cols(molecular_subtype = readr::col_character())) %>%
+  filter(experimental_strategy !="RNA-Seq" &
```
sjspielman, line 78: This filtering indeed ends up retaining only tumor samples, although not explicitly. *Potentially* worth changing the last filter in pipe to this `filter(broad_composition == 'tumor')` ? Not clear to me if necessary, though.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+  group_by(Kids_First_Participant_ID ) %>% 
+  summarise(count=n(),subtype=toString(molecular_subtype)) %>%
+  filter(count>1)
+
+# these patients might affect downstream analysis for survival and would need to be removed
+count_patient_subtype
+
+```
+
+#### Let's use independent_primary to use primary samples only and maybe will help filter cases where multiple subtypes are present per patient ID
```
sjspielman, line 96: I'd strongly recommend changing the headers. The goal for headers is to provide concrete and highly visible markers along the analysis path. Many headers in this file are instead casual descriptions of approach - would be better to reformulate headers as succinct steps and add the extra parts ("maybe will help filter...") as regular text below the header.

sjspielman: Ah, I think what is happening is that you are using `####` as a comment, but outside of an R chunk this becomes a header in markdown. If this is the case - instead just make these either comments in the code chunks, or regular sentences in markdown text.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+Uses a wrapper function (`survival_analysis`) from utils folder. 
+
+## Setup
+
+#### Packages and functions
+
+Read in set up script.
+
+```{r Set up library}
```
sjspielman, line 27: NOTE: This script does not knit in part to libraries not being used explicitly. Either must load `dplyr` explicitly, or add `dplyr::` to lines where it is relevant. There are some others bugs I note in other comments below.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+metadata <- readr::read_tsv(metadata_file,
+                            col_types = readr::cols(molecular_subtype = readr::col_character())) %>%
+  filter(experimental_strategy !="RNA-Seq" & 
+           broad_histology=="Diffuse astrocytic and oligodendroglial tumor" &
+           broad_composition !="cell-line") %>% 
+  as.data.frame()
+
+# find patients where multiple subtypes are present 
+count_patient_subtype<-metadata %>% 
+  select(Kids_First_Participant_ID ,molecular_subtype) %>%
```
sjspielman, line 85: This line is an example of where the code fails to run as stand-alone script - needs to be `dplyr::select()`.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+# these patients might affect downstream analysis for survival and would need to be removed
+count_patient_subtype
+
+```
+
+#### Let's use independent_primary to use primary samples only and maybe will help filter cases where multiple subtypes are present per patient ID
+
+```{r Read in independent}
+independent_primary<-read_tsv(independent_primary_file)
```
sjspielman, line 99: Again, since `readr` wasn't loaded, needs to be `readr::read_tsv` or load the library. The code runs when you execute line-by-line, but will not knit standalone otherwise.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+
+# find patients where multiple subtypes are present 
+count_patient_subtype<-metadata %>% 
+  select(Kids_First_Participant_ID ,molecular_subtype) %>%
+  unique() %>%
+  group_by(Kids_First_Participant_ID ) %>% 
+  summarise(count=n(),subtype=toString(molecular_subtype)) %>%
+  filter(count>1)
+
+count_patient_subtype
```
sjspielman, line 111: This approach will not really work safely in the long run (and see next line 114, I think you meant that to be a comment, but it appears as a header in Rmd document). Rather than interactively examining `count_patient_subtype` to see how many rows it has, I would strongly suggest an assertion of some kind - e.g.
`if (nrow(count_patient_subtype) > 0) stop("there are patients with multiple subtypes")`

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+survival curve shows the probability of survival over time.
+
+```{r Kaplan Meier}
+kap_fit <- survival_analysis(metadata,
+  ind_var = "molecular_subtype",
+  test = "kap.meier",
+  metadata_sample_col = "Kids_First_Biospecimen_ID"
+)
+# median and 0.95CI survival days
+kap_fit$model
```
sjspielman, line 128: I'm not an expert in survival analyses, but my gut is telling me we should probably *not* be considering any N=1 here as there is no possible estimate of uncertainty - CI's are NA. @cansavvy, since you prepped the module here, any thoughts?

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -0,0 +1,210 @@
...
+# Let's save the model for CI and median survial as well
+write_tsv(surv_median(kap_fit$model),kap_meier_model_file)
+```
+
+### Subtypes comapred to wildtype
+Lets look at Kap Meier plots for each mutant subtype to H3 wildtype
+
+```{r Kaplan Meier per subtype compared to wildtype}
+
+list_subtypes<-c("DMG, H3 K28","BRAF V600E","HGG, H3 G35","HGG, IDH","DMG, H3 K28, BRAF V600E")
```
sjspielman, line 171: This strikes me as a potentially concerning amount of multiple testing, but I'm still not the expert to ask. @jaclyn-taroni let's find one!

sjspielman:
Hi @kgaonkar6, 

I've gone through some of the code here line-by-line with comments throughout! To get at some of your questions here....

> I'm expecting the survival plot to be per patient so I checked if each patient has a single molecular subtype. Should I add a logic in the code to remove these patient samples or use the independent_primary list like I did sound good?

See comment I left - logic would be better, or as it seems ok without filtering, an assertion could work.

> I have plotted all subtypes here should there be a logic to only plot subtypes that have enough N?

I left a comment about this too - I'd definitely suggest logic for this, but again I'm not an expert in survival analyses. Either way, I can't imagine much good can come from N=1(?)

Another main comment I left is towards the end, when you do a bunch of pairwise comparisons. I'm concerned about the level of multiple testing there, and further some of these comparisons seem to include some of the N=1 subtypes. I would not think we can do any sort of robust analysis whenever N<3(?).  @jaclyn-taroni , can you tag some survival analysis experts here?

Finally, I'm getting lots of bugs knitting the script (usually works when running line by line, except for the `Read in metadata` chunk), so please be careful to check libraries are explicitly loaded/used. 

Thanks!
-Stephanie

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
sjspielman: NOTE: line 78 error "object 'experimental_strategy' not found" is what is causing CI to fail

kgaonkar6:
Hi @sjspielman , been busy with other stuff so will get to this this week sometime. Thanks for the code review!

sjspielman:
@kgaonkar6 no worries!! :)

kgaonkar6:
added libraries to wherever they are required and added the logic to use independent-primary file so if a patients have more than 1 molecular_subtypes (shouldn't a patient  1 molecular_subtype? )

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
kgaonkar6: So that bit was to show the reviewer that there are multiple patients which have 2 different molecular_subtype, which I'm not sure if that is expected or not. However, I did include your suggestion to add the warning and continue with using the independent-primary sample list which will always make sure that only 1 sample is selected per patient.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
kgaonkar6: changed this to have a specific title and the description below that.

jaclyn-taroni:
I have not totally dug into the discussion that came before but

> to use independent-primary file so if a patients have more than 1 molecular_subtypes (shouldn't a patient 1 molecular_subtype? )

Can you expand on this? I think we should be careful not to conflate participant with biospecimen - if you are using a consistent methodology for calling subtypes and there are multiple biospecimens from the same participant, it stands to reason that an individual could have biospecimens with different molecular subtype labels.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
jaclyn-taroni: I would only look at H3 K28 and H3 wildtype.

kgaonkar6:
So these participants have multiple subtypes
count_patient_subtype
Kids_First_Participant_ID |count | subtype                 
--- | ---    | --- 
  <chr>                     <int> <chr>                        
PT_KTRJ8TFY | 2 | DMG, H3 K28, HGG, H3 wildtype
PT_NK8A49X5 | 2 | DMG, H3 K28, HGG, H3 wildtype
PT_QA9WJ679  | 2 | DMG, H3 K28, HGG, H3 wildtype
PT_WGVEF96B  | 2 | DMG, H3 K28, HGG, H3 wildtype

If I only select the BS ids that are in independent-specimens.wgswxs.primary.tsv all the patients come up as DMG, H3 K28 . So it looks like the HGG H3 wildtype comes from the progressive/postmortem samples. 

My question here was for survival analysis specifically ,should we have  the survival plotted for these above patients as "DMG, H3 K28" or a combination of 2 subtypes as "DMG, H3 K28, HGG, H3 wildtype"

jaclyn-taroni:
It makes sense to me to use the independent specimen files at the moment (particularly because the subtype labels for HGG are mostly derived from the mutation data). I suspect there is methodology that accounts for multiple samples from the same patients in a survival analysis, but this is outside of my area of expertise.

kgaonkar6:
cool I'll edit to only use bs_ids in independent-specimens.wgswxs.primary.tsv per participant for this PR.

kgaonkar6:
So I investigated this a little bit more and looks like the multiple molecular_subtype participants for the below participants comes from **WGS** Initial Tumor being DMG, H3 K28 and **WXS** Initial Tumor being HGG, H3 wildtype

Kids_First_Participant_ID
--
PT_NK8A49X5
PT_QA9WJ679
PT_WGVEF96B


For example `PT_NK8A49X5` participant has WGS Initial CNS Tumor `BS_1MME7FBS` which is DMG, H3 K28 
but WXS `BS_6DT506HY` Initial CNS Tumor is HGG, H3 wildtype

Looking at the maf files directly for BS_6DT506HY it seems only mutect2 and strelka2 had the H3F3A K28M , lancet did not 👎 

Using v16 histology file I subsetted the 4 participants which have the multiple subtypes
[multiple_subtype_participants.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4669067/multiple_subtype_participants.txt)

I'm investigating this further separately from this PR

sjspielman:
@kgaonkar6 , I just initiated a master merge into this branch - can you make sure to pull down before your next commit here? I may have slightly jumped the gun and don't want my mediocre timing to induce conflicts :)

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
kgaonkar6: I missed this comment, I'll implement this

kgaonkar6:
I don't know anyone from the top of my head, I'll ask around here.

komalsrathi:
@jaclyn-taroni @kgaonkar6 

I have worked on a few projects/shiny applications leveraging survival curves. I am definitely not an expert but I have worked on those with Pichai, who is. From my experience (and from what I have observed in Pichai's code/other discussions), the code/plots look correct in terms of sample size used and application of methods.

kgaonkar6:
thanks @komalsrathi !

jaclyn-taroni:
@komalsrathi @kgaonkar6 any idea off-hand why we see `0.95UCL NA` in `molecular_subtype=HGG, H3 wildtype`?

komalsrathi:
> @komalsrathi @kgaonkar6 any idea off-hand why we see `0.95UCL NA` in `molecular_subtype=HGG, H3 wildtype`?

@jaclyn-taroni would this reference help? http://sphweb.bumc.bu.edu/otlt/MPH-Modules/BS/R/R7_LogisticRegression-Survival/R7_LogisticRegression-Survival4.html.

`NA == infinity` here. From the same link: "_A 95% upper confidence limit of NA/infinity is common in survival analysis due to the fact that the data is skewed_"

sjspielman:
> @komalsrathi @kgaonkar6 any idea off-hand why we see `0.95UCL NA` in `molecular_subtype=HGG, H3 wildtype`?

Seems that the package really should say Inf, but says NA just to mess us up: http://sphweb.bumc.bu.edu/otlt/MPH-Modules/BS/R/R7_LogisticRegression-Survival/R7_LogisticRegression-Survival4.html#:~:text=A%2095%25%20upper%20confidence%20limit,that%20the%20data%20is%20skewed.

jaclyn-taroni:
Thanks @komalsrathi and @sjspielman - this seems like a point that we probably want to include/comment on in the notebook itself.

sjspielman:
> Thanks @komalsrathi and @sjspielman - this seems like a point that we probably want to include/comment on in the notebook itself.

Further, we can also prove to ourselves this is the case - if we plot the curves with the CIs, the upper CI line should never reach as low as 50% survival. If indeed it's always above 50%, we can confirm that the NA means Inf. See here, bottom of page 3: https://myweb.uiowa.edu/pbreheny/5710/f14/labs/13.pdf

PR #675, opened 2020-04-09
#643 Part 2: Add multipanel plot for transcriptomic data

### LINKED ISSUE #643: Figure: Transcriptomic overview

jaclyn-taroni:
#### What was your approach?

The major addition here is `figures/scripts/transcriptomic-overview.R` which generates the figure below. I also made some changes to how the function we use to make the UMAP plot works so it could use the histologies palette, the instructions for using palettes (had to make some changes to get the gradient and divergent to work for me), and add `multipanelfigure` to the Docker image.

This only works for stranded data, which I think is enough to get started. We can talk about adapting this to make it more general, but I think that's beyond the scope of this pull request.

Closes #643 

* The updates to the figures documentation talking about palette use. I made changes I needed to get it to work for me.
* As with anything with figure generation, keep an eye out for anything that would be problematic if the underlying data changes.
* Annotation bar ordering for the heatmaps

#### Is there anything that you would like to discuss further?

I couldn't figure out how to make the built-in panel labeling from `multipanelfigure` to not meet the needs for this figure. We're adding a legend for all the panels - that doesn't need to be panel D and also couldn't figure out increasing the size. Maybe it requires adding text over the final PNG - thoughts?

Yes (but not necessarily for scientific content).

Transcriptomic overview figure

transcriptomic-overview

#### Documentation Checklist


cansavvy:
@jaclyn-taroni , since you are making fixes to the `figures/README.md` could you fix line 30? It should say `bash figures/generate-figures.sh`.

jaclyn-taroni:
I went on a whole `magick::image_read` journey this morning and can't get anything else that might automatically add labels to work well (e.g., preserve the quality of the panels).

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+short_histology_df <- histologies_df %>%
+  filter(experimental_strategy == "RNA-Seq",
+                RNA_library == "stranded") %>%
+  select(Kids_First_Biospecimen_ID,
+                short_histology) %>%
+  arrange(Kids_First_Biospecimen_ID) %>%
+  tibble::column_to_rownames("Kids_First_Biospecimen_ID") %>%
+  as.data.frame()
+
+gsva_scores_mat <- gsva_scores_mat[, order(colnames(gsva_scores_mat))]
```
cansavvy, line 146: Can this be moved to where the rest of the manipulation of `gvsa_scores_mat` is happening? In that 100 -118 range?

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+
+# We need a new color palette, etc.
+na_color <- gradient_palette %>%
+  filter(color_names == "na_color")
+
+gradient_palette  <- gradient_palette %>%
+  filter(color_names != "na_color")
+
+gradient_col_val <- seq(from = min(deconv_mat),
+                         to = max(deconv_mat),
```
cansavvy, line 212: I think this indent is off.

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# and stroma score
+# We're going to separate those rows out from the cell type scores
+scores_deconv_mat <- deconv_mat[grep("score", rownames(deconv_mat)), ]
+cell_type_deconv_mat <- deconv_mat[-grep("score", rownames(deconv_mat)), ]
+
+# To keep ordering consistent across the two xCell heatmaps, we'll order by
+# histology
+short_histology_df <- short_histology_df %>%
+  # We lose the rownames if we use arrange without this step
+  tibble::rownames_to_column("biospecimen") %>%
```
cansavvy, line 228: If you arrrange this before we use `short_histology_df` for the first time, can you avoid this bit and then just use `rownames(short_histology_df)`. Or is there a danger that this can get out of order  between that heatmap and this one?

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+
+#### Color palettes ------------------------------------------------------------
+
+palette_dir <- file.path(root_dir, "figures", "palettes")
+histology_palette <- read_tsv(file.path(palette_dir,
+                                               "histology_color_palette.tsv"))
+divergent_palette <- read_tsv(file.path(palette_dir,
+                                               "divergent_color_palette.tsv"))
+gradient_palette <- read_tsv(file.path(palette_dir,
+                                              "gradient_color_palette.tsv"))
```
jashapiro, line 49: Just because (guessing you took out `readr::` by the number of extra spaces)
```suggestion
histology_palette <- read_tsv(file.path(palette_dir,
                                        "histology_color_palette.tsv"))
divergent_palette <- read_tsv(file.path(palette_dir,
                                        "divergent_color_palette.tsv"))
gradient_palette <- read_tsv(file.path(palette_dir,
                                       "gradient_color_palette.tsv"))
```

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# We'll filter based on whether the ANOVA has a significant p-value
+included_genesets <- gsva_stats_df %>%
+  mutate(hallmark_name = stringr::str_replace_all(
+    stringr::str_remove(hallmark_name, "HALLMARK_"), "_", " ")
+  ) %>%
+  filter(significant_anova) %>%
+  pull(hallmark_name)
+
+# Filter to only those gene sets
+gsva_scores_mat <- gsva_scores_mat[included_genesets, ]
```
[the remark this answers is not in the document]
jashapiro: I would support this change, but it is certainly minor. Also the reordering of the matrix at this point as per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/675/files#r409099744

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# Annotation bar intended for the top of the heatmap
+column_heatmap_annotation <- HeatmapAnnotation(
+  df = short_histology_df,
+  col = annotation_colors,
+  name = "short histology",
+  show_legend = FALSE
+)
+
+## Heatmap itself!
+gsva_heatmap <- Heatmap(
```
jashapiro, line 163: I would love it if the column denrogram were a bit narrower here, but I don't think there is any way to achieve that directly. Perhaps by outputting to a larger png?

Alternatively, order by short histology to make the plot a bit simpler? (Ideally, dendrogram by histology, but I don't know if that is doable in any reasonably efficient way)

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+  "results",
+  "rsem_stranded_log_umap_scores_aligned.tsv"
+)
+
+umap_plot <- read_tsv(rsem_umap_file) %>%
+  select(X1, X2, short_histology) %>%
+  replace_na(list(short_histology = "none")) %>%
+  plot_dimension_reduction(point_color = "short_histology",
+                           x_label = "UMAP1",
+                           y_label = "UMAP2",
```
jashapiro, line 71: We will probably want these labels smaller; as it stands they seem disproportionately large. Same for the axis tick numbers, which also seem a bit large.

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# list of named vectors
+annotation_colors <- list(
+  short_histology = tibble::deframe(histology_palette)
+)
+
+# Annotation bar intended for the top of the heatmap
+column_heatmap_annotation <- HeatmapAnnotation(
+  df = short_histology_df,
+  col = annotation_colors,
+  name = "short histology",
```
jashapiro, line 158: Is it possible to move the name to the left? It kind of looks like it is the name for the row labels in its current position. Also, I note the included png does not reflect this name, but uses `short_histology` for some reason.

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+                                     divergent_palette$hex_codes)
+
+# Annotation bar for short histology
+# First get a data frame for the annotation and ensure samples are in the
+# same order
+short_histology_df <- histologies_df %>%
+  filter(experimental_strategy == "RNA-Seq",
+                RNA_library == "stranded") %>%
+  select(Kids_First_Biospecimen_ID,
+                short_histology) %>%
```
jashapiro, line 141: I think to make the `HeatmapAnnotation` have the correct label name you need to set it as the column name here, so something like?
```suggestion
         "short histology" = short_histology) %>%
```

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# Make the biospecimen ID the rownames again
+short_histology_df <- short_histology_df %>%
+  tibble::column_to_rownames("biospecimen")
+
+# Annotation bar intended for the top of the heatmap needs to be remade because
+# of the ordering but the colors are the same
+column_heatmap_annotation <- HeatmapAnnotation(
+  df = short_histology_df,
+  col = annotation_colors,
+  name = "short histology",
```
jashapiro, line 244: Same comment as above about location of this label looking like a header for the row labels, and the fact that the name isn't showing up. I think this is fixed by setting the data frame column names to the display name you want, as ugly as it is to have column names with spaces.

figures/scripts/transcriptomic-overview.R
jashapiro: Also, indentation here 😉

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+# We need a new color palette, etc.
+na_color <- gradient_palette %>%
+  filter(color_names == "na_color")
+
+gradient_palette  <- gradient_palette %>%
+  filter(color_names != "na_color")
+
+gradient_col_val <- seq(from = min(deconv_mat),
+                         to = max(deconv_mat),
+                         length.out = nrow(gradient_palette))
```
jaclyn-taroni, line 213:
```suggestion
gradient_col_val <- seq(from = min(deconv_mat),
                        to = max(deconv_mat),
                        length.out = nrow(gradient_palette))
```

figures/scripts/transcriptomic-overview.R
jaclyn-taroni: > (Ideally, dendrogram by histology, but I don't know if that is doable in any reasonably efficient way)

I'm going to call this beyond the scope of the current pull request, but I will try ordering by short histology.

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,336 @@
...
+                                              "gradient_color_palette.tsv"))
+
+#### UMAP plot -----------------------------------------------------------------
+
+dim_red_dir <- file.path(
+  analyses_dir,
+  "transcriptomic-dimension-reduction"
+)
+
+source(file.path(dim_red_dir, "util", "dimension-reduction-functions.R"))
```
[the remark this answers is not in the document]
jaclyn-taroni: The convention I'm using here is to call the directories, files, etc. relevant to each panel within that panels section. To do what you suggest, I'd have to move all of this out of the individual panel sections. Normally I would agree with you, but I think in this case having it arranged by panel is helpful in case you want to rip out a panel entirely.

figures/scripts/transcriptomic-overview.R
jaclyn-taroni: I will decrease the size in `theme(text = element_text(size = 15))`.

figures/scripts/transcriptomic-overview.R
```diff
@@ -0,0 +1,315 @@
...
+# list of named vectors
+annotation_colors <-  tibble::deframe(histology_palette)
+
+# Annotation bar intended for the top of the heatmap
+column_heatmap_annotation <- HeatmapAnnotation(
+  df = short_histology_df,
+  name = "short histology",
+  col = list("short histology" = annotation_colors),
+  na_col = na_color$hex_codes,
+  annotation_name_side = "left",
```
jashapiro, line 83: I do like the annotation left  change, but looking at the final output, this does not seem necessary for the Immune deconvolution plot, and adds a lot of unneeded white space to the final plot. If both plots are included, then I would think the best option would be to have the name only on the GSVA plot, and assume the reader can follow to the right with the same same colors and location.

figures/scripts/transcriptomic-overview.R
jashapiro: Agreed that this is beyond the scope, but I do like the short histology ordering.

figures/scripts/transcriptomic-overview.R
jaclyn-taroni: I both agree with you and am sad because I moved things around a bit so we only had to create the annotation once 😢 no big deal - will change.

PR #676, opened 2020-04-14
snv-callers: more explicit notes about relative file paths in README

### LINKED ISSUE #632: Documentation: Relative file paths for snv-callers

Documentation around how `snv-callers` and `chromosomal-instability` analyses take relative file paths should be made a bit more clear. Related to [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/625#issuecomment-598785681)
[And this is the original note in the documentation](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers#general-usage-of-scripts) about relative file paths that should be made more obvious / put in more places in these analyses. 

`snv-callers` and `chromosomal-instability` and check if there's any other instances where this applies. 

Today or tomorrow. 

@cansavvy

cansavvy:

There's been some confusion on how `snv-callers` uses file paths. I've added more explicit notes to its README so this is more clear that all file paths are relative to the root directory of `OpenPBTA-analysis`. 

This closes #632 

Is this wording clear enough? Are there other places I should add more notes?


jaclyn-taroni:
Documentation only change - merging without CI.

Issue #636, opened 2020-03-18
Mutational Signatures 


arpoe:
The current figure for mutational signatures is very nice. I am however worried that there may be problems with fitting. The "flat signatures" tend to have problems in this regard, of which Signature 3 is the most relevant one here. With the current methodology it is likely to be overestimated. 
I am also concerned that the smoking and UV signatures are picked up in non-negligible levels. 
May I raise a discussion of alternative strategies? I would suggest to perform de novo calling on the samples at hand, so that only signatures that are present in the dataset to a certain amount are actually used for fitting. Of course this does not solve all problems, but it reduces misassignment of signatures.

jaclyn-taroni:
Hi @arpoe, thanks for filing this! I am pinging @cansavvy who initially did the analysis.

I also wanted to direct you to the current version of the mutational signature analysis so you can take a closer look: https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/5f2468daf2756a62c7d4615a7660b0a52e5ad135/analyses/mutational-signatures

When you say:
> The "flat signatures" tend to have problems in this regard, of which Signature 3 is the most relevant one here. 

You are referring to the previously defined/derived signatures (e.g., COSMIC, Nature) that tend to be flat across datasets, is that correct?

> May I raise a discussion of alternative strategies? 

Yes please!

> I would suggest to perform de novo calling on the samples at hand, so that only signatures that are present in the dataset to a certain amount are actually used for fitting.

I had a couple questions regarding this point that I'd love to hear your thoughts on:

* Part of our rationale for using COSMIC signatures is that they seem to be often used in other datasets/analyses and that would perhaps facilitate some comparison to other datasets (e.g., adult datasets, related to #551). I am not an expert in this area, so it's possible that this line of thinking is misguided. If one were to perform de novo calling on pediatric and adult datasets, what would be the approach for comparison?
* Is there any floor to the number of samples that are required for that? On a related note, if you have a (unbalanced) mix of disease types like we do in OpenPBTA, do you generally analyze disease types separately?

arpoe:
Thanks for the link to the analysis. Here are the responses to your questions:
> 
> > The "flat signatures" tend to have problems in this regard, of which Signature 3 is the most relevant one here.
> 
> You are referring to the previously defined/derived signatures (e.g., COSMIC, Nature) that tend to be flat across datasets, is that correct?

[...]

PR #678, opened 2020-04-23
de novo mutational signature calling

arpoe:
Hi, 
As discussed previously, I have worked a bit on the mutational signature analysis. The reason is that with the current approach there are some artefacts created, such as smoking and UV signatures popping up in tumours that are not likely to have mutated through these mechanisms. This is caused by overfitting and a common problem when fitting mutational signatures from a general set, especially in tumours with low mutation rates. 
These problems would be expected to increase with newer signature selections as there are more signatures for the algortithm to choose from. As mentioned, these are common problems, but they become more apparent with lower mutation rates.
There is no perfect solution to this. What I offer is to apply de novo calling of signatures from this set and linking them back to the known signatures, including novel CNS specific ones and the newest set from COSMIC. 
I have used these de novo signatures to create the same kind of plots that were created previously. 
Generally it worked very well and all the signatures picked up make sense. To assign them to an ethiology some are obvious some are very alike other signatures from COSMIC. None of the signatures is new. To describe them properly, so more work should be done, but I first wanted to see whether you think this approach makes sense. 

I have not uploaded the signature calling output, as this is 8 GB. Please be aware that running that part of the script is quite time consuming. 

PS.: Please apologize if I have made some mistakes in the procedures. It is my first pull request ever :-)


jaclyn-taroni:
Hi @arpoe, thank you so much for submitting this! I am looking forward to digging in soon. The CCDL is in the middle of preparing for our first ever virtual workshop, so I apologize for the delay in review.

arpoe:
Thank you, that is no problem. Please let me know if there is something I could do in the meantime. 
Good luck with the workshop! I have advertised it in Europe :-)

jaclyn-taroni:
Hi @arpoe - sorry for the delay, we had some issues with the project Docker image (#689) last week that we had to take care of. I'm going to take a look at this today! 

It looks like when this ran in CI earlier (pre-Docker image issues), [this module failed for the following reason](https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/4201):

```
The downloaded source packages are in
	'/tmp/RtmpI75KCJ/downloaded_packages'
Quitting from lines 40-60 (mutational_signatures.Rmd) 
Error: Failed to install 'sigfit' from GitHub:
  System command error, exit status: 1, stderr empty

Execution halted

Exited with code exit status 1
```

We'll want to put `sigfit` on the project Docker image, which I can take care of right now and I will resolve the conflicts as well. This will allow us to test if the rest of your submission here, rather than just the install step, for execution. 

Then I will start going through what you're adding here! Thanks again for your patience!

analyses/mutational-signatures/mutational_signatures.Rmd
```diff
@@ -368,6 +369,247 @@ lapply(unique(nature_sigs_primary$short_histology),
 )
 ```
 
+## Determine Signatures de novo
```
jaclyn-taroni, line 372: My inclination would be to split this out into it's own notebook (+ rename and renumber the notebook you are altering here). My rationale is that this section is thematically different from the original notebook and I'd expect this nice description you have here to be at the top of the notebook. A few other considerations:

- This notebook ends up being over 600 lines long, which can be hard to digest if you're trying to understand what was done (or how this analysis evolved) for the first time
- I believe any _existing_ custom functions that you'd use are in `util/mut_sig_functions.R` and can be sourced into a new notebook
- You have some steps in lines 384-436 that you may want to have some more control over (more thoughts on that below)

analyses/mutational-signatures/mutational_signatures.Rmd
```diff
@@ -368,6 +369,247 @@ lapply(unique(nature_sigs_primary$short_histology),
...
+  
+  recmean <- reconstructions$mean
+  rownames(recmean)<- rownames(sigs_input) 
+  expmean <- exposures$mean
+  explow95 <- exposures$lower_95
+  rownames(expmean)<- rownames(sigs_input)
+  
+  # set exposures to 0, if lower confidence interval hits 1%
+  expmean[explow95<0.01]<-0
+```
```
jaclyn-taroni, line 437: If I'm understanding correctly - if we're in a situation where the underlying data files change (or the software dependencies!), I have the large files locally, and I know that everything related to mutational signatures needs to be rerun, I'd have to manually delete the files to accomplish that. I agree that skipping long-running, memory intensive steps by default makes a lot of sense, but I am worried about the situation I described.

To accommodate both situations, I propose we control the behavior here using [`params`](https://rmarkdown.rstudio.com/lesson-6.html), where there is some `params` that specifies whether or not to check for existing files and it checks for existing files by default.

I think before making any kind of specific recommendation, I'd like to understand this a bit more - is there ever a scenario where we would expect _some but not all_ of the files to exist?

analyses/mutational-signatures/mutational_signatures.Rmd
```diff
@@ -368,6 +369,247 @@ lapply(unique(nature_sigs_primary$short_histology),
...
+                                        nsignatures = 10,
+                                        iter = 10000, 
+                                        seed = 42)
+    
+  save(samples_extr10, file="results/sigfit_signatures_10.rdat")
+}
+```
+
+```{r}
+if(!file.exists("results/sigfit_signatures_10_signatures.rdat")){
```
jaclyn-taroni, line 402: I am unfamiliar with the `.rdat` file type - do you know how it compares to `.RDS` and `.RData`?

jaclyn-taroni:
Most recent failure reason is
```
label: unnamed-chunk-29
Quitting from lines 402-437 (mutational_signatures.Rmd) 
Error in dimnames(x) <- dn : 
Calls: <Anonymous> ... withCallingHandlers -> withVisible -> eval -> eval -> rownames<-

Execution halted
```

My guess is that the first long running step results in the 8GB file that is not committed to the repo is generated using the CI files, which only include a subset of samples to save time ([docs on CI](https://github.com/AlexsLemonade/OpenPBTA-analysis#continuous-integration-ci)). But the other files that _are_ committed to the repository, and therefore exist on the machine running CI when this repository gets checked out, are from the full dataset. I haven't tested this locally yet, but if that is what is happening it might be a good idea to revise the logic such that _all_ the files need to exist so there's no chance of a mismatch. 

Related note: I had to bump the timeout for this module to 90 minutes. This is likely because we're running `extract_signatures` twice. Does it make sense to take out the first one because of your comment re: hardcoding 10?

arpoe:
> Hi @arpoe - sorry for the delay, we had some issues with the project Docker image (#689) last week that we had to take care of. I'm going to take a look at this today!
> 
> It looks like when this ran in CI earlier (pre-Docker image issues), [this module failed for the following reason](https://circleci.com/gh/AlexsLemonade/OpenPBTA-analysis/4201):
> 
> ```
> The downloaded source packages are in
> 	'/tmp/RtmpI75KCJ/downloaded_packages'
> Quitting from lines 40-60 (mutational_signatures.Rmd) 
> Error: Failed to install 'sigfit' from GitHub:
>   System command error, exit status: 1, stderr empty
> 
> Execution halted
> 
> Exited with code exit status 1
> ```
> 
> We'll want to put `sigfit` on the project Docker image, which I can take care of right now and I will resolve the conflicts as well. This will allow us to test the rest of your submission here, rather than just the install step, for execution.
> 
> Then I will start going through what you're adding here! Thanks again for your patience!

No problem, I have been busy myself. I will be going through the issues you raise now. As for looking into the code, I could look into things from next week. This week is already quite full.

arpoe:
> Hi @arpoe! Based on what we talked about over on #636, the overall approach you've described seems reasonable to me! The comments I've left mostly have to do with organization, structure, and some design decisions in service of making things easier to follow and maintain down the road. Specifically, it makes sense to me to break out the de novo signature steps into their own notebook.
Yes, sounds good. I was just using what was there so that it is consistent and comparable. 
> 
> In your initial comment, you state:
> 
> > To describe them properly, so more work should be done, but I first wanted to see whether you think this approach makes sense.
> 
> Which makes me think a new notebook is best!
> 
> I may have some more detailed comments about individual sections in the future (often having to do with checking my understanding and documentation) but I wanted to send you some of the "bigger picture" comments first.
> 
> If you'd like/depending on what is on your plate, I can take the first pass at splitting things up and the logic for checking if files exist. Let me know!
My plate is quite full at the moment (moving to Germany in a month), so I'll do my best to keep the ball rolling, but I would be very happy, if you could proceed with splitting things up. 
> 
> I wanted to note that I've made a few changes to your branch here, so if you do go to address these comments with your local copy of the repository I would recommend using `git pull` first. If you have any trouble with that, please feel free to post here or over in Cancer Data Science Slack!
> 
> Please let me know if anything is unclear or if you have any questions. Thanks again for your contribution and your patience!

analyses/mutational-signatures/mutational_signatures.Rmd
arpoe: sounds good

analyses/mutational-signatures/mutational_signatures.Rmd
arpoe: I agree that a solution with params would be best. My solution is not ideal. 
If there are new data, everything would need to be rerun, because the signatures are derived from the cohort. This is the big disadvantage over using a fitting approach.

analyses/mutational-signatures/mutational_signatures.Rmd
arpoe: I have learned to use .rdat for saving R objects. I guess it is the same as .RData.

arpoe:
> Most recent failure reason is
> 
> ```
> label: unnamed-chunk-29
> Quitting from lines 402-437 (mutational_signatures.Rmd) 
> Error in dimnames(x) <- dn : 
>   length of 'dimnames' [1] not equal to array extent
> Calls: <Anonymous> ... withCallingHandlers -> withVisible -> eval -> eval -> rownames<-
> 
> Execution halted
> ```
> 
> My guess is that the first long running step results in the 8GB file that is not committed to the repo is generated using the CI files, which only include a subset of samples to save time ([docs on CI](https://github.com/AlexsLemonade/OpenPBTA-analysis#continuous-integration-ci)). But the other files that _are_ committed to the repository, and therefore exist on the machine running CI when this repository gets checked out, are from the full dataset. I haven't tested this locally yet, but if that is what is happening it might be a good idea to revise the logic such that _all_ the files need to exist so there's no chance of a mismatch.
> 
> Related note: I had to bump the timeout for this module to 90 minutes. This is likely because we're running `extract_signatures` twice. Does it make sense to take out the first one because of your comment re: hardcoding 10?

It is run twice, becuase one is with lower iteration to find the optimal number of signatures. This is the one that takes so long and the output is the graph with the connectected dots that shows the cosine similarities of the reconstructions.  
The second run is with a higher iteration and hardcoded to 10 signatures.
Once the numbers of signatures are picked, you dont really need to do the first run, but if there is a change in samples, the number 10 may have to be adjusted.

jaclyn-taroni:
Hi @arpoe - based on your responses, a few ideas come to mind for how we might use `params` in the new notebook:

* `is_ci` - a logical indicating whether or not something is being run in CI which is `FALSE` by default (or `0` and `1` depending on how it is implemented). When this is `TRUE`, we skip the long running first `extract_signatures` step where you supplied a range of values to the `nsignatures` argument of this function. This will be super helpful in keeping the CI runtime down! Pulling relevant lines from an example from elsewhere in the project for your reference:
  * In `params`: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L15 
  * A bit of error handling: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L51
  *  Logic in action: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L138
  *  Default in shell script: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/run-gsea.sh#L20
  *  Running the notebook via shell script: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/run-gsea.sh#L45
  *  Running in CI: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/.circleci/config.yml#L172

* `num_signatures` - supplied to `nsignatures` the second time you use `extract_signatures`, using 10 as the default much the same way that you do here. I'm less attached to this idea than `is_ci`, because ideally you'd have context about your choice of the number of signatures recorded in the notebook anyway.

* `check_existing_files` - as previously discussed in [this thread](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/678/files#r427384426) - including some pseudocode below with my initial thoughts on implementation:

   ```
   if (check_existing_files) {
      # get a logical that checks if all required files exist
      all_files_exist <- all(c(file.exists(first_file),
                               file.exists(second_file),
                           	 ...
                            ))
      if (all_files_exist) {
         read in all the files!
      } 
   }
   
   if (!(objects you would read in exist in the environment)) {
     generate those objects with `extract_signatures`, etc.
     and save to file
   }
   
   ```
   
   (Trying to avoid repeating code in two `else` statements - may be a better way to do this, though.)

Re:

> It is run twice, becuase one is with lower iteration to find the optimal number of signatures. This is the one that takes so long and the output is the graph with the connectected dots that shows the cosine similarities of the reconstructions.
> The second run is with a higher iteration and hardcoded to 10 signatures.
Once the numbers of signatures are picked, you dont really need to do the first run, but if there is a change in samples, the number 10 may have to be adjusted.

I would include the plot you describe in the notebook itself and put some context around your choice of 10. Not sure if `extract_signatures` will output the plot in the notebook by itself, but you have the option of using the PNG via Markdown or HTML methods.

Looking forward to your thoughts!

analyses/mutational-signatures/mutational_signatures.Rmd
```diff
@@ -368,6 +369,247 @@ lapply(unique(nature_sigs_primary$short_histology),
...
+                                        nsignatures = 5:15,
+                                        iter = 1000, 
+                                        seed = 42)
+  dev.off()
+  
+  samples_extr10 <- extract_signatures(counts = sigs_input,
+                                        nsignatures = 10,
+                                        iter = 10000, 
+                                        seed = 42)
+
```
jaclyn-taroni, line 396: I forgot to ask if this section needs a seed to be set to be fully reproducible in my initial review.

arpoe: I am not sure what you mean. There is a seed set "42" and therefore it should be reproducible.

jaclyn-taroni: Ah, sorry - totally missed that. I was wrapped up in thinking in terms of `params` that could be passed to `set.seed()`.

arpoe:
> Hi @arpoe - based on your responses, a few ideas come to mind for how we might use `params` in the new notebook:
> 
> * `is_ci` - a logical indicating whether or not something is being run in CI which is `FALSE` by default (or `0` and `1` depending on how it is implemented). When this is `TRUE`, we skip the long running first `extract_signatures` step where you supplied a range of values to the `nsignatures` argument of this function. This will be super helpful in keeping the CI runtime down! Pulling relevant lines from an example from elsewhere in the project for your reference:
>   
>   * In `params`: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L15
>   * A bit of error handling: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L51
>   * Logic in action: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/02-model-gsea.Rmd#L138
>   * Default in shell script: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/run-gsea.sh#L20
>   * Running the notebook via shell script: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/gene-set-enrichment-analysis/run-gsea.sh#L45
>   * Running in CI: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/.circleci/config.yml#L172
> * `num_signatures` - supplied to `nsignatures` the second time you use `extract_signatures`, using 10 as the default much the same way that you do here. I'm less attached to this idea than `is_ci`, because ideally you'd have context about your choice of the number of signatures recorded in the notebook anyway.
> * `check_existing_files` - as previously discussed in [this thread](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/678/files#r427384426) - including some pseudocode below with my initial thoughts on implementation:
>   ```
>   if (check_existing_files) {
>      # get a logical that checks if all required files exist
>      all_files_exist <- all(c(file.exists(first_file),
>                               file.exists(second_file),
>                           	 ...
>                            ))
>      if (all_files_exist) {
>         read in all the files!
>      } 
>   }
>   
>   if (!(objects you would read in exist in the environment)) {
>     generate those objects with `extract_signatures`, etc.
>     and save to file
>   }
>   ```
>   
>   
>   (Trying to avoid repeating code in two `else` statements - may be a better way to do this, though.)
> 
> Re:
> 
> > It is run twice, becuase one is with lower iteration to find the optimal number of signatures. This is the one that takes so long and the output is the graph with the connectected dots that shows the cosine similarities of the reconstructions.
> > The second run is with a higher iteration and hardcoded to 10 signatures.
> > Once the numbers of signatures are picked, you dont really need to do the first run, but if there is a change in samples, the number 10 may have to be adjusted.
> 
> I would include the plot you describe in the notebook itself and put some context around your choice of 10. Not sure if `extract_signatures` will output the plot in the notebook by itself, but you have the option of using the PNG via Markdown or HTML methods.
> 
> Looking forward to your thoughts!

I will have a look at this as soon as I can. Organising this with Params is a bit out of my comfort zone, so I will have a look. 
Just uploading the picture should be fine, but once the samples change it will not be correct anymore. 
The graph is automatically generated by the function.

jaclyn-taroni:
> I will have a look at this as soon as I can. Organising this with Params is a bit out of my comfort zone, so I will have a look.

Happy to help out in anyway I can!

> Just uploading the picture should be fine, but once the samples change it will not be correct anymore.
The graph is automatically generated by the function.

Yea, I think this speaks to a tension that arises when using notebooks more generally. They're great for documenting the `nsignatures = 10` decision, but then running the entire pipeline automatically becomes a problem because then all the prose is out of date. I don't have a good answer!

jaclyn-taroni:
Chatted with @arpoe - we think it's best to have someone else (likely from the CCDL) incorporate my comments and have @arpoe review a new pull request because it's important to include these improvements to mutational signature calling.

Here's what I think the process of that will look like:

1. Resolve the conflicts here (right now `Dockerfile` is the only conflicting file)
2. Someone from the CCDL will create a new branch in their own fork
3. Merge `arpoe:master` (this branch!) into the new branch
4. Response to my comments
5. File a new pull request and close this one
6. Request @arpoe for review

jaclyn-taroni:
@arpoe I am working on this today. Would it be possible for you to provide an estimate of what is required to run the `sigfit` steps for `nsignatures = 5:15` in terms of RAM, cores, and compute time?

arpoe:
I is a bit hard to say, because it really depends on the number of samples and the iterations. I think those I ran on an ordinary server without parallelisation and the number of iterations that are specified. I think it ran over night. 
Sorry for being so vage..

jaclyn-taroni:
> I think those I ran on an ordinary server without parallelisation and the number of iterations that are specified. I think it ran over night. 

No worries! Do you have a best guess about how much RAM was available on the server?

arpoe:
> > I think those I ran on an ordinary server without parallelisation and the number of iterations that are specified. I think it ran over night.
> 
> No worries! Do you have a best guess about how much RAM was available on the server?

I think it was the one with 136 CPU cores ≥190GB RAM

jaclyn-taroni:
I wanted to make a note about requirements. I was able to run [this script](https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a2701d83b0adb554ff3c4095b69cb6e96ab1d6a0/analyses/mutational-signatures/scripts/de_novo_signature_fitting.R) with the following arguments ([shell script](https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a2701d83b0adb554ff3c4095b69cb6e96ab1d6a0/analyses/mutational-signatures/run_mutational_signatures.sh)):

```sh
Rscript --vanilla \
  scripts/de_novo_signature_fitting.R \
  --maf_file ${data_dir}/pbta-snv-consensus-mutation.maf.tsv.gz \
  --nsignatures_floor 5 \
  --nsignatures_ceiling 15 \
  --num_iterations 1000 \
  --seed 42 \
  --output_file results/sigfit_signatures.RDS
```

on a Docker image built from [this `Dockerfile`](https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/a2701d83b0adb554ff3c4095b69cb6e96ab1d6a0/Dockerfile), with 30 GB of RAM available to Docker and it ran overnight. I assume that I can figure out from the `sigfit` output exactly how long it took. Here's that output: [sigfit-output.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5347730/sigfit-output.txt)

One thing to note - the last line of that output is

```
Estimated best number of signatures: 6
```

Which is different from what is included here (k = 10).  I'm not sure if that's the main (or best) way to pick the number of signatures to move forward with or not. I have yet to dig into the results.

arpoe:
Thank you very much! This is a very good summary and it is good to know what resources are needed. I have never really been limited, so I did not quite pay attention. 

The "Estimated best number of signatures: 6" is a bit of a problem. sigfit determines the optimal number of signatures based on the slope of the residuals after fitting. This automatic assignment is hardly ever appropriate and it is usually better to just make an educated guess based on the signatures and the QC plots. In this case: 10. 

I think the best solution is to hard-code a number based on the size of the cohort and expected signatures to look for in the cohort. I don't have a good solution to determine this number automatically. It might be something worth optimising in the future.

jaclyn-taroni:
I glanced at `htop` while this was going (so not very exact) and I doubt it required all 30 GB of RAM.

Thanks for the info about the number of signatures! Now that I have the results file, I'm going to adapt what you have here into a notebook for plotting, etc. 

Hopefully I will have a draft pull request shortly. I will probably also upload the results RDS file somewhere accessible at the same time so you can take a look during review.

arpoe:
Thank you so much!
The memory needs are very variable. The more signatures the more is needed, so it might spike from time to time.

jaclyn-taroni:
The code in this pull request was adapted in #799, we've gotten some PRs through, and the next steps are tracked in #818. I am going to close this PR and leave #799 open in its place.

PR #683, opened 2020-05-12
Remove d3r dependency (temporarily?)

sjspielman:

The `d3r` package (obtained from CRAN) has recently, for pending reasons, caused CI build fails related to deprecated use of `tidyselect::one_of`. Current strategy is to remove this dependency and comment out code that uses it, specifically the sunburst plots made in `analyses/sample-distribution-analysis/02-multilayer-plots.R`.

- [X ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [X ] This analysis has been added to continuous integration.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,6 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     --deps TRUE \
     R.utils \
     treemap \
-    d3r \
     hexbin \
     VennDiagram \
     Rtsne \
@@ -45,7 +44,9 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     pheatmap \
     RColorBrewer \
     viridis \
-    data.table
+    data.table # \
+   # d3r 
+
 
 # maftools for proof of concept in create-subset-files
 RUN R -e "BiocManager::install(c('maftools'), update = FALSE)"
--- a/Dockerfile
+++ b/Dockerfile
@@ -86,7 +86,7 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
 
 # Install java and rJava for some of the snv plotting comparison packages
 RUN apt-get install -y \
-    default-jdk \
+    default-jdk 
 RUN install2.r --error \
     --deps TRUE \
     rJava
--- a/Dockerfile
+++ b/Dockerfile
@@ -151,7 +151,7 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
 RUN Rscript -e "library(bedr)"
 
 # Install for mutation signature analysis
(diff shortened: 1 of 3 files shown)


jaclyn-taroni:
> Should we take this moment to clean things up a bit?

If someone has the bandwidth, yes, that makes a lot of sense to me! If not, this is intended to fix CI so I am anxious to get it through.

sjspielman:
> As this change is fairly high up in the Dockerfile, this might be an opportunity to consolidate/rearrange some of the other dependencies. Should we take this moment to clean things up a bit? Thoughts @jaclyn-taroni?

Agreed, this was a temp quick fix to unbreak CI that @jaclyn-taroni and I discussed. I have to head to various afternoon meetings now, but tomorrow I could take the time to clean this up more appropriately.

analyses/sample-distribution-analysis/02-multilayer-plots.R
```diff
@@ -145,20 +140,34 @@ interactive_tm <-
...
-    colors = color
-  )
-
-# Create an interactive sund2b plot
-p <- sunburstR::sund2b(tmnest, colors = color, valueField = "vSize")
-
 # Create HTML outputs for the interactive plots
 mapview::mapshot(interactive_tm, url = file.path(plots_dir,
                                                  "histology-treemap.html"))
 mapview::mapshot(p, url = file.path(plots_dir, "histology-pie.html"))
```
jashapiro, line 146:
```suggestion
# mapview::mapshot(p, url = file.path(plots_dir, "histology-pie.html"))
```

sjspielman:
Build now failing due to `survival_analysis/util/survival_models.R`, specifically from using `set_names` function on line 140. Trying a new commit with explicit `rlang::set_names`.
**That said**, this PR clearly needs to be done properly; we need to find the package version conflict..

jaclyn-taroni:
I am also building the Dockerfile in `master` locally after running `docker system prune --all` and then I will try running the sample distribution module that failed in the spirit of gathering more information.

Dockerfile
```diff
@@ -85,11 +85,11 @@ RUN apt-get update -qq && apt-get -y --no-install-recommends install \
     e1071
 
 # Install java and rJava for some of the snv plotting comparison packages
-RUN apt-get -y update && apt-get install -y \
-   default-jdk \
-   r-cran-rjava \
-   && apt-get clean \
-   && rm -rf /var/lib/apt/lists/
+RUN apt-get install -y \
+    default-jdk \
```
jashapiro, line 89:
```suggestion
    default-jdk 
```

jashapiro:
Superseded by #689, which is now merged. Closing.

PR #686, opened 2020-05-12
Find recurrent focal CN calls

cbethell:

The purpose of this PR is to find the _recurrent_ focal CN calls using the output of PR #644 .

This PR is working to isolate and save the _recurrent_ most focal CN calls.

#### What was your approach?

Using the `consensus_seg_most_focal_cn_status.tsv.gz` output file from PR #644, the focal CN calls were tallied by region (counting how many samples has the same call for the same region).

The final data frame was then created by keeping observations that have more than 3 samples with the same call in the same region.

Related: #486

- As mentioned above, I defined the cutoff for recurrence in this PR to be > 3. In the notebook, I provide further evidence for this decision, including a plot of the distribution of counts which should probably receive a particularly close look.
- Is the final file in an ideal format? The current format being: `status`, `region`, `region_type`, `count` -- we may/may not want to remove the count column

- I started to cross check the cytoband regions in the final recurrent data frame with the cytoband region calls in the GISTIC results but did not want to go further down that rabbit hole unless that is something we would want at this stage.

Yes. Note however that the `README.md` file will be updated once PR #644 is merged to avoid conflicts.

The output of this PR includes a TSV file with the _recurrent_ focal CN calls by region found in the module's directory at:
`results/consensus_seg_recurrent_focal_cn_units.tsv`

The output html notebook can be seen [here](https://cbethell.github.io/open-pbta-output/06-find-recurrent-calls.nb.html)

The final results includes the most focal CN calls made in PR #644 for regions that have more than 3 samples in agreement.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date. Will update once #644 is merged.


analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,95 @@
...
+  html_notebook:
+    toc: TRUE
+    toc_float: TRUE
+author: Chante Bethell for ALSF CCDL
+date: 2020
+---
+
+This notebook determines the recurrent focal copy number dominant status calls by region using the output of `05-define-most-focal-cn-units.Rmd`.
+
+Recurrence here has been arbitrarily defined based on the plotting of the distribution of status calls and a [similar decision](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d197f2717eb0d067fe35644db91c4910b6d# 46c06/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L111) made in `analyses/fusion_filtering/06-recurrent-fusions-per-histology.R`
```
cansavvy, line 13: This link isn't working for me.

cansavvy: Is the similar decision you are referring to here the distribution plot or something else? If it's something else can you describe this here?

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,95 @@
...
+
+## Find recurrent calls
+
+### Tally the sample counts of each unique status call by region.
+
+```{r}
+# Tally the number of samples that have the same status call by region
+final_df_tally <- focal_cn_df %>%
+  group_by(status, region, region_type) %>%
+  tally() %>%
```
cansavvy, line 53:
```suggestion
  tally(name = "count") 
```
Then you can get rid of the renaming step in the next line.

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,95 @@
...
+  arrange(desc(count))
+```
+
+### Plot the distribution of counts
+
+```{r}
+# Plot the distribution of counts
+ggplot(final_df_tally, aes(x = count)) +
+  geom_histogram(bins = 100) +
+  geom_vline(aes(xintercept = 3), linetype = "longdash", color = "red", size = .4)
```
cansavvy, line 67: May want to change the color here so it's easier to see the line.

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,95 @@
...
+```
+
+### Isolate recurrent calls
+
+```{r}
+# Here, we arbitrarily define recurrence to be greater than 3 based on 
+# the plot above and on a similar decision made in
+# `analyses/fusion_filtering/06-recurrent-fusions-per-histology.R`
+# https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d197f2717eb0d067fe35644db91c4910b6d# 46c06/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L111
```
cansavvy, line 76: This link also doesn't work for me.

analyses/focal-cn-file-preparation/run-prepare-cn.sh
```diff
@@ -44,58 +47,64 @@ Rscript --vanilla 04-prepare-cn-file.R \
...
-#           --independent_specimens_file $independent_specimens_file \
-#           --metadata $histologies_file \
-#           --goi_list $goi_file \
-#           --filename_lead ${filename}_${chromosome_type}_${strategy}
-#       done
-#     done
-#   done
-# 
-# fi
+# Define most focal units of recurrent CNVs
```
cansavvy, line 50: Why is this being uncommented now?

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
cansavvy: Can you add a bit more rationale about how the cutoff has come to be? You may also want to add a a thing in the set up where we can easily change the cutoff if needed.

cbethell: The similar decision made here is the cutoff. I can fix the link and add a better description here.

analyses/focal-cn-file-preparation/run-prepare-cn.sh
cbethell: This was done in PR #644 (it was originally only commented out for developmental purposes).

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
cbethell: I updated this description in the last commit and in the rendered notebook [here](https://cbethell.github.io/open-pbta-output/06-find-recurrent-calls.nb.html). Let me know if this is along the lines of what you were looking for!

cbethell:
@cansavvy I am re-requesting your review because I adjusted the final output file named `results/consensus_seg_recurrent_focal_cn_units.tsv` to include the biospecimen IDs (instead of the tallied data frame without the sample IDs).

The reason behind this change is that this format would likely be more useful for downstream analyses, as I have found in the oncoprint analysis in PR #666.

The updated rendered notebook can be seen [here](https://cbethell.github.io/open-pbta-output/06-find-recurrent-calls.nb.html).

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -83,17 +83,27 @@ The `recurrence_cutoff` defined as 3 appears to be reasonable as seen in the dis
...
 
 # Display table
-final_df_recurrent
+recurrent_df
+```
+
+### Filter focal CN data frame to include only recurrent observations
+
+```{r}
+# Filter the focal CN data frame to include only the recurrent observations
```
cansavvy, line 96: I don't think this is exactly what you want. I * think* you want to have `Kids_First_Biospecimen_ID` to be in your `group_by()` step in the first place. That way the tally's are also by sample.

cbethell: If I do that, the count will be 1 for all observations. I _believe_ we want to find out which status calls occur in multiple samples for a specific region.

cansavvy: Ah I see. Yes you're right. I lost track of the goal here.

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,114 @@
...
+# Display table
+recurrent_df
+```
+
+### Filter focal CN data frame to include only recurrent observations
+
+```{r}
+# Filter the focal CN data frame to include only the recurrent observations
+final_df_recurrent <- focal_cn_df %>%
+  select(-region_type) %>%
```
cansavvy, line 98: Why are we getting rid of region_type here?

cbethell: I did this to avoid the duplicate column that would come from the `inner_join` step below (there's probably a better way to do this here).

cansavvy: This is fine to me. Maybe just add a comment to explain why.

cbethell:
@cansavvy would you mind taking one last look at this? I added the comment as you suggested in your [previous review](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/686#discussion_r425767036) and also updated the `README.md` file with a few lines regarding the notebook in this PR.

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
jaclyn-taroni: I _think_ including `region_type` in your `by` below would accomplish the same thing and potentially be a little friendlier to your reader
```R
   inner_join(recurrent_df, by = c("region", "status",  "region_type"))
```

cbethell: Ah, I'll give that a try @jaclyn-taroni, thank you!

jaclyn-taroni:
I was looking to merge #666 and noticed `analyses/focal-cn-file-preparation/results/consensus_seg_recurrent_focal_cn_units.tsv` being added there as well as in this pull request. Can you explain the format of this file @cbethell?

cbethell:
> I was looking to merge #666 and noticed `analyses/focal-cn-file-preparation/results/consensus_seg_recurrent_focal_cn_units.tsv` being added there as well as in this pull request. Can you explain the format of this file @cbethell?

@jaclyn-taroni right, this file is in the same format as `analyses/focal-cn-file-preparation/results/consensus_seg_most_focal_cn_status.tsv` but including only those calls in regions that are common amongst more than three samples. Should I instead use the more general `analyses/focal-cn-file-preparation/results/consensus_seg_most_focal_cn_status.tsv` file in PR #666? There was a very slight difference in the visualization when I ran the plot using each of the two files.

jaclyn-taroni:
I think it's okay to filter to recurrent CNVs for the OncoPrint (you may want to control that through the gene list argument). I would expect the OncoPrint to accept something that has the format:
 
| ID | status | region |
|---|--------|-------|

And I would expect a table that keeps track of recurrence to be more like:

| region | status | count |
|-------|-------|--------|


Which you could filter to n > 3. What I am describing is similar to this file (but in this example things are counted by _histology_): https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0985505eb18402ecac925b2038764fb2ddc2a712/analyses/fusion_filtering/results/pbta-fusion-recurrent-fusion-byhistology.tsv

You could probably incorporate histology information into that table that also includes total counts (labeled `All` below):

| region | status | short_histology | count |
|-------|--------|----------------|------|
| 7q | gain | LGAT | n |
| 7q | gain |  All | n |

cbethell:
> I think it's okay to filter to recurrent CNVs for the OncoPrint (you may want to control that through the gene list argument to `01-plot-oncoprint`). I would expect the OncoPrint to accept something that has the format:
> 
> ID	status	region
> And I would expect a table that keeps track of recurrence to be more like:
> 
> region	status	count
> Which you could filter to n > 3. What I am describing is similar to this file (but in this example things are counted by _histology_): https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0985505eb18402ecac925b2038764fb2ddc2a712/analyses/fusion_filtering/results/pbta-fusion-recurrent-fusion-byhistology.tsv
> 
> You could probably incorporate histology information into that table that also includes total counts (labeled `All` below):
> 
> region	status	short_histology	count
> 7q	gain	LGAT	n
> 7q	gain	All	n

Gotcha, so I'll change the output file in this PR to not include the sample IDs but to include the histology information. But first, I will change the file used in PR #666 to be the `analyses/focal-cn-file-preparation/consensus_seg_most_focal_cn_status.tsv` file.

jaclyn-taroni:
> But first, I will change the file used in PR #666 to be the `analyses/focal-cn-file-preparation/consensus_seg_most_focal_cn_status.tsv` file which will then be updated with the list of recurrent regions once this PR is merged.

The first change sounds good. For the list of recurrent regions, I'd encourage you to (in an upcoming PR) do either one of these options:

* Accept a comma-delimited set of gene lists that get essentially concatenated in `01-plot-oncoprint`
* Have a step that puts all the gene lists together - this could be good as a notebook so you put lots of explanation around the lists.

I would pick an option based on what would be easier to adapt if we were to then split the OncoPrints by histology, rather than having them all in one.

jaclyn-taroni:
A few things I noticed when looking at the notebook locally:

* I would filter out the uncallable regions. We would expect those to be recurrent and they are not necessarily of interest here.
* Without in-depth knowledge of the steps to obtain the uncallable status calls, I might expect that uncallable counts = total number of samples. That is not always the case - is this expected based on upstream steps?

cbethell:
> * I would filter out the uncallable regions. We would expect those to be recurrent and they are not necessarily of interest here.

Good call, will do 👍 

> * Without in-depth knowledge of the steps to obtain the uncallable status calls, I might expect that uncallable counts = total number of samples. That is not always the case - is this expected based on upstream steps?

I'll take a deeper look into the upstream steps because the expected behavior is as you described above, uncallable counts for a region should be equal to the total number of samples because if an arm is uncallable, all samples for that arm should be uncallable. This would also mean that all samples in any cytoband on that arm should also have an uncallable status.

analyses/focal-cn-file-preparation/05-define-most-focal-cn-units.Rmd
```diff
@@ -155,7 +155,7 @@ filtered_consensus_cytoband_status <- consensus_seg_cytoband_status %>%
   ) %>%
   # Filter the annotated CN data to include only neutral chromosome arms and disagreements
   filter(
-    dominant_arm_status %in% c("neutral", "uncallable", "unstable") |
+    dominant_arm_status %in% c("neutral", "unstable") &
```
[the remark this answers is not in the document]
cbethell: The filtering step here was taking the all `neutral`, `unstable`, and `uncallable` arms and providing the cytoband status for these arms. Instead, we wanted to take provide only the `neutral` and `unstable` arms to this step so that we can get the cytoband status for these arms and leave the `uncallable` arms.

That being said, for each uncallable region, all samples are now labeled `uncallable`.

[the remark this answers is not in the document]
jashapiro: As I originally understood the logic here, it was possible to have an `uncallable` arm that contained callable cytoband (i.e. an arm that was 70% uncallable could still have many callable bands), so we wanted to include those.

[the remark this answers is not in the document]
cbethell: Ah okay, that's reasonable. I'll revert the change to the logic. Re the comment above:
> Without in-depth knowledge of the steps to obtain the uncallable status calls, I might expect that uncallable counts = total number of samples. That is not always the case - is this expected based on upstream steps?

I believe the reason the total number of samples for an uncallable region may not add up is because all samples may not have data for that region? (I am trying to track back and look at the raw data and how many samples that have data for each region)

Were the inconsistent regions the X and/or Y chromosomes? The `Xp` arm is where I noticed the inconsistency described.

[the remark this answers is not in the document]
jaclyn-taroni: > Were the inconsistent regions the X and/or Y chromosomes? The Xp arm is where I noticed the inconsistency described.

The short answer is I don't remember - looks like you just reverted this change, is there anything you would consider an inconsistency?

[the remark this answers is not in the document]
cbethell: So thinking about the logic a little more and @jashapiro's comment,

For sample X, cytoband 11q11 may be `uncallable` because more than 50% of the region is uncallable for that sample.

For sample Y on the other hand, cytoband 11q11 may be `loss` because more than 90% of the region for that sample is loss.

This can only be done for samples that have copy number data for that specific region (hence some regions may not have data for all 915 samples). I used the cytoband data for an example, but the same goes for the arm level calls, although arm level calls would be expected to have more agreement between samples due to the region being broader? (this is just a guess).

So based on that logic, I do not believe that the uncallable counts would necessarily be equal to the total number of samples, does that make sense?

That being said, it appears that the `Xp` and `Xq` arms are the only chromosome arms to have less samples than the total number of samples that have been labeled `uncallable`, but they are no other competing calls for these arms (by this I mean, `uncallable` is the only unique label in the `Xp` and `Xq` regions).

analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd
```diff
@@ -0,0 +1,189 @@
...
+# analyses
+top_recurrent_genes <- final_df_recurrent %>%
+  # Ungroup data from previous step
+  ungroup() %>%
+  # Arrange by descending count
+  arrange(desc(count)) %>%
+  # Filter to only gene regions
+  filter(region_type == "gene") %>%
+  # Select the top genes by count
+  top_n(20, count) %>%
```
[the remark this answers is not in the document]
cbethell: @jaclyn-taroni if I move this step before filtering to only gene regions, it will return the top `n` for all regions. Can you further clarify what you mean here?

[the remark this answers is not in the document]
jaclyn-taroni: The number 20 is hardcoded here, assign 20 to a variable up top.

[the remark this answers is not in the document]
cbethell: Ah gotcha 👍

Issue #400, opened 2020-01-03
Proposal to update docker container


sjspielman:
I am proposing we modify the project to use `rocker/tidyverse-3.6.1` rather than `3.6.0`. The primary benefit to this is access to features under [`tidyverse 1.0.0`](https://blog.rstudio.com/2016/09/15/tidyverse-1-0-0/). This version is almost entirely backwards compatible, while allowing new and very useful features (such as [tidyr 1.0.0](https://www.tidyverse.org/blog/2019/09/tidyr-1-0-0/) functions ). 

If we were to make this change, I volunteer to address any emerging bugs. If not, is there any specific reason we want to keep `3.6.0`?

PR #687, opened 2020-05-12
WIP: Update R to 3.6.2

jashapiro:

Since we seem to have run into some new dependency troubles, we might take the opportunity to update to a more current base image, including R 3.6.2. 

#### What was your approach?

I changed the base image in the Dockerfile, and while I was at it, moved a number of the more commonly used R packages to earlier layers.

#400 

Does everything run? I am assuming there will not be major changes in results, but that might be worth checking.

More reorganization might be warranted, but at this stage, the question is mostly about how much this will break. If this is successful, it might be worth a bigger reorg before merging.

Nope. Also No. Certainly not.

#### Documentation Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,9 +9,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends apt-utils dialo
 # Add curl
 RUN apt-get -y --no-install-recommends install curl
 
-# Install pip3
+# Install pip3 and instalation tools
 RUN apt-get -y --no-install-recommends install \
     python3-pip  python3-dev
+RUN pip3 install "setuptools==46.2.0" "wheel==0.34.2" "six==1.14.0"
 
 # Required for installing htslib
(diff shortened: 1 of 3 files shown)


sjspielman:
Let's see how the current build for PR #683 goes, and then perhaps worth closing it to centralize efforts here to fix the surprising mystery version conflicts? @jashapiro , any preference?

jashapiro:
> Let's see how the current build for PR #683 goes, and then perhaps worth closing it to centralize efforts here to fix the surprising mystery version conflicts? @jashapiro , any preference?

I think we want to get #683 done if we can figure it out quickly, as properly merging this will require rerunning _everything_. So I would vote to maintain these in parallel at least for now.

sjspielman:
If the issue turns out to be `r-cran`, #683 is moot anyways and should be deleted, with this one taking precedent.

jashapiro:
Closing for now. May return in the future.

PR #689, opened 2020-05-14
Install Bioconducter packages with MRAN lock

jashapiro:
#### What scientific question is your analysis addressing?

We have been using MRAN snapshots to improve reproducibility of installed R packages. 
Unfortunately, `BiocManager::install()` ignores the set CRAN repository, and installs from https://cran.rstudio.com when a package or dependency is not available in the Bioconductor repository. 
This means that dependencies of Bioconductor packages can change versions without warning, breaking things.

#### What was your approach?

I have added a script, `install_bioc.r` that uses the set Bioconductor repository and CRAN mirror (in our case MRAN) to install packages listed on the command line. 
The Dockerfile then uses this for installation, replacing invocations of `BiocManager::install()` invocations.

This PR also includes some cleanup and organization of the Dockerfile, removing some redundancies and adjusting the `.dockerignore` file to ignore the `.git` directory for dramatically faster build start (full builds still take a long time!)

#### What GitHub issue does your pull request address?

Everything was broken. But this is all a direct result of #684 

#### Which areas should receive a particularly close look?

The `install_bioc.r` script is currently quite minimal, and probably could use some error handling. 
I considered modifying install2.r, but decided to go for the minimal solution for licensing reasons in this repo.

#### Is there anything that you want to discuss further?

If it passes CI, I will be so happy.

#### Reproducibility Checklist

- [ ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

PR #695, opened 2020-05-18
WIP: Update R to 3.6.2

jashapiro:

After all the work that went into #689, I figured I would try to do a cleaner update to 3.6.2 to see how that went.

#### What was your approach?

Change the base rocker image to 3.6.2. 
Make changes as necessary for compatibility with the updated version of debian that image includes. 
Most notably, this includes updating some python packages for python 3.7 compatibility.
While I was doing that, I made all python packages explicitly versioned.

#400

Any new warnings in modules and testing should be noted and fixed either by updating code for compatibility with new package versions, or by changing package versions for improved compatibility.

The changes here will ultimately require full rerun of all analyses to ensure that any underlying code changes do not affect results (or do so only minimally). 

#### Documentation Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -237,24 +237,90 @@ RUN R -e "remotes::install_github('wilkox/treemapify', ref = 'e70adf727f4d13223d
 # Install python libraries
 ##########################
 
-# Install python3 data science tools
+# Install python3 data science tools and all dependencies (versioned)
 RUN pip3 install \
-    "cycler==0.10.0" "kiwisolver==1.1.0" "pyparsing==2.4.5" "python-dateutil==2.8.1" "pytz==2019.3" \
+    "MarkupSafe==1.1.1" \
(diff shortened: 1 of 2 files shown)


jashapiro:
Current error: 

```
R[write to console]: Error: package or namespace load failed for ‘methods’ in dyn.load(file, DLLpath = DLLpath, ...):
 
 unable to load shared object '/usr/local/lib/R/library/methods/libs/methods.so':
 
  libR.so: cannot open shared object file: No such file or directory
```

Strangely, this did not seem to happen in my local build. Will require more careful debugging to figure out why `rpy2` is failing in this case. 

Ideas for future exploration:
- Is the user different between `run_in_ci.sh` and running in the image via rstudio login?
- Are R-related environment variables different?

jaclyn-taroni:
I'm going to close this for now. We can always revisit.

PR #694, opened 2020-05-18
CN status heatmap: Final figure

cansavvy:
#### What was your approach?
In the `generate-figures.sh` script, I added two more steps to a new section for CN status heatmap: 
1) Run the `copy_number_consensus_call/run_consensus_call.sh` script
2) Run `cn_status_heatmap.Rmd` with an added `final_figures` option which saves the output file to the `figures/png` folder. 

#### What GitHub issue does your pull request address?
#594 

#### Which areas should receive a particularly close look?
- I still have it saving to a `pdf` but the rest of the images are pngs and are saved to a `png` folder? Should I change the image to be a `png` image or change the name of the folder? 
- Any other adds we need for the CN status figure? 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yup. 

#### What types of results are included (e.g., table, figure)?

`pdf` file in `figures/png`

#### Reproducibility Checklist
`generate-figures.sh` script has not been added to CircleCI.

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist
- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #696, opened 2020-05-18
CN status heatmap: Final figure

cansavvy:
Copied from original #694, but branches were messed up. 

#### What was your approach?
In the `generate-figures.sh` script, I added two more steps to a new section for CN status heatmap: 
1) Run the `copy_number_consensus_call/run_consensus_call.sh` script
2) Run `cn_status_heatmap.Rmd` with an added `final_figures` option which saves the output file to the `figures/png` folder. 

#594 

- I still have it saving to a `pdf` but the rest of the images are pngs and are saved to a `png` folder? Should I change the image to be a `png` image or change the name of the folder? 
- Any other adds we need for the CN status figure? 

Yup. 

`pdf` file in `figures/png`

`generate-figures.sh` script has not been added to CircleCI.

#### Documentation Checklist


jashapiro:
Copying my comment from #694:

I ran into the pdf/png question earlier. Manubot does not support pdfs, so we should probably either make both (with the pdf in a separate folder) or just the png for now, since that can be used in the manuscript.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -441,7 +443,7 @@ Save to PDF.
 
 ```{r}
 # Save plot as PDF
-pdf(file.path(plots_dir, "cn_status_heatmap.pdf"))
+png(file.path(plots_dir, "cn_status_heatmap.png"))
```
cbethell, line 446: Also, have you tried using the `res` argument of the `png` function with this plot as yet? The heatmap, as a `PNG`, is not as clear as we would necessarily want it I _think_ (the labels are a little difficult to read). I do not believe that the quality now is terrible but if we can improve it a bit I think it's worth improving.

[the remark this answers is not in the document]
cansavvy: Good call!

cansavvy: I did play with the `res` argument a bit. I don't think I pushed that change though. Will do that.

sjspielman:
Hey @cansavvy, question for you - I'm having a near impossible time getting this to run in a human time frame. Even with CI data, I'm experiencing so far >25 minute hang at Chunk 17 (`# Run call_bin_status for each biospecimen's segments.`). It runs `lapply()` so long time could be expected, but this is _long_, and no apparent end in sight yet. Do you also experience this when fully knitting?

cansavvy:
> Hey @cansavvy, question for you - I'm having a near impossible time getting this to run in a human time frame. Even with CI data, I'm experiencing so far >25 minute hang at Chunk 17 (`# Run call_bin_status for each biospecimen's segments.`). It runs `lapply()` so long time could be expected, but this is _long_, and no apparent end in sight yet. Do you also experience this when fully knitting?

It does take a bit of time to run the calculations (at Chunk 17). It's doing some GenomicRanges binning and overlapping for each sample. But I don't usually have much problem running locally. It does complete, just takes a long time.

cansavvy:
> It does take a bit of time to run the calculations (at Chunk 17). It's doing some GenomicRanges binning and overlapping for each sample. But I don't usually have much problem running locally. It does complete, just takes a long time.

I was debating about having the output calculations save to an RDS file so that step doesn't need to be run EVERY time. Let me know if you think that'd be helpful. 

For reference, chunk 17 is running this function for each sample: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d5b13f28a82bce0f098a9c3fb808a10a8e1aef48/analyses/cnv-chrom-plot/util/bin-coverage.R#L56 If you have ideas of how to make it quicker, I'm definitely on board, but I think it may just be a long step no matter what.

sjspielman:
I made it through the woods in 24 minutes over here w/ testing data. I think RDS file would helpful here, certainly not harmful. Progress messages would *definitely* help from a psychological standpoint ;)

sjspielman:
It may help to define variables in a top chunk associated w/ `levels = c(1:22, "X", "Y", "M")` used all around, and in certain places it's only `levels = c(1:22, "X", "Y")` - I'm not at all saying the lack of "M" is a bug as I haven't done a line-by-line run of the code, but cleaning up the hardcoding will make me extra sure it's not.

cansavvy:
@sjspielman , Thanks for this feedback! I've incorporated most of your suggestions. Let me know what you think! The thing I haven't addressed is:

>OR, maybe two versions of figure - one with all subtypes, and one with only the dominant ones? 

(I'm gonna table your OR OR suggestion 😄)

I think based on your feedback and some other feedback from our CCDL science team, my plan is to separate the heatmap into two panels: histologies with big sample sizes, and histologies with small sample sizes. My initial sloppy trials of this looked pretty good, however to make it actually come to life will require more work and changes than this PR set out to do. 

So for now, I'll write this down in an issue, and come back to it at a later point before I get too far down a rabbit hole.

sjspielman:
I am a big fan of that redesigned figure!!! It's much clearer and a great baseline for next steps!

jashapiro:
@cansavvy It looks like there is some kind of little bug in your code to draw the annotation lines that shows up for some of the smallest groups. _Chondrosarcoma_ is the first one it shows up for, but there are others: looks like the cases where n=2, with the addition of _CNS Rhabdomyosarcoma_, which says n = 4, but looks much narrower than _LGMT_, which is also n = 4, so is perhaps miscounted in the label?

cansavvy:
> draw the annotation lines that shows up for some of the smallest groups.

This looks like a ComplexHeatmaps problem, but I think it should be resolved by splitting up into two panels. But that will be the next PR.

cansavvy:
> > draw the annotation lines that shows up for some of the smallest groups.
> 
> @jashapiro This looks like a ComplexHeatmaps problem, but I think it should be resolved by splitting up into two panels. But that will be the next PR.

I am wrong about this. Looking into this problem some more.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -271,31 +279,39 @@ Run the bin status calling on each sample.
...
-bin_calls_list <- lapply(sample_ids,
-  call_bin_status,
-  bin_ranges = bins,
-  seg_ranges = filtered_seg_ranges,
-  uncallable_ranges = uncallable_ranges,
-  frac_threshold_val = frac_threshold,
-  frac_uncallable_val = frac_uncallable
-)
+# Read in the calculation file if it was already made and this isn't the final figure
+if (file.exists(file.path(results_dir, "cn_status_bp_per_bin.tsv")) & !params$final_figure) {
```
cansavvy, line 283: Can someone double check this logic for me? I basically am trying to make sure when we run this for the "final_figure" (see generate_figures.sh adds) that this always recalculates.

sjspielman:
> @cansavvy It looks like there is some kind of little bug in your code to draw the annotation lines that shows up for some of the smallest groups. _Chondrosarcoma_ is the first one it shows up for, but there are others: looks like the cases where n=2, with the addition of _CNS Rhabdomyosarcoma_, which says n = 4, but looks much narrower than _LGMT_, which is also n = 4, so is perhaps miscounted in the label?

Arg yes I see this now - do I need glasses finally? :(

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
sjspielman: Yep, on it.

cansavvy:
> > @cansavvy It looks like there is some kind of little bug in your code to draw the annotation lines that shows up for some of the smallest groups. _Chondrosarcoma_ is the first one it shows up for, but there are others: looks like the cases where n=2, with the addition of _CNS Rhabdomyosarcoma_, which says n = 4, but looks much narrower than _LGMT_, which is also n = 4, so is perhaps miscounted in the label?

I also expect that this should be fixed now with the most recent commits. `n = 1`'s don't show up but I will address that in a future PR before this one becomes too bloated.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
cansavvy: TY!

sjspielman:
Something else that I'm just now seeing that isn't specifically part of this PR but merits discussion - (current) line 198: 
`filtered_seg_ranges <- seg_ranges[which(seg_ranges@ranges@width < length_max)]`

This is an issue for consistency among all PBTA analyses, specifically the choice of `<` rather than `<=`. Similarly lines 143 - 149 in the utils script. @jaclyn-taroni @jashapiro has this been specifically discussed  for how thresholds should universally be treated in the project as a whole?

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
sjspielman: This hurt my brain way more than it should have. It seems right to me though - 

IMG_1922

jaclyn-taroni:
> @jaclyn-taroni @jashapiro has this been specifically discussed for how thresholds should universally be treated in the project as a whole?

I'm not sure I have all the context here, but I don't know how to have a universal discussion of thresholds - it depends on the analysis.

sjspielman:
> > @jaclyn-taroni @jashapiro has this been specifically discussed for how thresholds should universally be treated in the project as a whole?
> 
> I'm not sure I have all the context here, but I don't know how to have a universal discussion of thresholds - it depends on the analysis.

Good enough for me. The idea is pretty general. In my mind, if we want to exclude certain genomic regions from any analysis *and some analyses are focusing on analyzing the same genomic regions*, that should be consistent. E.g. if we have 5 analyses filtering out uncallable regions at threshold of 0.75, all should be "keep >=0.75" (aka remove < 0.75), and NOT some "keep > 0.75" (aka remove <=0.75). It seems to me that the project is *so* well-organized into analyses that this is unlikely to happen. Let me think more on this and if necessary I'll open an issue.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
cansavvy: Yes! This makes sense to me! Thanks for helping confirm!

jaclyn-taroni:
> E.g. if we have 5 analyses filtering out uncallable regions at threshold of 0.75, all should be "keep >=0.75" (aka remove < 0.75), and NOT some "keep > 0.75" (aka remove <=0.75). It seems to me that the project is so well-organized into analyses that this is unlikely to happen. 

Gotcha- my knee-jerk reaction to that is if we have a situation like that that you find, the solution is to come up with a "single source of truth" file that gets distributed as part of the data download and used in each of the analyses, rather than changing thresholds within analyses - does that make sense/have I expressed myself with sufficient clarity?

> Let me think more on this and if necessary I'll open an issue.

Sounds good!

sjspielman:
> Gotcha- my knee-jerk reaction to that is if we have a situation like that that you find, the solution is to come up with a "single source of truth" file that gets distributed as part of the data download and used in each of the analyses, rather than changing thresholds within analyses - does that make sense/have I expressed myself with sufficient clarity?

Yes, this was my thought. Clarity is not my forte right now, should have said that clearly :)

sjspielman:
@cansavvy small other request to followup on my apparently pre-mature review approval - should chunk 24 `all.equal` have an associated `stop()` ?

cansavvy:
> @cansavvy small other request to followup on my apparently pre-mature review approval - should chunk 24 `all.equal` have an associated `stop()` ?

Yes. I like this idea, but I will implement it in the next PR before this one gets too hairy!

sjspielman:
> Yes. I like this idea, but I will implement it in the next PR before this one gets too hairy!

Something else for next PR - save the `ComplexHeatmap::draw(heatmap, heatmap_legend_side = "bottom")` output to a variable which can be both printed out w/in the Rmd and saved to a file. Saves a little run time.

sjspielman:
> I also expect that this should be fixed now with the most recent commits. `n = 1`'s don't show up but I will address that in a future PR before this one becomes too bloated.

This is right on all counts (PUN ALERT) - numbers match width, n=1 and to some extent n=2 not visible, but that's a "later problem." I love later problems. How to deal with "Other" histology category is similarly a later problem.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -420,14 +451,19 @@ heatmap <- ComplexHeatmap::Heatmap(
...
   show_column_names = FALSE,
   show_row_names = FALSE,
+  row_labels = FALSE,
   bottom_annotation = chr_annot,
   right_annotation = hist_annot,
-  heatmap_legend_param = list(nrow = 1),
-  raster_quality = 8
+  heatmap_legend_param = list(nrow = 1, border = "black"),
+  raster_quality = 8, 
+  border = "black",
```
jashapiro, line 465: Changing this to `border=FALSE` may make the n=1 cases visible, at the expense of the pretty borders.

cansavvy: Yes. It does. But the question is do we still even want the n=1 on this full version anyway?

jashapiro: At this stage, I think we do want to include them. Alternatively, another quick fix might be to lump them into the "Other" group.

cansavvy: I like that for now. I'll push a quick PR for that.

Issue #690, opened 2020-05-14
Remove any R package installation from notebooks or scripts, especially with BiocManager


jaclyn-taroni:
Context: #689 - which is adding a script to specifically install Bioconductor packages on the project Docker container.

Having logic that checks whether or not a package is installed and if not installs it is reasonable if someone were to run an analysis outside the Docker container. But if we were to miss something in the Docker image and then run an analysis it could end up unpinning versions (as discovered on the journey of #684). So, we should remove any package installation from notebooks or scripts, especially any installation via `BiocManager`.

Here's a list from @sjspielman:

```
focal-cn-file-preparation/04-prepare-cn-file.R:  install.packages("BiocManager")
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("GenomicRanges", update = FALSE)
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("IRanges", update = FALSE)
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("annotatr", update = FALSE)
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("TxDb.Hsapiens.UCSC.hg38.knownGene", update = FALSE)
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("org.Hs.eg.db", update = FALSE)
focal-cn-file-preparation/04-prepare-cn-file.R:  BiocManager::install("AnnotationDbi", update = FALSE)
gene-set-enrichment-analysis/01-conduct-gsea-analysis.R:## Load and/or install libraries ##
gene-set-enrichment-analysis/01-conduct-gsea-analysis.R:  install.packages("msigdbr")
gene-set-enrichment-analysis/01-conduct-gsea-analysis.R:  install.packages("BiocManager")
gene-set-enrichment-analysis/01-conduct-gsea-analysis.R:  BiocManager::install("GSVA")
molecular-subtyping-ATRT/02-ATRT-molecular-subtyping-plotting.R:  install.packages("ComplexHeatmap")
molecular-subtyping-ATRT/02-ATRT-molecular-subtyping-plotting.R:  install.packages("matrixStats")
molecular-subtyping-ATRT/02-ATRT-molecular-subtyping-plotting.R:  install.packages("ggfortify")
oncoprint-landscape/01-plot-oncoprint.R:  install.packages("BiocManager")
[...]

PR #698, opened 2020-05-18
Remove R script/notebook package installs (Part 1 of N) 

cansavvy:
#### What was your approach?
I'm removing R package installations in the scripts in accordance with #690 
I'm keeping track of the PRs I file for that issue here and going through the modules alphabetically. 

This PR is me having gone through `chromosomal-instability` - `focal-cn-file-preparation`, here are the changes in this PR: 

- cnv-chrom-plot: In `gistic_plot.Rmd` got rid of:  `install.packages("BiocManager")` and `BiocManager::install("ggbio")` and refreshed `gistic_plot.Rmd` only. 
- cnv-comparison: is deprecated, so I didn't get rid of the install.packages there
- focal-cn-file-preparation: Got rid of series of installs in `04-prepare-cn-file.R`, re-ran ` bash run-prepare-cn.sh`.

Everything look okay? 

Yes, but the results shouldn't really be different. 

No changes are needed here. 


cansavvy:
> These changes look just about ready to merge 👍
> 
> I am just concerned about what appears to be the removal of a plot: `focal-cn-file-preparation/plots/consensus_seg_cn_autosomes_stranded_stacked_plot.png`
> Any idea why this may be the case here?

The script seemed to finish without any errors, but I'm also not super familiar with this code. This may be something I could use your help investigating, @cbethell . What notebook/script creates that plot?

cbethell:
> > These changes look just about ready to merge 👍
> > I am just concerned about what appears to be the removal of a plot: `focal-cn-file-preparation/plots/consensus_seg_cn_autosomes_stranded_stacked_plot.png`
> > Any idea why this may be the case here?
> 
> The script seemed to finish without any errors, but I'm also not super familiar with this code. This may be something I could use your help investigating, @cbethell . What notebook/script creates that plot?

The script that creates that plot is the `rna-expression-validation.R` script of the `focal-cn-file-preparation` module. It's strange to me that it is just that one plot that is missing. I can look into it on your branch to see if I could figure out what's going on there.

cansavvy:
@cbethell , I'm looking [at this line](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/focal-cn-file-preparation/run-prepare-cn.sh#L65): 

```
--filename_lead "consensus_seg_annotated_cn"_${chromosome_type}_${strategy}
```
And don't think this file `consensus_seg_cn_autosomes_stranded_stacked_plot.png`  which doesn't have the `annotated` bit in there, would be saved. Is there somewhere else that you expect this file would be saved from? 

As of right now, my best guess is that that file is from an older edition where before you added the `annotated` part in the name. Let me know if there's somewhere else I should be looking.

cansavvy:
My other guess is maybe you had previously included the consensus file in this next loop: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/8ac1772145833d015483a4ceb5249c77c9bd68a6/analyses/focal-cn-file-preparation/run-prepare-cn.sh#L92

Let me know if either of my guesses ring a bell, or if not, where in the code you would expect that file should be saved from.

cansavvy:
I don't think either of my theories regarding `consensus_seg_annotated_cn_autosomes_stranded_stacked_plot.png` are corrrect. I think it should have been saved. My best guess now is that a step got skipped because of memory intensive requirements. I think my plan is to run this on a bigger boat.

cansavvy:
@cbethell I re-ran the `run-prepare-cn.sh` script on AWS. It looks like those files were created this time, but can you take another look to make sure everything else with `focal-cn-file-preparation` looks as you would expect?

Issue #148, opened 2019-10-04
Proposed Analysis: quantify telomerase activity across pediatric brain tumors


syzheng:
#### Scientific goals

_The goal is to quantify telomerase activity and correlate them with telomere length and molecular alterations (TERTp mutation, ATRX mutation, etc)_

#### Proposed methods

_We will use our newly developed method EXTEND (EXpression based Telomerase ENzymatic activity Detection)_

#### Required input data

_Gene expression from RNAseq (either of TPM, RPKM, or counts)_

#### Proposed timeline

_One to two weeks._

#### Relevant literature

_Barthel et al. Nat Genet, 2017; Zheng et al. Cancer Cell, 2016; Ackermann et al. Science, 2016_

cgreene:
This sounds exciting! As I think about potential caveats, does it matter if the RNA-seq samples are poly-A selected or rRNA depleted?

syzheng:
> This sounds exciting! As I think about potential caveats, does it matter if the RNA-seq samples are poly-A selected or rRNA depleted?

that is a great point. we currently use data from regular polyA enriched protocol, mostly because our primary input data is TCGA. this does impact the method, because a key gene in our signature, TERC, is a non-coding RNA that is not properly captured by polyA methods. PCR shows this gene is abundantly expressed across tissues; however, RNAseq data from TCGA and GTEx only show very low expression of this gene. EXTEND demonstrates reasonable performance with both TCGA, CCLE and GTEx, but We have not tested data from total RNAseq or rRNA depletion. Great point.

cgreene:
@syzheng : Ok! This dataset contains both poly-A and rRNA depleted samples. #120 and https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/selection-strategy-comparison took a dive into the implications for gene expression analyses based on some earlier work by @cbethell.

I'm a bit confused by "a key gene in our signature, TERC, is a non-coding RNA that is not properly captured by polyA methods" and also "we currently use data from regular polyA enriched protocol, mostly because our primary input data is TCGA". Did you mean that you are better off with poly-A? There are many fewer poly-A samples here than rRNA-depleted, and @jharenza is looking to determine whether or not we can generate some that are matched (sequenced with both protocols).

syzheng:
[...]

PR #699, opened 2020-05-20
Medulloblastoma_EXTENDscores_Comparison

NNoureen:
The purpose here is to compare EXTEND scores for Medulloblastoma molecular subtypes

#### What was your approach?
We used our newly developed method EXTEND (Expression based Telomerase ENzymatic activity Detection) for telomerase activities prediction. We have compared medulloblastoma molecular subtypes scores using student’s t-test and visualized it using boxplots.

Issue #148 (Quantify telomerase activity across pediatric brain tumors)

We would like the reviewers to check script 05 comparing telomerase activities for medulloblastoma molecular subtypes.
YES
Plot is built using script -05 to compare the telomerase scores for medulloblastoma molecular subtypes.
Distribution of telomerase activities across different medulloblastoma molecular subtypes using boxplots whereas the p-values defines the differential patterns among them.

#### Documentation Checklist

This analysis module has a `README` and it is up to date.
This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
The analytical code is documented.


sjspielman:
This figure seems fine, *however* the implemented use of P-values needs to be changed.

1) First, we have a statistical issue of *multiple testing* - the code is implicitly performing four t-tests, so a correction needs to be applied. In this case, the P-values are mostly not significant at $\alpha=0.05$ except for the SSH/WNT comparison, but when corrected I suspect the significance will go away. The simplest approach in terms of correcting P-values is to use the Bonferroni correction (divide each P-value by the number of tests done). I gather this plotting aspect is performed by `ggpubr::stat_compare_means()`, but there does not appear to be an argument to deal with multiple comparisons for that function. This will take some more digging into.

2) Is there a specific reason only 4 comparisons were performed rather than all-to-all? This selection of comparisons is not clearly documented.

NNoureen:
Thanks. I will update the plot for all comparisons and multiple testing as well.

NNoureen:
Hi Dr. Stephanie
I have updated the code for all comparisons and have used the adjusted p-values for labeling, though they are no more significant

sjspielman:
Hi @NNoureen, looks good to me! I am not very concerned that significance has disappeared; the goal is more to apply robust statistics rather than to chase down significance. The data we have is the data we have, and if it's not significant that's just something else to conclude - certainly not a failed analysis!!

NNoureen:
Thanks for your help.

Issue #700, opened 2020-05-20
check pbta-snv-consensus-mutation.maf.tsv.gz for duplicated lines


kgaonkar6:
#### What data file(s) does this issue pertain to?
pbta-snv-consensus-mutation.maf.tsv.gz

#### What release are you using?

release-v16-20200320

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/03.methods.md#consensus-snv-calling

#### Put your question or report your issue here.
I think there are some duplicated lines in this file
 gzcat pbta-snv-consensus-mutation.maf.tsv.gz |sort|uniq -d > duplicated_lines_pbta-snv-consensus-mut.tsv
[duplicated_lines_pbta-snv-consensus-mut.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/4657777/duplicated_lines_pbta-snv-consensus-mut.txt)

jaclyn-taroni:
cc @cansavvy

cansavvy:
I'll look into this today.

jashapiro:
> I also had a random question about the snv-consensus analysis , the step1 takes a metadata file https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/snv-callers/scripts/01-setup_db.py but doesn’t seem to be using it anywhere in the analysis other than making a separate metadata samples table right? Asking because we were trying to review this code to perform consensus on some new samples which don't have cleaned histology file yet

That is correct. If no metadata file is included, it simply doesn't make or update that table in the sqlite database.

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d5b13f28a82bce0f098a9c3fb808a10a8e1aef48/analyses/snv-callers/scripts/01-setup_db.py#L317

kgaonkar6:
Thanks!

kgaonkar6:
Also is it possible to have no consensus calls for a sample?  BS_C3KDA5ZC seems to be missing in pbta-snv-consensus-mutation.maf.tsv.gz

Let me know if you want this to be a different issue, thanks!

cansavvy:
> I see it in strelka,mutect2,lancet merged files.

[...]

PR #701, opened 2020-05-20
Remove R script/notebook package installs from `snv-callers` (Part 2 of N)

cansavvy:
#### What was your approach?
I'm removing R package installations in the scripts in accordance with #690 
I'm keeping track of the PRs I file for that issue there. 

This PR is me having going through `snv-callers` by its own because its a large one and has notebooks that require a lot of memory usage. 

- Removed package installs from `compare_snv_callers_plots.Rmd` and `compare_snv_callers_plots-tcga.Rmd`
- Removed the VAF correlation plot from snv-callers, its very memory intensive and was crashing things out. Changed it to a table. 
- Changed the output file for pbta data plots to `pbta-comparison` so its more parallel to `tcga-comparison`.
- Set up AWS instance. 
- Reran both notebooks. 
- Committed and pushed results files

Note that I'm ignoring the `vaf_cutoff_experiment.Rmd` notebook since it is deprecated. 

Everything look okay? (Note this does not address #700 I'm looking into that separately). 

Yes, but the results shouldn't really be different. 

No changes are needed here. 


analyses/snv-callers/compare_snv_callers_plots.Rmd
```diff
@@ -347,19 +339,15 @@ ggplot2::ggsave(file.path(plots_dir, "vaf_violin_plot.png"))
...
-  ggplot2::theme_classic()
+vaf_df %>% 
+  dplyr::select(-VAF_consensus, -Variant_Classification) %>% 
+  cor() %>% 
+  knitr::kable()
 ``` 
 
-```{r}
-# Save this plot
-ggplot2::ggsave(file.path(plots_dir, "vaf_correlations_plot.png"))
```
jashapiro, line 360: Should presumably remove the associated output file as well.

jaclyn-taroni:
It looks to me like there are some differences in the plots - any ideas why that might be? Are these portions of the `snv-callers` pipeline that didn't get rerun after some data release?

cansavvy:
> It looks to me like there are some differences in the plots - any ideas why that might be? Are these portions of the `snv-callers` pipeline that didn't get rerun after some data release?

Yes. The compare notebook (because it takes so much memory) doesn't automatically get run in the pipeline scripts, there's an option to turn it on or off: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0985505eb18402ecac925b2038764fb2ddc2a712/analyses/snv-callers/run_caller_consensus_analysis-pbta.sh#L79
 
So I think it been a tad since these notebooks got run fully.

PR #707, opened 2020-05-22
Renamed analyses directories with underscores to have dashes

sjspielman:
Two analyses `fusion_filtering` and `tp53_nf1_scores` use underscores rather than the expected convention of dashes. This PR attempts (<- key word) to increase consistency by renaming directories, and any files referencing these directories, to use dashes and not underscores.


sjspielman:
Hi @jaclyn-taroni I did an OCD thing!

jaclyn-taroni:
In the interest of working through the backlog of PRs, I am going to close this. We can revisit making the module names more uniform in the future.

PR #712, opened 2020-05-28
Refresh the final figure png for `CN status heatmap`

cansavvy:
The last branch included updates to CN status heatmap, but it needs to be refreshed in the final figures `pngs` folders. 

I re-ran the notebook with the `final_figures=TRUE` bit as it is written in `generate-figures.sh`. 

#### What was your approach?
Re-run the notebook  as it is written in `generate-figures.sh` and commit the changes


jaclyn-taroni:
Can you elaborate on

> I re-ran the notebook with the final_figures=TRUE bit as it is written in generate-figures.sh.

To get the PNG both places you need it, you have to run the notebook twice?

cansavvy:
>To get the PNG both places you need it, you have to run the notebook twice?

Yes, the `final_figures=TRUE` options re-runs bin calculations and saves the png to the figures/png folder. While `final_figures=FALSE` runs the bin calculations only if the binned data file doesn't exist, and then it saves the png to the analysis module `plots` directory.

I'll add this to the README.

jaclyn-taroni:
No need to alter now, but for future reference it makes sense to me to decouple redoing the bin calculations from the file path (provided I'm understanding correctly). It stands to reason that you wouldn't want the two PNGs to get out of sync and you could copy the PNG from the module to the `figures` location. (That's what we do for the interaction plot and the oncoprint.)

jaclyn-taroni:
Nothing changed here is going to be tested via CI, merging without it!

cansavvy:
Tracking the decoupling `final_figures` idea here: https://github.com/cansavvy/openpbta-notebook-concept/issues/8

Issue #713, opened 2020-05-28
Proposed Analysis : Investigate top tp53_nf1_classifier scoring samples without TP53 alterations


kgaonkar6:
#### What are the scientific goals of the analysis?
Investigate top tp53_nf1_classifier scoring samples without TP53 alterations ("false-positives") to see if they fall into certain tumor types, have certain mutations, or have other evidence of TP53 loss. As per @cgreene's  [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/673#issuecomment-635058769)

#### What methods do you plan to use to accomplish the scientific goals?

- Use top scoring samples (tp53_score) and tp53_status==0 to find the sample set
- plot broad_histology distribution
- plot other alterations (need a genelist ? look into CNV data ?)

#### What input data are required for this analysis?
pbta-snv-consensus-mutation.maf.tsv.gz
pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
pbta-gene-expression-rsem-fpkm-collapsed.polya.rds
(maybe also ? ) consensus_seg_annotated_cn_autosomes.tsv.gz

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
2 days

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@kgaonkar6

jharenza:
going to close this, as step 5 of #837 will cover this

PR #697, opened 2020-05-18
Add guess_max to prevent readr errors

jashapiro:
#### What scientific question is your analysis addressing?

Fix errors/warnings related to readr not correctly guessing histology file data types as described in #682

#### What was your approach?

Search for all usage of `readr::read_tsv()` (explicitly scoped and not) that read in the `pbta-histologies.tsv` file and add `guess_max = 10000` where an alternative method was not already present.

#### What GitHub issue does your pull request address?

Closes #682

#### Which areas should receive a particularly close look?

Did I miss any?

#### What is your summary of the results?

Should not change any results, but should reduce warnings.
[histology file changes from v12 -> v13](https://app.gitkraken.com/glo/card/efdb8558f9e04e3a847c55c4e8ba42cd)

PR #720, opened 2020-06-15
Investigate top tp53_nf1_classifier results using SNV/SNV/histology features

kgaonkar6:
To plot a few visualizations to look at the tp53 inactivation score distributions in detail and their relation to other SNV alterations and broad histologies.

Investigate top tp53_nf1_classifier scoring samples without TP53 alterations ("false-positives") to see if they fall into certain tumor types, have certain mutations, or have other evidence of TP53 loss. As per @cgreene's comment

#### What was your approach?
 - I first used the samples with tp53 aberration status ==0 and ==1 and plotted the tp53 inactivation scores which look like they mostly aggregate around the 0.5 mark for stranded data 
TP53 status |freq
--- | ---
0   | 685
1   | 285

(polya has only 53 so isn't used in future distribution plots)
TP53 status | freq
--- | ---
1   | 29
0   | 29

- TP53 aberrations in stranded samples with tp53_status=1 were also visualized to plot scores for each kind of variant classifications

- I used top 250 high-scoring (highest probability of having tp53 inactivation as per classifier) samples (tp53_score) and tp53_status==0 to 
         - plot broad_histology distribution
         - plot other SNV alterations prevalent in these samples

#713

A lot of samples it seems don't have RNA - to - DNA matches so only 285 are actually paired does that look right?
I only used a few non-synonymous Variant_Classifications  to plot the SNVs is that ok as a first pass?

The TP53 loss scores seem to have a wide range should we look into this in more detail? 

Yes the notebook has all the figures.

notebook

Looks like a lot of interesting genes show up as mutated in tp53-non-mutant samples with high tp53-inactivation phenotype score. CTNNB1 is the most prevalent mutated genes, "is an important co-activator downstream of the oncogenic Wnt signaling pathway, so mutations in the β-catenin gene (CTNNB1) have been implicated in oncogenesis" from this review https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6354055/

FGFR1,ATRX and H3F3A are also here which mutations associated mostly with HGG?

#### Documentation Checklist


analyses/tp53_nf1_score/03-plot-distribution-scores.Rmd
```diff
@@ -0,0 +1,121 @@
...
+score_stranded_long<-reshape2::melt(score_stranded_df,id=c("tp53_status","sample_id","Kids_First_Biospecimen_ID","Kids_First_Participant_ID")) 
+score_stranded_long$tp53_status<-factor(score_stranded_long$tp53_status)
+
+score_polya_df<-read_tsv(file.path(analyses_dir,"results","score_status_polya.tsv")) %>% dplyr::select("TP53_SCORE","tp53_status","sample_id","Kids_First_Biospecimen_ID","Kids_First_Participant_ID") %>% unique()
+
+score_polya_long<-reshape2::melt(score_polya_df,id=c("tp53_status","sample_id","Kids_First_Biospecimen_ID","Kids_First_Participant_ID")) 
+score_polya_long$tp53_status<-factor(score_polya_long$tp53_status)
+
+
+histology<-read_tsv(file.path(data_dir,"pbta-histologies.tsv"))
```
sjspielman, line 35: This file has had some parsing issues due to recent-ish data release, so we're adding the argument `guess_max = 10000` to `read_tsv` as described the draft PR [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/697). Quick change before we get too far into this PR...

sjspielman:
Also, going to merge in master now before going too far with this code!

sjspielman:
A general comment, I think this analysis should be more systematic and use a threshold to define what counts as a "high score." For example, "false positives" are currently identified as samples w/out TP53 aberrations and whose score is in the top 250. But, this approach does not examine of what a high score _is_. Sorting scores in descending order doesn't guarantee that all scores at the top of the list are indeed "high." Indeed, the top 250 scores go as low as 0.4698566 - being below 0.5, this does not seem high to me. That said I don't have the domain expertise to determine an appropriate threshold here, but I feel confident one is definitely needed.

kgaonkar6:
> A general comment, I think this analysis should be more systematic and use a threshold to define what counts as a "high score."

That's a good point, I didn't look into that before! I'm adding @cgreene to this, since the [ticket](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/713) originated from a comment [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/673#issuecomment-635058769)

jaclyn-taroni:
Hey @kgaonkar6 - I filed https://github.com/kgaonkar6/OpenPBTA-analysis/pull/1 on your fork because I thought it would be useful to plot individual points for the score distributions as part of scientific review of this pull request. If you agree with the changes I've made there, we can get that merged and have the changes show up here!

jaclyn-taroni:
Hi @kgaonkar6,

I have been reviewing the discussion around this particular analysis as I thought about next steps, as the last discussion here was about picking a threshold for a high score.

First,

> A lot of samples it seems don't have RNA - to - DNA matches so only 285 are actually paired does that look right?

This seems low. I did a quick check by looking at the molecular subtyping tables for HGG and LGAT, where each row corresponds to a `Kids_First_Participant_ID`, `sample_id` pair and rows without `NA` in those tables will have both a DNA and RNA biospecimen. Those two histologies had 361 samples with a match based on that analysis. Where does the match/join step happen in this analysis module? Looks like it is upstream of the changes being introduced here.

Now, more related to the changes being introduced here – in the plot `Distribution of scores across aberrations` in this notebook, the distribution of _TP53_ scores for samples that are reported to have a loss looks _bad_, i.e., samples that have _TP53_ loss don't appear to mostly get _TP53_ scores > 0.5. This doesn't surprise me - it came up in earlier pull requests (though I think that was in the context of AUROC, specifically https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/653#issuecomment-610518874) and resulted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/673#issuecomment-632160320, where I recommended we used a file from `focal-cn-file-preparation` that accounted for broad CNA. This analysis doesn't use that file because there are no upstream changes in `run_classifier.sh`.

That makes sense because we were (I was? https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/673#issuecomment-635465431) interested in if the losses look bad (they do!) and I am convinced that we should re-evaluate how losses are included before we do what @cgreene was after (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/673#issuecomment-635058769):

> If you took the top "false positives" (those with 0 status but with very high confidence scores), it could help to investigate those to see if they fall into certain tumor types, have certain mutations, or have other evidence of TP53 loss. In other settings we've found this type of analysis quite fruitful.

if it is going to require us to pick a threshold in some way as suggested in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/720#issuecomment-644987600.

I think those two steps (determining what CNA are appropriate to include in a principled way, picking a threshold) could be somewhat time-consuming and, without looking into the upstream steps, I anticipate that the lower-than-expected number of matches is also impacting the other analyses in the module (e.g., ROC curve). 

Given the limited amount of bandwidth that folks have, both on the performing analyses/writing analytical code side and the review of said analyses and code side, I think we should be very cognizant of whether the potential benefit of completing this analysis outweighs the potential time cost. The option to deprecate the analysis and explain the rationale for doing so in the module README also exists.

Thoughts? Probably worth getting @cgreene's and @jharenza's impressions as well.

cgreene:
This topic is near and dear to my heart, and I'd love to see the TP53 scores make it into the data associated with the OpenPBTA project. On the other hand, I agree that there looks to be quite a lot of QC left to make that feasible. I think it's fine to deprecate or pause this for now if the bandwidth isn't in place to complete it.

jaclyn-taroni:
Based on the discussion above, I am going to close this pull request.

PR #725, opened 2020-07-13
Tmb d3b code

tkoganti:
## Purpose/implementation Section
Adding TMB code that CCDL team can review. 
- This code starts with MAF file and expects a consensus MAF file
- We have used consensus of 3 callers. - strlk2, mutect2 and lancet

#### What scientific question is your analysis addressing?z
- Tumor mutation burden for PBTA samples 

#### What was your approach?
- Using a consensus MAF file as input - calculate TMB
- Filter only variants within `friends` method(described in README)
- Determine target BED file based on `cohort` and `experimental_strategy` from histology file 
- Form target config file, get the correct BED file (For WGS samples, it is CCDS region. For WXS samples, it is target intersected with CCDS region)
- Calculate TMB and print every sample

- Mainly look for the differences in different consensus MAF files
- Check if `friends` filtering method is useful for TMB calculation (comparing adult and pediatric samples)

- Filtering methods and consensus MAF 

Ready for review 

### Results
I included some example in the outputs folder 

table in outputs folder 

Pediatric samples have lower TMB than TCGA 

#### Documentation Checklist


jaclyn-taroni:
Discussed via Slack @tkoganti is going to resolve those conflicts and merge in the master branch. We'll see if that resolves the EPN changes showing up.

jashapiro:
Because the filters in the [TMB harmonization paper cited](https://jitc.bmj.com/content/8/1/e000147#DC) are pretty buried, I am including what I think is the relevant table here. If there is a different section that is important, please let me know as I get ready to review.
Screen Shot 2020-07-13 at 9 53 02 AM

tkoganti:
> Because the filters in the [TMB harmonization paper cited](https://jitc.bmj.com/content/8/1/e000147#DC) are pretty buried, I am including what I think is the relevant table here. If there is a different section that is important, please let me know as I get ready to review.
> Screen Shot 2020-07-13 at 9 53 02 AM

Thanks @jashapiro We are only using the `types of mutation`  filtering in the script

analyses/tmb-compare-tcga/TMB_d3b_code/README.md
```diff
@@ -0,0 +1,50 @@
...
+          -o OUTFILENAME, --outfilename OUTFILENAME
+                        Out file name
+          -c CONFIGFILE, --configfile CONFIGFILE
+                        calculate_tmb.cfg.txt file with columns for disease,
+                        samplename, variant types etc.
+          -w TARGETCONFIG, --targetconfig TARGETCONFIG
+                        File with experimental strategy and path to BED file
+
+     `Inputs`  :
+          1. cohort based MAF files [here](https://s3.console.aws.amazon.com/s3/buckets/d3b-bix-dev-data-bucket/hgg-dmg-integration/openPBTA-consensus/?region=us-east-1&tab=overview)
```
jashapiro, line 38: We should use the MAF and  files that are included in the data download. These are canonical for all analyses in this repository. The file linked here is not public.

analyses/tmb-compare-tcga/TMB_d3b_code/config_files/target_cfg.targetcombos.txt
```diff
@@ -0,0 +1,6 @@
+WGS_PNOC003	/Users/kogantit/Documents/TMB/target_analysis/ccds_and_target/CCDS.bed
```
jashapiro, line 1: All files listed here need to be in the repository or downloads folder. Paths should be relative, not absolute.

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible_withbothexperstrt_and_cohort.py
```diff
@@ -0,0 +1,235 @@
...
+needed_packages = [
+    "pandas",
+    "numpy",
+    "pybedtools",
+]
+
+for package in needed_packages:
+    install_package(package, installed_packages)
+
+##################################################################
```
jashapiro, line 49: All code for this project should assume it is being run in the package docker container, which should already have these packages installed. If not, we will need to add them here https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/5b38033f25b64654b7a663b1b8c06398205a973c/Dockerfile#L312 (`pybedtools` is the only one not currently part of the container) I am happy to help if you need assistance with that section; we would want to specify version numbers for all packages and dependencies.

We do not want to encourage external installation, as that can lead to version conflicts.

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,121 @@
...
+needed_packages = [
+    "pandas",
+    "numpy",
+    "matplotlib",
+]
+
+for package in needed_packages:
+    install_package(package, installed_packages)
+
+##################################################################
```
jashapiro, line 46: See previous comment about package installation.

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible_withbothexperstrt_and_cohort.py
```diff
@@ -0,0 +1,235 @@
...
+#               commented line is absent
+########### 2. metadata file - This should have  disease column and
+#               experimental strategy. Sample should not be  repeated under
+#               different experimental strategies
+########### 3. target config file - This should have two columns,
+#               one with experimental strategy in metadata  and another with the
+#               corresponding BED file for processing variant counts
+
+
+# python3 01_calculate_tmb_targetflexible.py \
```
jashapiro, line 15: note that this does not match the file name (Personally, I like the shorter version here better.)

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible_withbothexperstrt_and_cohort.py
```diff
@@ -0,0 +1,235 @@
...
+            var_class = line.split("=")[1].split(",")
+        if line.startswith("disease_column="):
+            disease_col = line.split("=")[1].strip("\n")
+        if line.startswith("samplename_column="):
+            samplename_col = line.split("=")[1].strip("\n")
+        if line.startswith("cohort_column="):
+            cohort_col = line.split("=")[1].strip("\n")
+        if line.startswith("typeoftargetcolumn="):
+            typeoftargetcol = line.split("=")[1].strip("\n")
+
```
jashapiro, line 179: This may be out of scope for this code, but if you use a standard format like JSON for the config file, this section can be very much simplified to something like this:

```
# this goes at the top
import json

# here
with open(args.configfile) as f:
    config = json.load(f)
    var_class = config["Variants_to_use_for_TMB"] # this will already be a list!
    disease_col = config["disease_column"]
    ...
```

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,121 @@
...
+    help="Minimum samples from each histology/disease to plot",
+)
+args = parser.parse_args()
+
+######### Preparing inputs files and disease list ###########
+# Loading TMB scres text file into a  dataframe
+tmbfile = pd.read_csv(args.tmb_scores, sep="\t")
+
+# Choosing disease list to plot
+#tmbfile["new_TMB"] = tmbfile["TMB"]+1
```
cansavvy, line 71: Why is this part commented out?

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,121 @@
...
+for i, disease in enumerate(
+    sorted_diseases_final
+):  # i is index number, disease is disaese name
+    disease_df = perdisease.get_group(disease)  # Getting disease DF
+    TMB = disease_df["new_TMB"].sort_values()  # TMB will ne used for y-axis
+    x = (
+        np.arange(1, len(TMB) + 1) / len(TMB)
+    ) + i  # this generate an array in every loop that clusters
+    # all disease x-axis together and adds so the next disease
+    # x-axis numbers are spaced separaetly from previous disease
```
cansavvy, line 103:
```suggestion
    # x-axis numbers are spaced separately from previous disease
```

analyses/tmb-compare-tcga/TMB_d3b_code/README.md
```diff
@@ -0,0 +1,50 @@
...
+### Calculate TMB
+  1. Reads the MAF file and filters based on variant_classification column
+
+  2. Reads in config file and defines metadata specific fields/columns and variants  list etc.
+
+  3. Reads in a target config file and determines `experimental strategy` and `cohort` from metadata file. MAF  variants  are filtered within this target file and the BED length is calculated
+
+  4. Calculates TMB
+      - Calculates TMB based on `((# of variants)*1000000) / size of BED)`
+      - Prints out the samplename, TMB, counts, cohort and disease type for every sample
```
cansavvy, line 18:
```suggestion
      - Prints out the samplename, TMB, counts, cohort and disease type for every sample
```
Can we be more specific here and state that it is `Tumor_Sample_Barcode` that is printed out (if this is correct)? I think we've had confusion elsewhere about the different sample IDs `Participant_ID` , so being specific would be good.

analyses/tmb-compare-tcga/TMB_d3b_code/README.md
```diff
@@ -0,0 +1,50 @@
...
+          1. cohort based MAF files [here](https://s3.console.aws.amazon.com/s3/buckets/d3b-bix-dev-data-bucket/hgg-dmg-integration/openPBTA-consensus/?region=us-east-1&tab=overview)
+          2. config file for target  and calculating TMB  (in config folder)
+          3. Histology files for metadata  [here](https://github.com/d3b-center/d3b-bix-analysis-toolkit/tree/master/data/histologies)
+
+     `Outputs` :
+          `temp_pbta_wxs_OUTTMB`
+
+### Plot TMB scores
+
+    1. Takes an input file that has sample name, cohort and TMB
```
cansavvy, line 47: I have the same comment here. Is this taking a single `Tumor_Sample_Barcode`?

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible_withbothexperstrt_and_cohort.py
```diff
@@ -0,0 +1,235 @@
...
+
+########### Preparing target files ########################
+target_dict = get_target_dict(args.targetconfig)
+############################################################
+
+
+############  Groupby and calculate TMB #####################
+outfile = open(args.outfilename, "w")
+outfile.write("Samplename\texperimental_strategy\tcohort\tdisease\tcount\tbedlength\tTMB\n")
```
cansavvy, line 207: I may not be following this entirely, so corrrect me if I'm wrong, but if possible it would be nice to have the column names stay consistent with their origin files so it is easier to follow this back. So `Tumor_Sample_Barcode` again for example. (Sorry to keep harping on this same thing...)

analyses/tmb-compare-tcga/TMB_d3b_code/README.md
```diff
@@ -0,0 +1,50 @@
+## TMB analysis
+
+###### Authors : Teja Koganti for D3B
+
+This analysis computes tumor mutation burden for different disease types.
+The first script takes single MAF file and filters variants  based filtering strategies from [Friends of Cancer research](https://jitc.bmj.com/content/8/1/e000147#DC1). TMB is computed  based on
+`(filtered variant counts* 1000000) / target BED length`
+
+### Calculate TMB
+  1. Reads the MAF file and filters based on variant_classification column
```
cansavvy, line 10:
```suggestion
  1. Reads the MAF file and filters based on `Variant_Classification` column
```

analyses/tmb-compare-tcga/TMB_d3b_code/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+  2. Reads in config file and defines metadata specific fields/columns and variants  list etc.
+
+  3. Reads in a target config file and determines `experimental strategy` and `cohort` from metadata file. MAF  variants  are filtered within this target file and the BED length is calculated
+
+  4. Calculates TMB
+      - Calculates TMB based on `((# of variants)*1000000) / size of BED)`
```
cansavvy, line 17:
```suggestion
      - Calculates TMB based on `((# of nonsynonymous variants based on FoCR criterion within the BED ranges)*1000000) / size of BED)`
```
This may not be the best way to make this specification, but I want to make sure my understanding is correct.

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
tkoganti: I am implementing these lines instead - 

`value_to_add = np.min(tmbfile[tmbfile["TMB"] != 0]["TMB"])*0.8`
`tmbfile["new_TMB"] = tmbfile["TMB"].replace({0: value_to_add})`
`tmbfile = tmbfile.sort_values(by=['new_TMB'])`


When I looked at the values after adding one to all - there was not a lot of difference amongst the lower values. They did not look different on the plot. So instead of adding 1 to all values, 
- I take 80% of the smallest value (that is non-zero) 
- name it `value_to_add`
- Replace all zero values with this value - so taking log will not longer be a problem

tkoganti:
> This seems like a good start from a glance.
> 
> Can we have more information about the BED files themselves, specifically how are the caller's ranges being incorporated?
> 
> I am curious to see how @jashapiro 's review comments are incorporated. I'm happy to do another review later.

I am downloading the BED file from here - https://ftp.ncbi.nlm.nih.gov/pub/CCDS/current_human/CCDS.current.txt 
I have a python notebook on how I generated the BED file. Should I commit that under the tmb folder?

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible_withbothexperstrt_and_cohort.py
tkoganti: The initial idea was for this to be pretty generic. But since MAF file has this header as `Tumor_Sample_Barcode`, using the same in this output file is a good idea!

tkoganti:
@jashapiro and @cansavvy - 
I made the changes you guys requested, please look at the changes and let me know your feedback. 

I am using cohort + experimental_strategy to pick which target to use, so I cannot use the current pbta-histology file in the repo from V16. I am copying the new file with cohort column in inputs folder

jashapiro:
Hi Teja- Thanks for the updates! I'm quite busy with preparing for a training session this week, so I won't get a chance to look at this until mid next week at the earliest. 

One thing that you can do to speed things up for later is to add the commands to run your analysis to https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/.circleci/config.yml so they will be run by our continuous integration tests. (I expect them to fail now because of uninstalled python packages; those will need to be added to https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/Dockerfile)

I also noticed that you may still need to update the paths here that are specific to your machine: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/7af5d68a06c3c48537abfca33277db8a0e7192c3/analyses/tmb-compare-tcga/TMB_d3b_code/config_files/target_cfg.targetcombos.txt#L4-L6
I'm not sure if or how that file is used, but it seems worth updating.

jashapiro:
Hi @tkoganti! I have now added the required python package to the docker in this branch, as well as adding a script to call both of the included python scripts during CI. This script, `run_tmb_d3b.sh`, currently fails, as there are some non-portable paths in the config files that the python scripts use, as mentioned earlier (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/725#issuecomment-662404985). 
There may also be other errors that I have not run into as yet; you may need to modify the arguments to the scripts in `run_tmb_d3b.sh` to reflect the correct options.

Once we have things passing the CI tests, then we can start to compare the results to other analyses!

tkoganti:
@jashapiro I added a `basedir` for target files and copied all target BED files to `inputs/target_files`. Both the commands are now working. Please let me know if you have any questions.

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,94 @@
...
+# Choosing disease list to plot
+#tmbfile["new_TMB"] = tmbfile["TMB"]+1
+value_to_add = np.min(tmbfile[tmbfile["TMB"] != 0]["TMB"])*0.8
+tmbfile["new_TMB"] = tmbfile["TMB"].replace({0: value_to_add})
+tmbfile = tmbfile.sort_values(by=['new_TMB'])
+
+
+disease_number = tmbfile.groupby("cohort").count()["new_TMB"]
+top_diseases = disease_number[disease_number > int(args.minsamplestoplot)].index
+sorted_diseases = tmbfile.groupby("cohort").median().sort_values("new_TMB")
```
jashapiro, line 52: Can you add a comment to explain what these lines are doing?

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
jashapiro: This seems like a bit of an arbitrary solution; I think a more principled position might be to add 1 mutation to each sample, then recalculate TMB and log(TMB)... You have all of the data to do this in the file, I think, so you could calulate something like: 

```
tmbfile["new_TMB"] = ( tmbfile["count"] + 1 * 1000000) / tmbfile["bedlength"]
```

(If we want to be very principled, we would probably add something small to the denominator as well, but that won't matter much as long as the denominator is large!)

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,94 @@
...
+tmbfile = pd.read_csv(args.tmb_scores, sep="\t")
+
+# Choosing disease list to plot
+#tmbfile["new_TMB"] = tmbfile["TMB"]+1
+value_to_add = np.min(tmbfile[tmbfile["TMB"] != 0]["TMB"])*0.8
+tmbfile["new_TMB"] = tmbfile["TMB"].replace({0: value_to_add})
+tmbfile = tmbfile.sort_values(by=['new_TMB'])
+
+
+disease_number = tmbfile.groupby("cohort").count()["new_TMB"]
```
jashapiro, line 50: Do we want to plot by cohort or disease?

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -0,0 +1,94 @@
...
+    plt.plot(x, TMB, linestyle="none", marker="o")
+    plt.hlines(np.median(TMB), np.percentile(x, 30), np.percentile(x, 70))
+    # Takes y-axis median to know where to print median line
+    # 30 and 70 show how long the median line should go on x-axis
+    plt.yscale("log")
+plt.xticks(tick_locs, rotation=90, fontsize=15)
+ax = plt.gca()
+ax.set_xticklabels(sorted_diseases_final)
+ax.set_ylabel("Mutations per Mb", fontsize=20)
+ax.set_xlabel("Disease type", fontsize=20)
```
jashapiro, line 91: Since the current plot is by cohort, not disease type, we should either label it that way, or make the disease type plot as well (maybe color by cohort, but group by disease?). 
```suggestion
ax.set_xlabel("Cohort", fontsize=20)
```

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible.py
```diff
@@ -0,0 +1,204 @@
...
+    cohortcol):
+    samplename = np.unique(grouped_df["Tumor_Sample_Barcode"])[0]
+    meta_data = meta_data.set_index(samplenamecol)
+    cols = list(grouped_df.columns)
+    exp_strategy = meta_data.at[samplename, targetcol]
+    cohort = meta_data.at[samplename, cohortcol]
+    sample_target_identifier = exp_strategy+"_"+cohort
+    disease = meta_data.at[samplename, diseasecol]
+    if sample_target_identifier in target_dict.keys():
+        grouped_df["Start_Position"] = grouped_df.apply(
```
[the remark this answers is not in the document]
tkoganti: Yes to adjust for 0-based coordinates

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible.py
```diff
@@ -0,0 +1,204 @@
...
+def get_target_dict(target_config):
+    dict_to_return = {}
+    with open(target_config, "r") as target_cfg:
+        for line in target_cfg.readlines():
+            line = line.split()
+            dict_to_return[line[0]] = BASE_DIR+"/"+line[1]
+    return dict_to_return
+
+
+# This is the function where we calulate  the TMB, for every grouped samples,
```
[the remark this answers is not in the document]
tkoganti: Yeah filtering is done before this function starts - 

`maf_file = maf_file.loc[
    maf_file.apply(lambda x: x["Variant_Classification"] in var_class, axis=1)
]`

analyses/tmb-compare-tcga/TMB_d3b_code/code/01_calculate_tmb_targetflexible.py
```diff
@@ -0,0 +1,204 @@
...
+#   the function gets  the corresponding target  BED file and calulates the
+#   variants within those target files using pybedtools
+
+
+def calculate_tmb(
+    grouped_df, meta_data, target_dict, out, diseasecol, samplenamecol, targetcol,
+    cohortcol):
+    samplename = np.unique(grouped_df["Tumor_Sample_Barcode"])[0]
```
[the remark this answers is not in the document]
tkoganti: I am running the function in a loop after grouping based on sample name and every time you run it, `grouped_df` has all lines for each sample. 

 `grouped_df["Tumor_Sample_Barcode"]` looks like this - 
Name: Tumor_Sample_Barcode, dtype: object
922069    BS_0J5MCBZV
922172    BS_0J5MCBZV
922549    BS_0J5MCBZV
922556    BS_0J5MCBZV
922667    BS_0J5MCBZV
922702    BS_0J5MCBZV
922807    BS_0J5MCBZV
922906    BS_0J5MCBZV
922937    BS_0J5MCBZV
922959    BS_0J5MCBZV
922963    BS_0J5MCBZV
922989    BS_0J5MCBZV
923044    BS_0J5MCBZV
923095    BS_0J5MCBZV

In`samplename`, I want to save just the BS_ID as a string, so I had to take unique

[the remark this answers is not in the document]
jashapiro: Ah... I think you can do this a bit more simply with:
```
grouped_df.at[grouped_df.index[0], "Tumor_Sample_Barcode"]
```

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
tkoganti: Sure I am going to add comments, but to explain, 

`disease_number` calculates the number of samples in each disease
disease
ATRT                              30
Adenoma                            4
CNS EFT-CIC                        7
CNS Rhabdomyosarcoma               2
CNS lymphoma                       1
CNS neuroblastoma                  4

`top_diseases` gets a list of diseases that have higher number of samples than the one given  in input `minsamples`
(So only these will be in the plot.)

`sorted_diseases` - Sorting diseases based on median value to use this sorting in the plot

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
tkoganti: We were using both for PNOC003, but I will change this to disease for this code

tkoganti:
@jashapiro - I think I have addressed all your comments either in the comments above or in the last commit in the code. PLease let me know if I missed anything.

analyses/tmb-compare-tcga/TMB_d3b_code/code/02_cumulative_freq_TMBplot.py
```diff
@@ -41,15 +41,30 @@
...
-tmbfile["new_TMB"] = tmbfile["TMB"].replace({0: value_to_add})
-tmbfile = tmbfile.sort_values(by=['new_TMB'])
+#tmbfile["new_TMB"] = ( tmbfile["count"] + 1 * 1000000) / tmbfile["bedlength"]
+#tmbfile["new_TMB"] = tmbfile["count"] + 1 / tmbfile["bedlength"]
 
 
-disease_number = tmbfile.groupby("cohort").count()["new_TMB"]
+tmbfile["new_TMB"] = tmbfile.apply(lambda x:
+    (x["count"] + 1 * 1000000) / x["bedlength"] if x["count"]==0
+    else x["TMB"],axis=1)
```
jashapiro, line 50: For consistency of comparison, this should be applied to all TMB. 
```suggestion
    (x["count"] + 1 * 1000000) / x["bedlength"],
    axis=1)
```

tkoganti: That's fine, but we should add another parentheses around count + 1

`tmbfile["new_TMB"] = tmbfile.apply(lambda x:
    ((x["count"] + 1) * 1000000) / x["bedlength"],
    axis=1)`

I updated the code in the commit below

Issue #726, opened 2020-07-13
Updated analysis: filter to only non-synonymous mutations for TMB


jashapiro:
#### What analysis module should be updated and why?

`snv-callers`, specifically   
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/snv-callers/scripts/03-calculate_tmb.R   
and   
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/snv-callers/util/tmb_functions.R

#### What changes need to be made? Please provide enough detail for another participant to make the update.

Currently, the TMB we are using for comparisons includes *all* mutations in coding regions (with the caveat that there currently a bug as described in #724). To better conform with emerging standards, as implemented in #725, we should update our TMB calculations to use only the following *non-synonymous* changes: 
`Missense_Mutation`
`In_Frame_Del`
`Nonsense_Mutation`
`In_Frame_Ins`
`Frame_Shift_Del`
`Frame_Shift_Ins`

One note, is that this list of non-synonymous mutation types is slightly shorter than the one used in MAFtools, and implemented here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/5b38033f25b64654b7a663b1b8c06398205a973c/analyses/interaction-plots/scripts/02-process_mutations.R#L263-L274

We should probably standardize these two.

This change can be implemented by adding a `dplyr::filter()` step here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/5b38033f25b64654b7a663b1b8c06398205a973c/analyses/snv-callers/util/tmb_functions.R#L133
It is likely that we will want to make this optional in the enclosing function.

#### What input data should be used? Which data were used in the version being updated?

Input data is the same as previously.

#### When do you expect the revised analysis will be completed?

The actual changes are small and should be quick. Downstream consequences may take longer to sort out.

#### Who will complete the updated analysis?

TBD; could be implemented alongside the changes to fix #724, as both will require updating of downstream analyses, and we may not want to do that more than once!

cansavvy:
Now that #728 and #739 have been merged, I think this can be closed.

PR #728, opened 2020-07-13
Add filter for non-synonymous mutations and re-run TMB with bug fix

cansavvy:

- Adding an option to drop the synonymous mutations from the TMB calculations. 
- Re-running downstream TMB items after getting them updated. 

I would like to do some more in depth comparison of the old and new results, but I will save that for a different PR so as to not have too many things at once. 

#### What was your approach?
- Added a filtering option `--nonsynfilter` which is a TRUE/FALSE on whether mutations should be filtered to be only non-synonymous mutations. 
- I re-ran both TCGA and PBTA TMBs
- I re-ran the tmb-compare-notebook and adjusted some of the margins so it would be better aligned. 
- I re-ran the `fig2-mutational-landscape.R` figure again so it's updated. 

#726 

- Is there a way I can make the margins adjust more automatically? It was quite a bit of finagling of margins and text size to get the plots to be aligned. Perhaps this is just how it has to be. 
- Anything I'm missing? 

Yes. The TMBs for the adults do seem slightly higher? That's good. 

Still figuring this out. It's not changed as I suspected it would be. 
*Note that data files themselves are not pushed to GitHub but we will need to update them in the next data release. 

#### Documentation Checklist


cansavvy:
This is a reminder to myself to update the blog post with these updates to TMB. @cansavvy

analyses/snv-callers/scripts/03-calculate_tmb.R
```diff
@@ -146,6 +154,19 @@ join_cols <- c(
...
+  "Missense_Mutation",
+  "Frame_Shift_Del",
+  "In_Frame_Ins",
+  "Frame_Shift_Ins",
+  "Splice_Site",
+  "Nonsense_Mutation",
+  "In_Frame_Del",
+  "Nonstop_Mutation",
+  "Translation_Start_Site"
+)
```
jashapiro, line 168: As we have discussed, this does differ slightly from the standard proposed in https://jitc.bmj.com/content/8/1/e000147#DC by the *Friends of Cancer Research TMB Harmonization Project*. 

The relevant table is here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/725#issuecomment-657575733

In particular, the FoCR group does not include `Translation_Start_Site`, `Nonstop_Mutation` or `Splice_Site`. I tend to think all of those are worth including, but I guess one simple question is how common these mutations actually are in the data set. Either way, we will want to standardize our criteria across analyses.

cansavvy:
> This looks great. Do we want to include a version without the nonsyn filter as well, just for comparison? I would not put that in the final figure, but having the figure and/or data table might be nice.

I want to do this, but figured I should make it a separate PR so this one stays focused.

analyses/snv-callers/scripts/03-calculate_tmb.R
cansavvy: I think this question is something I can ask in a next PR that does some more exploration. I'll file an issue that lays out what kinds of exploration analyses I think should be included.

cansavvy: Filed the issue: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/729

cansavvy:
The follow up questions, I'll address on separate PRs, notoed on this issue here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/729

For wrapping up this current PR, I'll play around a bit with finding a better way to "automagically" align the tmb plots.

cansavvy:
I tried `patchwork`. But am encountering an error like what's documented here: https://github.com/thomasp85/patchwork/issues/148
I tried minimizing the `theme` arguments specified but it still didn't fix the problem.

cansavvy:
@jashapiro I actually use `cowplot` in the `tmb-compare-tcga` notebook, but for whatever reason the `align = "v"` argument doesn't work. Let me know if you have insight into why its broken/being ignored by cowplot: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tmb-compare-tcga/compare-tmb.Rmd#L141

cansavvy:
Not sure what else to try for the "automagic" y axis alignment (see above comments). But in the interest of moving this and the future issues (#729) forward, I've wrote this problem down here: https://github.com/cansavvy/openpbta-notebook-concept/issues/9 and will return to it if we have any breakthroughs. 

With this being noted, are we ready to merge, @jashapiro ?

jashapiro:
> With this being noted, are we ready to merge, @jashapiro ?
:shipit:

Issue #729, opened 2020-07-14
Exploratory Side Analysis: `Variant_Classification` breakdown


cansavvy:
#### What are the scientific goals of the analysis?
There's a few questions we want to explore in regards to the `Variant_Classification` breakdowns. 

1) How different are the synonymous vs nonsynonymous mutation counts? 

2) How often do these: `Translation_Start_Site`, `Nonstop_Mutation` or `Splice_Site` come up? (These are not included in the FoCR definitions, but are included in `maftools` definition. 

Related: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/728#discussion_r454401088
Related: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/728#pullrequestreview-448164485
Related: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/725#issuecomment-657575733

#### What methods do you plan to use to accomplish the scientific goals?
I'll plan on making two small notebooks that will probably just consist of some exploratory plots: 

1) A notebook that runs the `calculate_tmb.R` script with and without `--nonsynfilter` and creates some plots to get visualizes on how the participant data changes with filters. 

2) A second small notebook that looks at how often the discrepant mutation categories come up in PBTA and TCGA data: (`Translation_Start_Site`, `Nonstop_Mutation` or `Splice_Site`)

Both notebooks are exploratory. I will store them in their own folder in the`snv-callers` module. I think both would be retired pretty quickly since I don't see a use for these after we do a quick exploration. 

#### What input data are required for this analysis?
`snv_db_sqlite` and `tcga_snv_db.sqlite` that are based on the latest versions of snv consensus data. 

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
It will be two PRs

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@cansavvy 

#### What relevant scientific literature relates to this analysis?
https://jitc.bmj.com/content/8/1/e000147#DC by the *Friends of Cancer Research TMB Harmonization Project*.

jashapiro:
This sounds great. I would also suggest comparing results from these analyses to those that come in as part of #725

cansavvy:
With #741 being merged, this is wrapped up.

PR #741, opened 2020-08-20
Compare TMB calculation sets

cansavvy:
- We have two sets of TMB calculations. How do they compare? Are there things we should look into in either calculation?

-  Due to the expanded nature of `tmb-compare-tcga` folder not just being for `tcga` comparisons, I also renamed it `tmb-compare`. I did a project search and replace to get this done .

#### What was your approach?
I imported both `TMB_d3b_code/output/pbta-snv-consensus-TMB_intarget.txt` and `snv-callerrs/results/consensus/pbta-snv-mutation-tmb-coding.tsv` and did a series of correlations, plots, and calculations of percent differences. 

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/729#issuecomment-658259843

- I did a project search and replace for the module name change. I think everything is set.  If you notice anything missing, let me know. 
- Are there analyses I should include but haven't yet? 
- Are my thought processes clear? These analyses are quite literally in the order I did them and the questions I asked, but I can reorganize if this is not easily followable. 

Mutation counts seem to be fairly related but the bed sizes are very different and that seems to be what's causing the differences in TMB calculations ranges.
CCDL TMB is a much larger range (0 - ~300), while D3B TMB is smaller range (0 - ~5) but that is because the BED size for CCDL TMB is much smaller than the BED size for D3B TMB. 
This problem is exacerbated in (mostly) HGAT samples because they outliers and seem to have some high variance (we've seen this before). 

Here's the rendered notebook: [compare-tmb-calculations.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5109624/compare-tmb-calculations.nb.html.zip)

#### Documentation Checklist
I have not added this to the analysis READMEs but I can if we think its appropriate. 


jashapiro:
I had a look at this, and my main comment is really that the whole thing can be simplified down to your main point that the BED files are different. I don't think the TMB correlations are very informative, as they basically boil down to the counts being correlated. I think there are two main questions:

- Why are the BED regions so different? This has to be addressed and normalized before any meaningful discussion of TMB can be done as a proportion can be done.

- Where the counts differ, are we always counting more because we include more classes of mutations, and if so, what happens when we pare down to the set they used and compare again. I imagine this is in #740, which I still need to look at, but I might want to see these two analyses combined: I want to know what the counts are by various methods of ours, compared to the d3b counts. They should be able to match exactly, if we are doing the same things!

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,217 @@
...
+  dplyr::mutate(
+    bed_diff = abs(region_size - bedlength),
+    bed_perc_diff  = bed_diff/max(region_size, bedlength)
+    ) 
+```
+
+Let's make a table of this.
+
+```{r}
+table(all_tmb$region_size, all_tmb$bedlength)
```
jashapiro, line 207: I don't think a `table()` is what we want here. We really just want to see the parings of region size and bed length. Something like:
```
all_tmb %>%
  dplyr::select("region_size", "bed_length") %>%
  dplyr::distinct()
```

Or if you want counts:
```
all_tmb %>%
  dplyr::group_by("region_size", "bed_length") %>%
  dplyr::tally()
```

jashapiro (commented):
Looking at the results of the region_size calculations it seems that the D3B regions are very nearly the full genome, not restricted to coding regions (they are almost 2 billion bases). I did not catch that in my previous review of that submissions, but it is clearly not just the coding regions. I think this can be pointed out explicitly in your summary of these results.

cansavvy:
Per our in person discussion, @jashapiro I'll use the `focr` vs `maftools` definition option distinction added in #740 to add a `CCDL_focr` and `CCDL_maf` (which is already here but just referred to as CCDL_tmb here) to this same analysis. But I'll cut out the correlations and just take a look at mutation counts.

cansavvy:
Before I run the full TMB sets through the `compare-tmb-calculation.Rmd` (I need to obtain some of them by running an AWS instance). Can you take a big picture look at this notebook, @jashapiro and see if it addresses what we discussed on https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/741#issuecomment-686731925 ?

cansavvy:
The plots are only done with the testing set so far, will note on here when they are updated to the full set.

cansavvy:
Here's the rendered notebook with the full datasets: 

[compare-tmb-calculations.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5194957/compare-tmb-calculations.nb.html.zip)

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+  dplyr::rename(tmb_d3b = TMB, 
+                mutation_count_d3b = count)
+```
+
+## Percent Differences in TMB
+
+```{r}
+all_tmb <- all_tmb %>% 
+  dplyr::mutate(perc_diff_focr = abs(tmb_focr - tmb_d3b) / max(tmb_focr, tmb_d3b), 
+                perc_diff_maf = abs(tmb_maftools - tmb_d3b) / max(tmb_maftools, tmb_d3b))
```
jashapiro, line 130: This is an odd way to calculate the percent difference, as I think you would want to retain direction, and this masks it. You can use `abs()` below for your filter, but I would not use it here. I also would pick a reference calculation, rather than using `max()`.

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+```
+
+This is the same regardless of which definition (maftools/FoCR) is used.
```
jashapiro, line 133: What is "this"?

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+```
+
+This is the same regardless of which definition (maftools/FoCR) is used. 
+
+```{r}
+all_tmb %>% 
+  dplyr::filter(perc_diff_focr > .05)
+
+all_tmb %>% 
+  dplyr::filter(perc_diff_maf > .05)
```
jashapiro, line 140: I'm confused by these being such short lists, given that the counts are similar but the denominator is so very different.

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+## How different are the mutation counts?
+
+Select the counts columns only. 
+
+```{r}
+counts_only <- all_tmb %>% 
+  dplyr::select(mutation_count_focr, mutation_count_maftools, mutation_count_d3b)
+
+# Print this out
+counts_only
```
jashapiro, line 154: Here it looks like the difference is quite consistent that we count more mutations than d3b. maftools is greater than focr, as expected, but ccdl still is much higher than d3b, though at least in the same direction.

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+
+Plot this. 
+
+```{r}
+counts_only %>%
+  GGally::ggpairs(mapping = ggplot2::aes(alpha = 0.05)) +
+  ggplot2::theme_classic()
+```
+
+Counts look *mostly* the same, but those outliers make the plots not super useful.
```
jashapiro, line 165: I might try plotting `log1p()` values to make this more comparable. Adding a y=x line will also be informative, showing that d3b counts are always lower.

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,198 @@
...
+Let's take a closer look at the samples with larger differences in counts. 
+We'll compare things with `mutation_count_maftools` since that is what is in the current consensus file. 
+
+```{r}
+all_tmb %>%
+  dplyr::transmute(
+    Tumor_Sample_Barcode,
+    ccdl_filter_diff = mutation_count_maftools - mutation_count_focr,
+    d3b_diff = mutation_count_maftools - mutation_count_d3b) %>% 
+  dplyr::arrange(desc(ccdl_filter_diff))
```
jashapiro, line 176: An idea to add diffs on a relative scale (+1 to handle zero counts).
```suggestion
all_tmb %>%
  dplyr::transmute(
    Tumor_Sample_Barcode,
    ccdl_filter_diff = mutation_count_maftools - mutation_count_focr,
    ccdl_filter_diff_frac = ccdl_filter_diff / (mutation_count_maftools +1), 
    d3b_diff = mutation_count_maftools - mutation_count_d3b,
    d3b_diff_frac = d3b_diff / (mutation_count_maftools +1)
    ) %>% 
  dplyr::arrange(desc(ccdl_filter_diff))
```

jashapiro (commented):
A few suggestions below about calculations: I am somewhat surprised at how few samples seem to be coming up with > 5% difference. Given the order of magnitude difference in denominator, I was expecting more. 

I also think we could use a bit more text about the results to explain a bit more.

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: It's because of how I calculated them, which after changing that as you've noted here, it looks more like what you'd expect. https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/741#discussion_r485706337

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: This is a nonsensical sentence, is what this is.

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: Yes. Interesting. Not sure why that is.

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -122,29 +123,59 @@ all_tmb <- tmb_ccdl_maftools %>%
                 mutation_count_d3b = count)
 ```
 
-## Percent Differences in TMB
+## Fraction differences in TMB
```
jashapiro, line 126: I didn't even say anything! (The system works!)

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,235 @@
...
+
+```{r}
+counts_only <- all_tmb %>% 
+  dplyr::select(mutation_count_focr, mutation_count_maftools, mutation_count_d3b)
+
+# Print this out
+counts_only
+```
+
+Let' plot these counts correlations, but turn them into log1p values for easier visualization.
```
jashapiro, line 188:
```suggestion
Let's plot these counts correlations, but turn them into log1p values for easier visualization.
```

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,235 @@
...
+
+```{r}
+counts_only %>%
+  dplyr::transmute_all(log1p) %>%
+  GGally::ggpairs(mapping = ggplot2::aes(alpha = 0.05)) +
```
jashapiro, line 193: I was going to renew my suggestion for adding the y=x line, but then I looked at what a pain that would be in `ggpairs()`. Maybe plot one with that? focr vs d3b?

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,235 @@
...
+### Summary of findings: 
+
+Mutation counts seem to be fairly related but the bed sizes are very different and that seems to be what's causing the differences in TMB calculations ranges.
+CCDL TMB is a much larger range (0 - ~300), while D3B TMB is smaller range (0 - ~5) because the BED size for CCDL TMB is much smaller than the BED size for D3B TMB. 
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+    - [Usage](#usage)
```
jashapiro, line 34: Markdown doesn't seem to like starting with an interior node here, and treats this as a code block. I think changing to `### Usage` may be enough to fix this. *You probably also want to promote `Packages and functions` to H3.

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: I knew it was probably eating you up.

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: Okay. I'm making a separate plot so we can add that.

cansavvy: I used `maftools` one since that's what I claim I'm using as the comparison set, I figure I should try to be consistent with that. Is this what you are looking for?:
Screen Shot 2020-09-09 at 2 18 26 PM

jashapiro: Yes and maybe no. That looks like a fit line? I was thinking `geom_abline(intercept = 0, slope = 1)`

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,242 @@
...
+```{r}
+counts_only %>%
+  dplyr::transmute_all(log1p) %>%
+  GGally::ggpairs(mapping = ggplot2::aes(alpha = 0.05)) +
+  ggplot2::theme_classic()
+```
+
+The samples with high TMB, still usually have a higher TMB in the other datasets. 
+
+Let's take a closer look at one of the comparisons.
```
jashapiro, line 195:
```suggestion
Let's take a closer look at one of the comparisons, comparing the FoCR counts to the D3B counts, which should use the same criteria for inclusion. 
```

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,242 @@
...
+
+```{r}
+counts_only %>%
+  dplyr::transmute_all(log1p) %>%
+  ggplot2::ggplot(ggplot2::aes(x = mutation_count_maftools, y = mutation_count_d3b)) +
```
jashapiro, line 200:
```suggestion
  ggplot2::ggplot(ggplot2::aes(x = mutation_count_focr, y = mutation_count_d3b)) + 
```

analyses/tmb-compare/compare-tmb-calculations.Rmd
```diff
@@ -0,0 +1,242 @@
...
+
+```{r}
+counts_only %>%
+  dplyr::transmute_all(log1p) %>%
+  ggplot2::ggplot(ggplot2::aes(x = mutation_count_maftools, y = mutation_count_d3b)) + 
+  ggplot2::geom_point() + 
+  ggplot2::geom_abline(intercept = 0, slope = 1, color = "blue") + 
+  ggplot2::theme_classic()
+```
+
```
jashapiro, line 205: Assuming the statement below is true, I think this should be included.
```suggestion

The FoCR count always greater than or equal to the D3B count, despite the fact that these should be using the same criteria and data files. 
(The maftools count is also strictly greater than or equal to the FoCR count, as expected by its broader criteria for inclusion.)

```

jashapiro (approved):
Just a couple little changes, mostly to make the focal plot the most informative one (in my opinion) and to add a bit of detail to the conclusions. I don't think I should need to rereview this.

analyses/tmb-compare/compare-tmb-calculations.Rmd
cansavvy: What's your rationale for this? I like comparing to the current TMB which is maftools-defined filter.

cansavvy: Aka why do you find it more informative?

jashapiro: It is the closest thing we have to an apples-to-apples comparison. These two should be doing the same thing, so differences are due to data or code and can be investigated. Any maftools/focr difference can be explained by the choice of mutations, so that is not very interesting. Maybe include both?

cansavvy: I just put both. Why not?

Issue #742, opened 2020-08-21
Updated analysis: Take consensus of two classifiers for medulloblastoma subtype labels


jaclyn-taroni:
Related to #730 and #731 

#### What analysis module should be updated and why?

In #738, code and results for two medulloblastoma classifiers ([medulloPackage](https://github.com/d3b-center/medullo-classifier-package) and [MM2S](https://github.com/cran/MM2S)) are being added to the repository.

Some analyses of the classifier performance on the PBTA data, using the pathology labels where available (https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/731#issuecomment-663599379), suggest that the `medulloPackage` performs a bit better wrt Group 3 and Group 4 classification and the `MM2S` package tends to perform better on WNT ([notebook](https://jharenza.github.io/mb-subtypes/explore-medulloblastoma-unsupervised-clustering.nb.html) and [Gist](https://gist.github.com/PichaiRaman/cd441e31ecdb53160e3c595a73873fba)).

See the related issues #730 and #731. 

#### What changes need to be made? Please provide enough detail for another participant to make the update.

For this project, it seems reasonable to assign molecular subtype labels only when the two classifiers agree and leave all other samples as unclassified. That's the end goal, but it also makes sense to display the performance metrics in a notebook that's downstream of the two classifiers so it's in the repository and not just on issues.

#### What input data should be used? Which data were used in the version being updated?

`analyses/molecular-subtyping-MB/results/mb-classified.rds` being added in #738.

jaclyn-taroni:
I consider this closed by #743, with any remaining items tracked in #747

PR #743, opened 2020-08-21
Molecular subtypes (MB) summary notebook

komalsrathi:

Summarizes the following:

1.  Medulloblastoma subtype classification using the two classifiers [MM2S](https://github.com/cran/MM2S) and [medulloPackage](https://github.com/d3b-center/medullo-classifier-package) on batch corrected and uncorrected expression matrix along with the expected subtypes from pathology reports.
2. Associated performance metric in terms of % accuracy for each classifier.
3. Assignment of consensus subtype between the two classifiers as the `molecular_subtype`.

#### What was your approach?

As suggested in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/742, I have added some details to the notebook:

1. % Accuracy is currently being calculated by matching observed and expected subtypes where expected subtype info is available. In case of ambiguous subtypes, we treat it as a match if the observed subtype matches with any one of the expected subtypes
2. Consensus tabs: Molecular subtype labels are assigned only when the two classifiers agree, leaving all other samples as unclassified.

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/742

Accuracy calculation and consensus molecular subtype assignment.

Yes

.html output containing summary tables

#### Documentation Checklist


analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -0,0 +1,221 @@
...
+
+2. Consensus tabs: Molecular subtype labels are assigned only when the two classifiers agree, leaving all other samples as unclassified. 
+
+## {.tabset .tabset-fade}
+
+### MM2S (batch-corrected)
+
+```{r, echo = TRUE, warning = FALSE, message = FALSE}
+# merge observed and expected subtypes
+mm2s.corrected <- obs.class[[1]]
```
jaclyn-taroni, line 67: Using an index here means we're relying on an analyst to know or remember what order the observed results come back in. There are no names in `obs.class` at the moment. I'd recommend making alterations to the upstream step (`01-classify-mb.R`) such that there is information about the method and dataset used in this object. That way someone who did not author the module could use this object "off-the-shelf" without much digging.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -0,0 +1,221 @@
...
+                               scrollX = TRUE,
+                               dom = 'tpi',
+                               initComplete = JS("function(settings, json) {",
+                                                 "$(this.api().table().header()).css({'background-color':
+                                            '#004467', 'color': '#fff'});","}"))
+  )
+}
+```
+
+#### Details:
```
jaclyn-taroni, line 55: Can you add information about how many samples have pathology subtype labels and (briefly) what the process is for obtaining those labels (for you, not what goes into that for pathology! 🙂 ) please?

jaclyn-taroni: I would 'mirror' this information in the module README as you've done with these other points, too.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -0,0 +1,221 @@
...
+  filter(!is.na(pathology_subtype)) %>%
+  group_by(match) %>%
+  summarise(n = n()) %>%
+  mutate(Accuracy = paste0(n/sum(n)*100, '%')) %>%
+  filter(match) %>%
+  .$Accuracy
+print(paste0("Accuracy: ", medullo.classifier.corrected.acc))
+
+# output table
+viewDataTable(medullo.classifier.corrected)
```
jaclyn-taroni, line 129: A lot of this code is repeated between different classifier-dataset combinations. That suggests to me that it may be useful to wrap up joining with the expected subtype information and accuracy calculation in a custom function.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -0,0 +1,221 @@
...
+  group_by(match) %>%
+  summarise(n = n()) %>%
+  mutate(Accuracy = paste0(n/sum(n)*100, '%')) %>%
+  filter(match) %>%
+  .$Accuracy
+print(paste0("Accuracy: ", consensus.uncorrected.acc))
+
+# output table
+viewDataTable(consensus.uncorrected)
+```
```
jaclyn-taroni, line 221: I would expect the output from this notebook to include a table of the consensus labels (preferably as a TSV). 

In modules for subtyping other histologies, we've joined all the identifiers (DNA and RNA) into a single table: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e8fbc7dc6aa8a36d7236fbe10a038657e3782e09/analyses/molecular-subtyping-HGG/results/HGG_molecular_subtype.tsv This particular step can come in a subsequent notebook if you'd prefer. But I do not expect this notebook to be overly long, particularly with the custom function changes.

komalsrathi:
Hi @jaclyn-taroni thanks again, I will incorporate these and update latest by tomorrow.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -0,0 +1,221 @@
...
+```{r read_data, echo = TRUE}
+# read input
+exp.class <- readRDS('input/expected_class.rds')
+obs.class <- readRDS('results/mb-classified.rds')
+
+# format ambiguous expected classes 
+exp.class <- exp.class %>%
+  mutate(pathology_subtype = replace(pathology_subtype, 
+                                     pathology_subtype == "Group 3 or 4", "Group3, Group4")) %>%
+  mutate(pathology_subtype = gsub(" ", "", pathology_subtype))
```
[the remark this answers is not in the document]
komalsrathi: So for this, there is `Group 3` and `Group 4` in path reports and we have `Group3` and `Group4`  which is why the `gsub`.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: Yes totally makes sense - will add names to the list items.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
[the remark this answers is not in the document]
jaclyn-taroni: Ah, so `pathology_subtype` can have the values `Group 3 or 4` or `Group 3, Group 4` before you do any mutating - is that correct?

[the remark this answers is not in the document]
komalsrathi: Yes these are the unique values for expected types (pathology):

> unique(dat$pathology_subtype)
[1] NA             "WNT"          "Group 3 or 4" "SHH"          "non-WNT"     
[6] "Group 4" 

So I convert `Group 3 or 4` to `Group3, Group4` (so that we can match either Group3 or Group4 predicted subtype) and gsub converts `Group 4` to `Group4` (because for observed types we have SHH, WNT, Group3 and Group4 i.e. no spaces)

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: added names to the list items: 
https://github.com/komalsrathi/OpenPBTA-analysis/blob/mb-class-nb/analyses/molecular-subtyping-MB/01-classify-mb.R#L49

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: Updated README with breakdown of samples and subtypes from path report:  
https://github.com/komalsrathi/OpenPBTA-analysis/tree/mb-class-nb/analyses/molecular-subtyping-MB#02-compare-classesrmd

Just unsure of what the second part means: `process of obtaining those labels`. I am using a file that @jharenza provided as input.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: I have added individual functions here: 
https://github.com/komalsrathi/OpenPBTA-analysis/blob/mb-class-nb/analyses/molecular-subtyping-MB/02-compare-classes.Rmd#L18

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
jaclyn-taroni: I basically meant how do we get the labels from what I assume is a pathology report to the file you got and does it come from another database, for example. But you state it's from a pathology report, so I think that's sufficient for now. Thank you for the update!

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: Updated markdown to save consensus subtype output merged with RNA+DNA id to tsv files: https://github.com/komalsrathi/OpenPBTA-analysis/blob/mb-class-nb/analyses/molecular-subtyping-MB/02-compare-classes.Rmd#L256
and 
https://github.com/komalsrathi/OpenPBTA-analysis/blob/mb-class-nb/analyses/molecular-subtyping-MB/02-compare-classes.Rmd#L279

Also updated the README at the analyses and module level.

komalsrathi:
@jaclyn-taroni I am doing a bit of a QC just to make sure what I have is correct - I will update you shortly when this is ready to review.

komalsrathi:
@jaclyn-taroni I think this is ready now. I just added another tab with consensus comparison with some details. Total accuracy of consensus output remains the same for batch corrected and uncorrected consensus output with 25/32 matches to the reported pathology subtypes. Between the two consensus outputs, there are 24/25 matches. **BS_V96WVE3Z** is correctly predicted by consensus uncorrected output but not by batch-corrected output and **BS_HB03GSHF** is correctly predicted by consensus batch-corrected output but not by uncorrected output.

Added the above info to module README  as well.

Issue #758, opened 2020-09-02
Medulloblastoma subtype propagation question


jharenza:
#### What data file(s) does this issue pertain to?
pbta-histologies.tsv
#757 

#### What release are you using?
Preparation for #732 

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
MB BS_5BMNK8SY DNA has two associated RNA samples: BS_QEYPFYHD (classifiers did not agree) and BS_Z7PKVY9J (classifiers agreed == Group4). In #444, we propagated RNA subtypes to DNA labels per #379, but this sample would create an issue with doing so. How should we proceed? These biospecimens are from the same event ID (7316-14) and are both from the Initial CNS tumor, so I _**think**_ it could be safe to add logic as the following:
```
IF there are two RNA specimens from the same event (`sample_id`) with the same `tumor_descriptor`, and 
IF one of the two RNA specimens has a consensus match from the MB classifiers, and 
IF the RNA specimen with mismatched classifications includes a match to the subtype of the consensus of the other RNA specimen, 
THEN propagate the subtype to the second RNA specimen.
```
😬 ?

cc @jaclyn-taroni @komalsrathi @baileyckelly

jharenza:
> @jharenza can you clarify what you mean by
> 
> ```
> the RNA specimen with mismatched classifications includes a match to the subtype of the consensus of the other RNA specimen
> ```

Sure - example: 

**Situation 1**
| Specimen   | MB-classifier | MM2S classifier |
|------------|---------------|-----------------|
| Specimen A | subtype X     | subtype X       |
| Specimen B | subtype X     | subtype Y       |

vs. 

**Situation 2**
| Specimen   | MB-classifier | MM2S classifier |
|------------|---------------|-----------------|
| Specimen A | subtype X     | subtype X       |
| Specimen B | subtype Y     | subtype Z       |

Talking about situation 1.

jaclyn-taroni:
If the end result is that the DNA specimen in situation 1 gets subtype X and in situation 2 the DNA specimen would be unclassified, what you are proposing seems reasonable.

jharenza:
> If the end result is that the DNA specimen in situation 1 gets subtype X and in situation 2 the DNA specimen would be unclassified, what you are proposing seems reasonable.

[...]

PR #760, opened 2020-09-08
fix issue with BS_5BMNK8SY

komalsrathi:

MB BS_5BMNK8SY DNA has two associated RNA samples: BS_QEYPFYHD (classifiers did not agree) and BS_Z7PKVY9J (classifiers agreed == Group4). This pull request aims to address the issue with the above discrepancy.

#### What was your approach?

Suggested by @jharenza 

```
IF there are two RNA specimens from the same event (`sample_id`) with the same `tumor_descriptor`, and 
IF one of the two RNA specimens has a consensus match from the MB classifiers, and 
IF the RNA specimen with mismatched classifications includes a match to the subtype of the consensus of the other RNA specimen, 
THEN propagate the subtype to the second RNA specimen
```

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/758

Output .tsv and .html

NA

Yes

Table

NA

#### Documentation Checklist


analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -249,13 +318,8 @@ print(paste0("Accuracy: ", consensus.corrected.acc))
...
-consensus.corrected.out <- clin.merged %>%
-  inner_join(consensus.corrected %>% 
-               mutate(Kids_First_Biospecimen_ID_RNA = Kids_First_Biospecimen_ID) %>%
-               dplyr::select(sample_id, Kids_First_Biospecimen_ID_RNA, molecular_subtype), 
-             by = c('sample_id', 'Kids_First_Biospecimen_ID_RNA'))
-
+# merged output with clinical ids 
+consensus.corrected.out <- out$consensus.mat.out
 write.table(consensus.corrected.out, 
             file = file.path(results_dir, 'MB_batchcorrected_molecular_subtype.tsv'),
```
cgreene, line 324: Have you considered sorting everywhere the TSVs get output by the sample index or similar? I was surprised to see how long the TSV part of the diff was. It would be easier to verify if only the lines that were changed had a diff associated. Sorting won't help this PR, but it will help in the future and is potentially good to implement going forward where smaller TSV files are output.

komalsrathi: @cgreene oh right, that makes sense. I'll make that change.

komalsrathi: I also realized that for the first 4 tables in the html,  I have the `Biospecimen` id but for consensus tables I only have the `Participant` ID, so I am going to fix that as well.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
```diff
@@ -114,43 +118,102 @@ compute.accuracy <- function(expected.input, observed.input){
...
+    inner_join(consensus.mat %>% 
+                 mutate(Kids_First_Biospecimen_ID_RNA = Kids_First_Biospecimen_ID), 
+               by = c('sample_id', 'Kids_First_Biospecimen_ID_RNA'))
+  
+  # resolve discrepancy with multiple sample ids
+  # apply prop.mb.subtype function on same event (`sample_id`) with the same `tumor_descriptor`
+  consensus.mat.out <- plyr::ddply(consensus.mat.out, 
+                                   .variables = c("sample_id", "tumor_descriptor"), 
+                                   .fun = function(x) prop.mb.subtype(mat = x))
+  consensus.mat.out$sample_id <- as.factor(consensus.mat.out$sample_id)
```
jaclyn-taroni, line 190: Why does `sample_id` need to be a factor?

jaclyn-taroni (approved):
👍  I believe the logic in `prop.mb.subtype()` is correct. I had one question.

analyses/molecular-subtyping-MB/02-compare-classes.Rmd
komalsrathi: @jaclyn-taroni Good question - I was having some issues with viewing the `sample_id` column with the DT::datatable function in the sense that the hyphens in the `7316-` ids were dynamically getting removed and I am still not sure why. After converting it to a factor, that issue is resolved. In addition to this, earlier it would give a fuzzy match so if you searched for 7316-14 it would also show 7316-1460 as well. Now, you can search for exact matches.

jaclyn-taroni: Huh, the `-` thing is weird, thanks!

Issue #770, opened 2020-09-12
Updated analysis: Sort molecular subtyping tables for better diffs


jaclyn-taroni:
#### What analysis module should be updated and why?

Any (non-deprecated) molecular subtyping table that doesn't explicitly sort the output table, likely all of these:

```
molecular-subtyping-ATRT
molecular-subtyping-chordoma
molecular-subtyping-embryonal
molecular-subtyping-EPN
molecular-subtyping-EWS
molecular-subtyping-HGG
molecular-subtyping-LGAT
molecular-subtyping-pathology
```

Should be updated such that the output is sorted by identifiers (probably `Kids_First_Participant_ID` and `sample_id`) so it will easier to tell what's changed via the diff when we make post-v17 release changes. First brought up in the context of the MB module: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/760/files#r485623021

#### What changes need to be made? Please provide enough detail for another participant to make the update.

Add `dplyr::arrange(Kids_First_Participant_ID, sample_id)` to a step before writing the table that contains the molecular subtype labels to file. 

For example,  add the `arrange()` step in this code block: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/43e13134d3e9a6a5e0a126a63f68797c4ecb9b34/analyses/molecular-subtyping-HGG/07-HGG-molecular-subtyping-combine-table.Rmd#L524

To get a sorted `analyses/molecular-subtyping-HGG/results/HGG_molecular_subtype.tsv`.

Note that `molecular-subtyping-EPN` is in Python, but there will need to be an equivalent step.

#### When do you expect the revised analysis will be completed?

If this goes smoothly, I estimate this will take ~1-2 days. This should be addressed _prior to any other molecular subtyping issues_, so this should happen this week.

#### Who will complete the updated analysis?

Likely someone from the CCDL.

jaclyn-taroni:
We're going to plan to 1) use the v16 data (so wait to merge #764) and 2) spread these modules out across the CCDL team to make this go faster. @cbethell @cansavvy @jashapiro  if you pick up a module, can you comment about which one you're working on? If you're picking up a module, be sure to hard refresh just in case the most recent comment hasn't shown up!

jashapiro:
STarting at the top- I will do `molecular-subtyping-ATRT`

jaclyn-taroni:
I will do `molecular-subtyping-chordoma`!

[...]

PR #772, opened 2020-09-14
Sort embryonal molecular subtyping output

jashapiro:
This PR is part of #770, adding sorting to the the ATRT molecular subtyping output table. Interestingly, it seems to have already been sorted.

While I was at it, I also sorted the biospecimen IDs table, which was not previously sorted.


jaclyn-taroni (approved):
Good call on the bonus change!

PR #773, opened 2020-09-14
Sort EPN molecular subtyping files

jashapiro:
This PR adds sorting to the EPN subtyping intermediate and output files, as part of fulfilling #770


jaclyn-taroni (approved):
🐍

Issue #608, opened 2020-03-05
Updated analysis: chordoma subtyping


jharenza:
#### What analysis module should be updated and why?
We should add the "poorly differentiating" subtype

#### What changes need to be made? Please provide enough detail for another participant to make the update.
We have a data [table](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-chordoma/results/chordoma_smarcb1_status.tsv) of SMARCB1 copy number information and RNA expression, but this should be reviewed by a pathologist or chordoma expert for final determination of the poorly differentiating subtype. I was also reading that [IHC](https://www.ncbi.nlm.nih.gov/pubmed/29483606) can inform/distinguish this, so if possible, they should also review histology imaging for confirmation. 

#### What input data should be used? Which data were used in the version being updated?
This [table](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-chordoma/results/chordoma_smarcb1_status.tsv) and histology images available on KidsFirst.

#### When do you expect the revised analysis will be completed?
?

#### Who will complete the updated analysis?
?
D3b to pull IHC images and coordinate with a disease expert.

cc @yuankunzhu @allisonheath

jharenza:
CC @stefankies for coordination

jaclyn-taroni:
I'm going to label this as `blocked` because this is external to analyses contained within the repository.

jharenza:
I will coordinate with pathology to get these reviewed.

jharenza:
After my some pathology report translations by Laura Scolaro, my pathology report review, CN inspection, and discussion with pathologist, Mariarita Santi, I have the following to report:

| Kids_First_Participant_ID | Kids_First_Biospecimen_ID_RNA | Kids_First_Biospecimen_ID_DNA | pathology_diagnosis      | sample_id | tumor_descriptor  | focal_SMARCB1_status | SMARCB1_expression | Path report inspection                                                                               | Poorly differentiated? Comments                              |
|---------------------------|-------------------------------|-------------------------------|--------------------------|-----------|-------------------|----------------------|--------------------|------------------------------------------------------------------------------------------------------|--------------------------------------------------------------|
[...]

PR #775, opened 2020-09-14
Sort chordoma SMARCB1 table

jaclyn-taroni:

Addresses the chordoma portion of #770. There are some expression values that have changed in the output, but there were not changes to the files committed to the repository. I am not sure why that is. However this particular module has been blocked for quite some time (#608), so I do not want to dig into those changes that do not appear to change interpretation at the moment.


jaclyn-taroni:
I was wrong - the values look the same on the GitHub interface. I think this was an artifact of looking at the diff in GitKraken.

cgreene (approved):
Looks good! 🚀

PR #776, opened 2020-09-14
Trim down docker images a bit

jashapiro:
#### What question is your analysis addressing?

Our Docker image is pretty big, and it was maintaining some downloaded packages, apparently due to the use of `install2.R`. 

#### What was your approach?

I have replaced the calls to `install2.R` with calls to the script we use for installing bioconductor packages: `install_bioc.R` which ultimately uses the base R `install.packages()` function.

I also added some cleanup of python packages as they are installed.

The net result is that the docker image created here is 8.5GB, vs. 9.4GB in current `latest` image. This difference is smaller than I expected, but perhaps my expectations were too high given the fact that we have a few large annotation packages built in. Still, an almost 1GB savings may be worth implementing. 

If CI passes, I would assume that all of the intended packages are installed as expected. It is possible, but hopefully unlikely, that some package versions have changed. That should be checked as well.

Instructions in the readme should be updated to indicate installation of all packages with `install_bioc.R`. If this passes CI (after other more pressing changes are fixed), and we decide this is worth doing, I will make those changes.


jaclyn-taroni:
The changes to the Dockerfile seem fine, but I think this PR should also include a docs change to this section of the README: https://github.com/AlexsLemonade/OpenPBTA-analysis#docker-image

jashapiro:
> The changes to the Dockerfile seem fine, but I think this PR should also include a docs change to this section of the README: https://github.com/AlexsLemonade/OpenPBTA-analysis#docker-image

Agreed! I was planning to add that now that I can confirm that everything is working.

jaclyn-taroni (approved):
Looks good to me!

PR #777, opened 2020-09-14
Sort HGG subtyping output

cbethell:

This PR adds sorting to the HGG subtyping output files, as a part of addressing issue #770.

#### What was your approach?

I added the following step before the output file is saved this module's `.Rmd` files:
dplyr::arrange(Kids_First_Participant_ID, sample_id)

Note however, that `dplyr::arrange(Sample)` and `dplyr::arrange(Kids_First_Biospecimen_ID)` were used in two instances based on the ID values available to use in the `05` and `06` notebooks, respectively.

#770 

- Are there any more places where the `arrange()` step should be added?

Yes

No new results, just sorting updates to the TSV files in this module.


jaclyn-taroni (approved):
😎

PR #764, opened 2020-09-11
v17-Release

baileyckelly:
### Purpose/implementation Section
v17-Release to update the pbta-histologies file
closes #732 

1. updated files description
2. updated release notes
3. Updated download script to point to v16/v17

#### What scientific question is your analysis addressing?

n/a

#### What was your approach?
n/a

#### What GitHub issue does your pull request address?

#637 Updated Histologies File
#631 LGAT subtyping
#623 EWS subtyping
#742  consensus of two classifiers for medulloblastoma 

### Directions for reviewers. Tell potential reviewers what kind of feedback you are soliciting.
compare v16 & v17 histologies file

#### Which areas should receive a particularly close look?
all values were pulled fresh from the source databases so should be reviewed for changes vs v16

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
n/a

### Results
n/a

#### What types of results are included (e.g., table, figure)?
n/a

#### What is your summary of the results?
n/a

#### Reproducibility Checklist

n/a

- [ ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

n/a

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

### Data Release Checklist
- [x] Is the table in doc/data-file-descriptions.md up to date?
- [x] Is doc/data-format.md up to date?
- [x] Is doc/release-notes.md up to date?
- [x] Is download-data.sh up to date?

PR #780, opened 2020-09-15
Add EWS to histology_color_palette.tsv

cansavvy:
### Summary: 

Saw the error in CircleCI from #764: 

```
Quitting from lines 434-454 (cn_status_heatmap.Rmd) 
Error in col2rgb(colors, alpha = TRUE) : invalid color name 'EWS'
Calls: <Anonymous> ... do.call -> <Anonymous> -> ColorMapping -> t -> col2rgb
```

The problem was as simple as `EWS` was not in the `histology_color_palette.tsv. 
 I checked out v-17-release, ran the cnv_heatmap.Rmd notebook, got the same error, but then I added EWS to the color palette and re-ran and cnv_heatmap.Rmd ran fine.


cansavvy:
Ended up making a commit to v17-release: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/764/commits/994a6f235f7bc641ec901183c5b93988f0842546 So this PR isn't needed now.

Issue #784, opened 2020-09-18
Updated analysis: Overhaul molecular subtyping results compilation step in molecular-subtyping-pathology


jaclyn-taroni:
#### What analysis module should be updated and why?

The [`molecular-subtyping-pathology`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology) module compiles the results from other  molecular subtyping modules in the notebook `analyses/molecular-subtyping-pathology/01-compile-subtyping-results.Rmd`. 

With the updates in v17 (#748), the molecular subtyping modules are upstream of the `broad_histology`, `short_histology` and `integrated_diagnosis` fields and the code in the modules are being altered to use `pathology_diagnosis`. Put slightly more visually (and simplifying the relationship between the downstream fields):

`pathology_diagnosis` ---molecular subtyping in the repo---> `integrated_diagnosis`, `broad_histology`, and `short_histology`.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

Two changes to `analyses/molecular-subtyping-pathology/01-compile-subtyping-results.Rmd` should be considered:

1. There are two modules (`molecular-subtyping-embryonal` and `molecular-subtyping-HGG`) that involve a step where all samples are checked for molecular alterations as part of inclusion criteria and as a result, these modules will not rely on `pathology_diagnosis` alone. It is possible, then, that a sample could end up with two subtype labels from two different modules: one on the basis of `pathology_diagnosis` alone and the one based on the molecular alteration inclusion step that has typically been deemed the more appropriate label once we go back to pathology. If a sample has two labels, we should prefer the one from `molecular-subtyping-embryonal` and `molecular-subtyping-HGG` in this step.

2. With the exception of the EWS results, we join the `integrated_diagnosis`, `short_histology`, `broad_histology`, and `Notes` from the histology file included in the data release `release-v15-20200228` that was prior to the feedback we received from pathology. We can either skip the inclusion of these columns or instead add the `integrated_diagnosis`, `short_histology`, and `broad_histology` columns on the basis of `pathology_diagnosis` and `molecular_subtype`. _Note that it's entirely not clear that this notebook is the right place for this or if we should add it at the end of each individual module._

#### What input data should be used? Which data were used in the version being updated?

The subtyping files in `master` should be used. As mentioned above, this notebook currently uses `release-v15-20200228`.

#### When do you expect the revised analysis will be completed?

[...]

Issue #667, opened 2020-04-01
Updated analysis: Include EPN subtyping in `molecular-subtyping-pathology`


jaclyn-taroni:
#### What analysis module should be updated and why?

`molecular-subtyping-pathology` was established before the completion of #245 (Related PR: #555), but ultimately it will need to be included.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

1. Add EPN table to the compilation of all the molecular subtyping/classification results https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/65180c183c0828dd093dd99a47b689ff3a76999f/analyses/molecular-subtyping-pathology/01-compile-subtyping-results.Rmd#L190

2. Context is https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/663#discussion_r401712388. There was an EPN sample that was reclassified via the defining lesion criteria in `molecular-subtyping-HGG`, but this sample should remain EPN based on pathology feedback. So we should check this sample for the features outlined in #245. 

Relevant section starts here:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/65180c183c0828dd093dd99a47b689ff3a76999f/analyses/molecular-subtyping-pathology/02-incorporate-pathology-feedback.Rmd#L124

#### What input data should be used? Which data were used in the version being updated?

Final table with subtyping labels I anticipate will be located in `analyses/molecular-subtyping-EPN/results` - this likely relies on the current data release (`release-v15-20200228`).

#### When do you expect the revised analysis will be completed?

Blocked by #245/#555 - but ~1 day after that

#### Who will complete the updated analysis?

@jaclyn-taroni

jharenza:
>So we should check this sample for the features outlined in #245.

It looks like this sample was somehow missed in the selection for the EPN module, so I think this should be added back there first.

jharenza:
added with #854

Issue #609, opened 2020-03-05
Proposed Analysis: Investigate integrated diagnosis vs pathology diagnosis


jharenza:
#### What are the scientific goals of the analysis?
1. To determine via pathology reports, slides, images whether the molecular (`integrated_diagnosis`) was something missed, could be explained by pathology, and/or was not suggested by pathology. 
2. To determine whether therapy would have been different given the `integrated_diagnosis`.
3. If 2 is `yes, different`, to investigate clinical outcomes/survival. 

#### What methods do you plan to use to accomplish the scientific goals?
Manual inspection of pathology reports, IHC images, slides, clinical data, etc.

#### What input data are required for this analysis?
Table from #607, pathology reports, IHC images, slides, clinical data, etc. (available on [KidsFirst](kidsfirstdrc.org)).

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
~2-4 weeks (beginning the week of March 9 with CHOP pathology)

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
CHOP pathologists Maria Rita Santi and Angela Viaene
@taylordm

#### What relevant scientific literature relates to this analysis?
NA

cc: @yuankunzhu @allisonheath @stefankies @jvlilly

jvlilly:
Looping in the PMs

Jena V. Lilly MS CCRC
Director of Operations & Strategic Planning
Center for Data Driven Discovery in Biomedicine
Children’s Brain Tumor Tissue Consortium
The Roberts Center for Pediatric Research
12th floor, 12473
2716 South Street.
Philadelphia, PA 19146
M: (609) 922-1092
lillyj@email.chop.edu<mailto:lillyj@email.chop.edu> ● d3b.center<http://d3b.center/>

[cid:image002.jpg@01D3E787.6D25A060]
CBTTC.org <http://cbttc.org/>  Kidsfirstdrc.org <http://kidsfirstdrc.org/>  PNOC.us<applewebdata://38657B7D-0294-4061-912A-1C2E4CE8110E/pnoc.us> Cavatica.org<http://cavatica.org/>

jharenza:
Sent Mariarita and Angela the following files, but they had requested us to write up findings in paragraph form. @adamcresnick to determine who will do this.


[...]

PR #785, opened 2020-09-18
molecular subtyping EPN update

### LINKED ISSUE #755: Updated analysis: EPN subtyping to use pathology diagnosis
https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-EPN

Currently, this module uses `integrated_diagnosis` to search for the subset of samples to be subtyped. However, with the rework of rules to determine  `integrated_diagnosis` post molecular subtyping (#748), we should rework this module to search for samples using `pathology_diagnosis` and the new field for free text pathology diagnosis, which will be present in `pbta-histologies.tsv in #732. 

pbta-histologies.tsv from V17 instead of from V16

unsure

unsure

jashapiro:

With updates to `pbta-histologies.tsv`, we needed to redo molecular subtyping analysis to use the renamed columns and account for any data changes. 

#### What was your approach?

I substituted `pathology_diagnosis` for `integrated_diagnosis` wherever it occurred in the scripts, then reran  

Closes #755

The code changes are minimal, but some of the results have changed more than I expected. There are two additional participants included: `PT_7E3V3JFX` and `PT_80NVYCBS`

`PT_7E3V3JFX` was previously classified in `integrated_diagnosis` as `Diffuse midline glioma`
`PT_80NVYCBS` was previously classified in `integrated_diagnosis` as `CNS Embryonal Tumor`

Do these changes make sense? Do we expect these samples be reclassified in `integrated diagnosis` for other reasons? If so, what is the correct order in which these reclassifications should be performed? In other words, should these samples be excluded by some other criteria before proceeding with the subtyping here?

I do not think any final subtyping results here changed, but one could imagine that including or excluding samples could have such an effect, as the RNA expression levels used here are based on within-group z score normalization. (Should the z scores here actually be calculated for all samples, not just EPN samples?)

I did not use `pathology_free_text_diagnosis` as there did not seem to be any additional information in that field that related to EPN diagnosis: All samples seemed to be captured by `pathology_diagnosis`. Is it a safe assumption for the future that this will remain the case?

See above

#### Documentation Checklist


jharenza:
Hi @jashapiro! Thanks for working on this.

>PT_7E3V3JFX was previously classified in integrated_diagnosis as Diffuse midline glioma
PT_80NVYCBS was previously classified in integrated_diagnosis as CNS Embryonal Tumor

>Do these changes make sense? Do we expect these samples be reclassified in integrated diagnosis for other reasons? If so, what is the correct order in which these reclassifications should be performed? In other words, should these samples be excluded by some other criteria before proceeding with the subtyping here?
>

For `PT_7E3V3JFX`, this makes sense and will stay as EPN. What happened was early on, we found the H3 K28M histone mutation, but after [pathology review](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/609#issuecomment-602821376), it was deemed a rare EPN with H3 K28M. This is reflected in [molecular-subtyping-pathology](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-pathology). Since that module is done last, I think this subtype would get updated to include that mutation later.

For `PT_80NVYCBS`, this sample was re-classified in [embryonal subtyping](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-embryonal/results/embryonal_tumor_subtyping_relevant_data.tsv) due to the presence of MN1 fusions and subsequently verified as embryonal during pathology review and integrated into [molecular-subtyping-pathology](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-pathology). Same sentiment - since pathology subtyping occurs last, this sample would be safely updated later.

jaclyn-taroni:
@jharenza I think what you're describing is at least partly covered by the first point I raised in #784 but if you'd like to check and also expand on 

> Perhaps that is how we can keep track - another file for any changes to diagnosis that will accumulate as we do the modules?

That issue is probably a better place to keep track of those ideas. @jashapiro your feedback on #784 is welcome, too, of course 😁

jaclyn-taroni:
> I do not think any final subtyping results here changed, but one could imagine that including or excluding samples could have such an effect, as the RNA expression levels used here are based on within-group z score normalization. (Should the z scores here actually be calculated for all samples, not just EPN samples?)

The original issue (#245) describes certain genes being overexpressed relative to other EPN tumors, which is why within-group makes sense to me. There are about 90 of these samples and I expect only a handful (low single digits) would ever be removed in the context of this project so I'm inclined for the process to remain that we conduct the EPN subtyping upstream of where we compile all the other results - we can discuss that point more on #784! We also need to add EPN in in the first place (#667).

jashapiro:
> The original issue (#245) describes certain genes being overexpressed relative to other EPN tumors, which is why within-group makes sense to me. There are about 90 of these samples and I expect only a handful (low single digits) would ever be removed in the context of this project so I'm inclined for the process to remain that we conduct the EPN subtyping upstream of where we compile all the other results - we can discuss that point more on #784! We also need to add EPN in in the first place (#667).

Sounds good to me. I was mostly just surprised by the extent of changes (of small magnitude), so I was digging a bit to see what had happened. As long as the results still make sense (and as I said, I don't think any subtypes changed), we should be fine.

jaclyn-taroni (approved):
👍 looks good to me!

Issue #756, opened 2020-08-29
Updated analysis: MB subtyping to use pathology diagnosis


jharenza:
#### What analysis module should be updated and why?
https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-MB

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Currently, this module uses `integrated_diagnosis` to search for the subset of samples to be subtyped. However, with the rework of rules to determine  `integrated_diagnosis` post molecular subtyping (#748), we should rework this module to search for samples using `pathology_diagnosis` and the new field for free text pathology diagnosis, which will be present in `pbta-histologies.tsv in #732. 

#### What input data should be used? Which data were used in the version being updated?
pbta-histologies.tsv from V17 instead of from V16

#### When do you expect the revised analysis will be completed?
unsure

#### Who will complete the updated analysis?
unsure

jaclyn-taroni:
@cansavvy since #753 is apparently blocked, why don't we have you plan to pick up this issue? I've filed #786 (HGG) which it may be helpful to have a look at for inspiration re: first look at the `pathology_diagnosis` and `pathology_free_text_diagnosis`.

cansavvy:
Sounds good! I'll take a look at this today!

cansavvy:
I'll push a small notebook that summarizes this, but I did some poking and MB substyping seems to be a bit more straightforward than some of the other histologies: 

- Looks like all samples that have `short_histology` and `integrated_diagnosis` as `Medulloblastoma` also have `pathology_diagnosis` as `Medulloblastoma`. 
- Most but not all of these samples have `medulloblastoma` some place in `pathology_free_text_diagnosis`, but a couple have `PNET`. 
- There are no samples currently classified as `short_histology == "Other"` that have `medulloblastoma` as a part of `pathology_free_text_diagnosis`

jaclyn-taroni:
Closed via #787

Issue #748, opened 2020-08-28
Update/Create SQL rules for integrated diagnosis, broad, and short histologies


jharenza:
#### What data file(s) does this issue pertain to?
pbta-histologies.tsv

#### What release are you using?

will use V17 from #732 

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
https://alexslemonade.github.io/OpenPBTA-manuscript/#clinical-data-harmonization

#### Put your question or report your issue here.
@baileyckelly or @aadamk, correct me if I am wrong:

This is an overview of the new workflow for creating the `pbta-histologies.tsv` file:
![2020-06-23_Bixu_ADAPT_Workflow](https://user-images.githubusercontent.com/207034/91615461-a51d3900-e951-11ea-8407-2e4272190a32.png)

##### `integrated_diagnosis`:
Using updated `pathology_diagnosis` values:
1. Generate updated `integrated_diagnosis` using molecular subtyping modules in [openPBTA_analysis repo](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses#modules-at-a-glance), later to be implemented in [d3b analysis toolkit](https://github.com/d3b-center/d3b-bix-analysis-toolkit) (@kgaonkar6)
2. Then run [OpenPBTA pathology module](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-pathology) to update `integrated_diagnosis` using new pathological diagnoses, also part of the openPBTA_analysis repo as in (1) and to be implemented in the d3b analysis toolkit (@kgaonkar6)
3. Update CBTTC 'other' to new `integrated_diagnosis` if there are continued updates in the data warehouse (@baileyckelly)
4. For all remaining, leverage original `pathology_diagnosis` and populate as `integrated_diagnosis` (@baileyckelly)

##### `broad_histology`:
- Dependent on `integrated_diagnosis`
- Broad Histology: Broad 2016 WHO classification ([https://doi.org/f8mspx](https://doi.org/f8mspx), [https://alexslemonade.github.io/OpenPBTA-manuscript/](https://alexslemonade.github.io/OpenPBTA-manuscript/), [https://link.springer.com/article/10.1007/s00401-016-1545-1/tables/1](https://link.springer.com/article/10.1007/s00401-016-1545-1/tables/1))
    - Process
[...]

Issue #746, opened 2020-08-27
Medulloblastoma pathology subtypes: file for data release


jharenza:
#### What data file(s) does this issue pertain to?
A working GDoc of medulloblastoma subtypes being extracted from pathology reports [here](https://docs.google.com/spreadsheets/d/1lfEYhrKA2CJAKVTQVEvFRi2vj-1daUw4FGAehaHb358/edit#gid=0). 

#### What release are you using?

v16

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
Column B was used for #743, but:
1. We will want a stable version of this file, rather than a google doc, so we should include in a future release
2. We found some data entry errors, these are undergoing re-review, and the results may change the accuracies for #743.

jharenza:
cc @yuankunzhu just to be aware

jaclyn-taroni:
While addressing #756, specifically making the switch to use `pathology_diagnosis` and `pathology_free_text_diagnosis` instead of `short_histology`, we now have a sample that was not originally included in `molecular-subtyping-MB` being subtyped (changed from `Medulloblastoma` -> `ETMR` in `molecular-subtyping-embryonal`) . The code in `analyses/molecular-subtyping-MB/03-compare-classes.Rmd` is tied closely to the structure of the expected classes file. 

In https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787/commits/aea735ead26e05b21c2b2e0cb41a885688d6a9c3, I added a version of this file called `analyses/molecular-subtyping-MB/input/expected_class_v17.rds` with `NA` values in `pathology_subtype` and `other_molecular_findings` columns for this sample.

This additional sample should be taken into account as this issue is being addressed.

jharenza:
That sample `BS_QWNBZ9RJ` was mistakenly entered as medulloblastoma into our database, but will add it to this list, and try to see if we can get that updated on the backend. Thanks!

jharenza:
> That sample `BS_QWNBZ9RJ` was mistakenly entered as medulloblastoma into our database (path report had handwritten `medulloblastoma` but final dx was `ETMR`), but will add it to this list, and try to see if we can get that updated on the backend. Thanks!

[...]

PR #787, opened 2020-09-21
MB subtyping update to `pathology_diagnosis`

cansavvy:
Realigning Medulloblastoma subtyping to use labels based on `pathology_diagnosis` instead of `integrated_diagnosis` in accordance with #748  

#### What was your approach?

This is partially modeled after HGG subtyping changes #786, but its more straightforward when it comes to the terms.  

I made a `00-mb-select-pathology-dx.Rmd` notebook which explores the alignment of all the histology fields (e.g. `pathology_diagnosis`and creates a subset medulloblastoma histology file. Because the `short_histology`, `integrated_diagnosis`, and  `pathology_diagnosis` are pretty well in agreement for what biospecimen's are `Medulloblastoma` this is a pretty straightforward filtering step. But if we need to adjust this in the future, we can easily add medulloblastoma terms to the JSON file which is used downstream. 

I then rearranged all the subsequent files called in `run-molecular-subtyping-mb.sh` to use the JSON in terms in their filtering steps. This also involved some renumbering of the files and adding options that currently didn't exist. 

There were also a list of warnings printed out from the mb-classifier script having to do with the characters and string manipulations, so I added some steps that smoothed those steps a bit. 

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/756

- I am not the original author of this code, so I'm not super familiar with it. Are there any places that need to be updated to use the medulloblastoma terms JSON file that aren't? 
- Are there other steps that need adjusting that I missed? 
- Why are the results not changed at all with one extra sample? What did I miss? Or is this expected? 

I might have expected the results might change because of the one added sample that was `ETMR` when we used `short_histology` but with `pathology_diagnosis` is now `Medulloblastoma`.
But I also don't understand enough about what steps might have lost this biospecimen's data? 

#### Documentation Checklist


cansavvy:
I am a bit confused on where `input/expected_class.rds` comes from (as mentioned in my questions for reviewers above). It doesn't seem to be mentioned in the README for this analysis and I couldn't find information about it in the PR it was added in: `input/expected_class.rds` https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/738 Is there somewhere else it is described? Will we need to have updates to it now or sometime in the future?

jaclyn-taroni:
I think what you're looking for is tracked in #746 @cansavvy.

jaclyn-taroni:
Can you tell me more about why you would want to run the first part of the `analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd` notebook every time you run this module? I understand why you need to run the filtering step in that notebook every time, but the way #786 is set up is to effectively decouple the exploration of the pre #748 `pbta-histologies.tsv` from the filtering.

cansavvy:
> Can you tell me more about why you would want to run the first part of the `analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd` notebook every time you run this module? I understand why you need to run the filtering step in that notebook every time, but the way #786 is set up is to effectively decouple the exploration of the pre #748 `pbta-histologies.tsv` from the filtering.

My thought was that you need to look at the explorations to make a decision about filtering, so I found it handy to have the filtering step that would need to be changed as well as the results of the exploration in one place. This is of course dependent on the fact that because the exploration steps are very quick (if the exploration steps took a lot of time or memory or were very long, I guess it would be more necessary to separate them). 

Does that work or should I look to reformat this?

jaclyn-taroni:
In a situation where `integrated_diagnosis` and `short_histology` values are dependent on `molecular_subtype` (output from this module) in the next release, I'm not sure that the exploration you're including is appropriate anymore.

cansavvy:
> t on `molecular_subtype` (output from this module) in the next release, I'm not sure that the exploration you're including is appropriate anymore.

I see hadn't thought about that. The other steps/explorations would still be pertinent though (I think)?

jaclyn-taroni:
I looked back at https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/123#issuecomment-694403021 (which may inform changes to #786 🤔 ) - the steps that _always_ seem relevant to me are including samples based on 
  1. exact matching for `Medulloblastoma` in `pathology_diagnosis` 
  2. if `short_histology == "Other" & str_detect(pathology_free_text_diagnosis, <relevant strings>)`. 

Arguably you could perform these steps in `analyses/molecular-subtyping-MB/01-filter-and-batch-correction.R` as was done originally (originally `integrated_diagnosis == "Medulloblastoma"`). 

`00` could explore the v17 data and prep a JSON file with the strings used for matching and is therefore designed to never be rerun and pass the JSON file to `01`. (This is the model used in HGG but without command line arguments.) 

Or, if we expect that `pathology_diagnosis` and `pathology_free_text_diagnosis` are frozen as of v17 and this is pretty clear cut (it is in my opinion!), we do exploration in `00` that's tied to v17 (and never intended to be rerun) and the filtering as described above in the `01` step with hardcoding the filtering as it was previously done.

cansavvy:
I ended up doing the JSON strategy described above: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787#issuecomment-696422415

Not sure if we expect the terms to change, but if they do it will be easy enough to add terms to the JSON file in 00-mb-select-pathology-dx.Rmd. The 00-mb-select-pathology-dx.Rmd is not run automatically (with the run-molecular-subtyping-mb.sh script) but the JSON file it creates is used by the downstream files. 

This is described in the README updates in this PR.

jharenza:
> I ended up doing the JSON strategy described above: [#787 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787#issuecomment-696422415)

This looks great and terms are exactly what we should use for MB for now!

analyses/molecular-subtyping-MB/01-filter-and-batch-correction.R
```diff
@@ -42,10 +47,15 @@ polya <- readRDS(polya.file)
...
+clin <- read.delim(clin_file, stringsAsFactors = F)
 clin.mb  <- clin %>%
-  filter(experimental_strategy == "RNA-Seq",
-         integrated_diagnosis == "Medulloblastoma") %>%
+  # Inclusion on the basis of strings in pathology_diagnosis and 
+  # pathology_free_text_diagnosis
+  filter(str_detect(str_to_lower(clin$pathology_diagnosis), 
+                    paste0(path_dx_list$include_path_dx, collapse = "|")) | 
+           str_detect(str_to_lower(clin$pathology_free_text_diagnosis), 
+                      paste0(path_dx_list$include_free_text, collapse = "|")),
```
jaclyn-taroni, line 57: This doesn't appear to be doing exact matching per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787#issuecomment-696422415. It doesn't surprise me that this is giving us the expected result, but I think since we have the ability to be more stringent here (compared to what shows up in `pathology_diagnosis` for the HGAT samples) we can go ahead and do that.

analyses/molecular-subtyping-MB/03-compare-classes.Rmd
```diff
@@ -33,15 +33,32 @@ suppressPackageStartupMessages(library(plyr))
...
+# exclude samples for subtyping - see 00-HGG-select-pathology-dx
+path_dx_list <- jsonlite::fromJSON(terms.file)
 ```
 </details>
 
 <details><summary>Read input</summary>
 ```{r read_data, echo = TRUE}
 # read input
-clin <- read.delim(file.path(data_dir, "pbta-histologies.tsv"))
+subset.clin <- read.delim(clinical_file) %>%
```
jaclyn-taroni, line 56: To me, this suggests that part of the filtering step in `01` could be to output a subset version of `pbta-histologies.tsv` that can be used here so you don't have to change the filtering code in two places should you need to change the filtering code. I took some similar steps on #786 in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/786/commits/e8480cd94bda49ef6e390aea7575870ba2c4e37d and https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/786/commits/c808a512ddce6e975ff4c63e4335e78ba3faa2ba.

You'd first filter based on the criteria that make something a sample included for subtyping, output that file, and then filter to RNA-seq samples to be used for the rest of that script. Side note: I don't know that you actually need to filter to RNA-seq samples looking at the rest of the code in `analyses/molecular-subtyping-MB/01-filter-and-batch-correction.R` because there's an `intersect()` step prior to `dplyr::select()`.

analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,138 @@
...
+```
+
+There's no samples in `Other` that need to be added at this time, but if they do later, we can add them to the filter statements in the next section. 
+
+### Save the strings we'll use downstream
+
+Create a list with the strings we'll use for inclusion.
+
+```{r}
+terms_list <- list(include_path_dx = "medulloblastoma",
```
jaclyn-taroni, line 123:
```suggestion
terms_list <- list(include_path_dx = "Medulloblastoma",
```
For exact matching reasons - you'd then use `%in%` when you implement matching downstream for future-proofing reasons.

jaclyn-taroni (commented):
Overall structure is coming along! This does not quite yet do what I would expect based on the two conditions for inclusion laid out in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787#issuecomment-696422415. I am not surprised that the current implementation works as expected, but if we have the opportunity be more stringent than the approach in HGG (#786), I think we should be more stringent.

analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,138 @@
...
+- [Background](#background)
+- [Set up](#set-up)
+- [Read in data](#read-in-data)
+- [Explore the pathology diagnoses](#explore-the-pathology-diagnoses)
+- [Pathology diagnosis strings for inclusion](#pathology-diagnosis-strings-for-inclusion)
+- [Any `Other` samples that should be included based on `pathology_free_text_diagnosis`](#any-other-samples-that-should-be-included-based-on-pathology_free_text_diagnosis)
+- [Subset to Medulloblastoma based on `pathology_diagnosis`](#subset-to-medulloblastoma-based-on-pathology_diagnosis)
+- [Session Info](#session-info)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
```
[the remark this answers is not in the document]
cansavvy: I can delete it entirely. That's fine.

analyses/molecular-subtyping-MB/03-compare-classes.Rmd
cansavvy: Oh yeah, I hadn't noticed that. Do we want to drop the whole thing? in `01-` and only do filter in `03-`?

cansavvy: I have determined that a later step apparently does require the filtering be exact.

jaclyn-taroni: I didn't follow that, but I will say that I believe you do have to filter based on the criteria we've been talking about in `01` to get the correct expression matrices committed to the repo (if you don't commit those to the repo, this will fail CI because there's not a guaranteed number of MB samples in the CI files), you just don't need `experimental_strategy == "RNA-Seq"` step there as far as I can tell and you don't want to have the filtering code in two places (`01` and `03`).

jaclyn-taroni: Oh I see, the batch correction step requires the filtering to RNA-seq. Anyway, the points about not repeating oneself still stand.

cansavvy: Yeah, I did some rewriting of that section see how you feel about it.

cansavvy: I think it is less redundant now overall.

analyses/molecular-subtyping-MB/03-compare-classes.Rmd
```diff
@@ -33,15 +33,28 @@ suppressPackageStartupMessages(library(plyr))
...
+
+# this file was created in the 01-filter-and-batch-correction.R
+subset_clin_file <- file.path("input", "subset-mb-clinical.tsv")
+```
+
+```{r}
+# Read in the JSON file that contains the strings we'll use to include or
+# exclude samples for subtyping - see 00-HGG-select-pathology-dx
+path_dx_list <- jsonlite::fromJSON(terms.file)
 ```
```
jaclyn-taroni: I think this can go now with your most recent update.

cansavvy: I just found that. It's deleted in the next commit.

cansavvy:
I'm noticing the `medulloblastoma-exprs.rds` file is changed, but I think this is only due to sample order.

analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,122 @@
...
+In an upcoming release, `integrated_diagnosis`, which can be updated as the result of subtyping, will be used to populate the `short_histology` column (see [#748](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/748)).
+Thus, molecular subtyping modules need to be upstream of `short_histology` and use the `pathology_diagnosis` and `pathology_free_text_diagnosis` fields.
+
+This change for this module is tracked in [#756](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/756).
+ 
+_As of v17 release, this is pretty straightforward for Medulloblastoma samples:_  
+
+- `short_histology`, `integrated_diagnosis` and `pathology_diagnosis` are fairly in agreement as far as which samples are `Medulloblastoma`  
+- Most but not all of these samples have `medulloblastoma` some place in `pathology_free_text_diagnosis`, but a couple have `PNET`.   
+- There are no samples currently classified as `short_histology == "Other"` that have `medulloblastoma` as a part of `pathology_free_text_diagnosis`.
```
jaclyn-taroni, line 22:
```suggestion
- There are no samples currently classified as `pathology_diagnosis == "Other"` that have `medulloblastoma` as a part of `pathology_free_text_diagnosis`.
```

jaclyn-taroni:
I got to the bottom of

> Why are the results not changed at all with one extra sample? What did I miss? Or is this expected?

The problem was that the one additional sample was not included in the `expected_class.rds` file. I added a new version of this file (`expected_class_v17.rds`) in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/787/commits/aea735ead26e05b21c2b2e0cb41a885688d6a9c3 and added information about that step to #746.

jaclyn-taroni (approved):
With the most recent changes, this looks good to me!

Issue #732, opened 2020-07-23
Planned Data Release: V17


baileyckelly:
What data file(s) does this issue pertain to?
pbta-histologies.tsv

What release are you using?
V15/V16

Put your question or report your issue here.
Update histologies file to include additional information per tickets:

was planned for v16:
~#607~
~#610~
#637
#509
#608
#245
new:
#631 LGAT subtyping
#623 EWS subtyping
Who is going to work on this:
@baileyckelly

baileyckelly:
Note: This release will ONLY contain histology file changes

jharenza:
@baileyckelly @allisonheath - can we add an estimated timeframe for this on the ticket? Thanks :)

jharenza:
@jaclyn-taroni, since #743 was merged, would you like us to add these results to the `pbta-histologies.tsv` file for V17 release? If yes, should we use corrected or uncorrected results?

jaclyn-taroni:
I suppose my preference would be to use uncorrected, but that opinion is not strongly held. 

Can we get a time frame for this? There are items listed here (#610) that have not been completed. There are outstanding things like #665 that have not been addressed, which could inform the results that were initially included in #631. If we complete #667, we can use whatever is in [`molecular-subtyping-pathology`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/9efbddf883ca0fe5805fb1cc1cc27eefcfb9ac72/analyses/molecular-subtyping-pathology), which compiles all the subtyping results, as input into the histologies file and may give us an opportunity to address #735 if that's something we'd like to do.

I also know (but I didn't find this documented - apologies if I missed something) there are changes to `broad_histology` coming that may impact the subtyping modules. 

What order should these items be addressed in? Should we include anything that's ready now and deal with the `broad_histology` change + anything outstanding that I mentioned above at the same time -> V18?

jharenza:
#608 is blocked, so that will be added to V17

kgaonkar6:
I ran the molecular-subtyping-pathology ( last run on July 2nd  ) module for v17 release which contains subtyping results compiled from the following modules and pathology updates:
molecular-subtyping-EWS
molecular-subtyping-HGG
molecular-subtyping-LGAT
molecular-subtyping-embryonal

cansavvy:
I've realized the TCGA consensus and tmb files have never been added to the data release and it would be ideal to keep these tracked.

[...]

PR #788, opened 2020-09-22
Update molecular subtyping embryonal module to use pathology diagnosis fields

### LINKED ISSUE #752: Updated analysis: embryonal subtyping to use pathology diagnosis
https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-embryonal

Currently, this module uses `broad_histology` to search for the subset of samples to be subtyped. However, with the rework of rules to determine `broad_histology` post `integrated_diagnosis` (#748), we should rework this module to search for samples using `pathology_diagnosis` and the new field for free text pathology diagnosis, which will be present in `pbta-histologies.tsv in #732. 

`pbta-histologies.tsv` from V17 instead of from V16

unsure

unsure

cbethell:

As noted on the issue, with updates to `pbta-histologies.tsv` in #732, we needed to rework molecular subtyping embryonal module to use the renamed columns and account for any data changes. 

#### What was your approach?

I used some of the information in the HGG update PR #786 as a guide for my approach in this PR.

I added the `00-embryonal-select-pathology-dx.Rmd` to explore the values in the `pathology_diagnosis` and `pathology_free_text_diagnosis` fields when the histologies file is filtered for `broad_histology` (as this is the variable that is currently being filtered on in this module).

A JSON file is saved with the pathology diagnosis fields to include and exclude, and used to subset/filter in the `01-samples-to-subset.Rmd` and `04-table-prep.Rmd` files.
 
This was done following the guidance in the [original comment for this issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issue-688417429), which states that “we should rework this module to search for samples using pathology_diagnosis and the new field for free text pathology diagnosis” and the [2016 WHO classification](https://link.springer.com/article/10.1007/s00401-016-1545-1/tables/1)  for further validation. 

This PR closes #752

- Does the workflow of the `00` notebook, mainly the definition of which values in the `pathology_diagnosis` and `pathology_free_text_diagnosis` fields that should be included/excluded make sense? (based on the previous workflow and the 2016 WHO Classification)
- Many of the result files notably changed, do these changes make sense?

Yes.

No new results, just updated TSV result files.

There were notably some embryonal samples added to the `molecular_subtype` column of the final `embryonal_tumor_molecular_subtypes.tsv` file, but also a few removed that are a bit worrisome: `PT_BDM0M83S`, `PT_0T2EWNYR` and `PT_FZHGKJ0H`. Do these changes make sense?

#### Documentation Checklist


analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,177 @@
...
+histologies_df %>% 
+  filter(broad_histology == "Embryonal tumor",
+         !(pathology_diagnosis %in% c("Medulloblastoma",
+                                   "Atypical Teratoid Rhabdoid Tumor (ATRT)"))) %>%
+  count(pathology_diagnosis) %>%
+  arrange(desc(n))
+```
+
+For the most part, this is as we would expect given the 2016 WHO classifications.
+There is an ependymoma sample and a high-grade glioma sample that should probably be excluded.
```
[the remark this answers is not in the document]
jaclyn-taroni: Notably - the EPN sample appears to be changed to embryonal due to the presence of an _MN1_ fusion https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/785#issuecomment-695015488. We only look at _TTYH1_ fusions in `analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd` so this may be a good opportunity to update that notebook to include samples on the basis of the presence of _MN1_ fusions.

jaclyn-taroni (commented):
Based on https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/123#issuecomment-694403021, your results here, and my own poking around, I think we need a different strategy here than we have in HGG. Specifically, we should filter `pathology_diagnosis` to relevant labels and then if a specimen has `pathology_diagnosis == "Other"` then we can look at `pathology_free_text_diagnosis`. I left a more detailed comment on `analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd` to this effect. 

Let me know if you have questions!

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -112,32 +120,41 @@ fusion_biospecimens <- c(
...
+  # pathology_free_text_diagnosis
+  filter(str_detect(str_to_lower(pathology_diagnosis), 
+                    paste0(path_dx_list$include_path_dx, collapse = "|")), 
+           str_detect(str_to_lower(pathology_free_text_diagnosis), 
+                      paste0(path_dx_list$include_free_text, collapse = "|")),
+         # Exclude samples on the basis of this string
+         str_detect(str_to_lower(pathology_diagnosis),
+                    paste0(path_dx_list$exclude_path_dx, collapse = "|"),
+                    negate = TRUE)
+  )
```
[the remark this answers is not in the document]
jaclyn-taroni: I forgot to note re: 

> The strings I've included above come from filtering the histologies file based on `broad_histology == "Embryonal tumor"` and `pathology_diagnosis == "Other"` and taking a look. 

I also looked at the **Other embryonal tumors** section of Louis et al.

jaclyn-taroni:
@cbethell given this comment on the original issue https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066 and my feedback here, can you put together a plan for how you would address both and comment here?

analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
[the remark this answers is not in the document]
jaclyn-taroni: @jharenza wanted to bring your attention to this comment specifically - should we update the step where we include samples (looking at the entire cohort) on the basis of the presence of fusions to include _MN1_ fusions and _TTYH1_ fusions (only the latter is addressed currently)? I suspect yes - I can not see why this EPN sample would be included otherwise.

cbethell:
> @cbethell given this comment on the original issue [#752 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066) and my feedback here, can you put together a plan for how you would address both and comment here?

Based on your above review comments @jaclyn-taroni and the above mentioned [comment on the original issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066), the plan moving forward for this PR is to:

- Filter labels that are in `pathology_diagnosis` to include Neuroblastoma and Supratentorial or Spinal Cord PNET.
- If the `pathology_diagnosis` == "Other", the next step will be to look at the free text field (pathology_free_text_diagnosis) for the following values: `embryonal tumor with multilayer rosettes, ros (who grade iv)`, `embryonal tumor, nos, congenital type`, `ependymoblastoma`, `medullooepithelioma`.
- The following logic will also be implemented to ensure that we capture CNS neuroblastomas and not peripheral: `pathology_diagnosis == CNS Neuroblastoma && primary_site != Other locations NOS && !(str_detect(pathology_free_text_diagnosis, "peripheral"))`
- A step to explore the histologies file based on broad_histology == "Embryonal tumor" and pathology_diagnosis == "Other" will also be added.

jaclyn-taroni:
Taking a quick look at the histologies file – I don't think there are instances of `CNS Neuroblastoma` in `pathology_diagnosis`; those are in `broad_histology` which will be downstream. So I think the move is to _detect_ `neuroblastoma` in `pathology_diagnosis` because there are instances of `Metastatic secondary tumors;Neuroblastoma` in that column and then your other stated steps (`primary_site`, `pathology_free_text_diagnosis`) sound good to me!

analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,171 @@
...
+
+We know we need to drop medulloblastoma and ATRT samples.
+We can also exclude LGG and neurofibroma samples on the basis of the values in `pathology_diagnosis`.
+
+```{r}
+exclude_path_dx_terms <- c("medulloblastoma",
+                           "atypical teratoid rhabdoid tumor (atrt)",
+                           "low-grade glioma/astrocytoma",
+                           "neurofibroma/plexiform;other")
+```
```
jaclyn-taroni, line 149: _This is a more general comment that could have gone in several places in line._

My interpretation of https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066 is that we can do the following:

1. Exact matching for `Supratentorial or Spinal Cord PNET`, e.g., `pathology_diagnosis  == "Supratentorial or Spinal Cord PNET"` - but use `%in%` for future proofing (more on that below). 
2. If `pathology_diagnosis == "Other"`, then detection in `pathology_free_text_diagnosis` of
```
embryonal tumor with multilayer rosettes, ros (who grade iv)
embryonal tumor, nos, congenital type
ependymoblastoma
medullooepithelioma
```
(`str_detect()`, `str_to_lower()`, `paste0(..., collapse = "|")` all working together like we have)

3. Detection of `neuroblastoma` in `pathology_diagnosis` (the `str_detect()`, etc. strategy again ☝️ ) and `primary_site != "Other locations NOS"` and `pathology_free_text_diagnosis` does not contain `"peripheral"` (`str_detect(str_to_lower(...), negate = TRUE)`) (Side note: Do we want `&` or `&&` throughout? I have not given much thought about how `&&` behaves with filtering with `dplyr`.)

Which I think means we will not need to do any exclusion on the basis of terms that indicate that a specimen is medulloblastoma or ATRT. 

Instead you can have a set of terms for exact matching in the JSON file that you use in conjunction with `%in%` to filter to exact matches and then the free text terms for inclusion. Here's an example in the MB module for putting together a JSON file to be used in this way: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/a0ae052487a72790f7eaec318815f47eae13b466/analyses/molecular-subtyping-MB/00-mb-select-pathology-dx.Rmd#L106

And here's the filtering: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/a0ae052487a72790f7eaec318815f47eae13b466/analyses/molecular-subtyping-MB/01-filter-and-batch-correction.R#L53

I agree with your choice to keep the NBL steps separate/just in the filtering.

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -112,32 +149,39 @@ fusion_biospecimens <- c(
...
+  # Inclusion on the basis of strings in pathology_diagnosis and 
+  # pathology_free_text_diagnosis
+  filter(str_detect(str_to_lower(pathology_diagnosis), 
+                    paste0(path_dx_list$include_path_dx, collapse = "|")) |
+           pathology_diagnosis == "Other" &&
+           str_detect(str_to_lower(pathology_free_text_diagnosis), 
+                      paste0(path_dx_list$include_free_text, collapse = "|")) |
+           # "neuroblastoma" will also need to be detected in the `pathology_diagnosis` column, but we need to ensure that the `primary_site` != "Other Locations NOS" and the `pathology_free_text_diagnosis` column does note contain "peripheral"
+         str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma") & !(str_detect(str_to_lower(primary_site), "other locations nos")) & !(str_detect(str_to_lower(pathology_free_text_diagnosis), "peripheral"))
+  )
```
jaclyn-taroni, line 173: This code block is a lot! And that makes sense because there are a lot of considerations in what has been described for this issue and what I described in another comment in this pull request. You could instead create three separate data frames - one of each condition for inclusion - and bind them together and filter to unique entries. That should accomplish the same thing, but be easier to follow. You can see an example of this approach in the HGG filtering step in what's open on #786 (I'm permalinking to the branch): https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/37a07d2657d92bd1a06168c0a514642d7f9566b6/analyses/molecular-subtyping-HGG/02-HGG-molecular-subtyping-subset-files.R#L121

jaclyn-taroni (commented):
As promised, I had a few comments on the logic that I wanted to pass along before you dig into why we're seeing the results change. Let me know if you have any questions!

cbethell:
> As promised, I had a few comments on the logic that I wanted to pass along before you dig into why we're seeing the results change. Let me know if you have any questions!

@jaclyn-taroni, I applied the suggested changes you made above and dug into the changes in the results files as discussed. I decided it would most likely be best to represent the changes that I found in a table and discuss them in the last column of the table:

| `Kids_First_Participant_ID` | `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | new `molecular_subtype` | old `molecular_subtype` | status change |
|-----------------|------------|--------------------|--------------------|-----------------|----------------|-------|
| PT_5BWZA0NT |  7316-1102	| BS_BFDEZK1C | BS_MB7WN0ZB | CNS Embryonal, NOS | DMG, H3 K28 |added (on the basis of `pathology_diagnosis == "Supratentorial or Spinal Cord PNET"`) |
| PT_9BZETM0M |  7316-158 | BS_STNH7YSX | BS_5P1TN10Z | CNS Embryonal, NOS | HGG H3 G35 | added (on the basis of `pathology_diagnosis == "Supratentorial or Spinal Cord PNET"`) |
| PT_9BZETM0M |  7316-158 | BS_STNH7YSX | BS_TV5B86ZD | CNS Embryonal, NOS | HGG H3 G35 | added (on the basis of `pathology_diagnosis == "Supratentorial or Spinal Cord PNET"`)|
| PT_E3ADF4ZB |	7316-2664 | BS_C1RRDNF5 | BS_ZVZDDW2G	 | CNS Embryonal, NOS | LGG, BRAF wildtype | added (on the basis of `MN1` fusions)|
| PT_X71D7G5S	| 7316-1105	| BS_P8WN8XEQ |	BS_C6ZZ3FY5 |	CNS Embryonal, NOS | HGG H3 G35 | added (on the basis of `pathology_diagnosis == "Supratentorial or Spinal Cord PNET"`)|
| PT_BDM0M83S |	7316-3072 |	BS_M16CDR44 | BS_CYEY58NY	| NA | CNS Embryonal NOS | deleted (`pathology_diagnosis` does not fulfill the criteria; nor does the fusion data) |
|PT_FZHGKJ0H| 7316-4033 |	BS_NHMPQ8Q1	| BS_JBGZ7HHZ	| NA | CNS Embryonal, NOS | deleted (`pathology_diagnosis` does not fulfill the criteria; nor does the fusion data) |
|PT_FZHGKJ0H| 7316-4034 |	BS_5J5VH3X0	| BS_2SHWPB5P	| NA | CNS Embryonal, NOS | deleted (`pathology_diagnosis` does not fulfill the criteria; nor does the fusion data) |


Does the adjusted logic satisfy your suggestions and does this summary results seem to make sense?

analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,199 @@
...
+title: "Select pathology diagnoses for inclusion"
+output: 
+  html_notebook:
+    toc: TRUE
+    toc_float: TRUE
+author: Chante Bethell for ALSF CCDL
+date: 2020
+---
+
+Code for this notebooke was adapted from the `analyses/molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd`(https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/f65c803706d9c1eea7f2892cb8b638f3d47d15c6/analyses/molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd) notebook.
```
jaclyn-taroni, line 11:
```suggestion
Code for this notebook was adapted from the `analyses/molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd`(https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/f65c803706d9c1eea7f2892cb8b638f3d47d15c6/analyses/molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd) notebook.
```

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -77,14 +91,37 @@ ttyh1_fusions <- fusion_summary_df %>%
...
-fusion_biospecimens <- ttyh1_fusions %>%
+ttyh1_fusion_biospecimens <- ttyh1_fusions %>%
   # add a column that counts how many TTYH1 fusions are present
   mutate(fusion_count = rowSums(select(., starts_with("TTYH1")))) %>%
   # if there's any TTYH1 fusion - pull out the biospecimen ID
   filter(fusion_count > 0) %>%
   pull(Kids_First_Biospecimen_ID)
 ```
 
+### _MN1_ fusions
```
jaclyn-taroni, line 102: @jharenza this was put in place to specifically make sure we captured the EPN specimen you mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/785#issuecomment-695015488, but adding this in nets us an LGG sample:

| `Kids_First_Participant_ID` | `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | new `molecular_subtype` | old `molecular_subtype` | status change |
|-----------------|------------|--------------------|--------------------|-----------------|----------------|-------|
| PT_E3ADF4ZB |	7316-2664 | BS_C1RRDNF5 | BS_ZVZDDW2G	 | CNS Embryonal, NOS | LGG, BRAF wildtype | added (on the basis of `MN1` fusions)|

Per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-698598761 - thoughts? It is `MN1--PATZ1` if that's helpful information.

jaclyn-taroni (commented):
Thanks for the table in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-698598761! In the future, for the samples that are _removed_, if you can state why they were included prior to this release by taking a look at the `pbta-histologies.tsv` file in `release-v16-20200320` (I usually have a local copy hanging around) that can be helpful for tracking down any unexpected results. I took a look myself and those changes are expected.

I had a few comments to get this closer to what was described in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r493646018.

analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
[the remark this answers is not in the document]
jharenza: Sorry, I missed this, but yes to this:

>should we update the step where we include samples (looking at the entire cohort) on the basis of the presence of fusions to include MN1 fusions and TTYH1 fusions (only the latter is addressed currently)?

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
jharenza: Quickly looked this up and found [this paper](https://onlinelibrary.wiley.com/doi/epdf/10.1111/nan.12626) which suggests this is an entity **separate** of the HGNET-MN1 tumors, so I would keep that sample/fusion out of this analysis:

>In conclusion, MN1‐PATZ1 fusions may define a novel histological and immunohistochemical polyphenotypic malignant paediatric brain tumour, not showing epigenetic similarities with CNS HGNET‐MN1, but closely related to PATZ1‐sarcomas.

cbethell:
@jaclyn-taroni, I believe I have now incorporated the changes to the logic as you suggested. I also included logic to explicitly keep the fusion `MN1--PATZ1` fusion excluded as suggested in the [last comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r495212879). 
The results in the [comment with the summary of results table](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-698598761) mostly remained the same with exception of the sample that was added due to that fusion (`PT_E3ADF4ZB`) is no longer added.

Worth noting however is the addition of the following:

| `Kids_First_Participant_ID` | `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | new `molecular_subtype` | old `molecular_subtype` | status change | `broad_histology` |
|-----------------|------------|--------------------|--------------------|-----------------|----------------|-------|--------|
| PT_E3ADF4ZB |	7316-23 | NA | BS_TD1ZEQ7R | CNS Embryonal, NOS | NA (according to histologies file) | added (on the basis of the neuroblastoma logic -- `pathology_diagnosis == Neuroblastoma`, `primary_site == Spinal Cord- Lumbar/Thecal Sac;Spinal Cord- Thoracic` which is not "Other Locations NOS" and `pathology_free_text_diagnosis == neuroblastoma)` which is not "peripheral" | CNS neuroblastoma |

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -17,8 +17,14 @@ To summarize - we will identify biospecimen IDs that should be included for down
 1. An RNA-seq biospecimen includes a _TTYH1_ fusion (5' partner).
    We can consume the files from the [`fusion-summary` module](../fusion-summary/) for this step.
    (See also: [this comment on `AlexsLemonade/OpenPBTA-analysis#401`](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/401#issuecomment-573669727).)
-2. Any sample with `"Embryonal tumor"` in the `broad_histology` column of `pbta-histologies.tsv` that is not labeled `"Medulloblastoma"` or `"Atypical Teratoid Rhabdoid Tumor (ATRT)"` in `pathology_diagnosis` per [this comment on `AlexsLemonade/OpenPBTA-analysis#251`](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/251#issuecomment-568220913).
-   We'll look at the `integrated_diagnosis` column as well.
+   
+2. An RNA-seq biospecimen sample includes a _MN1_ fusion (5' partner) [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/785#issuecomment-695015488).
```
jaclyn-taroni, line 21: This documentation should reflect that we are excluding the MN1--PATZ1 fusion and why.

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -17,8 +17,14 @@ To summarize - we will identify biospecimen IDs that should be included for down
...
+
+3. Any sample with "Supratentorial or Spinal Cord PNET" in the `pathology_diagnosis` column of the metadata `pbta-histologies.tsv` (also [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066)).
+
+4. Any sample with "Neuroblastoma" in the `pathology_diagnosis` column, `primary_site != "Other locations NOS"`, and `pathology_free_text_diagnosis` does not contain "peripheral".
```
jaclyn-taroni, line 25: I would probably take out `primary_site != "Other locations NOS"` here. I would expect the code to perform filtering on the basis of `primary_site != "Other locations NOS"` the way this is written and that's not what the code does.

analyses/molecular-subtyping-embryonal/02-generate-subset-files.R
```diff
@@ -3,8 +3,8 @@
 # This script subsets the files required for subtyping non-MB and non-ATRT
 # embryonal tumors. A sample will be included if either of the following
 # conditions are met: 1) the sample is labeled as an embryonal tumor
-# (broad_histology) but NOT an MB or ATRT tumor (pathology_diagnosis) OR 2)
-# the sample contains a TTYH1 fusion (5' partner)
+# but NOT an MB or ATRT tumor (pathology_diagnosis) OR 2) the sample contains
+# a TTYH1 fusion (5' partner)
```
jaclyn-taroni, line 7: I think if you are not going to point a reader to the `01` notebook here and leave it at that, this needs to be more detailed - it's a lot more complicated than this now!

analyses/molecular-subtyping-embryonal/README.md
```diff
@@ -27,14 +27,25 @@ Files will be regenerated using the symlinked files in `data` by default when ru
 bash run-embryonal-subtyping.sh
 ```
 
+`00-embryonal-select-pathology-dx.Rmd` is not run via this module's shell script, as it should be run locally, tied to `release-v17-20200908`, and should not be re-rendered when there are changes to the underlying `pbta-histologies.tsv` file in future releases (see [Folder content](#folder-content) and [#748](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/748)).
```
jaclyn-taroni, line 30: Not easily commented on in this PR, but you should update line 5 to reflect the release that you used here.

analyses/molecular-subtyping-embryonal/README.md
```diff
@@ -27,14 +27,25 @@ Files will be regenerated using the symlinked files in `data` by default when ru
...
 
 ## Folder Content
 
+[`00-embryonal-select-pathology-dx.Rmd`](https://alexslemonade.github.io/OpenPBTA-analysis/analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.nb.html) is a notebook used to explore the `pathology_diagnosis` and `pathology_free_text_diagnosis` fields in the `release-v17-20200908` version of `pbta-histologies.tsv`. 
+Prior to `release-v17-20200908`, this module used `broad_histology == "Embryonal tumor"` to identify samples to be included for subtyping.
```
jaclyn-taroni, line 35: Maybe mention that ATRT and MB samples were filtered out, too?

analyses/molecular-subtyping-embryonal/README.md
```diff
@@ -27,14 +27,25 @@ Files will be regenerated using the symlinked files in `data` by default when ru
...
+2. An RNA-seq biospecimen sample includes a _MN1_ fusion (5' partner) [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/785#issuecomment-695015488).
+3. Any sample with "Supratentorial or Spinal Cord PNET" in the `pathology_diagnosis` column of the metadata `pbta-histologies.tsv` [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066).
+4. Any sample with "Neuroblastoma" in the `pathology_diagnosis` column, `primary_site != "Other locations NOS"`, and `pathology_free_text_diagnosis` does not contain "peripheral" [per the same comment as above](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066).
+5. Any sample with "Other" in the `pathology_diagnosis` column of the metadata, and with "embryonal tumor with multilayer rosettes, ros (who grade iv)", "embryonal tumor, nos, congenital type", "ependymoblastoma" or "medullooepithelioma" in the `pathology_free_text_diagnosis` column [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066).
```
jaclyn-taroni, line 48: I have the same comments about this documentation as I did about the documentation in `analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd`.

jaclyn-taroni (commented):
Updates look good @cbethell! My remaining comments are related to documentation. It seems like the inclusion of `7316-23` (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700046688) is exactly what we want to capture with our implementation. Tagging @jharenza just to be sure.

jharenza:
> Updates look good @cbethell! My remaining comments are related to documentation. It seems like the inclusion of `7316-23` ([#788 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700046688)) is exactly what we want to capture with our implementation. Tagging @jharenza just to be sure.

jharenza:
> Taking a quick look at the histologies file – I don't think there are instances of `CNS Neuroblastoma` in `pathology_diagnosis`; those are in `broad_histology`, etc. which will be downstream. So I think the move is to _detect_ `neuroblastoma` in `pathology_diagnosis` because there are instances of `Metastatic secondary tumors;Neuroblastoma` in that column and then your other stated steps (`primary_site`, `pathology_free_text_diagnosis`) sound good to me!

yes, you're right!

jaclyn-taroni (approved):
👍  looks good to me

analyses/molecular-subtyping-embryonal/04-table-prep.Rmd
```diff
@@ -507,6 +511,27 @@ all_data_df %>%
...
+### CNS Neuroblastoma
+
+> * CNS Neuroblastoma tumor 
+> * Primary site does not contain "Other locations NOS"
+> * Pathology free text diagnosis does not contain "peripheral"
+
+```{r}
+neuroblastoma_df <- all_data_df %>%
+  filter(
+    pathology_diagnosis == "Neuroblastoma",
```
cbethell, line 523: I also want to point out the difference in logic here in `04-table-prep.Rmd` from the `01` notebook. This logic excludes the `Ganglioneuroblastoma` sample that was included in the `01` notebook [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700717514), which states that `Neuroblastoma` samples that meet this logic should be classified as `CNS Neuroblastoma` (which is being done here), while the `Ganglioneuroblastoma` samples should be designated as `Embryonal NOS` for now due to being characterized histologically without clear molecular lesions.

cbethell (commented):
In reference to a discussion with @jaclyn-taroni, 

I wanted to point out the difference in the Neuroblastoma logic in the `01-samples-to-subset.Rmd` notebook where samples with `Neuroblastoma` and `Ganglioneuroblastoma` (in the `pathology_diagnosis` column) satisfy the logic, versus in the `04-table-prep.Rmd` notebook, where just the `Neuroblastoma` sample satisfies the logic to be classified as `CNS Neuroblastoma` (further explained below).

Does this method satisfy both [the original comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/752#issuecomment-697000066) and the [additional comment on this PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700717514) re classification or should the logic be kept the same across notebooks?

analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.Rmd
```diff
@@ -0,0 +1,197 @@
...
+
+These are the terms that we'll use to look for exact matches in the `pathology_free_text_diagnosis` column.
+
+```{r}
+free_text_dx_terms <- c(
+  "embryonal tumor with multilayer rosettes, ros (who grade iv)",
+  "embryonal tumor, nos, congenital type",
+  "ependymoblastoma",
+  "medulloepithelioma",
+  "medullooepithelioma" # It was noted that this misspelling was included in the `pathology_free_text_diagnosis` column of the histologies file so we will add it to the JSON file [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700717514)
```
jharenza, line 116: @jaclyn-taroni do we want to be strict here to include only what we see now and update as new terms come along or be pre-emptive and also add general terms such as embryonal tumor, PNET, primitive neuroectodermal tumor?

jaclyn-taroni: I think update as new terms come along - that method seems more _intentional_ to me than adding terms we don't see yet.

analyses/molecular-subtyping-embryonal/01-samples-to-subset.Rmd
```diff
@@ -112,40 +151,65 @@ fusion_biospecimens <- c(
...
-Is there anything in the `integrated_diagnosis` column we should be concerned about removing?
+Any sample with "Neuroblastoma" in the `pathology_diagnosis` column, `primary_site != "Other locations NOS"`, and `pathology_free_text_diagnosis` does not contain "peripheral".
 
 ```{r}
-embryonal_df %>%
-  group_by(integrated_diagnosis) %>%
-  count()
+neuroblastoma_df <- embryonal_df %>%
+  filter(
+    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
```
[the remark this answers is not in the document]
jaclyn-taroni: Including the `Ganglioneuroblastoma` samples such that they are designated as `CNS Embryonal, NOS` makes sense to me based on the discussion, but there's nothing in the comments here or the documentation to note that the `Ganglioneuroblastoma` samples are being captured. There are also instances of `Metastatic secondary tumors;Neuroblastoma` in `pathology_diagnosis`, so otherwise we'd have to split based on `;` (this strategy would have applied to LGAT I believe). I'm inclined to add a detailed comment here and update the documentation throughout the module to make the inclusion of `Ganglioneuroblastoma` samples more clear.

jharenza (changes requested):
@cbethell thanks so much for working on this. I made a few edits, but I realized when going through the notebooks that I was actually unclear about `CNS Neuroblastoma`. This is actually one in the same as the `CNS NB-FOXR2` entity and as such, the interrogation of the neuroblastoma samples, along with `FOXR2` expression and fusions should be combined. Sorry about this - definitely my fault for not connecting them in the beginning.

Based on that, it looks like the one sample subtyped as `CNS Neuroblastoma`, `BS_TD1ZEQ7R`, actually does not fit into `CNS NB-FOXR2` based on molecular alterations. As such, I would suggest leaving this sample non-subtyped.

Otherwise, I think this looks good!

cbethell:
@jaclyn-taroni, I took a look at the items we discussed and summarized it as follows:

1. The `CNS NB-FOXR2` samples that get included/subtyped, are included on the basis of `pathology_diagnosis %in% ""Supratentorial or Spinal Cord PNET"` (with exception of the `BS_TD1ZEQ7R` Neuroblastoma sample mentioned in the comments above). Should we change the logic in the `04` notebook for labeling `CNS NB-FOXR2` samples to:

```
all_data_df %>%
  filter(
    FOXR2_fusions != "None" | FOXR2 > 3 |
      pathology_diagnosis == "Neuroblastoma",
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "peripheral",
      negate = TRUE
    )
  ) %>%
  select(contains("ID")) %>%
  mutate(
    molecular_subtype = case_when(
      Kids_First_Participant_ID == "PT_MRTHX2AQ" ~ "NA",
      TRUE ~ "CNS NB-FOXR2"
    )
  )
```

for an output of

| `Kids_First_Participant_ID` | `sample_id` | `pathology_diagnosis` | `pathology_free_text_diagnosis` | FOXR2 | FOXR2_fusions |
|-----------------|------------|--------------------|--------------------|-----------------|----------------|
| PT_NESAQHB1 | 7316-3689 | Supratentorial or Spinal Cord PNET | cns embryonal tumor, nos (who grade iv) | 14.83557 | None |
| PT_NESAQHB1 | 7316-3844 | Supratentorial or Spinal Cord PNET | treated recurrent / residual cns embryonal tumor, nos | 14.30523 | None |
| PT_MRTHX2AQ | 7316-23 | Neuroblastoma | neuroblastoma | -0.1190762 | None |
| PT_9CSZBV2P | 7316-1473 | Supratentorial or Spinal Cord PNET | pnet | 10.18064 | ZC4H2--AL627224.1/FOXR2, ZC4H2--FOXR2 |

In which case the sample `PT_MRTHX2AQ` that we do not want subtyped [per this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#pullrequestreview-502513095) would be labeled as `NA` with a comment pointing to this PR and the other samples would be labeled as `CNS NB-FOXR2`.

**OR** should we stick to the logic for subtyping `CNS NB-FOXR2`: 

```
all_data_df %>%
  filter(FOXR2_fusions != "None" | FOXR2 > 3)
```

for an output of

| `Kids_First_Participant_ID` | `sample_id` | `pathology_diagnosis` | `pathology_free_text_diagnosis` | FOXR2 | FOXR2_fusions |
|-----------------|------------|--------------------|--------------------|-----------------|----------------|
| PT_NESAQHB1 | 7316-3689 | Supratentorial or Spinal Cord PNET | cns embryonal tumor, nos (who grade iv) | 14.83557 | None |
| PT_NESAQHB1 | 7316-3844 | Supratentorial or Spinal Cord PNET | treated recurrent / residual cns embryonal tumor, nos | 14.30523 | None |
| PT_9CSZBV2P | 7316-1473 | Supratentorial or Spinal Cord PNET | pnet | 10.18064 | ZC4H2--AL627224.1/FOXR2, ZC4H2--FOXR2 |

in which case, we can then use the logic we used to capture the CNS Neuroblastoma sample:
```
all_data_df %>%
  filter(
    pathology_diagnosis == "Neuroblastoma",
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "peripheral",
      negate = TRUE
    )
  )
```

or an even simpler logic:
```
all_data_df %>%
  filter(
    pathology_diagnosis == "Neuroblastoma")
```

and label the `BS_TD1ZEQ7R` Neuroblastoma sample `NA` in the `molecular_subtype` column.

**Note** that I went with the *first option* I provided here in the last commit (as to keep the logic for future samples that may meet the criteria and have molecular alterations in agreement with the subtype label).

Is `NA` a good label to use when we want a sample left non-subtyped? Would we prefer to go with the second option I mentioned here or an alternative option?

2. As for excluding Neuroblastoma samples with "metastatic" in the `pathology_diagnosis` or `pathology_free_text_diagnosis` columns per [the last comment on this PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r499948141), I adjusted the logic from:

```
histologies_df %>%
  filter(
    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "peripheral",
      negate = TRUE
    )
  )
```
to

```
histologies_df %>%
  filter(
    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "peripheral",
      negate = TRUE
    ),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "metastatic",
      negate = TRUE
    )
  )
```

noting that this did not change the samples that are captured here.

I also added a section to the `00` notebook at [line 97](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788/files#diff-ad662ddddcd97c91ce984ad69517dd2aR97) exploring the samples with this logic and why we need to exclude `metastatic` on the basis of both `pathology_diagnosis` and `pathology_free_text_diagnosis`. I included this based on conversation with @jaclyn-taroni and to address [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r499950252) but can remove this section if it is not helpful.

I also wanted to mention that I attempted to save "peripheral" and "metastatic" in the JSON file as `pathology_free_text_diagnosis` exclusion terms (as I am not sure what I committed is the best strategy), but the logic did not obey these terms when I attempted:
```
histologies_df %>%
  filter(
    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      path_dx_list$exclude_free_text,
      negate = TRUE
    )
  )
```

3. Lastly, with the current neuroblastoma logic, we capture two ganglioneuroblastoma samples:

| `Kids_First_Participant_ID` | `sample_id` | `pathology_diagnosis` | `pathology_free_text_diagnosis` | `Kids_First_Biospecimen_ID` |
|-----------------|------------|--------------------|--------------------|-----------------|
| PT_RXT6KRR6 | 7316-3147 | Ganglioneuroblastoma | ganglioneuroblastoma | BS_NFRXNAFE |
| PT_RXT6KRR6 | 7316-3147 | Ganglioneuroblastoma | ganglioneuroblastoma | BS_CK7CPM73 |

Per [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-700717514), I left these samples to be classified as `CNS Embryonal, NOS` instead of their own molecular subtype label, which I wanted to confirm is expected here based on @jharenza's explanation in that comment.

I also updated the comments/documentation in the `00` and `01` notebooks, as well as the `README.md` file, to highlight exactly where the logic captures the ganglioneuroblastoma samples per [your comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r499950252) @jaclyn-taroni.

jaclyn-taroni:
@jharenza should we include the `Ganglioneuroblastoma` samples and set `molecular_subtype` to `CNS Embryonal, NOS` in this module? Or would our preference be to exclude them? This is not clear to me based on our discussion and determines how we might move forward.

jaclyn-taroni:
@jharenza another question that impacts implementation at this point - would we want to make sure to exclude samples where `pathology_diagnosis == "Metastatic secondary tumors;Neuroblastoma"`?  In specimens that meet that criterion, `pathology_free_text_diagnosis` is `metastatic secondary tumor, neuroblastoma`. So these would be excluded on the basis of (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#discussion_r499948141)

>  `pathology_free_text_diagnosis` does not contain "metastatic"

It's not clear to me if this is the desired behavior since the `Metastatic secondary tumors;Neuroblastoma` specimens seem like they may have two distinct dx? (My interpretation is that the samples we _definitely want to exclude_ have `pathology_free_text_diagnosis` such as `metastatic neuroblastoma` and `neuroblastoma, metastatic`.)

jharenza:
> Metastatic

I asked Cassie Kline about this one and even though they do have two diagnoses, she suggested we lump those in with metastatic tumors.

jaclyn-taroni:
Thanks for your replies @jharenza! As of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788/commits/028dedbb17271fb269cde0d157da87b6dc60da59, we are using `pathology_diagnosis == "Neuroblastoma"` rather than _detecting_ the string `neuroblastoma` in the `pathology_diagnosis` column. This excludes both `Ganglioneuroblastoma` and `Metastatic secondary tumors;Neuroblastoma` specimens as is desired. 

@cbethell I believe with the following (taken from https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#issuecomment-704638078)

```R
histologies_df %>%
  filter(
    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "peripheral",
      negate = TRUE
    ),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      "metastatic",
      negate = TRUE
    )
  )
```

You are excluding on the basis of the presence of `"peripheral"` **and** `"metastatic"`, when my understanding is that we want to exclude if either are present (**or**). 

I think changing your last code chunk in that same comment to (untested)

```R
histologies_df %>%
  filter(
    str_detect(str_to_lower(pathology_diagnosis), "neuroblastoma"),
    str_detect(str_to_lower(primary_site), "other locations nos", negate = TRUE),
    str_detect(
      str_to_lower(pathology_free_text_diagnosis),
      paste0(path_dx_list$exclude_free_text, collapse = "|"),
      negate = TRUE
    )
  )
```
would have done the trick. I've implemented something similar, but I didn't include it in the JSON file because I consider this a more complex case (the same reason `Neuroblastoma` isn't in the JSON file!). 

Regarding the point (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/788#pullrequestreview-502513095)

> Based on that, it looks like the one sample subtyped as CNS Neuroblastoma, BS_TD1ZEQ7R, actually does not fit into CNS NB-FOXR2 based on molecular alterations. As such, I would suggest leaving this sample non-subtyped.

What I did was perform the CNS NB-FOXR2 subtyping as we did before (one of @cbethell's suggested options) and then in the final step before producing the subtyping table in `04`, I added an exclusion step that can be summarized as: if there were neuroblastoma samples included based on `pathology_diagnosis`, etc. (what's in `01`) but didn't meet the CNS NB-FOXR2 criteria, don't include them in the final table for subtyping.

I'm going to rerequest @jharenza's review. @cbethell I know this is your PR so you can't formally review, but please take a look at what I've done and see if you agree!

jharenza (approved):
This looks good to me now!

analyses/molecular-subtyping-embryonal/04-table-prep.Rmd
```diff
@@ -512,7 +489,30 @@ all_data_df %>%
...
+If neuroblastoma samples were included on the basis of `pathology_diagnosis`, `pathology_free_text_diagnosis`, and `primary_site` (see `01-samples-to-subset`), but do not meet the criteria for _FOXR2_ alterations (CNS NB-FOXR2), those samples should not get a subtype label. 
+
+```{r}
+other_nbl_df <- all_data_df %>%
+  filter(
+    # All other criteria would have been met to be included in all_data_df
+    pathology_diagnosis == "Neuroblastoma",
+    # Exclude samples that *do* meet FOXR2 criteria
+    !(sample_id %in% nb_foxr2_df$sample_id)
+  )
```
cbethell, line 505: Ah yes, I like this logic!

cbethell (commented):
Looks great to me! 🚀

Issue #789, opened 2020-09-22
Updated analysis: fusion filtering


kgaonkar6:
#### What analysis module should be updated and why?
fusion_filtering needs some filtering value updates. 

With the cutoff of 10 we were able to get ~96% sensitivity for PPTC (100 read length ) data.  However, when we tested TCGA data which has (50-75 read length) the cutoff of 10 wasn't sufficient (~70% sensitivity) but at 100 we get about ~96% sensitivity (percentage overlap to what is previously published for these samples)
<img width="633" alt="Screen Shot 2020-09-11 at 2 30 06 PM" src="https://user-images.githubusercontent.com/34580719/92960363-7ddb6700-f43b-11ea-988d-b830cbf52a39.png">

So we believe we should change the cutoff to 100 to capture all read length datasets , tagging @jharenza.

We also want to remove `ConjoinG` from artifact filtering since we believe these are true fusions between neighboring genes forming completely novel proteins in some cases. https://metasystems.riken.jp/conjoing/index

#### What changes need to be made? Please provide enough detail for another participant to make the update.
 - update to `default=100`
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/9e1996ad486aa1cf20bc4fa172b3f99620c0faee/analyses/fusion_filtering/02-fusion-filtering.R#L51 

 - update to `artifact_filter="GTEx_Recurrent|DGD_PARALOGS|Normal|BodyMap"`
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/9e1996ad486aa1cf20bc4fa172b3f99620c0faee/analyses/fusion_filtering/run_fusion_merged.sh#L31

#### What input data should be used? Which data were used in the version being updated?
pbta-fusion-arriba.tsv.gz
pbta-fusion-starfusion.tsv.gz

#### When do you expect the revised analysis will be completed?
1 day

#### Who will complete the updated analysis?
@kgaonkar6

jaclyn-taroni:
Closed via #791

PR #744, opened 2020-08-23
Cleaning fusion_filtering paths

sjspielman:
#### What scientific question is your analysis addressing?

Issue #737 

#### What was your approach?

Update paths in bash script to run in working directory rather than top-level of OpenPBTA project

#### What GitHub issue does your pull request address?

Issue #737 

#### Which areas should receive a particularly close look?

There remain some hard-coded paths in some of the R scripts. Is this worth updating? My gut is to let it be so long as the code runs from its wdir.

#### Is there anything that you want to discuss further?

No

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

N/A

### Results

N/A

PR #791, opened 2020-09-24
Update fusion filtering

kgaonkar6:

 In this PR I'm updating the spanningFragCountFilter to 100 and removing ConjoinG from artifact_filter. 

With the earlier cutoff of 10 we were able to get ~96% sensitivity for PPTC (100 read length ) data.  However, when we tested TCGA data which has (50-75 read length) the cutoff of 10 wasn't sufficient (~70% sensitivity) but at 100 we get about ~96% sensitivity (percentage overlap to what is previously published for these samples) So we believe we should change the cutoff to 100 to capture all read length datasets , tagging @jharenza.

Screen Shot 2020-09-11 at 2 30 06 PM

From additional review of fusions we believe true fusions between neighboring genes forming completely novel proteins in some cases are annotated with [ConjoinG](https://metasystems.riken.jp/conjoing/index). For example FIP1L1--PDGFRA https://ghr.nlm.nih.gov/gene/FIP1L1#conditions

#### What was your approach?
The following 2 filters have been updated in `run_fusion_merged.sh`
artifact_filter="GTEx_Recurrent|DGD_PARALOGS|Normal|BodyMap"
spanningFragCountFilter=100

#789 

I've only updated the run script and README. 02-fusion-filtering.R  has the default value spanningFragCountFilter=10 which I'm overwriting in `run_fusion_merged.sh`. Didn't want to make changes in multiple files in a PR. Let me know if it seems ok to edit the default value as well.

Yes 

tables

#### Documentation Checklist


analyses/fusion_filtering/02-fusion-filtering.R
```diff
@@ -170,6 +170,9 @@ fusion_filtering_QC<-function(standardFusioncalls=standardFusioncalls,readingFra
 # QC filter: artifact and read support
 QCFiltered<-fusion_filtering_QC(standardFusioncalls = standardFusioncalls,junctionReadCountFilter = junctionReadCountFilter,spanningFragCountFilter = spanningFragCountFilter,readingFrameFilter = readingFrameFilter,artifactFilter = artifactFilter,readthroughFilter = readthroughFilter)
 
+saveRDS(QCFiltered,paste0(opt$outputfile,"_QC_filtered.RDS"))
```
jaclyn-taroni, line 173: This add doesn't result in a new file being committed to the repository. Is this leftover from development?

analyses/fusion_filtering/run_fusion_merged.sh
```diff
@@ -28,8 +28,10 @@ standard_arriba_file="${scratch_path}/arriba.tsv"
 standard_starfusion_file="${scratch_path}/starfusion.tsv"
 
 # general filtering parameters
-artifact_filter="GTEx_Recurrent|DGD_PARALOGS|Normal|BodyMap|ConjoinG"
+artifact_filter="GTEx_Recurrent|DGD_PARALOGS|Normal|BodyMap"
 reading_frame_filter="in-frame|frameshift|other"
+spanningFragCountFilter=100
```
jaclyn-taroni, line 33: If you think the default for this project should be `100` (sounds like you do!), it makes sense to me to make the quick 1 line change to `02-fusion-filtering.R` in this PR. You can leave the changes to the shell script, too. That change, even if you change the default,  makes it a little easier to quickly track down that info in the code in my opinion!

analyses/fusion_filtering/results/pbta-fusion-recurrent-fusion-bysample.tsv
```diff
@@ -71,6 +71,7 @@ BS_2ADZNNV4	0	0	1	0	0
 BS_2AR3AP0N	0	0	0	1	0
 BS_2EMW8ZG3	0	0	0	1	0
 BS_2EN3X6HB	0	0	0	1	0
+BS_2HN96129	0	0	0	1	0
```
jaclyn-taroni, line 74: Doing a quick spot check of these changes - this appears to be due to the random selection of an RNA-seq specimen from samples/participants with multiple RNA-seq specimens. I assume that the selected specimens have changed because `analyses/fusion_filtering/results/pbta-fusion-putative-oncogenic.tsv` changed the ordering of the specimens in some way.

jaclyn-taroni (commented):
Thanks for sending this @kgaonkar6! I had one question. I think it's okay to change the default `spanningFragCountFilter` in `02-fusion-filtering.R` as part of this PR.

analyses/fusion_filtering/02-fusion-filtering.R
kgaonkar6: yeah I'm saving this intermediate file in scratch, should I commit that file as well ?

analyses/fusion_filtering/run_fusion_merged.sh
kgaonkar6: ok sounds good, I'll update the default values in `02-fusion-filtering.R`

analyses/fusion_filtering/results/pbta-fusion-recurrent-fusion-bysample.tsv
kgaonkar6: thanks for pointing this out, I'll review the changes in the RNA-seq sample selection from this new update.

analyses/fusion_filtering/02-fusion-filtering.R
jaclyn-taroni: Ah, nope - if it's in `scratch` this makes sense! I forgot the output from this (`opt$outputfile`) is going into scratch.

analyses/fusion_filtering/results/pbta-fusion-recurrent-fusion-bysample.tsv
kgaonkar6: old_sample_selected | new_sample_selected | Kids_First_Participant_ID
---- | ---- | ----
BS_RR12T74P | BS_2HN96129 | PT_DTP4MMRA
BS_5P1TN10Z | BS_TV5B86ZD | PT_9BZETM0M
BS_6VPKXXMR | BS_M85CXHDV | PT_6PP78VNE
BS_7ZV3DPGT | BS_M7QJ8VF0 | PT_QD6KKKJH
BS_5CTVXVRX | BS_8ZS9F31R | PT_DW2XRPC9
BS_8GW1TMV2 | BS_RSRPX3S8 | PT_GR8P8NY5

Since I didn't make changes in `06-recurrent-fusions-per-histology.R` my guess  was maybe order in input file but I can'r recreate the old selection with v16 as well.  I see some of the sample selection changed in #744 as well so not exactly sure how/if we can select the exact same samples each time. 

I looked into set.seed with sample() a bit. 

```
> set.seed(1234)
# sample () selects different in every run if set.seed() is not called exactly before the sample() call
# so it seems to me that the set.seed() at the start of the script doesn't actually apply for all sample() calls in a script?
> sample(LETTERS, 6)
[1] "C" "P" "O" "X" "S" "N"
> sample(LETTERS, 6)
[1] "A" "F" "P" "L" "X" "W"
> sample(LETTERS, 6)
[1] "H" "X" "Z" "T" "G" "F"
```

```
# same selection when set.seed(1234) is run in before each sample() call
> set.seed(1234)
> sample(LETTERS, 6)
[1] "C" "P" "O" "X" "S" "N"
> 
> set.seed(1234)
> sample(LETTERS, 6)
[1] "C" "P" "O" "X" "S" "N"
```

So for these 2 sample() calls in the script I guess I should set.seed() before each of them
https://github.com/kgaonkar6/OpenPBTA-analysis/blob/eacda77df1c372617ba826995e45c6a189dfe39b/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L46-L53 

and 

https://github.com/kgaonkar6/OpenPBTA-analysis/blob/eacda77df1c372617ba826995e45c6a189dfe39b/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L67-L73

jaclyn-taroni: Yea, I would expect the behavior above - if you start a new session, set the seed, and sample `LETTERS` 3x, you'd get the same results for all three sequentially, but they wouldn't be identical to one another.

Can you remind me, at a high level, why do we need to do sampling _twice_?  

An idea that comes to mind for keeping things more consistent is sorting the putative oncogenic fusions file by probably a combination of identifiers before doing the sampling (probably upon reading it in).

kgaonkar6: 1st Sampling is done in histology file for where there is a matching WGS/WXS for RNA-Seq sample ( that is sample id matches in independentSpecimensFile with has both initial and progreesive/relapse etc )

2nd Sampling  is done  in histology file for samples where we only have RNA-Seq  ( but we need to separate this subset into Initial and Progressive/Relapse since we group for Kids_First_Participant_ID )

kgaonkar6: I ordered the input fusion and clinical files and ran the sampling (3 times and compared each time to the previous one) and that I think does fix the different selection of samples in each run issue. 

This way it'll be release specific selection and we will see random selection changes between release .. right?

jaclyn-taroni: If you're ordering the file you are sampling by identifiers before sampling and the biospecimens included don't change (and setting a seed), I would _expect_ that you are guarding against changes between releases. Seems like this won't matter much if #795 gets implemented, so I am inclined to get this merged.

kgaonkar6: Thanks for the clarification. .  I'll work on #795

jaclyn-taroni (approved):
👍  will merge once CI finishes

PR #798, opened 2020-10-09
adding clinical review and updates to pnoc003

kgaonkar6:
#### What scientific question is your analysis addressing?
As part of molecular-subtype-HGG analysis we assign a HGG or DMG subtype from looking for K28M histone variants, in this PR we wanted to identifying samples which have differences between the clinically reviewed subtype and subtype from molecular-subtype-HGG module in PNOC003 and update those in the OpenPBTA subtyping module results

#### What was your approach?

The table from the original [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/751) added by @jharenza is added as file in `input` folder. I used the file to check for changes and updated 3 WXS samples from PT_NK8A49X5, PT_QA9WJ679, PT_WGVEF96B by matching to the values in the reviewed files

#### What GitHub issue does your pull request address?
#751 

#### Which areas should receive a particularly close look?
I've created the file for v17 (which doesn't have subtypes for cell-lines) so I believe that will need to be updated once #797 

#### Is there anything that you want to discuss further?
I've coded the checks specifically to only `substitute` and not `assign` H3 wildtype to DMG, H3 K28 if clinically the subtype was K28M mutant 
https://github.com/kgaonkar6/OpenPBTA-analysis/blob/6e70a6795719007cd0233074b3fff7218d84bbb3/analyses/molecular-subtyping-HGG/10-pnoc003-clinical-review-update.Rmd#L145-L155

I did that because some subtypes have additional subtype like HGG IDH or BRAF V600 that need to be retained

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
table

#### What is your summary of the results?
results/HGG_molecular_subtype-clinical-review-update.tsv

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #797, opened 2020-10-07
adding function to find independent rna samples

kgaonkar6:
#### What scientific question is your analysis addressing?
Independent sample selection for RNA samples as we have for DNA samples. The RNA samples need to match the independent dna samples if matching DNA exists and if not select unique RNA only sample per participant. 

#### What was your approach?
I thought of this as 3 step process to identify 1-1  Biospecimen_ID to Participant_ID for (primary_plus tumore description option which will choose primary if available or other) for each participant in cohort. 

Step 1. Find RNA samples that match the given independent dna sample list
Step 2. Find only-RNA (ie. the participant doesn't have DNA sample so was not added to independent dna sample selection) initial tumor samples
Step 3. Find only-RNA samples for participants which have no initial tumor RNA sample 

Then *sample()* Kids_First_Biospecimen_ID`  grouped by `Kids_First_Participant_ID` as is done in DNA independent sample selection function.

#### What GitHub issue does your pull request address?
#795 

#### Which areas should receive a particularly close look?
- The steps for the unique sample selection  for each options provided
- should I add other options/filters

#### Is there anything that you want to discuss further?

 - Should we run this separately for polya and stranded (like I'm running below for polya only samples) or prefer 1 over the other?

```
independents <- histology_df %>%
   filter(sample_type == "Tumor", RNA_library=="poly-A",
          composition == "Solid Tissue"
          ) %>%
   independent_rna_samples(independent_dna_sample_df = independent_dna_sample_df,histology_df = .,match_type = "independent_dna_plus_only_rna",tumor_description_rna_only = "primary",seed = 123)
```

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
No

### Results
NA
#### What types of results are included (e.g., table, figure)?
NA

#### What is your summary of the results?
NA

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

Issue #751, opened 2020-08-28
Updated analysis: pathology subtyping to include PNOC003 histone mutations from report data


jharenza:
#### What analysis module should be updated and why?
[pathology subtyping
](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-pathology)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
We will add histone mutations:  **H3F3A K28M, HIST1H3B K28M, HIST1H3C K28M, HIST2H3C K28M, H3F3A.G35R** from pathology (or clinical?) reports of PNOC003 trial patient samples. In some cases, as noted in #735, subtyping was different because of mutations missed during consensus calling. However, this information should be available in clinical and/or pathology reports, so we can add.

#### What input data should be used? Which data were used in the version being updated?
This particular subtyping is not currently a part of pathology subtyping, so it would be added based on a doubly-reviewed input file or table provided by someone at D3b.

**UPDATE 2020-10-05** 
Below is a double-reviewed (by Cassie Kline and myself) table of histone mutations found in the specified patient tumor per the TGEN genomic reports. Technically, these do not come from pathology, but from the clinical trial reports, so @jaclyn-taroni, we can either keep these in `pathology-subtyping` and broaden the module to something like `clinician-subtyping-review` or create a new module for this specific subtyping based on clinical trial reporting. Thoughts?

Kids_First_Participant_ID | H3 Status | tumor_descriptor
-- | -- | --
PT_M23Q0DC3 | H3F3A K28M | Initial CNS Tumor
PT_9GKVQ9QS | H3F3A K28M | Initial CNS Tumor
PT_KBFM551M | H3 WT | Initial CNS Tumor
PT_KBFM551M | no report | Progressive   Disease Post-Mortem
PT_V1HNAC2Q | H3F3A K28M | Initial CNS Tumor
PT_NK8A49X5 | H3F3A K28M | Initial CNS Tumor
PT_NK8A49X5 | H3F3A K28M | Progressive
PT_KZ56XHJT | H3F3A K28M | Initial CNS Tumor
PT_KZ56XHJT | H3F3A K28M | Progressive
PT_KZ56XHJT | no report | Progressive   Disease Post-Mortem
PT_QA9WJ679 | H3F3A K28M | Initial CNS Tumor
PT_WGVEF96B | H3F3A K28M | Initial CNS Tumor
[...]

PR #800, opened 2020-10-12
Add pnoc003 clinical review

kgaonkar6:
As part of molecular-subtype-HGG analysis we assign a HGG or DMG subtype from looking for K28M histone variants, in this PR we wanted to identifying samples which have differences between the clinically reviewed subtype and subtype from molecular-subtype-HGG module in PNOC003 and update those to the clinical reviewed subtypes.

#### What was your approach?

The table from the original [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/751) added by @jharenza is added as file in `input` folder. I used the file to check for changes and updated 3 WXS and 3 RNA-Seq samples from PT_NK8A49X5, PT_QA9WJ679, PT_WGVEF96B by matching to the values in the reviewed files.

As discussed in #798 , I'm moving the code from `molecular-subtyping-HGG` to `molecular-subtyping-pathology` 
- as a separate notebook before [incorporating path feedback](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology/02-incorporate-pathology-feedback.Rmd) 

This order was decided after discussing with @jharenza, since the pathology feedbacks are major changes to the diagnosis , so once the clinical feedbacks are incorporated  (as step 2) the subtype/diagnosis would need additional review from pathology as a final step in this module ( as step 3).

#751 

I've created the file for v17 (which doesn't have subtypes for cell-lines) so I believe that will need to be updated once #797 

I've coded the checks specifically to only `substitute` and not `assign` H3 wildtype to DMG, H3 K28 if clinically the subtype was K28M mutant 
https://github.com/kgaonkar6/OpenPBTA-analysis/blob/6e70a6795719007cd0233074b3fff7218d84bbb3/analyses/molecular-subtyping-HGG/10-pnoc003-clinical-review-update.Rmd#L145-L155

I did that because some subtypes have additional subtype like HGG IDH or BRAF V600 that need to be retained

Yes

table

results/compiled_molecular_subtypes_with_pathology_feedback_clinical_review.tsv

#### Documentation Checklist


analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
```diff
@@ -0,0 +1,200 @@
...
+```{r}
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+data_dir <- file.path(root_dir, "data")
+results_dir <- "results"
+```
+
+
+#### Input
+
+When we run this locally, we want to tie it to a specific version of the histologies file _prior_ to reviewed clinical subtypes for PNOC003 samples and pathology feedback [#751](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/751)
```
jaclyn-taroni, line 89: Hmmmm... _do_ we want to tie this to a specific release? 🤔  I believe the rationale for tying `analyses/molecular-subtyping-pathology/01-compile-subtyping-results.Rmd` to a specific release was to preserve the `integrated_diagnosis`, `broad_histology`, and `short_histology` fields before they got updated. But now, all of the subtyping uses `pathology_diagnosis` and `pathology_free_text_diagnosis` fields and `integrated_diagnosis`, `broad_histology`, and `short_histology` will be downstream of subtyping (#748). Given that we're going to redo the `01` notebook (#784), I expect tying that to v15 very well will come out.

What was your rationale for tying this to v15? Was it something besides consistency with the rest of the module?

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
```diff
@@ -0,0 +1,200 @@
...
+pnoc003_review <- readr::read_tsv(file.path("input","pnoc003-clinical-review.tsv"))
+  
+# subtyping from molecular-subtype-HGG
+subtype <- readr::read_tsv(file.path(results_dir,
+                         "compiled_molecular_subtypes.tsv")) %>%  
+  # ids to match
+  left_join(clinical[,c("Kids_First_Participant_ID",
+                        "Kids_First_Biospecimen_ID",
+                        # need tumor_descriptor to match to clinical review
+                        "tumor_descriptor")],by=c("Kids_First_Biospecimen_ID",
```
jaclyn-taroni, line 117: I think this is the only spot where we use the `pbta-histologies.tsv` file, is that correct? If so, I don't think we need to tie this to a specific release.

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
```diff
@@ -0,0 +1,200 @@
...
+    # checking for samples where clinically reviewed subtype is H3 K28M
+    # but molecular subtype is not DMG H3 K28
+    (grepl("K28M",.$`H3 Status`) & !grepl("DMG, H3 K28",molecular_subtype)) | 
+    # checking for samples were clinical review was not found  
+    grepl("no report",.$`H3 Status`) )  %>%
+  dplyr::select(Kids_First_Participant_ID,
+                sample_id,
+                `H3 Status`,
+                molecular_subtype,
+                tumor_descriptor)
```
jaclyn-taroni, line 147: It looks like there are essentially duplicates in `diff_subtype`. I expect this is because `compiled_molecular_subtypes.tsv` is laid out with one biospecimen per row and you're joining by participant identifier and `tumor_descriptor`. So I'm thinking it could be useful to include a `dplyr::distinct()` step somewhere in here? Maybe just for the display below.

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
```diff
@@ -0,0 +1,200 @@
...
+    molecular_subtype =  case_when(
+      (grepl("H3 WT",.$`H3 Status`) & grepl("HGG",molecular_subtype)) |
+        (grepl("K28M",.$`H3 Status`) & grepl("DMG, H3 K28",molecular_subtype)) 
+      ~ molecular_subtype,
+      # Don't think we would need to ever change from K28 to H3 wildtype
+      # since there would be evidence for K28 from snv data for the sample
+      #(grepl("H3 WT",.$`H3 Status`) & !grepl("HGG",molecular_subtype)) ~ 
+      #  gsub("DMG, H3 K28","HGG, H3 wildtype",molecular_subtype),
+      (grepl("K28M",.$`H3 Status`) & !grepl("DMG, H3 K28",molecular_subtype)) ~ 
+        gsub("HGG, H3 wildtype","DMG, H3 K28",molecular_subtype),
```
jaclyn-taroni, line 191: As I'm looking at this, I am wondering if, for the purposes of this notebook, we might want to standardize the labels between the `H3 Status` and `molecular_subtype` so you can essentially ask "are these the same?" That logic might be more straightforward. Thoughts? We'd probably also want to filter out the no reports - I think this is essence what this is accomplishing as well.

jaclyn-taroni: I suppose this extends to the `Notes` step, too.

jaclyn-taroni (commented):
Thanks for moving this to this module @kgaonkar6 ! I think the functionality is where it needs to be. I had some questions about tying this to a specific release and I also had a thought about how to make some of the logic more readable.

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
kgaonkar6: just wanted to keep it consistent in the module, forgot to mention this in the PR

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
kgaonkar6: I thought about changing the values in H3 Status to what we have in molecular_subtype, but since we have some additional values  from IDH/BRAF alterations I assumed that it would be the same logic of `grepl` because of the IDH/BRAF alterations subtyping that is not part of clinical review. For example: 

Kids_First_Participant_ID | sample_id | Kids_First_Biospecimen_ID | molecular_subtype | integrated_diagnosis | short_histology | broad_histology | Notes | tumor_descriptor
-- | -- | -- | -- | -- | -- | -- | -- | --
PT_0MXPTTM3 | 7316-3220 | BS_8SYN7GXG | HGG, IDH | High-grade glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor
  | PT_0MXPTTM3 | A09410 | BS_R7NTZR4C | HGG, IDH | High-grade glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor
  | PT_0MXPTTM3 | A09410 | BS_H3TYF6GD | HGG, IDH | High-grade glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor
PT_HGM20MW7 | 7316-3221 | BS_FK3B5SDH | DMG, H3 K28, BRAF V600E | Diffuse midline glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor
  | PT_HGM20MW7 | A09446 | BS_5VPM0F36 | DMG, H3 K28, BRAF V600E | Diffuse midline glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor
  | PT_HGM20MW7 | A09446 | BS_52ETE050 | DMG, H3 K28, BRAF V600E | Diffuse midline glioma | HGAT | Diffuse astrocytic and oligodendroglial tumor | Updated via OpenPBTA subtyping | Initial CNS Tumor


Maybe these IDH, BRAF alterations should also be clinical reviewed so we can match the values?

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
jaclyn-taroni: Gotcha, I think it's okay to take it out at this point. That will also allow you to simplify things (always good!) since there won't be a need to distinguish between running in CI and running locally.

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
```diff
@@ -0,0 +1,205 @@
...
+    # but molecular subtype is not DMG H3 K28
+    (grepl("K28M",.$`H3 Status`) & !grepl("DMG, H3 K28",molecular_subtype)) | 
+    # checking for samples were clinical review was not found  
+    grepl("no report",.$`H3 Status`) )  %>%
+  dplyr::select(Kids_First_Participant_ID,
+                sample_id,
+                `H3 Status`,
+                molecular_subtype,
+                tumor_descriptor) %>%
+  distinct()
```
jaclyn-taroni, line 152: I don't see `dplyr` loaded anywhere, so I expect this needs to be 
```suggestion
  dplyr::distinct()
```

jaclyn-taroni: Nope I'm wrong - line 105.

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
jaclyn-taroni: I see - so in the `PT_0MXPTTM3` case, H3 status is wildtype which agrees with the molecular subtyping status and in the `PT_HGM20MW7` H3 status and the subtyping is in agreement, it's just the matter of BRAF V600E also being included.

Taking a step back - my understanding of this addition is that we're essentially looking for any K28M mutations that were _missed_ because they were not detected in the SNV consensus for whatever reason (from #751 and some additional context in #735). I had some ideas about how we might approach this differently for clarity reasons, but I think we can go with this solution and have you add comments throughout at each condition within calls to `case_when()` (lines 180-197).

analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd
kgaonkar6: thanks for the review, will remove the specific release chunk

kgaonkar6: hmm there seem to be some updates to tumor descriptor values from v15 to v17, that should be ok?

jaclyn-taroni: Ah, yes, that should be okay. I think we can think of the v17 tumor descriptor values as more correct!

jaclyn-taroni (approved):
This looks good to me! Do we need @jharenza to take a look before merging?

jharenza (approved):
Looks good, thanks @kgaonkar6 !

PR #802, opened 2020-10-13
De novo mutational signatures: rename and rerun known signatures notebook

jaclyn-taroni:

I am adapting de novo mutational signatures code (see #636, #678, and #799). We have a notebook in the existing `mutational-signatures` module called `mutational_signatures`. In this notebook I am renaming it to 1. reflect numbering because there will be other scripts/notebooks/etc. added and 2. to reflect that this is working with known signatures, rather than de novo signatures.

I reran this notebook, so we have some new/updated plots. There are some short histology values that have changed since the last time this was run, which we can see in the TSV output as well (no change in values as far as I can tell).

Not particularly, my intention is to get this out of the way and start cranking on the de novo signatures code 🏃  Documentation will be out of date for a bit, as I am planning to take a pass once a significant chunk of the de novo code has gone in.

#### Documentation Checklist


cansavvy (approved):
This looks good to me!

Issue #803, opened 2020-10-14
Updated analysis: MB subtyping


jharenza:
#### What analysis module should be updated and why?
Minor update to medulloblastoma subtyping to enable inference of `pathology_diagnosis` from `molecular_subtype` for #748 

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Update output from:
>Group3
Group4
SHH
WNT

to:
>MB, Group3
MB, Group4
MB, SHH
MB, WNT

#### What input data should be used? Which data were used in the version being updated?
no additional data required

#### When do you expect the revised analysis will be completed?
1-2 days

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
closed with #809

PR #809, opened 2020-10-14
adding MB to subtype for consistency

kgaonkar6:
To make subtyping values consistent with other broad histology subtyping value example HGG wildtype, DMG K28M etc we are adding "MB" to the subtyping values 

#### What was your approach?
just add "MB" to the subtyping value

#803 

There is a lot of red and green from the value change but just eye-balling it looks like only "MB, " is added to each value if not NA.
Please let me know if something changed drastically which I might have missed.

Table

"MB, " added to all subtypes, the actual subtyping should remain same.

#### Documentation Checklist


jaclyn-taroni (approved):
👍 satisfies #803 ✅

komalsrathi (approved):

jharenza (approved):

PR #811, opened 2020-10-15
De novo mutational signatures: Only extract using WGS samples

jaclyn-taroni:

I was looking at the [`sigfit` vignette](https://htmlpreview.github.io/?https://github.com/kgori/sigfit/blob/master/doc/sigfit_vignette.html) and at [the _Converting between genome- and exome-relative representations of signatures_ section](https://htmlpreview.github.io/?https://github.com/kgori/sigfit/blob/master/doc/sigfit_vignette.html#converting-between-genome--and-exome-relative-representations-of-signatures) in particular:

> In some analyses, signatures that have been inferred from whole-genome mutation data need to be fitted to mutation data from whole-exome sequencing samples. For instance, if the cohort of samples includes both whole genomes and exomes, it is normally advisable to extract signatures from the whole-genome samples (provided that there are enough of them) and re-fit the resulting signatures to the whole-exome samples

A key piece being "provided there are enough of them." In our dataset, the vast majority of specimens are assayed with WGS (only 31 are WXS). The vignette goes on to say:

>  (An alternative approach would be to provide a matrix of sample-specific mutational opportunities that reflects the origin of each catalogue, as explained in the previous section.)

It's not clear to me that `deconstructSigs::mut.to.sigs.input()` approach we use accomplishes that. So I thought I would separate out the WGS biospecimens and plan to use them for the de novo signature extraction. Inituitively, I might expect this has very little impact on extraction given the low number of WXS samples, but may improve some downstream steps?

#### What was your approach?

This pull request

* Adds an Rscript that splits up the consensus "MAF-like" file into WGS and WXS files, which are written to scratch (`analyses/mutational-signatures/02-split_experimental_strategy.R`). All of this is hardcoded in the Rscript, since I expect this to be _highly specific_ to this project and its structure. 
* Moves the steps for running `extract_signatures()` with a range of `nsignatures` values into its own numbered shell script 
(`analyses/mutational-signatures/03-de_novo_range_of_nsignatures.sh`).
* Revises the main module shell script to reflect the above changes.

#636 - building off of #799, which is an adaptation of #678.

Bash tips always welcome. The documentation for this module is out of date and will remain out of date until more of the de novo signatures code goes through.

#### Documentation Checklist


analyses/mutational-signatures/run_mutational_signatures.sh
```diff
@@ -12,35 +12,12 @@ cd "$(dirname "${BASH_SOURCE[0]}")"
...
-  --maf_file "${data_dir}/pbta-snv-consensus-mutation.maf.tsv.gz" \
-  --nsignatures_floor $FLOOR \
-  --nsignatures_ceiling $CEILING \
-  --num_iterations $NUM_ITER \
-  --seed 42 \
-  --output_file "results/denovo_sigfit_signatures.RDS" \
-  --plot_output "${denovo_plot_dir}/denovo_sigfit_${FLOOR}_to_${CEILING}_gof.pdf"
+# Run the shell script that is for determining the number of signatures to use
+# with a low number of iterations
+QUICK_MUTSIGS=$ABBREVIATED_MUTSIGS bash 03-de_novo_range_of_nsignatures.sh
```
jashapiro, line 23: I guess my question here is why do this step in a separate bash script rather than just calling de_novo_signature_extraction.R  as part of this one? I don't have a strong opinion, but I am not quite sure of the modularity advantage here, weighed against the indirectness. 

Phrased differently, why does `de_novo_signature_extraction.R` have so many required options that have to be set at the command line, while `02-split_experimental_strategy.R` uses default write locations, etc.?

analyses/mutational-signatures/02-split_experimental_strategy.R
```diff
@@ -0,0 +1,67 @@
...
+maf_file <- file.path(data_dir, "pbta-snv-consensus-mutation.maf.tsv.gz")
+maf_df <- data.table::fread(maf_file, data.table = FALSE) %>%
+  dplyr::select(Tumor_Sample_Barcode,
+                Chromosome,
+                Start_Position,
+                Reference_Allele,
+                Allele)
+
+# Read in the metadata file, we only need the biospecimen ID which is what is
+# used in the
```
jashapiro, line 44: used in the what?

jashapiro (approved):
My only comment is about how you heard we love bash so you put bash in our bash.

Other than that (and an incomplete comment), LGTM!

analyses/mutational-signatures/run_mutational_signatures.sh
jaclyn-taroni: There is nothing specific about `de_novo_signature_extraction.R` to this project, in my opinion. (All the column names used are standard MAF stuff as far as I know.) Having to join the `pbta-histologies.tsv` file to the `Tumor_Sample_Barcodes` and using `WXS` to refer to whole exome sequencing - these all seem pretty specific to the project to me. I could make splitting into experimental strategies step more general but decided the development time cost outweighed the benefits. I made the opposite decision for `de_novo_signature_extraction.R` because I thought it had more potential for future use and because we will need to run that at least one more time in the context of this module with different parameters (probably more than that with different seeds...). 

My reasoning for the move to numbering a bash script is almost certainly unclear from this PR, because it's future-looking in a way (and I didn't say anything about why when filing 😄 ).

In the absence of moving the step where we are looking at a range of values for `nsignatures` to its own numbered script, the module structure would probably will have been something like:

```
01 look at known signatures
02 split by strategy
03 pick the number of signatures
04 match to known signatures
05 bubble plot, etc.
run_mutational_signatures.sh
```

Where between `02` and `03` we'd run the range of `nsignatures` values and between `03` and `04` we'd run signature extraction with a set number of signatures and a high number of iterations.

I thought moving to a separate shell script would be more clear from that view than having those steps in `run_mutational_signatures.sh`. On the other hand, there will probably be a step in `run_mutational_signatures.sh` that runs `de_novo_signature_extraction.R` with the selected number of signatures for a high number of iterations. Based on discussion on #799, though, I think we might need to do that with multiple values for the `seed` argument which might be a rationale for moving it into it's own script, too. 

That was a long winded way to say that the own bash script decision is not set in stone and I feel more strongly about the design decisions I made with the R scripts given the information I have.

analyses/mutational-signatures/02-split_experimental_strategy.R
jaclyn-taroni: Thanks for catching this!

Issue #804, opened 2020-10-14
Updated analysis: EWS molecular subtyping


jharenza:
#### What analysis module should be updated and why?
[molecular-subtyping-EWS](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-EWS)

Currently, this module currently recodes `integrated_diagnosis` and `short_histology`, but I think we should update this module in preparation for #748.
 
#### What changes need to be made? Please provide enough detail for another participant to make the update.
1. Pull all samples with hallmark EWS fusions
2. Match RNA `sample_id` with DNA `sample_id`
3. Add `EWS` to `molecular_subtype` for the specimens in 2. --> this will enable updating of `integrated_diagnosis`, `broad_histology`, and `short_histology` with #748 

We can remove the `CNS EFT-CIC` bit of code since this deals with recoding `integrated_diagnosis`.

#### What input data should be used? Which data were used in the version being updated?
`fusion_summary_ewings_foi.tsv`
`pbta-histologies.tsv`

#### When do you expect the revised analysis will be completed?
2-3 days

#### Who will complete the updated analysis?
@kgaonkar6

jaclyn-taroni:
Closed via #815

PR #815, opened 2020-10-16
adding molecular_subtype for EWS

kgaonkar6:
Currently, this module currently recodes integrated_diagnosis and short_histology, in this PR we will only update the molecular_subtype for EWS samples so it's similar to other subtyping

#### What was your approach?
We used sample_id from RNA-Seq samples with EWSR1 hallmark fusion and match to WGS Tumor samples to update the molecular_subtype for each sample_id as "EWS"

#804 

There are 3 sample_ids that are Ewings Sarcoma in pathology_diagnosis which and were also part of the original results file to reclassify integrated_diagnosis, short and broad histology.
Should we have them as "EWS,To be classified" ?

I `git mv` the old 01 script to rename that to 01-run-subtyping-ewings.Rmd ( was 01-reclassify_as_ewings.Rmd before ) and also renamed the run script which I've updated in CI config.yml and module's bash script. Is there something I missed where these changes will be affected?

Yes

tables

results/EWS_results.tsv

#### Documentation Checklist


kgaonkar6:
hmm the 'integrated_diagnosis_reclassified' column from the resulting file from this module is needed in 01-compile-subtyping-results.Rmd in `molecular_subtyping_pathology` module. Should I change  01-compile-subtyping-results.Rmd in this PR?

jaclyn-taroni:
> hmm the 'integrated_diagnosis_reclassified' column from the resulting file from this module is needed in 01-compile-subtyping-results.Rmd in `molecular_subtyping_pathology` module. Should I change 01-compile-subtyping-results.Rmd in this PR?

I expect that notebook to be drastically different in the future (#784). So my general advice having not looked at this code yet would be to make the decision that makes the most sense here/with the information you have because `01-compile-subtyping-results.Rmd` is subject to change. You can comment out that module from CI so long as you make a note that you've done that either on #784 or in a new issue.

kgaonkar6:
Ok sounds good, I'll comment of #784 and comment out that module (molecular-subtyping-pathology) from CI

jharenza (approved):
Thanks @kgaonkar6! This looks good. Regarding below:
>Should we have them as "EWS,To be classified" ?

I think yes for now! Thanks!

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
+---
+title: "Add molecular subtype for samples that have hallmark Ewings Sarcoma fusions"
+output: html_notebook
+author: K S Gaonkar for D3b
+date: January 2020
+---
+
+Identify sample IDs with hallmark _EWSR1_ fusions and subtype as `EWS`
```
cansavvy, line 8: Can you add a bit more specifics in this preview here? Things like, what is your input and what is your intended output.

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+
+
+#### Save to file
+
+```{r}
+ews_subtype %>%
+  readr::write_tsv(file.path(results_dir, "EWS_results.tsv"))
+```
+
+### List which samples are molecular_subtype == "EWS"
```
cansavvy, line 102: You may want to do these previews before you write to file. In other words, if something in these previews informs us about our resulting output we might want to know that before we write to file.

analyses/molecular-subtyping-EWS/results/EWS_results.tsv
```diff
@@ -1,13 +1,10 @@
...
-PT_J50NZYZF	BS_PSNBFWQV	BS_J52D4XQY	7316-128	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
-PT_J50NZYZF	BS_B9V8RGTA	BS_MWZCP1XW	7316-956	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
-PT_NBSRQAMS	BS_R4Y180DM	BS_Q9NHTR7G	7316-1079	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
-PT_PJFTMJ6H	BS_WJGFDRWK	BS_SSYNTXDC	7316-2150	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
-PT_SA94083W	BS_8S86CM61	NA	7316-1856	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	NA
-PT_VHS5RM2K	BS_0448A413	BS_FNM4BB83	7316-364	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
-PT_X8S0H6TG	BS_1N9XZXGE	BS_ZECKZTFC	7316-876	Ewings Sarcoma	EWS	Mesenchymal non-meningothelial tumor	Reclassified due to presence of hallmark EWS fusions
+Kids_First_Participant_ID	Kids_First_Biospecimen_ID_RNA	Kids_First_Biospecimen_ID_DNA	sample_id	molecular_subtype
```
cansavvy, line 1: Is this expected that we are no longer following the `Mesenchymal non-meningothelial tumor` labeled biospecimens? 

Also just checking, we no long need those other columns that have been dropped with the new results?

.circleci/config.yml
```diff
@@ -61,12 +61,13 @@ jobs:
...
 
       - run:
-         name: Molecular Subtyping / Reclassification - EWS
-         command: ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-EWS/run_reclassification.sh
+         name: Molecular Subtyping - EWS
+         command: ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-EWS/run_subtyping.sh
 
-      - run:
-         name: Molecular Subtyping - Compile and incorporate pathology feedback
-         command: OPENPBTA_TESTING=1 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-pathology/run-subtyping-aggregation.sh
```
cansavvy, line 69: I don't know enough about what output from here `run-subtyping-aggregation.sh`. Just wanna double check that the new output in `EWS_results.tsv` still has all the necessary components that `run-subtyping-aggregation.sh` will ultimately need?

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
cansavvy: Also maybe a sentence about your strategy.

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+```
+
+Are there samples that where `pathology_diagnosis` is Ewings Sarcoma and they are not in list above?
+
+```{r}
+
+readr::read_tsv(file.path(dataDir, "pbta-histologies.tsv"),
+                            guess_max = 10000) %>%
+  dplyr::filter(str_detect(pathology_diagnosis,"Ewings Sarcoma"),
+                !sample_id %in% ews_subtype$sample_id)
```
cansavvy, line 116: If I'm interpreting your test correctly, it looks like there are 5 samples that are `pathology_diagnosis` that are `"Ewings Sarcoma"`? Is this a concern (I do not know, so just want to check).

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+
+# Get RNAseq ids because we have fusion information for RNA-seq samples
+ews_rnaseq <- clinical %>%
+  filter(Kids_First_Biospecimen_ID %in% ews_rnaseq_bsids,
+         sample_type == "Tumor") %>%
+  select(Kids_First_Biospecimen_ID, 
+         Kids_First_Participant_ID, 
+         sample_id) 
+
+# we're going to join the WGS samples to RNA-seq with sample_id
```
cansavvy, line 69:
```suggestion
# we're going to filter to the WGS samples to RNA-seq with sample_id
```

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+ews_subtype <- ews_rnaseq %>% 
+  full_join(ews_wgs,
+             by = c("Kids_First_Participant_ID",
+                    "sample_id"),
+             suffix = c("_RNA", "_DNA")) %>%
+  select(Kids_First_Participant_ID,
+         Kids_First_Biospecimen_ID_RNA,
+         Kids_First_Biospecimen_ID_DNA,
+         sample_id,
+         everything()) %>%
```
cansavvy, line 90: Unless this changes, we don't need this `everything()` bit because I think you only have the four columns already named at this point.

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+```{r}
+# get a vector of biospecimen IDs for those biospecimens that should be subtyped
+have_ews_fusion <- rowSums(ewings_foi_bsids[, -1]) > 0
+
+# get rna BS_IDs which have hallmark fusion 
+ews_rnaseq_bsids <- ewings_foi_bsids %>%
+  filter(have_ews_fusion) %>%
+  pull(Kids_First_Biospecimen_ID)
+
+# Get RNAseq ids because we have fusion information for RNA-seq samples
```
cansavvy, line 61: This comment is a bit unclear; can you reword it? Looks like you are filtering to rna seq biospecimen ids for EWS fused biospecimens only.

cansavvy (commented):
I think this looks good. But because I am unfamiliar with this analysis and the some of the downstream analyses' needs I have some questions about the changes as well as some requests for further clarity in some of the comments and documentation.

But from what I can see so far, it looks like this is fulfilling the intents of #804 , just want to make sure so I've posted some questions.

analyses/molecular-subtyping-EWS/results/EWS_results.tsv
kgaonkar6: Thanks for brining this up , I'm going to reorganize the code to add back these samples as "EWS, To be classified" since in the current state it's only subtyping the samples as EWS if fusions exists and not keeping samples that are "Ewings Sarcoma" in pathology_diagnosis.

.circleci/config.yml
kgaonkar6: Yeah I think it does have all columns mainly BS IDs with matching molecular_subtype. Earlier we were "reclassifying" the samples but now as per #804  we will subtype only and so I've tried to keep the output similar to molecular-subtyping-HGG module

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
kgaonkar6: Yes, they need to be added back..  these will be added as "EWS, To be classified" in the code update shortly.

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+ews_rnaseq <- clinical %>%
+  filter(Kids_First_Biospecimen_ID %in% ews_rnaseq_bsids,
+         sample_type == "Tumor") %>%
```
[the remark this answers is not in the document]
kgaonkar6: I'll remove that since RNA-Seq are always "Tumor" we don't have "non-Tumor" samples for EWS I believe, 

I had that added since we also have "Normal" WGS samples but for RNA-Seq I don't need it, thanks!

analyses/molecular-subtyping-EWS/01-run-subtyping-ewings.Rmd
```diff
@@ -0,0 +1,120 @@
...
+# get subset folder
+results_dir <- "results"
+
+# create if doesn't exist
+if (!dir.exists(results_dir)) {
+  dir.create(results_dir)
+}
+```
+
+### Add "EWS" to molecular_subtype for samples that have hallmark ewings sarcoma EWSR1 fusion in RNA and match to DNA per sample_id
```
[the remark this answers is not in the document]
kgaonkar6: sounds good, I'll break it into smaller chunks

cansavvy (approved):
Okay. I think this looks like you've addressed most of my comments and it meets the goals of #804 as far as I can tell. :+1:

Issue #812, opened 2020-10-15
Updated analysis: Fusion Filtering - add reciprocal and kinase domain


jharenza:
#### What analysis module should be updated and why?
[fusion_filtering](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering)
With #790, we will be adding subtypes based on presence of fusions other than _KIAA1549--BRAF_, however, we would like to ensure that they retain the kinase domain and if the kinase is the 5' gene, that a reciprocal fusion exists. This functionality is in [annoFuse](https://github.com/d3b-center/annoFuse), but not yet in OpenPBTA

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Add functionality to assess kinase domain retention and reciprocality and add column for kinase domain retained and add column for is reciprocal to putative oncogenic file.

#### What input data should be used? Which data were used in the version being updated?
V17

#### When do you expect the revised analysis will be completed?
1-1.5 weeks

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
closed with #816 and #821

PR #816, opened 2020-10-19
adding annoFuse for kinase status

kgaonkar6:
Adding annoFuse to docker to get pfam domain annotation and kinase domain status

#### What was your approach?
Added 
RUN R -e "BiocManager::install(c('EnsDb.Hsapiens.v86', 'ensembldb'))"
RUN R -e "remotes::install_github('d3b-center/annoFuse', dependencies = TRUE)"

#812 

Adding annoFuse will make adding kinase domain status 1 line command by running something similar to the example in the function here: 
https://github.com/d3b-center/annoFuse/blob/master/R/fusion_driver.R

So do we want to use this function from annoFuse or replicate the internal functions *fusion_driver()* and *get_Pfam_domain()* in OpenPBTA? 
Asking mainly for 
 - some aggregating steps will need to be updated to keep `LeftBreakpoint` and `RightBreakpoint` columns so domain retention status can be added with respect to where the breakpoints are within the genes
 - consistency, since the other scripts in fusion_filtering are base functions and not functions called from annoFuse?

NA

No

NA

NA

#### Documentation Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -334,6 +334,7 @@ RUN R -e "remotes::install_github('kgori/sigfit', ref = '209776ee1d2193ad4b682b2
 
 # Package for kinase domain retention for fusions
 RUN R -e "BiocManager::install(c('EnsDb.Hsapiens.v86', 'ensembldb'))"
+RUN R -e "remotes::install_github('tidyverse/tidyr')"
 RUN R -e "remotes::install_github('d3b-center/annoFuse', dependencies = TRUE)"
 
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -334,7 +334,7 @@ RUN R -e "remotes::install_github('kgori/sigfit', ref = '209776ee1d2193ad4b682b2
 
 # Package for kinase domain retention for fusions
 RUN R -e "BiocManager::install(c('EnsDb.Hsapiens.v86', 'ensembldb'))"
-RUN R -e "remotes::install_github('tidyverse/tidyr',dependencies = TRUE)"
+RUN R -e "install.packages('tidyr',dependencies = TRUE)"
 RUN R -e "remotes::install_github('d3b-center/annoFuse', dependencies = TRUE)"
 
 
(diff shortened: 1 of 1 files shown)


jharenza:
>So do we want to use this function from annoFuse or replicate the internal functions fusion_driver() and get_Pfam_domain() in OpenPBTA?

I think we can use from annoFuse directly.

jashapiro:
Please follow the instructions in https://github.com/AlexsLemonade/OpenPBTA-analysis#docker-image for adding files to the docker image. In particular, note that `BiocManager::install()` should not be used. Please use `install_bioc.R` instead. Also, any packages installed from github must include a specific commit id. 

If `annoFuse` requires a more recent version of `tidyverse` than we currently have in the image, this will require special accommodations in order to not break other modules.

kgaonkar6:
Thanks for the link @jashapiro 

Yeah it seems annoFuse imports tidyr::one_of() which was only added to tidyr1.02 or later but we have tidyr0.8.3 in openpbat docker so I was trying to update the version but without luck. 

Do you think we can update the tidyr to 1.0.2 or I was gonna try changing the `@importFrom` to dplyr::one_of() in a branch of annoFuse to check if that works in the docker.

jashapiro:
> Yeah it seems annoFuse imports tidyr::one_of() which was only added to tidyr1.02 or later but we have tidyr0.8.3 in openpbat docker so I was trying to update the version but without luck.
> 
> Do you think we can update the tidyr to 1.0.2 or I was gonna try changing the `@importFrom` to dplyr::one_of() in a branch of annoFuse to check if that works in the docker.

Unfortunately, updating `tidyr` is a pretty big project given the other dependencies across the project and the major changes that came with 1.0. We might be able to come up with a solution, but it may not be straightforward. If you do need `one_of` then using the `dplyr` version is probably the best solution for now, and that can be installed pretty easily by pointing `install_github()` to a commit on that branch.

kgaonkar6:
Importing one_of from dplyr seems to have worked! Thanks!

jashapiro (approved):
Looks good! Looking forward to seeing the analysis this enables!

Issue #822, opened 2020-10-26
Updated analysis: craniopharyngioma results table remove duplicates


jharenza:
#### What analysis module should be updated and why?
[cranio subtyping](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-CRANIO) 
The [results table](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-CRANIO/results/CRANIO_molecular_subtype.tsv) contains duplicates because RNA and DNA samples are both in the [SNV lesion dataframe ](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-CRANIO/results/CRANIO_defining_lesions.tsv) as separate rows per patient.

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Rearrange the code to select DNA and RNA specimens first, join them, then do the subtyping. Also, add a classification ("to be classified") for RNA-only samples.

#### What input data should be used? Which data were used in the version being updated?
same- pbta-histologies.tsv and pbta-snv-consensus-mutation.maf.tsv.gz (V17)

#### When do you expect the revised analysis will be completed?
today/tomorrow

#### Who will complete the updated analysis?
@jharenza

jharenza:
closed with #823

PR #823, opened 2020-10-26
Update craniopharyngioma subtyping

jharenza:
Updating craniopharyngioma subtyping results per #822 

#### What was your approach?
- Updated code to remove duplicate patient rows in results file. Changed order of dataframe creation to merge DNA and RNA specimens before subtyping occurred.
- Added subtype "To be classified" for RNA only sample which does not have associated SNV information.
- Add README

#822 

NA

No

yes

- defining lesions table
- results table

De-duplicated results files from 74 to 39 rows.

#### Documentation Checklist


jaclyn-taroni (approved):
This looks good to me. Before we merge, can you add steps that sort the output file by identifier before writing to TSV? We made that change in all of the pre-v17 subtyping modules so that it will be easier to tell if something has truly changed (vs. the order of the samples changing) in diffs (see #770).

jharenza:
> This looks good to me. Before we merge, can you add steps that sort the output file by identifier before writing to TSV? We made that change in all of the pre-v17 subtyping modules so that it will be easier to tell if something has truly changed (vs. the order of the samples changing) in diffs (see #770).

yep, done!

Issue #827, opened 2020-11-04
Updated analysis: survival_analysis_HGG_DMG


kgaonkar6:
#### What analysis module should be updated and why?

survival-analysis/survival-analysis_HGG_DMG.Rmd needs to be updated so that we don't need to use broad_composition in any module in OpenPBTA. 
`broad_composition` is based on integrated_diagnosis so we do not want to use it in any analysis modules.

#### What changes need to be made? Please provide enough detail for another participant to make the update.
change:
```
 broad_composition == “tumor” 
```

to 
```
composition == "Solid Tissue"
```

#### What input data should be used? Which data were used in the version being updated?
pbta-histologies.tsv
independent-specimens.wgswxs.primary.tsv

#### When do you expect the revised analysis will be completed?
1/2 day

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
closed with #828

PR #828, opened 2020-11-04
chnaging broad_composition to composition=="Solid Tissue"

kgaonkar6:
Mainly changing broad_composition=="tumor" to composition =="Solid Tissue" to remove cell-line. broad_composition is based on integrated_diagnosis so we don't want it included in the analysis modules.

#### What was your approach?
change:
```
 broad_composition == “tumor” 
```

to 
```
composition == "Solid Tissue"
```

#827 

It seems the result is also updated with v17 pbta-histologies.tsv without adding any changes in selection criteria (broad_composition !="tumor").

Checking for samples that are cand not broad_composition !="tumor" the molecular_subtype is NA so that couldn't have affected H3 K28M /H3 wildtype subtype numbers:

```
metadata %>% dplyr::filter(composition == "Solid Tissue",broad_composition !="tumor") %>% pull(molecular_subtype)
 [1] NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA NA
```

So I've added a comment on which version of histologies file was used in the notebook to keep track of the input data. Please let me know if this is required/can be handled differently.

 umm should there be a README for this module or not yet?

Yes

table: https://github.com/kgaonkar6/OpenPBTA-analysis/blob/change_to_composition/analyses/survival-analysis/results/survival_model_DMG_H3_K28_v_HGG_H3_wildtype.tsv

figures in notebook

Lesser overall survival for K28M patients compared to H3 wildtype

#### Documentation Checklist


jaclyn-taroni:
> umm should there be a README for this module or not yet?

I'd file an issue about this.

analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd
```diff
@@ -53,7 +53,9 @@ if (!dir.exists(plots_dir)) {
 }
 ```
 
-Declare input file paths and useful variables.
+Declare input file paths and useful variables
+
+Using v17 histologies file
```
jaclyn-taroni, line 58: I'm a little bit worried about this phrasing, because the code itself is not tying it to a particular release. My inclination would be to put a note at the very top of the notebook that it has most recently been run with v17. However, I think it's very easy for this to get out of date (which is not ideal!). There are ways to use `params` here to tie this to v17 when run locally (or not in CI) but change the subdirectory to `testing` in CI, but I'm not sure how much we want to get into that here.

jaclyn-taroni (approved):
Besides moving where we talk about what release was used, this looks good!

kgaonkar6:
I've now moved the version of pbta-histologies comment to the top of the notebook

Issue #808, opened 2020-10-14
Updated analysis: fusion summary to add fusions for LGG


jharenza:
#### What analysis module should be updated and why?
[fusion summary](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion-summary)
Currently, the [LGAT subtyping module](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-LGAT) only looks for _BRAF_ fusions, so it utilizes the `pbta-fusion-putative-oncogenic.tsv` file directly. Since we are adding many more subtypes and considering additional fusions, we should create a new fusion summary file containing this information.

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/793c60417354496d0350838355aad6b6c0e6feb0/analyses/molecular-subtyping-LGAT/02-make-lgat-final-table.Rmd#L54-L60

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Create and add `fusion_summary_lgg_foi.tsv` file consisting of the following fusions and genes with fusions:

**Fusions:**
KIAA1549--BRAF
FGFR1-TACC1
MYB-QKI

**Genes with fusions:**
BRAF
ALK
ROS1
NTRK1
NTRK2
NTRK3
PDGFRA
FGFR2
FGFR1
MYB
MYBL1
RAF1

For kinase fusions with a 5' kinase, we should use an adaptation of #665 to confirm likely functionality and if these conditions are satisfied, then we will add those fusions to the final file:
    1. Is reciprocal == yes
    2. Kinase domain retained == yes
    3. inframe == yes

#### What input data should be used? Which data were used in the version being updated?
`pbta-fusion-putative-oncogenic.tsv`

#### When do you expect the revised analysis will be completed?
2-3 days

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
@kgaonkar6, I will work on this once #821 is merged

jharenza:
closed with #830

PR #793, opened 2020-09-25
Update LGAT molecular subtyping module to use pathology diagnoses

jaclyn-taroni:
### Purpose/implementation Section

In later releases, `short_histology` will be derived from the pathology diagnosis information and molecular subtyping results (#748). The `molecular-subtyping-LGAT` module used to use `short_histology == "LGAT"` as the only inclusion criterion. `short_histology` will need to be downstream of subtyping in the future, so we needed to update the inclusion/exclusion criteria to be based off of `pathology_diagnosis`.

Ganglioglioma samples also needed to be included in this module (https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/753#issuecomment-697008356); those specimens have `short_histology == "Ganglioglioma"` so we would not have included them previously.

#### What was your approach?

I'm going to quote https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/753#issuecomment-697008356 re: criteria for inclusion/exclusion

> **pathology_diagnosis**: 
> Ganglioglioma
> Ganglioglioma;Low-grade glioma/astrocytoma (WHO grade I/II)
> Low-grade glioma/astrocytoma (WHO grade I/II)
> Low-grade glioma/astrocytoma (WHO grade I/II);Other
> 
> There were no LGG in the tumors diagnosed as `Other` as of right now.
> 
> NOTES:
> - Ganglioglioma also shows up with `Dysembryoplastic neuroepithelial tumor (DNET);Dysplasia/Gliosis;Ganglioglioma` and `Dysembryoplastic neuroepithelial tumor (DNET);Ganglioglioma` and these are classified as `neuronal and mixed neuronal-glial tumor`, so we **_do not_** want to include those.

This is to say that samples we very much would _like to include or specifically exclude_ can sometimes have multiple pathology dx separated by `;`. I can imagine a situation where the contents (and therefore _interpretation_) of the `pathology_diagnosis` column remains the same but the order changes. For that reason, I'm electing to use a string detection approach here, rather than exact matching.

#### What GitHub issue does your pull request address?

Closes #753 

#### Which areas should receive a particularly close look?

* The new notebook being included here (`analyses/molecular-subtyping-LGAT/00-LGAT-select-pathology-dx.Rmd`). It is being added in the interest of documenting the changes over time in the analytical code base itself. 
[...]

PR #830, opened 2020-11-06
add fusion summary for LGAT

jharenza:
Adding fusion summary for LGAT tumors in preparation for #793 

#### What was your approach?
1. Added Fusions of interest: (_KIAA1549--BRAF, FGFR1--TACC1, and MYB--QKI_) and Genes of interest with fusions:
BRAF
ALK
ROS1
NTRK1
NTRK2
NTRK3
PDGFRA
FGFR2
FGFR1
MYB
MYBL1
RAF1
2. Added all 3' kinase fusions which are in-frame and retain the kinase domain 
3. For 5' kinase fusions, added those which are in-frame and retain the kinase domain
4. For 5' kinase fusions that don't meet 3., check whether it has a reciprocal fusion and whether that reciprocal is in-frame and the kinase domain is retained. Then, add those 5' kinase fusions.

#808 

Double check the logic

I kept all of this analysis in the same Rmd as the other fusion summary analyses because I am using / modifying those functions. I almost think some of this could be its own notebook since it is unique to LGAT, but could go either way. Thoughts on this?

Yes

table

**Numbers of unique patient fusions:**
Non-kinase fusions: 11
3' kinase fusions which are in-frame and retain the kinase domain: 167
5' kinase fusions, added those which are in-frame and retain the kinase domain: 1
5' kinase fusions that don't meet 3., has a reciprocal fusion in-frame and the kinase domain is retained: 12 

Combined these into fusion_summary_lgat_foi.tsv 

#### Documentation Checklist


analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -67,10 +91,14 @@ prepareOutput <- function(fuseDF, bioid) {
...
 dataDir <- file.path("..", "..", "data")
+fusDir <- file.path("..", "..", "analyses", "fusion_filtering", "results")
+annotDir <- file.path("..", "..", "analyses", "fusion_filtering", "references")
 #' The putative oncogenic fusion file is what we'll use to check for the 
+#' annotation file
+annot <- read.delim2(file.path(annotDir, "genelistreference.txt"), sep = "\t", header = T)
 #' presence or absence of the fusions.
 putativeOncogenicDF <- 
-  read_tsv(file.path(dataDir, "pbta-fusion-putative-oncogenic.tsv"))
+  read_tsv(file.path(fusDir, "pbta-fusion-putative-oncogenic.tsv"))
```
jharenza, line 101: @jaclyn-taroni - some heads up:

I used `read.delim2()` here because `read_tsv()` gave me an error in that the first column was rownames and the last column was missing. Maybe you will know a workaround if you don't like read.delim2 :)

The putative oncogenic path will need to be changed with v18 update once we have the required columns, but for now, I pulled from the latest analysis.

jaclyn-taroni: @cansavvy is going to provide a review, so maybe she will have some ideas!

cansavvy: Where does `genelistreference.txt` come from? [The README says what it contains](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering#inputs-used-as-reference) but its unclear to me where its from -- there's a lot of references listed which is great but not info about how this file came to be. Anyway, `genelistreference.txt`  has numbering saved as a row name type of column which is why this is giving an error (it doesn't have a column name). It would be best if we just got rid of that numbering column since its not doing anything anyway. 

Screen Shot 2020-11-06 at 9 21 45 AM

This might need to be its own issue (probably not something we should address in this PR).

cansavvy: If we cannot or should not change the file, and we want to use readr::read_tsv(), then we can get around this by providing col_names and telling readr to skip whatever column name we give to the number column, but the better solution if possible is just to drop this column that doesn't give any info.

kgaonkar6: The file was created from different sources and formatted [code](https://gist.github.com/kgaonkar6/02b3fbcfeeddfa282a1cdf4803704794) I can probably re-do that to save with write_tsv instead of write.table which I believed added the rownames (default row.names==TRUE)

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -67,10 +91,14 @@ prepareOutput <- function(fuseDF, bioid) {
 
 ```{r}
 dataDir <- file.path("..", "..", "data")
+fusDir <- file.path("..", "..", "analyses", "fusion_filtering", "results")
+annotDir <- file.path("..", "..", "analyses", "fusion_filtering", "references")
 #' The putative oncogenic fusion file is what we'll use to check for the 
+#' annotation file
+annot <- read.delim2(file.path(annotDir, "genelistreference.txt"), sep = "\t", header = T)
```
cansavvy, line 98: If we don't change to readr::read_tsv(), we will need to add one more argument here: 
```suggestion
annot <- read.delim2(file.path(annotDir, "genelistreference.txt"), sep = "\t", header = TRUE, stringsAsFactors = FALSE)
```

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+out_frame_reciprocals <- setdiff(five_prime_reciprocals[,c("Sample", "FusionName")], five_prime_reciprocals_full[,c("Sample", "FusionName")])
+out_frame_reciprocals
+
+#Check these do not retain the kinase domain. They do not, so will leave out
+out_frame_reciprocals_full <- out_frame_reciprocals %>%
+  left_join(allFuseLGAT, by = c("Sample", "FusionName")) %>%
+  select(Sample, FusionName, Gene1A, Gene1B, Fusion_Type, DomainRetainedGene1A, DomainRetainedGene1B)
+
+# Rbind lists for final table of LGAT fusions of interest
+subsetFuseLGAT <- bind_rows(nonkinaseLGAT, three_prime_kinase_inframe[,c("Sample", "FusionName")],
```
kgaonkar6, line 335:
```suggestion
subFuseLGAT <- bind_rows(nonkinaseLGAT, three_prime_kinase_inframe[,c("Sample", "FusionName")],
```
In the next chunk it seems the object `subsetFuseLGAT` should be actually named `subFuseLGAT` ?

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+
+# Pull LGAT non-kinase fusions
+nonkinaseLGAT <- filterFusion(df = putativeOncogenicDF,
+                             fuses = nonkinase_lgatFuses,
+                             genes = nonkinase_lgatGenes) %>%
+  distinct()
+
+# First, filter all fusion dataframe for 3' kinases which are in-frame and retain the kinase domain - keep these
+three_prime_kinase_inframe <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1B_anno) & Fusion_Type == "in-frame" & DomainRetainedGene1B == "Yes") %>%
```
cansavvy, line 288: If the other data frames have `"Kinase"` capitalized then we should probably change `lgatFuses_df` to have a capitalized `"Kinase"` for consistency.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+  select(Sample, FusionName, Gene1A, Gene1B) %>%
+  distinct()
+
+# Which fusions are not in-frame?
+three_prime_kinase_outframe <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1B_anno) & Fusion_Type != "in-frame") %>%
```
cansavvy, line 294: Basic question: If we have an in-frame fusion but the domain is not retained, we aren't counting it as in frame? So are there fusions that are in-frame but not retaining the domain that will not be in either dataframes being set up here?

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+  distinct()
+# Check they don't retain the kinase domain. They do not, so leave out.
+table(three_prime_kinase_outframe$DomainRetainedGene1B)
```
cansavvy, line 297: If we have things being printed out, it would be easier if this were at the end of a chunk so we can see the output printed nearer to the code that's responsible for it. 

In general it might be nice to have this one very long chunk broken up into smaller chunks that are more focused.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+
+# Keep those with kinase domain retained and fusion in-frame - keep this list
+five_prime_domain_intact <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE") %>%
+  select(Sample, FusionName, Gene1A, Gene1B, Fusion_Type, DomainRetainedGene1A) %>%
+  filter(DomainRetainedGene1A == "Yes") %>%
+  distinct()
+
+# Identify reciprocal matched fusions which are in-frame
+five_prime_domain_lost$reciprocal <- paste(five_prime_domain_lost$Gene1B, five_prime_domain_lost$Gene1A, sep ="--")
```
cansavvy, line 314: Can add this on to the set of tidyverse manipulations you have above.

cansavvy: I'll put a suggestion up there.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+  distinct()
+# Check they don't retain the kinase domain. They do not, so leave out.
+table(three_prime_kinase_outframe$DomainRetainedGene1B)
+
+# Filter all fusions for 5' kinase fusions which have reciprocal fusions and have lost the kinase domain. 
+five_prime_domain_lost <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE") %>%
+  select(Sample, FusionName, Gene1A, Gene1B, Fusion_Type, DomainRetainedGene1A) %>%
+  filter(DomainRetainedGene1A == "No") %>%
+  distinct()
```
cansavvy, line 304:
```suggestion
  distinct() %>% 
  mutate(reciprocal = paste(Gene1B, Gene1A, sep ="--"))
```

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+# Keep those with kinase domain retained and fusion in-frame - keep this list
+five_prime_domain_intact <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE") %>%
+  select(Sample, FusionName, Gene1A, Gene1B, Fusion_Type, DomainRetainedGene1A) %>%
+  filter(DomainRetainedGene1A == "Yes") %>%
+  distinct()
+
+# Identify reciprocal matched fusions which are in-frame
+five_prime_domain_lost$reciprocal <- paste(five_prime_domain_lost$Gene1B, five_prime_domain_lost$Gene1A, sep ="--")
+five_prime_reciprocals <- five_prime_domain_lost[,c("Sample", "reciprocal")] %>%
```
cansavvy, line 315: What are these different and new data frames for?

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+  rename(FusionName = reciprocal)
+five_prime_reciprocals_full <- five_prime_reciprocals %>%
```
cansavvy, line 317: If you do this, you can delete the lines above. 
```suggestion
five_prime_reciprocals_full <- five_prime_domain_lost %>% 
     select(Sample, FusionName = reciprocal) %>%
```

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
   write_tsv(ewingsFile)
 ```
 
+#### Perform selection for LGAT fusions
```
kgaonkar6, line 260: Maybe we can have smaller chunks per selection criteria for LGAT below or do we want it 1 chunk per broad_histology ?
 - Added all 3' kinase fusions which are in-frame and retain the kinase domain
 - For 5' kinase fusions, added those which are in-frame and retain the kinase domain
- For 5' kinase fusions that don't meet 3., check whether it has a reciprocal fusion and whether that reciprocal is in-frame and the kinase domain is retained. Then, add those 5' kinase fusions.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+five_prime_domain_lost$reciprocal <- paste(five_prime_domain_lost$Gene1B, five_prime_domain_lost$Gene1A, sep ="--")
+five_prime_reciprocals <- five_prime_domain_lost[,c("Sample", "reciprocal")] %>%
+  rename(FusionName = reciprocal)
+five_prime_reciprocals_full <- five_prime_reciprocals %>%
+  left_join(allFuseLGAT, by = c("Sample", "FusionName")) %>%
+  filter(Fusion_Type == "in-frame")
+# Keep the list of the original 5' kinase fusions with matched reciprocals
+five_prime_kinase_keep <- five_prime_reciprocals_full[,c("Sample", "Gene1A", "Gene1B")] %>%
+  distinct()
+  five_prime_kinase_keep$FusionName <- paste(five_prime_kinase_keep$Gene1B, five_prime_kinase_keep$Gene1A, sep ="--")
```
cansavvy, line 323: You can use the same suggestion as above to shorten this.

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: @kgaonkar6 you had cleaned up that file in [annofuse](https://github.com/d3b-center/annoFuse/issues/62) - will you add an issue and quick PR here?

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+
+# Which ones are not in-frame? These are mostly genes fused to self
+out_frame_reciprocals <- setdiff(five_prime_reciprocals[,c("Sample", "FusionName")], five_prime_reciprocals_full[,c("Sample", "FusionName")])
```
cansavvy, line 326: I've lost how this setdiff step helps us identify outframe reciprocals?

Does it make sense instead to make a column that designates this and filter by it? Or to go the original dataframe that says "inframe" or "outframe"? That might be clearer and easier to track back if we need to.

kgaonkar6 (commented):
Looks good to me, just a few questions/ suggestions I had are commented below.

I wonder if the CI is errors out with the test data?

analyses/fusion-summary/01-fusion-summary.Rmd
kgaonkar6: ok will do

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+# Filter all fusions for 5' kinase fusions which have reciprocal fusions and have lost the kinase domain. 
+five_prime_domain_lost <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE") %>%
+  select(Sample, FusionName, Gene1A, Gene1B, Fusion_Type, DomainRetainedGene1A) %>%
+  filter(DomainRetainedGene1A == "No") %>%
+  distinct()
+
+# Keep those with kinase domain retained and fusion in-frame - keep this list
+five_prime_domain_intact <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE") %>%
```
cansavvy, line 308: Since these two steps look identical, does it make sense to have this as one bigger data frame and instead of splitting it up into two data frames?

cansavvy (commented):
From what I can tell, it looks like this implements what's suggested in #808. :+1:

My suggestions are mainly about where we might be able to shorten this up and add clarity in certain parts.

analyses/fusion-summary/01-fusion-summary.Rmd
cansavvy: Sounds great! :+1:

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: thanks for catching that - this is what happens when you write code at midnight! I was renaming to make things clearer, so just now, I actually updated the latter to `subsetFuseLGAT`

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -44,6 +44,30 @@ filterFusion <- function(df, bioid, fuses, genes) {
   return(df %>% select(Sample, FusionName))
 }
 
+#create function specifically for lgat, since we want more information from the putative oncogenic file
+filterLGATFusion <- function(df, bioid, fuses, genes) {
```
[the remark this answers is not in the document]
jharenza: The bioid is pulling `Sample` which harbor a fusion or gene from the list from the putative_oncogenic file (lines 24-26)
```
filterFusion <- function(df, bioid, fuses, genes) {
  if (!missing(bioid)) {
    df <- filter(df, Sample %in% bioid)
  }
```

>It looks like the main difference is the select() step? Is this right? Since this function is only used once maybe we can just use filterFusion and add a select step after that (no function).

This is correct, but `select()` cannot be used after the function because the function itself has its own select step in line 44 
```
return(df %>% select(Sample, FusionName))
}	
```

What I could do is add the extra columns into the original function so that there is only one function, and subset out the sample and FusionName with those other histologies in the end. That would get rid of the other function if that sounds good to you.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -44,6 +44,30 @@ filterFusion <- function(df, bioid, fuses, genes) {
...
+    df <- filter(df, FusionName %in% fuses)
+  } else if (!missing(genes)) {
+    df <- filter(df,
+                 Gene1A %in% genes |
+                   Gene2A %in% genes |
+                   Gene1B %in% genes |
+                   Gene2B %in% genes)
+  }
+  return(df %>% 
+           select(Sample, FusionName, Fusion_Type, Gene1A, Gene1B, Gene2A,  Gene2B, Gene1A_anno, Gene1B_anno, LeftBreakpoint, RightBreakpoint,      reciprocal_exists, DomainRetainedGene1A, DomainRetainedGene1B) %>%
```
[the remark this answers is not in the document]
jharenza: For this, since we are specifically looking for in-frame (`Fusion_Type`), kinase (`Gene1A_anno`, `Gene1B_anno`), whether a reciprocal exists (`reciprocal_exists `), and fusions with kinase domains retained (`DomainRetainedGene1A`, `DomainRetainedGene1B`), these columns required for LGAT fusion compilation. I added the left and right breakpoint columns to make sure that I was seeing unique fusions, although the rows may be unique enough with the combination of columns above. Let me check on this.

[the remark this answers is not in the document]
jharenza: Looks like the number of unique fusions without breakpoints == 265 and with breakpoints == 300, but the results are identical. I will remove those two columns!

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
 
+#### Perform selection for LGAT fusions
+
+```{r}
+# Which genes/fusions are not kinases, but in the list?
+# Separate LGAT fusions into genes, combine with gene list, check for not kinase
+lgatFuses_df <- as.data.frame(lgatFuses) %>%
+  separate(lgatFuses, into = c("Gene1A", "Gene1B"), remove = F)
+kinases <- annot %>%
+  filter(type == "Kinase")
```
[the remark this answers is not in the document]
jharenza: good tip!

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+lgatFuses_df$Gene1A_anno <- ifelse(lgatFuses_df$Gene1A %in% kinases$Gene_Symbol, "kinase", "")
```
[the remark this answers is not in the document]
jharenza: no reason they are not being labeled - I can do that.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+nonkinase_lgatFuses <- lgatFuses_df$lgatFuses
+
+# Identify non-kinase genes in LGAT goi list
+nonkinase_lgatGenes <- setdiff(lgatGenes, kinases$Gene_Symbol)
+
+# Pull LGAT non-kinase fusions
+nonkinaseLGAT <- filterFusion(df = putativeOncogenicDF,
+                             fuses = nonkinase_lgatFuses,
+                             genes = nonkinase_lgatGenes) %>%
+  distinct()
```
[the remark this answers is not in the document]
jharenza: yes, these are repetitive because there are many fusions with multiple breakpoints or frame-ness, which aren't being captured in the columns (not needed here since looking for exact fusions and not looking for kinases)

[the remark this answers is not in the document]
jharenza: Went from 19 --> 11

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +257,104 @@ allFuseEwing %>%
...
+# Separate LGAT fusions into genes, combine with gene list, check for not kinase
+lgatFuses_df <- as.data.frame(lgatFuses) %>%
+  separate(lgatFuses, into = c("Gene1A", "Gene1B"), remove = F)
+kinases <- annot %>%
+  filter(type == "Kinase")
+lgatFuses_df$Gene1A_anno <- ifelse(lgatFuses_df$Gene1A %in% kinases$Gene_Symbol, "kinase", "")
+lgatFuses_df$Gene1B_anno <- ifelse(lgatFuses_df$Gene1B %in% kinases$Gene_Symbol, "kinase", "")
+# Only pull fusions that do not contain kinase genes, as ones with kinases will be dealt with sepaately later
+lgatFuses_df <- lgatFuses_df %>%
+  filter(lgatFuses_df$Gene1A_anno != "kinase" & lgatFuses_df$Gene1B_anno != "kinase") %>%
```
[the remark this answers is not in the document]
jharenza: also good tip!

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: I just added some checks to check these and commented on the code. You are correct, there are two fusions which are in-frame, but do not retain the kinase domain. In case 1, there is another in-frame fusion with the kinase domain, so it is being captured in the `three_prime_kinase_inframe` list, but in case 2, the in-frame fusion does not retain the kinase domain and so we are not keeping.

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: will break these up some

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: I actually don't think this will work because I want to pull the reciprocals of these using `allFuseLGAT`, but I have combined steps to do this.

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: Actually I have added the inframe + kinase domain retained selection above in `five_prime_reciprocals` so this check is no longer necessary and combined `five_prime_reciprocals` and `five_prime_reciprocals_full` into `five_prime_reciprocals`

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: updated to smaller chunks

analyses/fusion-summary/01-fusion-summary.Rmd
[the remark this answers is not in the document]
jharenza: consolidated functions in [this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/830/commits/6a74fd707742fc0df05d42836ab41be570c66156)

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: >I wonder if the CI is errors out with the test data?

This may be due to fusions not being present in the test data. @jaclyn-taroni I think we need to add more fusions to the test data. I only see the below fusions in the test fusion file, so we would have zeroed out dataframes in this module, which may be what is causing the CI failure.

```
intersect(subsetFuseLGAT[,c("Sample", "FusionName")], test.fus[,c("Sample", "FusionName")])
# A tibble: 4 x 2
  Sample      FusionName    
  <chr>       <chr>         
1 BS_49852ZBH KIAA1549--BRAF
2 BS_7P83Z28P KIAA1549--BRAF
3 BS_BMMY0H73 NTRK2--NTRK2  
4 BS_F6V1Y4QS GOPC--ROS1  
```

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: yup

jharenza:
@cansavvy and @kgaonkar6 - just made all of the updates and requested re-review. I actually found a bug in the 5' in-frame fusion filtering and had missed 3 `FGFR1--TACC1` fusions, so the new total for all LGAT fusions is 193.

@cansavvy thank you SO much for your helpful tips - this was my first tidyverse code :)

analyses/fusion-summary/01-fusion-summary.Rmd
jaclyn-taroni: Yes, we already have special consideration for fusions in how we generate the test data. It is a matter of updating those steps to reflect your changes here.

See the docs for [`create-subset-files`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b39ff2e25439a6c66a4f278a5314984da3eb8eb0/analyses/create-subset-files/README.md#special-considerations) and the code where we make that accommodation:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b39ff2e25439a6c66a4f278a5314984da3eb8eb0/analyses/create-subset-files/02-subset_files.R#L112

If you want to file a new pull request with the required changes, we can probably regenerate the files and get them uploaded to the right spot before this gets merged.

jharenza: @jaclyn-taroni quick question - I have updated the fusions/genes to select. Do you just want the change in `OpenPBTA-analysis/analyses/create-subset-files/02-subset_files.R` or do you want me to also do the subsetting of files? Sounds like you would do this on your end and upload to s3?

jharenza: Nevermind, I did a quick subset of the putative oncogenic and tested it, and it works. PR coming in a min!

analyses/fusion-summary/01-fusion-summary.Rmd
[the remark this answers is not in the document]
cansavvy: Okay. Sounds good.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+```
+
+Rbind lists for final table of LGAT fusions of interest
+```{r}
+# Rbind lists for final table of LGAT fusions of interest
+subsetFuseLGAT <- bind_rows(nonkinaseLGAT,
+                            three_prime_kinase_inframe[,c("Sample", "FusionName")], 
+                            five_prime_domain_intact[,c("Sample", "FusionName")], 
+                            five_prime_kinase_keep[,c("Sample", "FusionName")]) %>%
+  select(Sample, FusionName) %>%
```
cansavvy, line 341: It looks like you've already selected all the data frames to `c("Sample", "FusionName")` so is this `select()` here is a bit redundant except for `nonkinaseLGAT`. Since bind_rows go by column names, you could keep this `select()` here and delete all the instances of `[,c("Sample", "FusionName")]` above and it will also be fine but shorter to read.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+# Then, select the in-frame reciprocals which have the kinase domain in tact.
+five_prime_reciprocals <- five_prime_domain_lost %>% 
+  select(Sample, FusionName = reciprocal) %>%
+  left_join(allFuseLGAT, by = c("Sample", "FusionName")) %>%
+  filter(Fusion_Type == "in-frame" & DomainRetainedGene1B == "Yes")
+
+# Add back the reciprocal of the reciprocal (the 5' kinase fusion) and keep this list
+five_prime_kinase_keep <- five_prime_reciprocals %>%
+  mutate(reciprocal = paste(Gene1B, Gene1A, sep ="--")) %>%
+  select(Sample, reciprocal) %>%
```
cansavvy, line 329:
```suggestion
  select(Sample, FusionName = reciprocal) %>%
```
You can rename things in a select step so this way we can get rid of line 330.

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+  rename(FusionName = reciprocal) %>%
```
cansavvy, line 330:
```suggestion
```

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+# Keep those with kinase domain retained and fusion in-frame - keep this list
+five_prime_domain_intact <- allFuseLGAT %>%
+  filter(grepl("Kinase", Gene1A_anno) & DomainRetainedGene1A == "Yes" & Fusion_Type == "in-frame") %>%
+  select(Sample, FusionName) %>%
+  distinct()
+```
+
+Next, filter all fusions for 5' kinase fusions which have lost the kinase domain for reciprocal fusions which have a kinase domain intact and are in-frame.
+Keep these for the final output file.
+```{r}
```
cansavvy, line 311: I'm thinking this chunk can still be simplified. In particular, I'm noticing that `five_prime_domain_lost` is from `allFuseLGAT` and `five_prime_reciprocals` is from `five_prime_domain_lost` but then we are adding back info from `allFuseLGAT`.

So basically both `five_prime_domain_lost` and `five_prime_reciprocals` are based on `allFuseLGAT` but we have extra manipulation steps. I think we can do this more directly.

cansavvy: As I'm trying to make this shorter, I'm a bit confused on these steps. 

I tried to boil it down to what looks like the essential steps to get `five_prime_kinase_keep`, but as of now it does not get the same result. I think part of this is I still don't really understand all that is happening in this chunk. Can you take a look at my attempt as well evaluating your original code to see how we can make these steps more efficient and clear? 

In particular I know the differences have to do with the paired up  "select() then distinct()" steps that are done a few timees, but its unclear to me why each of these "select() then distinct()" steps are necessary. 

```
five_prime_kinase_keep <- allFuseLGAT %>%
  # Make reciprocal variable
  mutate(reciprocal = paste(Gene1B, Gene1A, sep ="--")) %>%
  filter(# Keep the in-frame reciprocals which have the kinase domain in tact.
        grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE" & DomainRetainedGene1A == "No" |
        # Also keep select the in-frame reciprocals which have the kinase domain in tact.
        Fusion_Type == "in-frame" & DomainRetainedGene1B == "Yes") %>% 
  # Select to which variables we want here
  select(Sample, FusionName = reciprocal) %>% 
  distinct()
```

cansavvy (commented):
Hi @jharenza, thanks for these changes, I think this is coming along nicely! I just have one more area that I think we should further evaluate as far as what the code is doing.

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: #836 submitted

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+# Separate LGAT fusions into genes, combine with gene list, check for not kinase
+lgatFuses_df <- as.data.frame(lgatFuses) %>%
+  separate(lgatFuses, into = c("Gene1A", "Gene1B"), remove = F)
+
+kinases <- annot %>%
+  filter(type == "Kinase") %>%
+  pull(Gene_Symbol)
+
+lgatFuses_df <- lgatFuses_df %>% 
+  dplyr::mutate(Gene1A_anno = case_when(Gene1A %in% kinases ~ "Kinase"),
```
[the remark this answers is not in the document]
jharenza: oh, another good tip!

analyses/fusion-summary/01-fusion-summary.Rmd
```diff
@@ -203,9 +232,133 @@ allFuseEwing %>%
...
+  separate(lgatFuses, into = c("Gene1A", "Gene1B"), remove = F)
+
+kinases <- annot %>%
+  filter(type == "Kinase") %>%
+  pull(Gene_Symbol)
+
+lgatFuses_df <- lgatFuses_df %>% 
+  dplyr::mutate(Gene1A_anno = case_when(Gene1A %in% kinases ~ "Kinase"), 
+                Gene1B_anno = case_when(Gene1B %in% kinases ~ "Kinase")) %>%
+   replace_na(list(Gene1A_anno = "Non-Kinase", Gene1B_anno = "Non-Kinase"))
```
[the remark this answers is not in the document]
jharenza: got it!

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: ahh, cool. got it!

analyses/fusion-summary/01-fusion-summary.Rmd
jharenza: Ahh
```filter(# Keep the in-frame reciprocals which have the kinase domain in tact.
        grepl("Kinase", Gene1A_anno) & reciprocal_exists == "TRUE" & DomainRetainedGene1A == "No" |
        # Also keep select the in-frame reciprocals which have the kinase domain in tact.
        Fusion_Type == "in-frame" & DomainRetainedGene1B == "Yes")
```
This filtering you have here is filtering on the frame-ness/domain retention of the 5' kinase fusions and not their reciprocals (3' kinase fusions), which is why I was making that variable, then re-joining `five_prime_reciprocals` with the `allFuseLGAT`. 

What I am trying to do is find any 5' kinases without the domain retained, but then go back to that original dataframe `allFuseLGAT` and collect the annotations of their reciprocal fusions (ie 3' kinase fusions' frameness and domain retention), assess whether _those_ are in-frame and retain the kinase domain, then if there is a reciprocal that satisfies that, join _both_ the 5' kinase fusion dataframe (domain lost) and 3' kinase fusion dataframe (inframe/domain retained).

Actually, I can remove:
```
# Add back the reciprocal of the reciprocal (the 5' kinase fusion) and keep this list
five_prime_kinase_keep <- five_prime_reciprocals %>%
  mutate(reciprocal = paste(Gene1B, Gene1A, sep ="--")) %>%
  select(Sample, FusionName = reciprocal) %>%
  distinct()
```

and use:
```
# Then, select the in-frame reciprocals which have the kinase domain in tact. Retain 5' kinase fusion information and update 3' fusion column name to FusionName for merging with allFuseLGAT.
five_prime_kinase_keep <- five_prime_domain_lost %>% 
  select(Sample, five_prime_kinase = FusionName, FusionName = three_prime_kinase) %>%
  left_join(allFuseLGAT, by = c("Sample", "FusionName")) %>%
  filter(Fusion_Type == "in-frame" & DomainRetainedGene1B == "Yes") %>%
  select(Sample, FusionName = five_prime_kinase) %>%
  distinct()
```

`distinct()` is needed because in some cases, there are multiple breakpoints for each fusion and sometimes multiple can have the same frame-ness, so this reduces them to one fusion call per sample.

jharenza: updated with [commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/830/commits/ec13333862792e8b02e5726987c5156aeec33666) - @cansavvy let me know if this clarifies!

cansavvy: Okay! Thanks for those updates! I think this looks good!

cansavvy (approved):
This looks good to me! Thanks for the clarification!

Issue #832, opened 2020-11-06
Updated File and Documentation : Fusion gene and fusion reference lists


kgaonkar6:
#### What analysis module should be updated and why?
fusion_filtering : the files currently have rownames which causes issues reading the file as discussed in the [thread](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/830#discussion_r518758733)

Screenshot from @cansavvy 
https://user-images.githubusercontent.com/23458084/98376682-d8dfa380-2011-11eb-8202-8c031c6736d6.png

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Remove the row.names while saving the file , save with write_tsv()

#### What input data should be used? Which data were used in the version being updated?
https://gist.github.com/kgaonkar6/02b3fbcfeeddfa282a1cdf4803704794

#### When do you expect the revised analysis will be completed?
1/2 day

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
updated via #836

PR #836, opened 2020-11-09
removing rownames from reference files

kgaonkar6:
Updating format for input gene and fusion list to be compatible with read_tsv and updating README 

#### What was your approach?
I saved the file with write_tsv() so now the files can be read with read.delim as well as read_tsv()

#832 

I'm linking to the code I'd used originally to format the and create the file does that sound ok?

NA

NA

NA

NA

#### Documentation Checklist


jharenza (commented):
@kgaonkar6 thanks for this - will it break anything in fusion-filtering right now?

kgaonkar6:
It doesn't break the script, it's only used https://github.com/kgaonkar6/OpenPBTA-analysis/blob/23a116afbd2da387fbc1e59143df3549aa94460c/analyses/fusion_filtering/02-fusion-filtering.R#L241-L255 which works fine now as well

jharenza (approved):
Looks good to me!

Issue #837, opened 2020-11-09
Updated analysis: investigate putative functionality of TP53 SNV and CNV calls


jharenza:
#### What analysis module should be updated and why?
[tp53_nf1_score](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tp53_nf1_score) was never fully completed because of bandwidth.

Since we would like to assess whether or not _TP53_ alterations are likely functional within #807, it seems like a better place to put this would be within the [tp53_nf1_score](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tp53_nf1_score) module.

For reference of where this was left, see this [PR comment ](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/720#issuecomment-690246491) by @jaclyn-taroni and [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/720#issuecomment-644987600) by @sjspielman and #720 

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Referencing #807, we should assess the likely functionality of _TP53_ alterations before calling them "altered" or "wildtype" for AUROC assessment of the classifier.

### First
QC: Perform a correlation between RNA-Seq expression values and _TP53_ classifier scores. Are these inversely correlated as we would expect?

### Second
Reduce the CNV list by focusing on CNVs which delete one or more of _TP53's_ [functional domains](https://www.nature.com/articles/4401904#:~:text=The%20functional%20domains%20in%20p53%20family%20proteins%20exhibit%20both%20common%20and%20distinct%20properties,-K%20L%20Harms%20%26&text=p53%20is%20a%20sequence%2Dspecific,and%20other%20p53%2Ddependent%20activities.). 
- TAD = trans-activating domain (essential for function): AD1 (residues 1–42) and AD2 (residues 43–92)
- DBD = DNA-binding domain (residues 102–292)
- TD = tetramerization domain (residues 326–356)

Something to keep in mind for samples which may have low TP53 scores, but have alterations: _TP63_ and TP73 are homologues which can be functionally redundant and rarely mutated, so if in tact, these might compensate. In addition, for samples which may have high _TP53_ scores but no alterations, we can check for MDM2 amplification, TP53's most potent negative regulator. For reference from the above paper: 
>The p63-AD1 (residues 1–59) and p73-AD1 (residues 1–54) are 22 and 29% identical to AD1 in p53, respectively

[...]

PR #839, opened 2020-11-10
qc script to find correlation between classifier score and expression 

kgaonkar6:

Is there a  negative correlation between TP53 inactivation classifier score and TP53 expression 

#### What was your approach?

Read in classifier score and expression RDS file to match by BS_ID and then cor.test (tp53_score, expression_value)

#837 

I have not applied any filter on the classifier score nor on expression in the code. I did run separately a cutoff of 0.7 below but seems to not change the correlation drastically do we want this in the code as well for different cutoff?

```r
> stranded_exp_score_more_than_0.7 <- stranded_exp_score[stranded_exp_score$tp53_score>0.7,]
> cor(scale(stranded_exp_score_more_than_0.7$expression_value),stranded_exp_score_more_than_0.7$tp53_score)
          [,1]
[1,] 0.1200886
```

This is 1 of 5 tasks in #837  which I think will all be steps before we run 03-evaluate-classifier.py so evaluation plots will be updated as we go through these steps.

No

NA

There is negligible correlation between the scores and expression of TP53 

#### Documentation Checklist


analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
```diff
@@ -0,0 +1,106 @@
...
+
+---
+    
+In this notebook we will find correlation between RNA-Seq and classifier score
+We will check if the values are inversely correlated as we would expect?
+
+### Setup
+```{r}
+library("ggpubr")
+library("ggthemes")
```
jharenza, line 14: Right now, you aren't using these. However, I do want to request scatterplots so we can see the data, as poor as the correlations may be.

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
```diff
@@ -0,0 +1,106 @@
...
+
+ 1) Read in scores from tp53-nf1-classifier for stranded
+```{r}
+
+score_stranded_df <- read_tsv(file.path(results_dir, "pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv")) %>%
+  dplyr::select(
+    "tp53_score",
+    "sample_id"
+  ) %>%
+  unique() %>%
```
jharenza, line 34:
```suggestion
```
I think this shouldn't be necessary since there should be only one score per BS_ID already, right?

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
```diff
@@ -0,0 +1,106 @@
...
+```
+
+ 2) Read in scores from tp53-nf1-classifier for polya
+```{r}
+score_polya_df <- read_tsv(file.path(results_dir, "pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv")) %>%
+  dplyr::select(
+    "tp53_score",
+    "sample_id"
+  ) %>%
+  unique() %>%
```
jharenza, line 46:
```suggestion
```
same comment as above

jharenza (changes requested):

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
kgaonkar6: Thanks for the review! I've added the scatterplots to the notebook now.

jharenza (commented):
Thanks for adding the correlation plots. So what this is telling me is that we cannot, in a straightforward way, utilize TP53 score + baseline expression to infer TP53 function. Will you document some comments in the notebook about this and we can get it merged?

kgaonkar6:
Just added a comment, in the end of the notebook to mention that we don't have a direct inference of functionality from expression+ classifier score since we have negligible correlation between them

jharenza (approved):

PR #841, opened 2020-11-13
Tp53 cnv domain loss

kgaonkar6:
Find cnv losses that overlap domains in TP53  
 

- TAD = trans-activating domain (essential for function)
-  DBD = DNA-binding domain (residues 102–292)
-  TD = tetramerization domain (residues 326–356)

We want to subset CNV calls where the domain are lost which will possibly lead to loss of function to use for evaluation of TP53  inactivation score at a later step.

#### What was your approach?
I read in `consensus_seg_with_status.tsv` which we generate as a scratch file while running `02-add-ploidy-consensus.Rmd` in focal-cn-file-preparation which adds ploidy information to the consensus SEG file and adds a status column that defines gain and loss broadly. I

```
 Rscript -e "rmarkdown::render('analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd', clean = TRUE)"
```

We needed the start and end location of CNVs which I didn't find in `consensus_seg_annotated_cn_autosomes.tsv.gz` so from discussion with @jharenza we decided to use  the seg file. This file `consensus_seg_with_status.tsv` provided the location,ploidy and status so I'm using this file here.

#837 

I'm running a direct overlap for domain and CNVs locations.  Should CNVs be filtered for length? 

image

Loss are distributed as , for next steps we would need to check for double mutation so are 1 copy loss ok to keep for next steps or should we only look for more than 1 copy loss :
image

I'm trying to make a staggered PR for #837 
1 ) #839 
2) This PR with TP53 cnv domain overlap

So when #839 gets merged I'd have to pull in the changes. Also I specifically wanted to know does this work with renaming 03-evaluate-classifier.py to 04-evaluate-classifier.py in this PR ?

Yes table

tp53-subset/loss_overlap_domains_tp53.tsv

276 unique bs_id have losses that overlap functionally relevant domains in TP53 .

#### Documentation Checklist


jharenza (changes requested):
@kgaonkar6 thanks for this!

I am going to request you create a notebook for this, since there will be a lot of scientific discussion around this exploration and it will be useful to have it organized.

To address your questions:
>I'm running a direct overlap for domain and CNVs locations. Should CNVs be filtered for length?

I don't think we need to filter for CNV length. However, in looking at your `loss_overlap_domains_tp53.tsv` file, it does appear that all samples that have loss of any domain have loss of all domains. Perhaps you can print the full table in the notebook (combining pfam_id and NAME since these are redundant), but summarize the output more simply for use in the later steps. I would suggest combining the data into one row per BS_ID with columns: `TP53_status`, `TP53_CN`, `ploidy`.

>Loss are distributed as , for next steps we would need to check for double mutation so are 1 copy loss ok to keep for next steps or should we only look for more than 1 copy loss :

I think we might need to state this differently. The way I am thinking about this is more: how many copies (alleles) are being retained, knowing that both alleles are required for function. Based on this, I suggest we create a subset of TP53 loss as our "high-confidence" calls, which would be CN <= 1. This looks to be 56 patients. 

Then, will you plot TP53 scores (violin with jitter?) by samples with copy numbers of 0, 1, 2, 3? Of course, this is not the only alteration (we have SNV and LFS to consider), but perhaps we can be assured that those with 0 or 1 copy tend to have higher scores?

I would remove the `tp53-subset` folder and put all of your output tables into `results`.

kgaonkar6:
@jharenza I incorporated the changes you requested above and also resolved the conflict with master.

jharenza (approved):
@kgaonkar6 this looks great and the "high-confidence" CNVs of 0 or 1 copy TP53 check out!

I just ask that you add some comments under the last header of your notebook re: the plots.

"Do samples with copy numbers of 0 and 1 copy of TP53 tend to have higher scores? Yes"

analyses/tp53_nf1_score/README.md
```diff
@@ -21,14 +21,20 @@ The analysis can be run with the following (assuming you are in the root reposit
...
 _NF1_ positive examples are additionally filtered to remove missense mutations, as these are not annotated with OncoKB ([#381 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/381#issuecomment-570748578)).
 
 `01-apply-classifier.py` produces  `results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv`  and `results/pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv`, which contains all 3 classifier scores for the stranded data and for shuffled stranded (e.g., random) data.
 
+`02-qc-rna_expression_score.Rmd` here expression of TP53 gene was compared to TP53 classifier score but we didn't find any correlation between TP53 expression is supported by tp53 inactivation score.
+
+`03-tp53-cnv-loss-domain.Rmd` here copy_number of consensus CNV reginos overlapping TP53 were compared to TP53 classifier score. We find TP53<=1 is supported by higher TP53 classifier score so we only keep TP53 <= 1 as high confidence TP53 loss biospecimen list.
```
jharenza, line 34:
```suggestion
`03-tp53-cnv-loss-domain.Rmd` here copy_number of regions overlapping TP53 functional domains were compared to TP53 classifier score. We find tumors with TP53 copies <=1 have higher TP53 classifier scores, so we only retain biospecimens with <= 1 copy TP53 as high confidence TP53 loss. 
```

analyses/tp53_nf1_score/README.md
```diff
@@ -21,14 +21,20 @@ The analysis can be run with the following (assuming you are in the root reposit
...
-#### Output
+### Order of analysis
 
 `00-tp53-nf1-alterations.R` produces `TP53_NF1_snv_alteration.tsv`, which contains information about the presence or absence of coding SNVs in _TP53_ and _NF1_ for the purpose of evaluating the classifier results.
 For evaluation purposes, a coding SNV 1) is found in a CDS region and 2) is not a silent mutation or in an intron as indicated by the `Variant_Classification` column of the consensus mutation file.
 _NF1_ positive examples are additionally filtered to remove missense mutations, as these are not annotated with OncoKB ([#381 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/381#issuecomment-570748578)).
 
 `01-apply-classifier.py` produces  `results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv`  and `results/pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv`, which contains all 3 classifier scores for the stranded data and for shuffled stranded (e.g., random) data.
 
+`02-qc-rna_expression_score.Rmd` here expression of TP53 gene was compared to TP53 classifier score but we didn't find any correlation between TP53 expression is supported by tp53 inactivation score.
```
jharenza, line 32:
```suggestion
`02-qc-rna_expression_score.Rmd` here expression of TP53 gene was compared to TP53 classifier score. We didn't find a strong correlation between TP53 expression and TP53 inactivation score, thus, expression and classifier score together cannot predict function.
```

jharenza (approved):

PR #847, opened 2020-11-20
#790 Part2: adding Fusion based subtyping for LGAT

kgaonkar6:
⚠️ #842 needs to be merged before this PR, 02-subset-fusion-files-LGAT.R and output can be reviewed independently though

### Purpose/implementation Section
LGAT subtyping is being revamped as per issue. I'm diving into this with staggered PRs per alteration as:

- [x] SNV #842 
- [x]  Fusion #847
- [x]  CNV #848 

#### What scientific question is your analysis addressing?

As per [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/790) we will be subtyping LGAT based on fusion  in the following genes:

 - LGG, KIAA1549-BRAF
contains KIAA1549-BRAF fusion

 - LGG, other MAPK
contains non-canonical BRAF fusion other than KIAA1549-BRAF
contains RAF1 fusion

- LGG, RTK
harbors a fusion in ALK, ROS1, NTRK1, NTRK2, or NTRK3 or
harbors a PDGFRA fusion

- LGG, FGFR
harbors FGFR1-TACC1 fusion 
harbors FGFR1 or FGFR2 fusions

- LGG, MYB/MYBL1
harbors either a MYB-QKI fusion or other MYB or MYBL1 fusion

#### What was your approach?
- I used a [list](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/lgat_add_subtyping_Fus/analyses/molecular-subtyping-LGAT/input/fusionOI_list.json) of genes to look for in presence of fusion or fused genes per subtype. 
- I'm using file from lgat fusion-summary , #830 checks for kinase overlap and reciprocal fusion for LGAT biospecimen fusions.

The fusion status per subtype is saved as lgat-subset/LGAT_fusion_subset.tsv

  | Kids_First_Biospecimen_ID | KIAA_BRAF_fus | MAPK_fus | RTK_fus | FGFR_fus | MYB_fus
-- | -- | -- | -- | -- | -- | --

#### What GitHub issue does your pull request address?
#790 

#### Which areas should receive a particularly close look?
We will have to review/push through #842 since SNV subtyping was done in the previous PR.

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
lgat-subset/LGAT_fusion_subset.tsv

#### What is your summary of the results?
[...]

PR #842, opened 2020-11-16
#790 Part1: adding new SNV subtypes for LGAT

kgaonkar6:
### Purpose/implementation Section

LGAT subtyping is being revamped as per [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/790). I'm diving into this with staggered PRs per alteration as:

- [x] SNV #842 

- [x] Fusion #847 

- [x] CNV #848 

#### What scientific question is your analysis addressing?
As per [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/790) we will be subtyping LGAT based on SNV in the following genes:

 - LGG, NF1
somatic loss of NF1 via either missense, nonsense mutation 

 - LGG, BRAF V600E
contains BRAF V600E or V599 SNV or non-canonical BRAF alterations such as p.V600ins or p.D594N

 - LGG, other MAPK
contains KRAS, NRAS, HRAS, MAP2K1, MAP2K2, MAP2K1, ARAF SNV or indel

 - LGG, RTK
harbors a MET SNV 
harbors a KIT SNV or
harbors a PDGFRA SNV 

- LGG, FGFR
harbors FGFR1 p.N546K, p.K656E, p.N577, or p. K687 hotspot mutations or

- LGG, IDH
harbors an IDH R132 mutation

 - LGG, H3.3
harbors an H3F3A K28M or G35R/V mutation

 -  LGG, H3.1
harbors an HIST1H3B K28M
harbors and HIST1H3C  K28M

#### What was your approach?
I used a [list](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/lgat_add_subtyping_SNV/analyses/molecular-subtyping-LGAT/input/snvOI_list.json) of genes to look for in consensus SNV per subtype ( with additional information about hotspots, canonical mutations ). 

The mutation status per subtype is saved as in lgat-subset/LGAT_snv_subset.tsv : 

  | Tumor_Sample_Barcode | BRAF_V600E_mut | FGFR_mut | IDH_mut | H3F3A_mut | HIST1H3B_mut | HIST1H3C_mut | MAPK_mut | RTK_mut | NF1_mut
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

#### What GitHub issue does your pull request address?
#790 

#### Which areas should receive a particularly close look?
 - I wanted to save the alterations as a list since there are different conditions per gene and subtype , let me know if you prefer some other way to organize the genes/mutations lists.
[...]

Issue #790, opened 2020-09-23
Updated analysis: LGAT - add additional subtypes


jharenza:
#### What analysis module should be updated and why?
[LGAT subtyping
](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-LGAT)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
From conversation with Cassie Kline, @adamcresnick, and inspired by this [pediatric LGG review](https://actaneurocomms.biomedcentral.com/articles/10.1186/s40478-020-00902-z) and this [pLGG review](https://www.sciencedirect.com/science/article/abs/pii/S1535610820301513), I am updating the subtypes we should add for LGAT tumors. <s>However, I am marking this as `in progress` because I want to discuss with @adamcresnick. I believe there is more about LGAT in the WHO 2016 book (rather than the 2016 publication [here](https://braintumor.org/wp-content/assets/WHO-Central-Nervous-System-Tumor-Classification.pdf)) of which I do not have a copy.</s>

With #631, we currently have the following subtypes for LGAT:
- LGG BRAF fusion
- LGG BRAF V600E
- LGG BRAF WT 

From the Ryall et al., 2020 paper above:
[...]

PR #848, opened 2020-11-20
#790 Part3: adding CNV based subtyping for LGAT

kgaonkar6:
⚠️ #842 #847  needs to be merged before this PR, 03-subset-cnv-files-LGAT.Rmd and output can be reviewed independently though

LGAT subtyping is being revamped as per issue. I'm diving into this with staggered PRs per alteration as:

- [x] SNV #842 
- [x]  Fusion #847  
- [x]  CNV #848 

As per [issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/790) we will be subtyping LGAT based on CNV in the following genes:
 - LGG, FGFR
 harbors FGFR1 TKD (tyrosine kinase domain tandem duplication)
 
 - LGG, CDKN2A/B
 harbors focal CDKN2A and/or CDKN2B deletion

#### What was your approach?
I used 
- manta pbta-sv-manta.tsv.gz to capture tandem duplication
- consensus seg pbta-cnv-consensus.seg.gz to capture consensus seg CNV region
- consensus annotated consensus_seg_annotated_cn_autosomes.tsv.gz to capture annotated genes per consensus CNV with ploidy 

I only keep CNV and SV that overlap within FGFR1 gene since the tandem duplication event is a exon level duplication.

CNV status is saved in lgat-subset/LGAT_cnv_subset.tsv as: 

  | biospecimen_id | FGFR_DUP_TANDEM | FGFR_DUP | CDKN2A_DEL | CDKN2B_DEL
-- | -- | -- | -- | -- | --

#790 

For CDKN2A/B I only used copy_number <=1 
```r
> cnv_focal_anno_lgat %>%
   dplyr::filter(gene_symbol %in% c("CDKN2A","CDKN2B")) %>%
   dplyr::pull(copy_number) %>%
   table()

  0   1   2   3 
  2  16 134   6 
```

For FGFR1 I looked for "tandem duplication" in manta file where it's annotated as "DUP:TANDEM" within FGFR1 gene region that also overlaps tyrosine kinase domain

biospecimen_id<chr> | ALT<chr> | NAME<chr> | hgnc_symbol<chr> | FGFR_DUP_TANDEM<chr>
-- | -- | -- | -- | --
BS_3108RS7G | <DUP:TANDEM> | Pkinase_Tyr | FGFR1 | Yes
BS_ETTZM63Y | <DUP:TANDEM> | Pkinase_Tyr | FGFR1 | Yes
BS_R726FKHP | <DUP:TANDEM> | Pkinase_Tyr | FGFR1 | Yes
BS_MW6YMRBJ | <DUP:TANDEM> | Pkinase_Tyr | FGFR1 | Yes

Similar to the SV I also filtered consensus seg CNV within FGFR1 and overlaps tyrosine kinase domain with copy_number >2

biospecimen_id<chr> | copy_number<dbl> | NAME<chr> | FGFR_DUP<chr> |  
-- | -- | -- | -- | --
BS_4G9Y2VC6 | 3 | Pkinase_Tyr | Yes |  
BS_D6XHKZDZ | 3 | Pkinase_Tyr | Yes |  
BS_M2MSWKBC | 3 | Pkinase_Tyr | Yes |  
BS_MW6YMRBJ | 3 | Pkinase_Tyr | Yes |  
BS_V79WFJ6N | 3 | Pkinase_Tyr | Yes |  

I did look at exons that overlap the manta/ consensus seg calls that are within FGFR1 and it seems 4 biospecimens with tandem duplication and 5 biospecimens with duplication have exon overlap between 10 and 18.

There are a few calls that overlap <10 exons but it seems from the check that all 4 biospecimens with tandem duplication in FGFR1 and 5 biospecimens with duplication in FGFR1 have `atleast 1 SV/CNV` overlap where all exons are between 10 and 18. 
image

Let me know if we need more checks on this

Yes

table

FGFR_DUP_TANDEM =  4 
FGFR_DUP = 5 
BS_MW6YMRBJ is annotated with a tandem duplication(SV) as well as duplication(CNV)

CDKN2A_DEL = 54 
CDKN2B_DEL = 54 

#### Documentation Checklist


analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+
+# plot exon rank for manta and cnv
+ggplot(rbind(data.frame("exon_number"=unlist(overlaps_manta_fgfr_exon$exon_rank),
+                  "type"="tandem_dup"),
+             data.frame("exon_number"=unlist(overlaps_cnv_fgfr_exon$exon_rank),
+                  "type"="cnv_dup"))
+       ,aes(x=exon_number,color=type))+geom_density()
+
+```
+Majority SV and CNV overlap >5 exons
```
jharenza, line 229:
```suggestion
Majority SV and CNV overlap exons downstream of exon 5
```

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+
+any SV overlap which has exon overlap for exon less than 10
```
jharenza, line 231:
```suggestion
any SV overlap which overlaps exons 10 through 18 
```

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+Majority SV and CNV overlap >5 exons
+
+any SV overlap which has exon overlap for exon less than 10 
+```{r}
+overlaps_manta_fgfr_exon[all(overlaps_manta_fgfr_exon$exon_rank>10  &
+                        overlaps_manta_fgfr_exon$exon_rank<18)==TRUE,"Kids.First.Biospecimen.ID.Tumor"] %>% unique()
+
+```
+
+any CNV overlap which ha exon overlap for exon less than 10
```
jharenza, line 238:
```suggestion
any CNV overlap which overlaps exons 10 through 18 
```

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+any CNV overlap which ha exon overlap for exon less than 10 
+```{r}
+
+overlaps_cnv_fgfr_exon[all(overlaps_cnv_fgfr_exon$exon_rank>10 & overlaps_cnv_fgfr_exon$exon_rank<18)==TRUE,"ID"] %>% unique()
+```
+
+All 4 biospecimens with tandem duplication in FGFR1 and 5 biospecimens with duplication in FGFR1 have atleast 1 SV/CNV overlap where all exons are between 10 and 18. 
+
+
+### CDKN deletion
```
jharenza, line 247:
```suggestion
### CDKN2A/B deletion 
```

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+# merge all lgat WGS samples
+lgat_wgs_df %>%
+  left_join(merge_CNV_substype,by=c("Kids_First_Biospecimen_ID"="biospecimen_id")) %>%
+  # change NA to value==No
+  replace_na(list(FGFR_DUP_TANDEM="No",
+                         FGFR_DUP="No",
+                         CDKN2A_DEL="No",
+                         CDKN2B_DEL ="No"
+                         ))%>%
+  write_tsv(file.path(subset_dir,"LGAT_cnv_subset.tsv"))
```
jharenza, line 282:
```suggestion
LGAT_cnv_subset <- lgat_wgs_df %>%
  left_join(merge_CNV_substype,by=c("Kids_First_Biospecimen_ID"="biospecimen_id")) %>%
  # change NA to value==No
  replace_na(list(FGFR_DUP_TANDEM="No",
                         FGFR_DUP="No",
                         CDKN2A_DEL="No",
                         CDKN2B_DEL ="No"
                         ))
#check frequency - does this align with the LGG paper of 9.6%? We get ~3.3%, however, the paper also states that this deletion most frequently co-occurs with BRAF V600E, so if we have a lower prevalence of that SNV, then this could make sense. Let's circle back when we put all of the data together.
CNV_by_part <- LGAT_cnv_subset %>%
  left_join(lgat_specimens_df) %>%
  select(Kids_First_Participant_ID, CDKN2A_DEL) %>%
  distinct()

CNV_by_part %>%
  tally(CDKN2A_DEL == "Yes")/nrow(CNV_by_part)*100
 
  write_tsv(LGAT_cnv_subset, file.path(subset_dir,"LGAT_cnv_subset.tsv"))
```

I wanted to check the frequency of this CDKN2A/B deletion (A/B are co-deleted in these samples; they are right next to each other) with the reported frequency by patient to be sure we were not over-estimating with our CNV results. Feel free to update if you have a better way :)

jharenza (changes requested):
Few update requests, but looks good after that.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,288 @@
...
+cnv_anno <- read_tsv(file.path(root_dir,
+                          "data",
+                          "consensus_seg_annotated_cn_autosomes.tsv.gz"))
+
+# Gene location and domain overlap file 
+bioMartDataPfamFGFR1 <- 
+  readRDS(system.file("extdata", "pfamDataBioMart.RDS", package = "annoFuse")) %>%
+  # to check if biospecimen harbors FGFR1 TKD (tyrosine kinase domain tandem duplication) 
+  dplyr::filter(hgnc_symbol == "FGFR1",
+                DESC == "Protein tyrosine kinase")
```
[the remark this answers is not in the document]
kgaonkar6: Yes it is hg38, added comments to specify input files  and code links.

jharenza (approved):

kgaonkar6:
I think all conflicts are now resolved. The conflicts mostly arose because I'd staggered this PR from Part1-2 LGAT which have now been merged with additional comments/edits that needed to  bee propagated in this PR.

jaclyn-taroni:
@cansavvy has done more with `GenomicRanges` recently than I have, so I am going to remove myself from review and request her instead!

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+overlaps_manta_fgfr_df <- data.frame(
+  biospecimen_id = overlaps_manta_fgfr$Kids.First.Biospecimen.ID.Tumor,
+  ALT = overlaps_manta_fgfr$ALT,
+  NAME = overlaps_manta_fgfr$NAME,
+  hgnc_symbol = overlaps_manta_fgfr$hgnc_symbol,
+  # annnotation for dup tandem
+  FGFR_DUP_TANDEM = "Yes",
+  stringsAsFactors = FALSE
+) %>%
+  dplyr::distinct()
```
cansavvy, line 154: It doesn't look like this distinct() is doing anything. Is there a particular reason its here?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+4 biospecimens have tandem duplication that overlap tyrosine kinase domain in FGFR1
+
+
+### Distribution of copy number
+CNV that are within FGFR1 gene region only and overlap "Protein tyrosine kinase" domain in FGFR1 in biospecimens
+
+```{r}
+
+# gene region overlap
+fgfr_gene = GenomicRanges::GRanges(seqnames=paste0("chr",domain_gr$chromosome_name),
```
cansavvy, line 168:
```suggestion
fgfr_gene <- GenomicRanges::GRanges(seqnames=paste0("chr",domain_gr$chromosome_name),
```

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+          ranges=IRanges::IRanges(start = domain_gr$gene_start, end = domain_gr$gene_end))
+
+# subset cnv+domain grange to be within FGFR1 gene region
+cnv_fgfr_gr <- IRanges::subsetByOverlaps(cnv_gr, fgfr_gene,type="within")
```
cansavvy, line 172: Same kinds of clarification questions here since this and the next steps are important. Can you explain the goal here. You are intending to retrieve only the CNVs that are completely contained within the fgfr gene? Can you explain why you only want genes that are completely within the gene? What about if a gene is like 90% overlapped do we still want it? If so, this won't be kept so we should switch strategies.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+
+# overlap consensus seg with exon
+overlaps_fgfr_cnv <- IRanges::mergeByOverlaps(cnv_fgfr_gr,domain_gr)
```
cansavvy, line 175: Same types of clarifications as before. Can you explain what you are intending to get out of this step?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+```{r}
+# Define the annotations for the hg38 genome
+txdb <- GenomicFeatures::makeTxDbFromGFF(
+    file = file.path(root_dir,"data",
+                     "gencode.v27.primary_assembly.annotation.gtf.gz"),
+    format = "gtf"
+  )
+# get exons and exon_id
+tx_exons<-GenomicFeatures::exons(txdb, 
+                                 columns = list("exon_rank"),
```
cansavvy, line 216: What's this argument for?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+
+```{r}
+# Define the annotations for the hg38 genome
+txdb <- GenomicFeatures::makeTxDbFromGFF(
+    file = file.path(root_dir,"data",
+                     "gencode.v27.primary_assembly.annotation.gtf.gz"),
+    format = "gtf"
+  )
+# get exons and exon_id
+tx_exons<-GenomicFeatures::exons(txdb,
```
cansavvy, line 215: It looks like you've filtered this to only to this particular FGFR1 transcript, so the name `tx_exons` is a bit misleading since its only got FGFR1 exons for `ENST00000447712.6`.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+txdb <- GenomicFeatures::makeTxDbFromGFF(
+    file = file.path(root_dir,"data",
+                     "gencode.v27.primary_assembly.annotation.gtf.gz"),
+    format = "gtf"
+  )
+# get exons and exon_id
+tx_exons<-GenomicFeatures::exons(txdb, 
+                                 columns = list("exon_rank"),
+                                 # genocode canonical transcript https://genome.ucsc.edu/cgi-bin/hgGene?db=hg38&hgg_gene=FGFR1
+                                 filter=list(tx_name = "ENST00000447712.6"))
```
cansavvy, line 218: Is there a particular reason that only this transcript for FGFR1 is being used? It looks like there's a lot of them. Is it because this one contains all the exons?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+             data.frame("exon_number"=unlist(overlaps_cnv_fgfr_exon$exon_rank),
+                  "type"="cnv_dup"))
+       ,aes(x=exon_number,color=type))+geom_density()
+
+```
+Majority SV and CNV overlap exons downstream of exon 5
+
+any SV overlap which overlaps exons 10 through 18 
+```{r}
+overlaps_manta_fgfr_exon[all(overlaps_manta_fgfr_exon$exon_rank>10  &
```
cansavvy, line 243: Because you are using `>` and `<` this would technically not include 10 and 18 but actually 11 - 17. I think you want these to be `>=` and `<=`.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+                        overlaps_manta_fgfr_exon$exon_rank<18)==TRUE,"Kids.First.Biospecimen.ID.Tumor"] %>% unique()
+
+```
+
+any CNV overlap which overlaps exons 10 through 18 
+```{r}
+
+overlaps_cnv_fgfr_exon[all(overlaps_cnv_fgfr_exon$exon_rank>10 & overlaps_cnv_fgfr_exon$exon_rank<18)==TRUE,"ID"] %>% unique()
```
cansavvy, line 251: Same comment here.

cansavvy (commented):
This looks like a good start! I have some comments and questions. Some of my questions are trying to double check that the steps being employed here are what we expect/what we want.

kgaonkar6:
Thanks for the review @cansavvy! 

For the FGFR1 exon overlap issue I looked up the #790, I can't clarify if we are looking for overlap including 10 and 18 exons or just between those 2. This figure in 1 of the references says intron of 10 and part of 18 https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3951336/figure/F3/?report=objectonly. @jharenza for clarification?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: ENST00000447712.6 seems to be the canonical transcript as per UCSC and also has all exons I'd decided to use this ttranscript.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,304 @@
...
+  dplyr::rename(chr = SV.chrom, start = SV.start, end = SV.end,
+                ALT = ALT) %>%
+  # format with chr to match to cnv
+  dplyr::mutate(chr=paste0("chr",chr)) %>%
+  # make genomic ranges
+  GenomicRanges::makeGRangesFromDataFrame(keep.extra.columns = TRUE,
+                                          starts.in.df.are.0based = FALSE)
+
+# overlap manta and domain
+overlaps_manta <- IRanges::mergeByOverlaps(manta_gr, domain_gr)
```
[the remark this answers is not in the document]
kgaonkar6: > `mergeByOverlaps` keeps both the original ranges from manta_gr and domain_gr that overlap with each other

I had intended to keep columns from both domain and manta so that we can save ALT from `manta_gr` and (domain)NAME from `domain_gr` in `overlaps_manta_fgfr_df` to do checks of the type of ALT that overlap the domain in the output printed in the notebook.

I'll change it to `subsetOverlaps` to make it a shorter code since domain_gr only has "Pkinase_Tyr" domain forr FGFR1 anyways from the filtering in previous chunks.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: no, I'll remove that line,

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: I believe the condition was to keep CNVs that overlap the boundaries of FGFR1 exon 10 to 18 so I've added the type="within".

cansavvy: My point is about using `type = "within"` that means that: `the query interval must be wholly contained within the subject interval`https://www.rdocumentation.org/packages/GenoGAM/versions/1.0.2/topics/subsetByOverlaps%2CGenomicTiles%2CGRanges-method

Which seems pretty stringent and I'm not sure its what we want? If the concern is making sure there is enough overlap, we should instead use `maxgap`.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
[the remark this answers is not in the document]
cansavvy: My point here was trying to clarify what types of sequences are you looking for this to return? OR, do you not care about the sequences themselves, but more just want to know about which rows in `manta_gr` and `domain_gr` overlap without caring much about the sequences that are returned? 

But it looks like in the next steps you mainly care about the information and not the sequences themselves so it probably doesn't matter. (Unless you are returning to use `overlaps_manta` later -- which I will have to check).

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: I think I figured this out later. These are the exon numbers?

kgaonkar6: yes the number values in exon_rank are the exon numbers for the transcript.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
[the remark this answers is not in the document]
kgaonkar6: I don't use sequences from `overlaps_manta` mainly looking for rows in manta that overlap domian, `subsetByOverlaps` didn't change the resulting output that much so keeping the `subsetByOverlaps` suggestion.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: ENST00000447712.6 seems to be the canonical transcript as per UCSC and also has all exons I'd decided to use this ttranscript

cansavvy: My point is here, maybe rename this so its clear that its not all `tx_exons`, its actually `fgfr1_exons`.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: Can you add a comment that says that thats what that is extracting?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: I think we do want the query(CNV) to be wholly within the subject(gene) . Only with this type of overlap I would be able to get CNVs overlapping exactly/parts of the exon 10 to 18 (kinase) region, right? Am I missing something?

```r
# 1. subset cnv+domain grange to be within FGFR1 gene region
cnv_fgfr_gr <- IRanges::subsetByOverlaps(cnv_gr, fgfr_gene,type="within")

# 2. subset CNV granges within FGFR1 to only kinase domain overlapping rows
overlaps_fgfr_cnv <- IRanges::subsetByOverlaps(cnv_fgfr_gr,domain_gr) 

```

kgaonkar6: With the "within" condition we are able to gather CNVs that exactly overlap the 10-18 exon regions.
image

Without the "within" condition it seems we gather CNVs that span the complete gene body and not just the 10-18 exon region of interest.
image

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: ah ok, will do

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: added a comment

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: thanks, updated the variable name

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: If you don't use within, the default is `type="any"` which means any overlap (even an overlap of 1bp) will be included. Hence why my suggestion includes to set what amount of overlap you think is reasonable (This is what I meant when I referenced `maxgap` but `minOverlap` -- they are somewhat two sides of the same coin). You would have to use the `minOverlap` argument to specify a certain amount of overlap that you think is deemed enough overlap to be certain the CNV is affecting the exon. 

But using type="within" as is right now means that 100% of base pairs in that exon must be covered by the CNV. 

What if there is one base pair in the exon not included in the CNV? To me, It seems like that exon-CNV overlap should still be considered an overlap, but using `type="within"` means that in that scenario, it would not be considered an overlap. 

This is an odd overlapping scenario to interpret `type="within"` for in my opinion because overwhelmingly the CNVs are much longer than the exons meaning that this default rule from their docs applies: `For within, it is the maximum amount by which the subject may be wider than the query.`

Note that I do not know how much scenarios like `What if there is one base pair in the exon not included in the CNV` that I mentioned above occur. But it seems worth looking into.

cansavvy: Basically my fundamental question here is: How many base pairs does our CNVs and the FGFR1 exons need to overlap for us to care about it?

jharenza: hi @cansavvy and @kgaonkar6! Thanks for reviewing! What we want to do is find all instances of the tyrosine kinase domain (TKD, exons 10-18) that are duplicated in tandem (right next to the other). I do think that the first analysis @kgaonkar6 had performed gets at this information, especially after seeing the image above with the large CNVs spanning the entire gene. This would be a gene-gene duplication, rather than the smaller TKD itself duplicated).

jharenza:
> For the FGFR1 exon overlap issue I looked up the #790, I can't clarify if we are looking for overlap including 10 and 18 exons or just between those 2. This figure in 1 of the references says intron of 10 and part of 18 https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3951336/figure/F3/?report=objectonly. @jharenza for clarification?

Unless they have coordinates, I think it is safe to do exons 10-18.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: For the specific number of base overlap, if the CNV has 100 bp we want all 100 bp to overlap 100 bps "within" the *gene*. These tandem duplication of a group of exons will thus be smaller than the gene length, which I believe we can capture with :

```r
# 1. subset cnv+domain grange to be within FGFR1 gene region
cnv_fgfr_gr <- IRanges::subsetByOverlaps(cnv_gr, fgfr_gene,type="within")

```

Not sure how if we can specify `maxgap` beforehand, let me know what you have in mind with that parameter. Thanks!

kgaonkar6:
> Unless they have coordinates, I think it is safe to do exons 10-18.

Thanks for the input @jharenza ! I've added >= 10 and <= 18 to include 10 and 18.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: I am not 100% certain how the coordinates from the CNV data are obtained, but this analysis is looking at whether coordinates overlap, and not whether the base pairs sequences of the gene itself is duplicated within the CNV. So I'm not saying that this approach is wrong, but we might want to be careful about what information we can specifically say from it.

cansavvy: Okay. I had to think about this a bit and go back to read a bit about the CNV data. I think what you are saying makes sense. And if we want to go with 100% overlap by using type="within" that seems fine.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+fgfr_gene <- GenomicRanges::GRanges(
+  seqnames = paste0("chr", domain_gr$chromosome_name),
+  ranges = IRanges::IRanges(start = domain_gr$gene_start, end = domain_gr$gene_end)
+)
+
+# subset cnv+domain grange to be within FGFR1 gene region
+cnv_fgfr_gr <- IRanges::subsetByOverlaps(cnv_gr, fgfr_gene, type = "within", maxgap = )
+
+# overlap consensus seg with exon
+overlaps_fgfr_cnv <- IRanges::subsetByOverlaps(cnv_fgfr_gr, domain_gr)
```
cansavvy, line 193: So is this overlap showing whether or not the previous overlap contains an protein tyrosine kinase exon?

If yes, can we make the note more specific?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+  ) %>%
+  dplyr::select(-num.mark, -seg.mean) %>%
+  # make genomic ranges
+  GenomicRanges::makeGRangesFromDataFrame(
+    keep.extra.columns = TRUE,
+    starts.in.df.are.0based = FALSE
+  )
+
+# Genomic range for gene location and domain overlap file
+domain_gr <- bioMartDataPfamFGFR1 %>%
```
cansavvy, line 117: So this range only contains the Protein tyrosine kinase exon, is that correct?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+  # format with chr to match to cnv
+  dplyr::mutate(chr = paste0("chr", chr)) %>%
+  # make genomic ranges
+  GenomicRanges::makeGRangesFromDataFrame(
+    keep.extra.columns = TRUE,
+    starts.in.df.are.0based = FALSE
+  )
+
+# overlap manta and domain
+overlaps_manta <- IRanges::subsetByOverlaps(manta_gr, domain_gr)
```
cansavvy, line 158: So this one is by default set to "any" overlap, which I don't think is what we want. We probably at the least should set a "minOverlap" parameter. (And we can probably base this on how long `domain_gr` is).

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+
+# filter for fgfr only overlap
+overlaps_manta_fgfr <- overlaps_manta[which(overlaps_manta$Gene.name == "FGFR1" &
+  # filter for tandem duplication
+  overlaps_manta$ALT == "<DUP:TANDEM>"), ]
```
cansavvy, line 163: If we only care about tandem duplicates, it might make sense to put a filter for tandem duplicates before we even make the GenomicRange for the SV data. This has the benefit of making it easier to read and probably slightly more computationally efficient (though this second part we aren't really encountering a problem with as of now).

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+
+```{r}
+
+# Genomic range for manta file
+manta_gr <- manta_lgat %>%
+  # format
+  dplyr::rename(
+    chr = SV.chrom, start = SV.start, end = SV.end,
+    ALT = ALT
+  ) %>%
```
cansavvy, line 148: Related to another comment, if we only care about tandem duplicates, let's filter to those here before we even make a GenomicRange. No need to make more comparisons than needed.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+```{r}
+
+# gene region overlap
+fgfr_gene <- GenomicRanges::GRanges(
+  seqnames = paste0("chr", domain_gr$chromosome_name),
+  ranges = IRanges::IRanges(start = domain_gr$gene_start, end = domain_gr$gene_end)
+)
+
+# subset cnv+domain grange to be within FGFR1 gene region
+cnv_fgfr_gr <- IRanges::subsetByOverlaps(cnv_gr, fgfr_gene, type = "within", maxgap = )
```
cansavvy, line 190: In the context of `within`: `maxgap... is the maximum amount by which the subject may be wider than the query.` So is there an extent to which the CNV repeated region would be so large that we should treat it differently? If yes, then we should set maxgap, if no, then I think we can leave off maxgap.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: Is there a reason we can't do this in one overlap? I think we should be able to do that. It doesn't look like you return to using `cnv_fgfr_gr` for anything.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+
+# overlap consensus seg with exon
+overlaps_fgfr_cnv <- IRanges::subsetByOverlaps(cnv_fgfr_gr, domain_gr)
+
+# filter for fgfr only overlap
+overlaps_fgfr_cnv <- overlaps_fgfr_cnv[which(
+  # filter for duplication
+  overlaps_fgfr_cnv$copy_number > 2
```
cansavvy, line 198: Again, this filter might be better to do to the CNV file itself.  Unless there are instances where we care about copy numbers less than 2?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+) %>%
+  dplyr::distinct() %>%
+  # keep copy_number > 2
+  dplyr::mutate(FGFR_DUP = if_else(copy_number > 2, "Yes", "No"))
+
+overlaps_cnv_df
+```
+5 samples have copy_number changes > 2 that overlap with tyrosine kinase domain
+
+Are there overlap in SV and CNV calls?
```
cansavvy, line 215: This is kind of a its own question, might be good to be its own chunk.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+  ),
+  aes(x = exon_number, color = type)
+) +
+  geom_density()
+```
+Majority SV and CNV overlap exons downstream of exon 5
+
+any SV overlap which overlaps exons 10 through 18 
+```{r}
+overlaps_manta_fgfr_exon[all(overlaps_manta_fgfr_exon$exon_rank >= 10 &
```
cansavvy, line 270: One thing I'm not clear on here: is there a reason we want to exclude if more exons than 10 -18 are duplicated? aka if a CNV or SV has 8- 20 duplicated we would be excluding that here, is there a reason we only want 10 -18 but not more or less than that?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
```diff
@@ -0,0 +1,329 @@
...
+  overlaps_manta_fgfr_exon$exon_rank <= 18) == TRUE, "Kids.First.Biospecimen.ID.Tumor"] %>% unique()
+```
+
+any CNV overlap which overlaps exons 10 through 18 
+```{r}
+
+overlaps_cnv_fgfr_exon[all(overlaps_cnv_fgfr_exon$exon_rank >= 10 & overlaps_cnv_fgfr_exon$exon_rank <= 18) == TRUE, "ID"] %>% unique()
```
cansavvy, line 277: Same question as above.

cansavvy (commented):
Okay @kgaonkar6 I think we are close! I just have a few more questions. Thanks!

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: updated these to >= and <=  to capture 10 and 18 exons

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: updated these to >= and <=  to capture 10 and 18 exons

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: >So is this overlap showing whether or not the previous overlap contains an protein tyrosine kinase exon?
Right ^ I can update the comment to specify the same.

I  didn't quite figure out how to put it in 1 step such that we can find CNVS that overlaps domain_gr and are also fully within FGFR gene only (ie. CNV is smaller than FGFR gene body). 
I guess a maxgap of min[( domain_start - gene_start ), (gene_end - domain_end)] might work, checking ..

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: Correct, I filter to keep only protein tyrosine kinase *domain* region.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: good idea, I can add that param

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: 👍 cool, I'll implement the filter in manta granges itself

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: ok will implement in CNV file filter

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: This and the 1 below were just checks to see if the CNV/SV  regions captured from the steps above overlap 10-18 exon region as we see in previous references mentioned in #790. The CNV/SVs do actually cover regions but 2 bs_ids overlap 9 exon as well. 

Instead of these checks I've just used the ggplot to specify the same information.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: ah yes this was a typo from testing. I've implemented a type="within" and minOverlap= domain_gr length which I think will also make this a 1 step subsetByOverlaps command.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: thanks, implemented the tandem_dup filter in manta_gr itself.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: I update the 2 step above to 1 step: 

```r
# subset cnv+domain grange to be within FGFR1 gene region 
cnv_fgfr_gr <- IRanges::subsetByOverlaps(
  cnv_gr,
  fgfr_gene,
  # within FGFR1 gene
  type = "within",
  # has atleast min overlap of domain region
  minoverlap = width(ranges(domain_gr)))

```

So this way the subset if within "FGFR1" gene body and has minOverlap domain_gr. We can argue that the minOverlap region could also be outside the actual domain_gr , but this is the only way i could think of making it 1 step. Let me know if you had any other way.

kgaonkar6:
Thanks for the additional review @cansavvy  and validation @jharenza.

I just wanted to mention, other than review updates there is a code update for the fgfr1_exons. Earlier filtering exons for the canonical transcript was *also* keeping exon rank(order) of other transcripts as well. So know I have a very specific filtering step in: 
https://github.com/kgaonkar6/OpenPBTA-analysis/blob/c9151f16759c846fd9c60ac3963bd5b9bab76f06/analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd#L232-L243

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: > CNVS that overlaps domain_gr and are also fully within FGFR gene only

Does this mean you do not want to identify a CNV that is longer than the FGFR gene? Why is that? But yes you could should be able to use maxgap for that.

type="within" (based on my reading of the documentation) won't exclude things based on the CNV being longer than the FGFR gene either. So even if you keep both steps, I think you'd still need to specify a maxgap parameter.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
jharenza: I think this sounds resolved, right @cansavvy ?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
jharenza: This one also resolved @cansavvy ?

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
kgaonkar6: >Does this mean you do not want to identify a CNV that is longer than the FGFR gene? Why is that? But yes you could should be able to use maxgap for that.

CNVs need to be smaller that FGFR1 gene body, actually we only want to eventually retain CNVs that are duplication of exon10 to exon18 Or the kinase domain because duplication of this specific region defines this subtype.  

>type="within" (based on my reading of the documentation) won't exclude things based on the CNV being longer than the FGFR gene either. So even if you keep both steps, I think you'd still need to specify a maxgap parameter. 

Does a maxgap of gene body minus domain sound like a good maxgap value? It seems to give the same 5 bs ids as overlap

cansavvy: Okay. Then yes, I think that plan makes sense.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: Yes. Sounds good.

analyses/molecular-subtyping-LGAT/03-subset-cnv-files-LGAT.Rmd
cansavvy: Sounds good.

cansavvy (approved):
@kgaonkar6! This looks good! Looks like all the changes we discussed have been made. I like the move to the plot for the check. :+1: 🚀

Issue #853, opened 2020-11-24
Updated analysis: remove BRAF from HGG subtyping


jharenza:
#### What analysis module should be updated and why?
HGG subtyping 

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Step 1a from [this issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/807) (Remove BRAF from subtyping) should be done before #784  

#### What input data should be used? Which data were used in the version being updated?
```
pbta-snv-consensus-mutation.maf.tsv.gz
```

#### When do you expect the revised analysis will be completed?
tuesday 11/24

#### Who will complete the updated analysis?
@kgaonkar6

kgaonkar6:
closed with #856

Issue #807, opened 2020-10-14
Updated analysis: HGG subtyping to remove BRAF and include TP53


jharenza:
#### What analysis module should be updated and why?
[molecular-subtyping-HGG](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-HGG)

1. Per discussion with @adamcresnick, we should remove the _BRAF_ V600E subtyping from HGG and replace with _TP53_ subtypes for HGG, as _TP53_ is the top co-occurring gene with H3 mutations in DMGs. Neither are currently in WHO, but the latter may be able to inform clinical decision making. 
2. We should remove LGG samples which contain histone mutations during the initial search for samples with histone mutations, since there have been recent reports of LGG with these mutations. (See literature below).

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Step 1 - cleanup
- Remove the _BRAF_ V600E subtypes - This is being done in #856 
- <s>Remove LGG tumor samples which have histone mutations.</s> - This is being done in #851 

Step 2 - annotate as _TP53_ altered if: 
- There are two alterations (SNV+SNV, CNV+SNV)
- There is one alteration + cancer_predispositions == "Li-Fraumeni syndrome"
- There are no TP53 alterations, but cancer_predispositions == "Li-Fraumeni syndrome" **_and_** _TP53_ classifier score for matched RNA-Seq > 0.5
- There are one of _TP53_ activating mutations R273C and R248W (see literature below)

Step 3 - assess and potentially annotate as _TP53_ altered if:
- There is an SNV  **or** Focal CNV loss **_and_** _TP53_ classifier score for matched RNA-Seq > 0.5
- There is an SNV **or** Focal CNV gain **_and_** _TP53_ classifier score for matched RNA-Seq < 0.5 (indicating activating mutation) **_and_** _TP53_ expression is high

#### What input data should be used? Which data were used in the version being updated?
```
pbta-snv-consensus-mutation.maf.tsv.gz
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/results/pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv
```
[...]

PR #796, opened 2020-10-07
Update subtyping cell lines hgg

kgaonkar6:
#### What scientific question is your analysis addressing?
Currently we don't have molecular subtyping for cell lines , which are all HGG.   Main issue was that the DNA and RNA samples couldn't be uniquely matched so were excluded in previous runs of the subtyping;  details have been discussed #610  and #510

#### What was your approach?
In this PR I'm removing `composition == "Solid Tissue" ` filters in the code to include cell-lines and also added an input file with the cell-line composition to help uniquely  match DNA and RNA samples in molecular-subtyping-HGG module.

#### What GitHub issue does your pull request address?
#509 #610

#### Which areas should receive a particularly close look?
I added a new file from #610 to add the cell-line composition ; do you want this to be added some other way? 
There are other cell-lines ( from PNOC003 ) not mentioned in #610 which are added 

Kids_First_Biospecimen_ID | sample_id
-- | --
BS_DVDT4VXQ |	7316-3234
BS_6JBE0947 |	7316-4448
BS_4DQAQFQH |	7316-4446
BS_QZRP3NSG |	7316-4447
BS_YZD4SSMA |	7316-3240
BS_VXDGXQKZ |	7316-3235

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
tables mostly in results folder have been updated with cell-lines BS ids

#### What is your summary of the results?
32 CBTTC and 6 PNOC003 cell line samples have been added to the analysis

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #856, opened 2020-11-24
Removing BRAF subtyping for HGG

kgaonkar6:
Step 1a from [this issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/807) (Remove BRAF from subtyping) should be done before we merge #784  

#### What was your approach?
Removed BRAF V600E subtyping steps from 04-HGG-molecular-subtyping-mutation.Rmd and BRAF V600E mutant checks from 07-HGG-molecular-subtyping-combine-table.Rmd where subtypes are combined and assigned to sample_ids

#853

In 07-HGG-molecular-subtyping-combine-table.Rmd where subtypes are combined and assigned to sample_ids I removed the checks for BRAF V600E do we still want to check if BRAF V600E does exist for mutual exclusivity checks just not assign the molecular subtype as "BRAF V600E"? 

 The plot in 09-HGG-with-braf-clustering.Rmd will have broad_histology/molecular_subtype from v17 I believe for  LGAT, HGAT so it still has the molecular_subtype of BRAF V600E that is being plot in the tsne. 

Should we specifically mention the release data use in this script OR this script needs to be re-run once the histology file is update with v17 molecular_subtyping.

yes

tables

PT_3AWKWXEV,PT_63F63QY0, PT_HGM20MW7,PT_P5HHJJPH,PT_T8V9ES93 and PT_Y8RNS9JE were coded as HGG, H3 wildtype instead of BRAF V600E

#### Documentation Checklist


jharenza:
> The plot in 09-HGG-with-braf-clustering.Rmd will have broad_histology/molecular_subtype from v17 I believe for LGAT, HGAT so it still has the molecular_subtype of BRAF V600E that is being plot in the tsne.

I reviewed this PR and it looks good for removal, but I do think we want to keep the plot labels, since this led to re-classification of a tumor from HGG to LGG. This means we would need to keep the annotation. I wonder if the best way to do this is to break this up into two different modules - one which handles subtyping and one which handles the clustering and BRAF annotation? Thoughts @jaclyn-taroni ?

jaclyn-taroni:
For the scatter plot produced in `09-HGG-with-braf-clustering.Rmd`, I understand that we need the following:

- Histone mutation status for HGAT
- BRAF V600E status for HGAT
- Pathology diagnosis information
- Relevant scores from `analyses/transcriptomic-dimension-reduction/`

I don't think we need to piece together this information in the same manner as what is currently in that notebook, though! Here's a general idea about how we could accomplish the same plot using different information and files:

- Retain the step where we look at BRAF V600E in `04-HGG-molecular-subtyping-mutation.Rmd`, but don't use it at all in `07-HGG-molecular-subtyping-combine-table.Rmd`
- Get the histone mutation status and BRAF V600E information from the `"HGG_cleaned_mutation.tsv"` file from `04`
- Get the pathology dx information from `pbta-histologies.tsv`
- Use the scores from `analyses/transcriptomic-dimension-reduction/` (they may be joined with metadata in there IIRC, we don't want to use that metadata).

If that sounds good/reasonable, we could revert the changes to `04` from this PR but continue not to use the BRAF V600E info in `07-HGG-molecular-subtyping-combine-table.Rmd`.

Thoughts @jharenza and @kgaonkar6 ?

kgaonkar6:
> I don't think we need to piece together this information in the same manner as what is currently in that notebook, though! Here's a general idea about how we could accomplish the same plot using different information and files:
> 
> * Retain the step where we look at BRAF V600E in `04-HGG-molecular-subtyping-mutation.Rmd`, but don't use it at all in `07-HGG-molecular-subtyping-combine-table.Rmd`
> * Get the histone mutation status and BRAF V600E information from the `"HGG_cleaned_mutation.tsv"` file from `04`
> * Get the pathology dx information from `pbta-histologies.tsv`
> * Use the scores from `analyses/transcriptomic-dimension-reduction/` (they may be joined with metadata in there IIRC, we don't want to use that metadata).
> 
> If that sounds good/reasonable, we could revert the changes to `04` from this PR but continue not to use the BRAF V600E info in `07-HGG-molecular-subtyping-combine-table.Rmd`.

This sounds good to me, only thing in addition to the above would be to use DNA BS ids from HGG_cleaned_mutation.tsv to capture potential broad_histology updates in base histology because of H3 mutant detection like for 7316-158 [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/609#issuecomment-602821376) which goes from PNET to HGAT because of h3f3A mutation. Match DNA to RNA to filter scores from analyses/transcriptomic-dimension-reduction/

jharenza:
sounds good, thanks @jaclyn-taroni

kgaonkar6:
In the latest commit, as per @jaclyn-taroni's suggestion I have :

- Reverted 04-HGG-molecular-subtyping-mutation.Rmd to file in master to retain BRAF mutation status
- Get the histone mutation status and BRAF V600E information from the "HGG_cleaned_mutation.tsv" file from 04
- Get the pathology dx to extract LGAT sample_ids information from pbta-histologies.tsv
- Use the scores from analyses/transcriptomic-dimension-reduction/

jharenza (approved):
@kgaonkar6 thanks for these changes. I think this looks great. I also checked out the two additional HGAT BRAF-mutated samples which are showing up in the tSNE and UMAP which we previously did not observe (`BS_QYPHA40N` and `BS_59ZJWJTF`). These have `composition == Derived Cell Line` and were not previously captured in the molecular subtyping module, so this change is now expected since #796 has gone in and wanted to note that here.

jaclyn-taroni:
@kgaonkar6 can you update this branch please? I think you are in a better position to (correctly) resolve these merge conflicts than I am, thanks! Requesting myself for review!

kgaonkar6:
Ah yes, I'll fix these!

jaclyn-taroni (approved):
Looks great!

Issue #859, opened 2020-12-01
Updated analysis: Ependymoma add WGS only samples to subtyping 


kgaonkar6:
#### What analysis module should be updated and why?
WGS only epndymoma samples are being missed out from the subtyping.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

Mainly from this section it seems only participant ids which have RNA are chosen  to subset WGS samples as well
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0a6e54370399e7df632059957186a631b381d23b/analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py#L49

4 WGS only samples which get missed out because of the participant_id selection :
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0a6e54370399e7df632059957186a631b381d23b/analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py#L53

In addition, since there will be NA's in the RNA bs id column if we remove the above selection, this snippet will also need to be updated 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0a6e54370399e7df632059957186a631b381d23b/analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py#L113-L118

#### What input data should be used? Which data were used in the version being updated?
pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
pbta-cnv-consensus-gistic.zip
analyses/chromosomal-instability/breakpoint-data/union_of_breaks_densities.tsv
analyses/fusion-summary/results/fusion_summary_ependymoma_foi.tsv
analyses/gene-set-enrichment-analysis/results/gsva_scores_stranded.tsv

#### When do you expect the revised analysis will be completed?
~1 day

#### Who will complete the updated analysis?
@kgaonkar6

jaclyn-taroni:
Closed via #860

PR #860, opened 2020-12-01
Ependymoma subtyping missing WGS only samples

kgaonkar6:
Ependymoma subtyping results file is missing WGS only samples , through this PR I'm attempting to add these (4) samples and corresponding values used in subtyping. 

#### What was your approach?
removed rna participant id selection and updated the fill_df_with_fpkm_zscorres to match to only rows where Kids_First_Biospeecimen_ID_RNA is not NA.

#859 

- I might have missed some edits downstream to adding WGS only samples.  
- CNV and SV density columns have changed which is expected because of new samples being added and the columns are calculated as 

```r
x["breaks_density-chromosomal_instability_SV"]-sv_median)/sv_iqr
x["breaks_density-chromosomal_instability_CNV"]-cnv_median)/cnv_iqr
```

EPN molecular_subtypes seem to deviated from the format of other subtyping values like HGG, wildtype/ LGG, BRAF V600E values do we want to update these? 
```r
"ST_EPN_RELA" "ST_EPN_YAP1" "PT_EPN_A"   
```

Yes

tables

97 participants are being added to EPN subtyping with additional 4 WGS only samples

#### Documentation Checklist


analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py
```diff
@@ -42,16 +42,10 @@ def group_disease(primary_site):
 # Filtering for ependymoma samples 
 EP = pbta_histologies[pbta_histologies["pathology_diagnosis"]=="Ependymoma"]
 EP_rnaseq_samples = EP[EP["experimental_strategy"] == "RNA-Seq"][["Kids_First_Biospecimen_ID", "primary_site",
```
jashapiro, line 44: v. minor comment: can we make the order of columns here and for `WGS_dneseqsamples` the same? Took me longer than I would have expected to confirm that they include the same data!

analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py
```diff
@@ -62,10 +56,13 @@ def group_disease(primary_site):
 # sample_id is common between both  datafarmes and also unique between RNA and DNA. 
 # Some DNA BSID's are missing for the corresponding RNA samples
 EP_rnaseq_WGS = EP_rnaseq_samples.merge(WGS_dnaseqsamples, 
-                                        on = ["sample_id", "Kids_First_Participant_ID"], 
+                                        on = ["sample_id", "Kids_First_Participant_ID","primary_site"],
```
jashapiro, line 59: I assume here that `primary_site` does not change anything about the way the tables are matched, but is in the merge statement to prevent creating a duplicate column?

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
```diff
@@ -114,7 +114,8 @@ def broad_CNA_fill_df(row, CNA, arm, loss_gain):
 def fill_df_with_fpkm_zscores(df, fpkmdf, gene_name):
```
jashapiro, line 114: I am concerned here that because `df` and `fpkmdf` are not necessarily the same size, they may also be out of order at times. Is there a check to be sure that the rows that satisfy `df["Kids_First_Biospecimen_ID_RNA"].notna()` have the same ids as `fpkmdf`?

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
```diff
@@ -176,9 +177,20 @@ def fill_df_with_fpkm_zscores(df, fpkmdf, gene_name):
...
-for gene in expression_cols:
-    EPN_notebook = fill_df_with_fpkm_zscores(EPN_notebook, fpkm_df, gene)
+# samples with rna matched with dna
+EPN_notebook_rna_dna = EPN_notebook[EPN_notebook["Kids_First_Biospecimen_ID_RNA"].notna()]
+# samples with only dna
+EPN_notebook_dna_only = EPN_notebook[EPN_notebook["Kids_First_Biospecimen_ID_RNA"].isna()]
 
+for gene in expression_cols:
+    # rna matched dna samples
+    EPN_notebook_rna_dna = fill_df_with_fpkm_zscores(EPN_notebook_rna_dna, fpkm_df, gene)
```
jashapiro, line 187: `EPN_notebook_rna_dna` has already been filtered for NA values on line 181, but then `fill_df_with_fpkm_zscores()` checks again? Do we need the changes to `fill_df_with_fpkm_zscores()` for usage elsewhere? Or can we eliminate the splitting and recombining here and do it all within `fill_df_with_fpkm_zscores()`?

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
```diff
@@ -176,9 +177,20 @@ def fill_df_with_fpkm_zscores(df, fpkmdf, gene_name):
...
+    column_name = gene + "_expr_zscore"
+    # add NA to expression zscore columns for dna only samples
+    EPN_notebook_dna_only.loc[EPN_notebook_dna_only["Kids_First_Biospecimen_ID_RNA"].isna(),column_name] = np.nan
+    
+# append dataframes    
+EPN_notebook = EPN_notebook_rna_dna.append(EPN_notebook_dna_only)
 
 # Replacing all Nan values with NA so they are not empty when writing to a file
 EPN_notebook = EPN_notebook.replace(np.nan, 'NA', regex=True)
```
jashapiro, line 196: Just noting here that in the two output files included here, `EPN_all_data.tsv` uses`NA`, but `EPN_all_data_withsubgroup.tsv` uses blank fields. I like the explicit `NA` here, but this should maybe be added to the other output?

jashapiro (commented):
Hi @kgaonkar6 ! Thank you for your patience while waiting for my review!

This looks good, and my comments are mostly minor. 

I can see from the output that my concern about variation in the order of rows is not a problem in this case, and perhaps not in any case, but I always worry about accidental merges, so it seems worth checking in the code if possible (or if this is already done and I just missed it, let me know!).

analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py
kgaonkar6: Right, I added primary_site here primarily to avoid the _x and _y duplicated columns,

```
Index(['Kids_First_Biospecimen_ID_RNA', 'Kids_First_Participant_ID',
       'sample_id', 'primary_site_x', 'Kids_First_Biospecimen_ID_DNA',
       'primary_site_y'],
      dtype='object')
```

analyses/molecular-subtyping-EPN/01-make_notebook_RNAandDNA.py
kgaonkar6: I edited the order of columns being selected for EP_rnaseq_samples and WGS_dneseqsamples to be the same.

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
kgaonkar6: hmm there is no checks for the order. I can look into it

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
kgaonkar6: Updated fill_df_with_fpkm_zscores() to add NA to WGS only samples and zscore to RNAseq samples instead of splitting and recombining. 
However the issue with possible change in order from your above comment remains to be worked on.

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
kgaonkar6: Thanks for the review @jashapiro ! I added some commits below to work on some of your suggestions and also added NA to EPN_all_data_withsubgroup.tsv.

analyses/molecular-subtyping-EPN/02_ependymoma_generate_all_data.py
kgaonkar6: I checked fpkm_df.loc["RELA",df.loc[df["Kids_First_Biospecimen_ID_RNA"].notna(),"Kids_First_Biospecimen_ID_RNA"]] which seems to have the same order as `df.loc[df["Kids_First_Biospecimen_ID_RNA"].notna(),"Kids_First_Biospecimen_ID_RNA"]` which is how I envision the order to be in line 115 in 02_ependymoma_generate_all_data.py. Is there a. better way to add a check in the code?

BS_4T0HPZDC     67.52
BS_H4NXBD2D    140.89
BS_S0175QKX     39.55
BS_KXMYBQ6N     38.31
BS_0BXY0F9N     26.66
                ...  
BS_C05FKR8X    162.01
BS_HZ707C55     41.73
BS_YE1MAQYJ     28.29
BS_PJPGVD62     79.53
BS_7B89YR6K     28.72

jashapiro: Oh, I missed that on my first read: the fact that you are using `x["Kids_First_Biospecimen_ID_RNA"]` inside your lambda function means that you will get the correct values in every case.

jashapiro (commented):
This looks good to me, but I am noting that final review of this is waiting for v18 release.

jaclyn-taroni:
This apparently fixes #860 per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/889#issuecomment-754842961 and all review comments have been addressed, so I'm going to merge this one _next_ to get the release unblocked.

jaclyn-taroni (approved):

Issue #861, opened 2020-12-01
Updated analysis: RNA summary files and base histology subtyping update for v18


kgaonkar6:
#### What analysis module should be updated and why?

**RNA-seq summary files generation modules part of download or used in subtyping**
| Module |  Reason |Brief Description | output
|--------|----------|--------|------
| [`collapse-rnaseq`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/collapse-rnaseq) | adding 8 samples #749 and used in subtyping  | Collapses RSEM FPKM matrices such that gene symbols are de-duplicated. | `results/pbta-gene-expression-rsem-fpkm-collapsed.polya.rds` (included in data download; too large for tracking via GitHub) <br> `results/pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds` (included in data download; too large for tracking via GitHub)
| [`fusion_filtering`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering) | adding 8 samples #749 and used in subtyping  | Standardizes, filters, and prioritizes fusion calls updated in #808 and #830 | `results/pbta-fusion-putative-oncogenic.tsv`(included in data download) <br> `results/pbta-fusion-recurrent-fusion-byhistology.tsv` (included in data download) <br> `results/pbta-fusion-recurrent-fusion-bysample.tsv` (included in data download) <br> `results/fusion_summary_lgat_foi.tsv` (included in data download)
| [`fusion-summary`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion-summary)| adding 8 samples #749 and used in subtyping  | Generate summary tables from fusion files updated in  #812 and #816 and #821 and [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/750#issuecomment-691529675) | `results/fusion_summary_embryonal_foi.tsv` (included in data download) <br> `results/fusion_summary_ependymoma_foi.tsv` (included in data download) <br> `results/fusion_summary_ewings_foi.tsv`
[...]

Issue #795, opened 2020-10-05
Updated analysis: independent samples list for RNA-seq


jharenza:
#### What analysis module should be updated and why?
[Independent samples](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples)

The current analysis creates files of independent (de-deplicated) samples as follows:
- Primary specimens only with whole genome sequence (WGS)
- Primary and secondary specimens with WGS
- Primary specimens only with either WGS or whole exome sequence (WXS)
- Primary and secondary specimens with WGS or WXS

However, there is no independent specimen RNA list and upon a quick check of the v17 data, it looks like we have 52 `Initial CNS Tumor` + `Progressive | Progressive Progressive Disease Post-Mortem | Recurrence` pairs (excluding cell lines). Currently, the only RNA-Seq module that utilizes an independent RNA-Seq specimen list is [fusion filtering](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering) as below.

**RNAseq sample matched by `sample_id` to `independent-specimens.wgswxs.primary.tsv`**
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ed6552df1f7b233ee7eece71b364ca561ca459a2/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L46-L53

**and if an RNA-Seq sample, but no DNA sample exists**
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ed6552df1f7b233ee7eece71b364ca561ca459a2/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L59-L73

I think it follows that we should update other RNA-Seq only modules which create summaries by a grouping variable to utilize independent RNA-Seq lists and avoid counting duplicate specimens in analyses:
- [gene-set-enrichment-analysis](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/gene-set-enrichment-analysis)
- [immune-deconv](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/immune-deconv)
- [telomerase-activity-prediction](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/telomerase-activity-prediction)

[...]

Issue #750, opened 2020-08-28
Planned Release: V18


jharenza:
#### What data file(s) does this issue pertain to?
`pbta-histologies.tsv`
all RNA-Seq files

#### What release are you using?

V17

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
Some future updates to include:
Add 8 RNA-Seq samples #749:
- [x] pbta-star-log-final.tar.gz
- [x] pbta-star-log-manifest.tsv
- [x] pbta-mend-qc-manifest.tsv
- [x] pbta-mend-qc-results.tar.gz
- [x] pbta-isoform-counts-rsem-expected_count.stranded.rds
- [x] pbta-isoform-expression-rsem-tpm.stranded.rds
- [x] pbta-fusion-arriba.tsv.gz
- [x] pbta-fusion-putative-oncogenic.tsv
- [x] pbta-fusion-recurrently-fused-genes-byhistology.tsv
- [x] pbta-fusion-recurrently-fused-genes-bysample.tsv
- [x] pbta-fusion-starfusion.tsv.gz
- [x] pbta-gene-counts-rsem-expected_count.stranded.rds
- [x] pbta-gene-expression-kallisto.stranded.rds
- [x] pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
- [x] pbta-gene-expression-rsem-fpkm.stranded.rds
- [x] pbta-gene-expression-rsem-tpm.stranded.rds
- [x] independent-specimens.rnaseq.primary-plus-stranded.tsv
- [x] fusion_summary_lgat_foi.tsv
- [x] md5sum.txt
 
- [ ] Add new medulloblastoma subtype pathology results file #746
- [ ] pbta-histologies.tsv refresh (V18 or maybe V19, post rework of subtyping modules post #732 to use `pathology_diagnosis` and `free_text_dx` due to being stuck in a loop for #748) 
- [x] Add `Extent of Tumor Resection` to pbta-histologies.tsv cc @adamcresnick and @baileyckelly

jaclyn-taroni:
We're going to need to update the following files, because the `tumor_descriptor` field is changing in v17:

```
independent-specimens.wgs.primary-plus.tsv
independent-specimens.wgs.primary.tsv
independent-specimens.wgswxs.primary-plus.tsv
[...]

PR #864, opened 2020-12-07
PBTA Histologies: Independent sample base (2 of N)

kgaonkar6:
Rna summary modules needed for subtyping need to be updated because of the 8 new stranded files in v18. In this PR  I'm updating independent-samples module to read from pbta-histologies-base.tsv which will be used to generate RNA summary files and run molecular subtyping modules.

| Module |  Reason |Brief Description | output
|--------|----------|--------|------
| [`independent-samples`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples) | adding 8 samples #749 and used in subtyping  | Generates independent specimen lists for WGS/WXS samples updated in  #795 and #797 and [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/750#issuecomment-691526896)| `results/independent-specimens.wgs.primary.tsv` (included in data download) <br> `results/independent-specimens.wgs.primary-plus.tsv` (included in data download) <br> `results/independent-specimens.wgswxs.primary.tsv` (included in data download) <br> `results/independent-specimens.wgswxs.primary-plus.tsv` (included in data download) <br> `results/independent-specimens.rnaseq.primary-plus-stranded.tsv`(included in data download) <br> `results/independent-specimens.rnaseq.primary-plus-polya.tsv` (included in data download)
|

#861 

I've added a condition to run independent-samples modules with pbta-histologies-base.tsv in run-independent-samples.sh
Most of the different samples selected in the output files is from random selection since I didn't update the actual code generating the list.
Other modules that need to be rerun for subtyping are in #861 

But some changes occurred because of different tumor_descriptors in v18 
PT_ID | BS_ID | Change
--- | --- | ---
PT_1VQWQ0TC |	BS_QXZTRMWM  | was Initial now Recurrence

Yes

tables

8 new samples added to stranded RNA independent sample list and updated tumor_descriptor used for WGS/WXS independent sample list

#### Documentation Checklist


jharenza (commented):
@kgaonkar6 It looks like the files from `collapse-rnaseq` were also included in this PR - were you staggering these PRs or should these not be here?

jharenza (approved):
Works as expected. Using the v18 data download, I ran with 
```
RUN_FOR_SUBTYPING=${BASE_SUBTYPING:-1}
```

and got the same file dims as you did, but did not compare directly because of the random selection.

Just check out my comment about about whether you should remove the files from the previous PR submitted.

kgaonkar6:
These are staggered, I used instructions in https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/CONTRIBUTING.md#creating-stacked-pull-requests let me know it I missed something. 

hmm CI seems to error out because it doesn't have a file named pbta-histologies-base.tsv should I add that to the CI subset files?

jharenza:
> These are staggered, I used instructions in https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/CONTRIBUTING.md#creating-stacked-pull-requests let me know it I missed something.
> 
> hmm CI seems to error out because it doesn't have a file named pbta-histologies-base.tsv should I add that to the CI subset files?

Ok great. No, I don't think there is a need for that right now because these PRs will not be merged until #857 goes in, which depends on CI file updates by @jaclyn-taroni and/or someone at CCDL.

jharenza (approved):
This looks good @kgaonkar6 - will you also update the READMEs for all of the modules to which you are adding the subtyping base run to note the option?

Thanks!

kgaonkar6:
Hi @jharenza  just updated the README

jharenza (approved):
thanks for the updates! looks good.

analyses/independent-samples/run-independent-samples.sh
```diff
@@ -22,3 +28,14 @@ Rscript 02-generate-independent-rnaseq.R \
   --output_directory results \
   --independent_dna_sample_df ../../data/independent-specimens.wgswxs.primary-plus.tsv  
 
+else 
+Rscript 01-generate-independent-specimens.R \
+  -f ../../data/pbta-histologies-base.tsv \
```
cansavvy, line 33: It looks like the only thing that is different is whether `../../data/pbta-histologies-base.tsv`or `../../data/pbta-histologies.tsv` is used, so a shorter way to do this is to use a smaller if/then specify a `HISTOLOGIES_FILE` variable and supply that to the `-f` and `--histology_file` options. Then you can get rid of this larger if/then and not have to repeat these Rscript calls.

analyses/independent-samples/README.md
```diff
@@ -25,11 +25,18 @@ independent-specimens.rnaseq.primary-plus-stranded.tsv
 
 To generate the independent sample lists and associated analysis of redundancies in the overall data set, run the following script from the project root directory:
 
+use BASE_SUBTYPING=1 to run this module using the pbta-histologies-base.tsv from data folder while running molecular-subtyping modules for release.
+```sh
+BASE_SUBTYPING=1 ../analyses/independent-samples/run-independent-samples.sh
```
cansavvy, line 30: It's unclear to me which histologies file should be used for the circle CI, but right now it doesn't have its own variable supplied and seems to be failing. 

You can follow the OpenPBTA instructions here https://github.com/AlexsLemonade/OpenPBTA-analysis#passing-variables-only-in-ci to add an explicit call to the circle CI file so its clear which file should be being used.

cansavvy (commented):
I think this mostly looks good, I just have two comments that should probably be addressed (and might help resolve that circle CI error).

analyses/independent-samples/README.md
jharenza: Hi @cansavvy! Thanks for your comments. This sequence of PRs will require #849 to be merged and #871 to be generated, as they are utilizing the `pbta-histologies-base.tsv` file. I think the idea is for these to be reviewed, as multiple of the files generated from these modules will be included in the release, then for us to merge #849, and have all of these run in CI for confirmation they pass.

kgaonkar6:
thanks for review @cansavvy @jharenza , latest commit was from a re-run with updated base histology(minor change to Normal sample rows) file which doesn't change any results files.

analyses/independent-samples/run-independent-samples.sh
kgaonkar6: ah that's a great idea, I'll update that, thanks!

analyses/independent-samples/README.md
kgaonkar6: Also I wanted to confirm that in the v18 release data folder we will have eventually have from #870  pbta-histologies.tsv so it should be able to run through .circleci/config.yml for checks as is, right? Or do I add the following to .circleci/config.yml?
```
BASE_SUBTYPING=1 ../analyses/independent-samples/run-independent-samples.sh 
```

kgaonkar6:
updated the run script with a simple if else for HISTOLOGY_FILE assignment instead of the previous longer code update as suggested by @cansavvy . I also updated reading `results/independent-specimens.wgswxs.primary-plus.tsv` instead of the data folder since we want to read the latest file generated for WGSWXS specimens from results folder.

analyses/independent-samples/README.md
cansavvy: Yes, but according to the instructions, you will want to start the variable with `OPENPBTA_` like the examples show: https://github.com/AlexsLemonade/OpenPBTA-analysis#passing-variables-only-in-ci 
```
Any environment variables prefixed with OPENPBTA_ are passed to the specified shell script. Environment variables without this prefix are not passed.
```

kgaonkar6: ah got it, I didn't realize we will be running the BASE_SUBTYPING=1 runs in CI. But since we will I'm going to update the name for the variable as you've suggested here and re-run

kgaonkar6: Updated to OPENPBTA_BASE_HISTOLOGY variable

jaclyn-taroni:
It looks to me like all of @cansavvy's comments have been address since she last reviewed. I will merge if/when CI passes.

Issue #878, opened 2020-12-23
Updated analysis molecular-subtype-pathology: clinically reviewed PNOC003 samples needs int_dx update ; add Notes to all pathology reviewed samples 


kgaonkar6:
#### What analysis module should be updated and why?

molecular-subtype-pathology is divided into 3 steps:
1) compile subtypes
2) add clinical review
3) update from pathology review 

 - update integrated_diagnosis column values for samples where initial molecular_subtype was "HGG, H3 wildtype" which got updated to molecular_subtype== "DMG K28M"  from clinical review in [step 2: add clinical review](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd) 

 - update [step 3. update from pathology review](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology/03-incorporate-pathology-feedback.Rmd) to add Notes to each pathology review updated sample_ids. 

#### What changes need to be made? Please provide enough detail for another participant to make the update.

- I missed to update integrated_diagnosis column values for samples where initial molecular_subtype was "HGG, H3 wildtype" which got updated to molecular_subtype== "DMG K28M"  from clinical review in [step 2: add clinical review](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology/02-incorporate-clinical-feedback.Rmd) which needs an update.
            * In addition, in **v18** we see "HGG, To be classified" for couple of newly sequenced RNA-Seq samples from PNOC which don't match WXS because of different sample_id and are left as "HGG, To be classified" but as per clinical review are K28M subtype. So I believe, to include these samples we also also need to be updated the code to include "HGG, To be classified" to "DMG K28M" conversion as well.

 - We also need to update [step 3. update from pathology review](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-pathology/03-incorporate-pathology-feedback.Rmd) to add Notes to each pathology review updated sample_ids because:
[...]

PR #876, opened 2020-12-22
Find breaking changes from v18 (part 2)

cansavvy:
### Purpose/implementation Section
This is the second attempt (#874 was the first) at finding breaking changes now with the metadata updates made and discussed on #875. 

I'm commenting out the broken modules and putting notes of what I find here. 

### Running list of errors: 
Checked off errors means they have been addressed. 
As I find the errors in Circle CI testing, I'm posting them here and will comment with more info as I investigate. 

- [ ] Molecular subtyping for Ependymoma needs to be adjusted because of new biospecimens added. See [CI run here](https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/4687/workflows/7d69e741-a37a-4506-8960-5450eb87751f/jobs/5259). 
Related issue #881 
Edit: This has been re-run and addressed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/876/commits/8f33c4a20eebaf89402003b91440b68f521fa4aa

```
Traceback (most recent call last):
  File "02_ependymoma_generate_all_data.py", line 180, in <module>
    EPN_notebook = fill_df_with_fpkm_zscores(EPN_notebook, fpkm_df, gene)
```
- [ ] Interaction plot does not run. I did some investigating. Issue #880 Only 1 LGAT sample shows up and that's because there's more `NA`s  for `integrated_diagnosis` than there were in v17. See comment here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/875#issuecomment-749705328
```
Error in combn(genes, m = 2) : n < m
Calls: coocurrence -> t -> combn
Execution halted
```

- [ ] Chromosomal instability breakpoints has a color palette issue because `Melanocytoma` is not in the color palette. Related Issue #882  Should be a straightforward fix. 
```
Quitting from lines 262-278 (02a-plot-chr-instability-heatmaps.Rmd) 
Error in col2rgb(colors, alpha = TRUE) : 
  invalid color name 'Melanocytoma'
Calls: <Anonymous> ... do.call -> <Anonymous> -> ColorMapping -> t -> col2rgb
```

- [ ] Comparative RNA-seq doesn't work. Related Issue: #883 Probably something also about samples lining up. 
```
^@^@Loading sample matrix ../../data/pbta-gene-expression-rsem-tpm.polya.rds
Traceback (most recent call last):
  File "analyses/comparative-RNASeq-analysis/01-correlation-matrix.py", line 231, in <module>
    main()
[...]

PR #879, opened 2020-12-23
Update integrated dx for 3 samples based on pathology review

kgaonkar6:

- update integrated_diagnosis column values for samples where initial molecular_subtype was "HGG, H3 wildtype" which got updated to molecular_subtype== "DMG K28M" from clinical review in 02-incorporate-clinical-feedback.Rmd

- update 03-incorporate-pathology-feedback.Rmd to add Notes to each pathology review updated sample_ids.

#### What was your approach?
- I am updating integrated_diagnosis column values for samples where initial molecular_subtype was "HGG, H3 wildtype"  OR " HGG, To be classified" which got updated to molecular_subtype== "DMG K28M" from clinical review 

- I'm also adding "Notes" = "Updated via OpenPBTA subtyping" in 03-incorporate-pathology-feedback.Rmd to cover all columns that need to be updated per updated molecular_subtype from pathology review. And specifically, since in v18 samples with "XYZ, To Be classified" the Notes and integrated_diagnosis is left NA (unlike in v17 histology file) so will create duplicate rows in v18 for same bs id (7316-506, 7316-765 and 7316-238) has Notes =c("NA", "Updated via OpenPBTA subtyping")  

#878 

Is there some condition I'm missing in clinical review update in 02?

The LGAT update to "Recurrence" was not expected, it happened because I had v18 files in data release. I've updated to v17 files in the latest commit 3fd999b

Yes

compiled subtypes with updated int_dx for PNOC003 samples 

Update compiled subtypes

#### Documentation Checklist


kgaonkar6:
hmm not sure why CI is erroring out with the EPN subtyping error that we see in v18 CI runs #876. In the "Checkout code" section of CI it does seem like we are using v17 histology file so not sure why it fails.

Regarding the actual error that fails CI : I believe (might be wrong) the error occurs because change in pathology_diagnosis from v17 to v18 discussed here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/876#issuecomment-749723223

jharenza:
> hmm not sure why CI is erroring out with the EPN subtyping error that we see in v18 CI runs #876. In the "Checkout code" section of CI it does seem like we are using v17 histology file, oh maybe OPENPBTA_RELEASE=testing now has v18 testing files?
> 
> Regarding the actual error that fails CI : I believe (might be wrong) the error occurs because change in pathology_diagnosis from v17 to v18 discussed here: [#876 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/876#issuecomment-749723223)

This looks good to me, but yes, we are now using v18 CI files, which I forgot about. I wonder if you should add then add this as an update within the v18 set of PRs?

jharenza (approved):
content looks right

jharenza:
Hi @jaclyn-taroni @jashapiro @cansavvy - When looking at the final v18 histologies file, I found the error prompting this PR. We did as a standalone PR instead of within the v18 release PRs since it was a bug in the code. However, it errors out with V18 CI files because of the EPN subset step that @jashapiro has subsequently [fixed](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/876/commits/8f33c4a20eebaf89402003b91440b68f521fa4aa), but we do not yet have merged. 

>hmm not sure why CI is erroring out with the EPN subtyping error that we see in v18 CI runs #876. In the "Checkout code" section of CI it does seem like we are using v17 histology file, oh maybe OPENPBTA_RELEASE=testing now has v18 testing files?

The easiest thing may be for us to switch back to v17 CI files to get this run through CI, merge this, switch to v18 CI files (unless you are OK with us merging this without it passing V18 CI, so that we can get this updated into the V18 histologies files. Please advise :).

analyses/molecular-subtyping-pathology/03-incorporate-pathology-feedback.Rmd
```diff
@@ -316,6 +323,10 @@ compiled_df <- compiled_df %>%
     molecular_subtype = case_when(
       sample_id == "7316-1102" ~ "DMG, H3 K28",
       TRUE ~ molecular_subtype
+    ),
+    Notes = case_when(
+      sample_id == "7316-1102" ~ "Updated via OpenPBTA subtyping",
```
cansavvy, line 328: I don't quite understand why it looks like there's recoding by individual samples -- it would probably be good to look into a more systematic way to do this, but that should be in a different PR, not now.

cansavvy (approved):
Seems reasonable to me and like it is doing what is described.

analyses/molecular-subtyping-pathology/03-incorporate-pathology-feedback.Rmd
jaclyn-taroni: This notebook is designed to handle individual cases based on feedback from pathologists / clinical expertise. I think it's possible that coding for individual cases with lots of context is more clear.

PR #886, opened 2020-12-24
Integrate tp53 alteration annotation as multi hits (molecular and/or classifier and/or germline predisposition)

kgaonkar6:
In this PR we will add TP53 alteration status as discussed in [#837](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/837):

**TP53 altered - loss**, if :

 - A sample contains a TP53 hotspot SNV mutation. (Cancer hotspot database and downloadable file available). Please also crosscheck that all mutations from this table are included.
 - A sample contains two TP53 alterations, suggesting (but not confirming) that both alleles are affected (SNV+SNV, CNV+SNV).
 - A sample contains one alteration (SNV or CNV) + has cancer_predispositions == "Li-Fraumeni syndrome", suggesting there is a germline variant in addition to the somatic variant we observe.
 - A sample does not have a TP53 alterations, but has cancer_predispositions == "Li-Fraumeni syndrome" and TP53 classifier score for matched RNA-Seq > 0.5 (or higher cutoff we decide upon later).
 
**TP53 altered - activated**, if:

- A sample contains one of the two TP53 activating mutations R273C and R248W. Reference and reference.

**Note** this annotation is also applicable to #807 HGG TP53 annotation

#### What was your approach?
Steps:
- gather consensus SNV for TP53 removing Silent and Intron
- gather filtered CNV losses 
- gather values for cancer_predispositions and tp53 classififer scores for TP53
- For annotation purposes I used counts of SNV and CNVs as values since we wanted to check for presence/absence and multi-hit TP53 variants. If AA position overlaps statistically significant SNV/Indels hotspots is assigned "Yes", if  AA position overlaps  gain-of-function sites according to literature activating column is assigned "Yes".

#837 

While matching by sample_id in these annotations should cell-lines be included in this annotation?

I think 1 condition to be discussed for annotation would be CNV/SNV loss + >0.5 TP53 score , should this also included as tp53_altered=="loss"?

I added a couple plots to visualize the distributions of scores, do we want to discuss those further in this PR?

yes

table , notebook figures

In OpenPBTA have 26 sample_ids with putative gain-of-function mutations, 96 with multiple loss of function mutations/CNVs and 980 as unknown/no evidence for tp53 altered status.

#### Documentation Checklist


jharenza:
Thanks for this @kgaonkar6 - looks like a great start!

I have a few comments:

>While matching by sample_id in these annotations should cell-lines be included in this annotation?

yes, please include

>I think 1 condition to be discussed for annotation would be CNV/SNV loss + >0.5 TP53 score , should this also included as tp53_altered=="loss"?

Yes, it looks like I forgot that category above.

>I added a couple plots to visualize the distributions of scores, do we want to discuss those further in this PR?

Yes, these are great!

It is interesting that the samples with activating mutations also have "high" scores. Based on the description of the classifier predicting inactivation, I would have thought that they would have been low scores. I wonder if this really means the high scores are more suggestive of "oncogenic" TP53? In your plot "Explore distribution of tp53_altered status vs tp53 inactivation scores
", will you plot RNA expression of these same samples and groups? I wonder if those with activating mutations will be high expression outliers? cc @cgreene and @gwaygenomics 

Re: "Check if other cancer predisposition have high TP53 inactivation scores" (image below)
- Samples with LFS and TP53 loss have very high scores (looks like ~0.75+) and there are two samples with no evidence for TP53 loss, which also have low scores. We should first check those samples to see if we can find TP53 alterations which were presumably the basis of the LFS diagnosis. They may be outside of the hotspot region, or may not have been caught in consensus calls, or just not exist in our data. I think this in itself is a really cool result, either way, because it could possibly suggest a misdiagnosis of LFS or a means of monitoring LFS patients and their TP53 functional status. Presuming, these patients still have non-oncogenic, functional TP53, which is an advantage. We should dig more into this. cc @cgreene and @adamcresnick
Screen Shot 2021-01-04 at 2 18 43 PM

The other plots look as expected - let's see how they look with the updated condition above (if we see more of the "non-altered" samples with high scores being removed from that group).

jharenza (changes requested):
See comments!

kgaonkar6:
Thanks for the review @jharenza! With regards to including cell-lines I also need to add the cell-line composition file like we used in molecular-subtyping-HGG since they have the same sample_id, matching by sample_id causes some duplicates in the outputs.  

And to explore the expression for activating/loss samples the direct visualization that came to my mind was a PCA but since it seems we are expecting to look at high outliers specifically were you expecting like a boxplot/scatter with specific labels of gene names instead?

jharenza:
> And to explore the expression for activating/loss samples the direct visualization that came to my mind was a PCA but since it seems we are expecting to look at high outliers specifically were you expecting like a boxplot/scatter with specific labels given a gene list instead?

I just was thinking of this plot with TP53 expression on the y-axis:
Screen Shot 2021-01-05 at 1 37 09 PM

gwaygenomics:
> It is interesting that the samples with activating mutations also have "high" scores. Based on the description of the classifier predicting inactivation, I would have thought that they would have been low scores. I wonder if this really means the high scores are more suggestive of "oncogenic" TP53? In your plot "Explore distribution of tp53_altered status vs tp53 inactivation scores", will you plot RNA expression of these same samples and groups? I wonder if those with activating mutations will be high expression outliers? cc @cgreene and @gwaygenomics

Cool! The classifier as an oncogenic TP53 detector seems like a reasonable hypothesis to me. In the original model, we did not disambiguate activating vs. inactivating mutations during training. One could go back to the original analysis and make the binary classifier multiclass. This would require bit of time annotating the TP53 mutations though, and it can get tricky since some variants are uncertain.

kgaonkar6:
Here's the distribution of TP53 for activated and loss annotated samples:
image

In addition, I did some comparison of genes which have non zero coefficient [tp53_classifier_coefficients.tsv](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/integrate_tp53_alt/analyses/tp53_nf1_score/reference/tp53_classifier_coefficients.tsv) which was used  while applying the classifier. This might give some insight in how the expression of these genes looks like for these subset of tp53 activated/loss status? For example these genes seem to have significant differences when tested with wilcox test:
Screen Shot 2021-01-06 at 1 26 27 PM

jharenza:
> Here's the distribution of TP53 for activated and loss annotated samples:
> image
> 
> In addition, I did some comparison of genes which have non zero coefficient [tp53_classifier_coefficients.tsv] (https://github.com/kgaonkar6/OpenPBTA-analysis/blob/integrate_tp53_alt/analyses/tp53_nf1_score/reference/tp53_classifier_coefficients.tsv)
> which was used while applying the classifier **plus TP53** in stranded samples . This might give some insight in how the expression of these genes looks like for these subset of tp53 activated/loss status? For example these genes seem to have significant differences when tested with wilcox test so it seems some genes might help with the differentiating different classes of TP53 status :
> Screen Shot 2021-01-06 at 1 26 27 PM

This looks good! Would you replot as a boxplot with jitter or use the Enhanced Volcano package to add a boxplot in the volcano to show the medians here? 

I may want to follow up with the samples annotated as GOF, and if so, will submit a new issue for that.

Did you have a chance to annotate these samples?
>I think 1 condition to be discussed for annotation would be CNV/SNV loss + >0.5 TP53 score , should this also included as tp53_altered=="loss"?

jharenza (commented):
> Here's the distribution of TP53 for activated and loss annotated samples:
> image

This looks good! Would you replot as a boxplot with jitter or use the Enhanced Volcano package to add a boxplot in the volcano to show the medians here? I may want to follow up with the samples annotated as GOF, and if so, will submit a new issue for that.

> In addition, I did some comparison of genes which have non zero coefficient [tp53_classifier_coefficients.tsv] (https://github.com/kgaonkar6/OpenPBTA-analysis/blob/integrate_tp53_alt/analyses/tp53_nf1_score/reference/tp53_classifier_coefficients.tsv)
> which was used while applying the classifier **plus TP53** in stranded samples . This might give some insight in how the expression of these genes looks like for these subset of tp53 activated/loss status? For example these genes seem to have significant differences when tested with wilcox test so it seems some genes might help with the differentiating different classes of TP53 status :
> Screen Shot 2021-01-06 at 1 26 27 PM

I think this could be something we work on later, but not within this paper.

Did you have a chance to annotate these samples?
>I think 1 condition to be discussed for annotation would be CNV/SNV loss + >0.5 TP53 score , should this also included as tp53_altered=="loss"?

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
```diff
@@ -0,0 +1,416 @@
...
+              # check if tp53 inactivating score for RNA_Seq is greater that 0.5
+              # and has Li-Fraumeni syndrome suggesting
+              # germline TP53 mutant as cancer predisposition 
+              (tp53_score > 0.5 & 
+                             grepl("Li-Fraumeni syndrome",cancer_predispositions))
+          ) ~ "loss",
+        # when activating == "Yes"
+        activating == "Yes" ~ "activated",
+        # if no evidence supports TP53 inativation or activation 
+        TRUE ~ "No evidence"
```
jharenza, line 267:
```suggestion
        TRUE ~ "Other"
```

Now that I think about this a bit more, I think maybe we should rename this - it is not that there is no evidence, it is that there is not high-confidence evidence to support this.

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
```diff
@@ -0,0 +1,416 @@
...
+  stat_compare_means() +
+  theme_bw() +
+  ggtitle("Distribution of scores across tp53 altered status") +
+  theme(axis.text.x = element_text(angle = 60, hjust = 1))+
+  xlab("tp53 altered status") 
+
+```
+TP53 mutants with gain-of-function (activating) mutants promote tumorigenesis according to literature. [Reference](https://pubmed.ncbi.nlm.nih.gov/17417627/) and [reference](https://pubmed.ncbi.nlm.nih.gov/24677579/) have similar distribution of classifier scores to that of potential loss-of-function (multi-allelic) Tp53 mutations. 
+
+"No evidence" annotated samples either don't have any loss/gain TP53 SNVs nor CNV losses OR DNA sample is not available.
```
jharenza, line 288:
```suggestion
"Other" annotated samples either don't have any high-confidence loss/gain TP53 SNVs nor CNV losses OR DNA sample is not available.
```

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
```diff
@@ -0,0 +1,355 @@
...
+  theme(axis.text.x = element_text(angle = 60, hjust = 1))+
+  xlab("tp53 altered status") 
+
+```
+TP53 mutants with gain-of-function (activating) mutants promote tumorigenesis according to literature. [Reference](https://pubmed.ncbi.nlm.nih.gov/17417627/) and [reference](https://pubmed.ncbi.nlm.nih.gov/24677579/) have similar distribution of classifier scores to that of potential loss-of-function (multi-allelic) Tp53 mutations. 
+
+"Other" annotated samples either don't have any high-confidence loss/gain TP53 SNVs nor CNV losses OR DNA sample is not available.
+
+### (stranded) Expression profile for activating vs loss TP53 status
+Using stranded because there are more samples in stranded
```
jharenza, line 295: Will you also plot polyA here (in a separate plot - you can facet library), even though there are fewer samples?

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
```diff
@@ -0,0 +1,355 @@
...
+  theme_bw() +
+  ggtitle("Distribution of TP53 expression across tp53 altered status") +
+  theme(axis.text.x = element_text(angle = 60, hjust = 1))+
+  xlab("tp53 altered status") 
+
+```
+
+
+
+### Check if other cancer predisposition have high TP53 inactivation scores
```
jharenza, line 330: Can you print the BS_IDs of the two LiFraumeni samples which show low TP53 scores? Will want to go back and check these diagnoses/outcomes.

jharenza (commented):
Thinking about this data going into #807, it looks like the the tp53_altered_status.tsv file will be used as input. Do we want to add some evidence columns? Eg: SNV_evidence would be the TP53 SNVs, CNV_evidence would be the consensus CNV?

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
kgaonkar6: done!

analyses/tp53_nf1_score/04-tp53-altered-annotation.Rmd
kgaonkar6: done!

kgaonkar6:
> Thinking about this data going into #807, it looks like the the tp53_altered_status.tsv file will be used as input. Do we want to add some evidence columns? Eg: SNV_evidence would be the TP53 SNVs, CNV_evidence would be the consensus CNV?

hmm yeah I think that would be good instead of just 1/0 for presence/absense. I'll implement this

kgaonkar6:
We currently have this filter `filter(!(Variant_Classification %in% c("Silent", "Intron")))` which is the filter we've used before in other modules but while looking at the HGVSp_Short values, there are some "." because of 3`Flank and 3`UTR region mutations is that something we should remove because of non amino acid change?

jharenza:
> We currently have this filter `filter(!(Variant_Classification %in% c("Silent", "Intron")))` which is the filter we've used before in other modules but while looking at the HGVSp_Short values, there are some "." because of 3`Flank and 3`UTR region mutations is that something we should remove because of non amino acid change?

Yes

jharenza (commented):
Hi @kgaonkar6 - just checking the numbers here. In `tp53_alterations_score_wide`, I see that we have 1033 Kids_First_Biospecimen_ID_RNA, but in the histologies file, we have 1036. We still have 1028 in the `classifier_score`, which was the N of RNA samples pre-v18. Looks like we are missing these samples, but also have some duplicate samples in the table (eg: 7316-1463). Perhaps when we have duplicate RNA specimens per DNA specimen, we can comma-separate them in the Kids_First_Biospecimen_ID_RNA field to have unique rows of Kids_First_Biospecimen_ID_RNA to ensure we are not plotting values twice, which it looks like we are right now. As an alternative, we could unique the Kids_First_Biospecimen_ID_RNA in the plots, but that may be harder to implement. Thoughts?

kgaonkar6:
good catch! I haven't run the classifier on v18 was just adding the TP53 annotations I'll re-run 01 to get the updated scores. 

For the multiple Kids_First_Biospecimen_ID_RNA , I could comma collapse the ids  then do we plot the mean value of tp53_score to be plotted? 7316-87 for example has quite different score for each bs id :

sample_id | Kids_First_Biospecimen_ID_DNA | Kids_First_Biospecimen_ID_RNA | cancer_predispositions | tp53_score | SNV_indel_counts | CNV_loss_counts | HGVSp_Short | copy_number | hotspot | activating | tp53_altered
 -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
7316-87 | BS_4DY83R02 | BS_1HQ76V6D | None documented | 0.44161220 | 0 | 0 | NA | NA |   |   | Other
7316-87 | BS_4DY83R02 | BS_JGA9BP3A | None documented | 0.06645831 | 0 | 0 | NA | NA |   |   | Other

With 7316-1463 type situation with multiple WGS and multiple RNA per sample_id I think maybe selecting ids from the independent might help?
sample_id | Kids_First_Biospecimen_ID_DNA | Kids_First_Biospecimen_ID_RNA | cancer_predispositions | tp53_score | SNV_indel_counts | CNV_loss_counts | HGVSp_Short | copy_number | hotspot | activating | tp53_altered
7316-1463 | BS_KH3859M5 | BS_2HN96129 | NF-2,Schwannomatosis | 0.1688424 | 0 | 0 | NA | NA |   |   | Other
7316-1463 | BS_KH3859M5 | BS_RR12T74P | NF-2,Schwannomatosis | 0.1513061 | 0 | 0 | NA | NA |   |   | Other
7316-1463 | BS_NASADC3P | BS_2HN96129 | NF-2,Schwannomatosis | 0.1688424 | 0 | 0 | NA | NA |   |   | Other
7316-1463 | BS_NASADC3P | BS_RR12T74P | NF-2,Schwannomatosis | 0.1513061 | 0 | 0 | NA | NA |   |   | Other

jharenza:
> 7316-87

Interesting - these are both initial tumor, but one is polyA and one is stranded, the latter having the higher score.

kgaonkar6:
I've now updated the tp53 annotation to v18. There are still duplicated RNA-Seqs in the ouptut either because samples have stranded and polya RNA or there are multiple bs_ids per sample_id (maybe because of multiple locations?).

With regards to the classifier score in the QC script we did plot the stranded and polya samples, the scores don't seem to have a drastically different trends per library. 
image

image

I can look into plotting sample_id-matched stranded and polya scores if that would be interesting to look at for selection?

jharenza:
>I can look into plotting sample_id-matched stranded and polya scores if that would be interesting to look at for selection?

Oh yes, this is what I meant - if we plot all pairs side by side, or just plot the polyA-scores vs stranded scores, is there a clear difference by library?

kgaonkar6:
@jharenza  here's a sample_id matched plot: 

image

On average it seems that polya samples on average have lower scores compared to sample_id matched stranded bs ids.

kgaonkar6:
Also there was a comment before to update the format of output so I've added the SNV evidence and CNS copy number as well in columns HGVSp_Short and copy_number respectively. Does ⤵️ look right? 

sample_id | Kids_First_Biospecimen_ID_DNA | Kids_First_Biospecimen_ID_RNA | cancer_predispositions | tp53_score | SNV_indel_counts | CNV_loss_counts | HGVSp_Short | copy_number | hotspot | activating | tp53_altered
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --

jharenza:
> Also there was a comment before to update the format of output so I've added the SNV evidence and CNS copy number as well in columns HGVSp_Short and copy_number respectively. Does ⤵️ look right?
> 
> sample_id	Kids_First_Biospecimen_ID_DNA	Kids_First_Biospecimen_ID_RNA	cancer_predispositions	tp53_score	SNV_indel_counts	CNV_loss_counts	HGVSp_Short	copy_number	hotspot	activating	tp53_altered

Yes - just a few questions: 
1. are `CNV_loss_counts` and `copy_number` sort of redundant information? Should we just keep one of those columns? 
2. Should we make all binary columns binary for consistency (eg `hotspot` and `activating` as well?)

kgaonkar6:
> 1. are `CNV_loss_counts` and `copy_number` sort of redundant information? Should we just keep one of those columns?
> 2. Should we make all binary columns binary for consistency (eg `hotspot` and `activating` as well?)

CNV_loss_counts is the number of copy number losses found for tp53 in the sample and copy_number is the actually value of copy loss for tp53. So it is redundant information but I guess I added `copy_number` as the evidence column for `CNV_loss_counts` . For example BS_ZV21J6YW is the only sample that has copy_number ==0 overlapping P53_tetramer, P53, P53_TAD domains of TP53 which would be lost with just binary column? Maybe I can change the name from `copy_number` to `CNV_loss_evidence`? 

```r
    SNV_indel_counts= length(unique(HGVSp_Short[!is.na(HGVSp_Short)])), 
    CNV_loss_counts = length(unique(copy_number[!is.na(copy_number)])),
    HGVSp_Short = toString(unique(HGVSp_Short)),
    copy_number = toString(unique(copy_number))
```

jharenza:
> > 1. are `CNV_loss_counts` and `copy_number` sort of redundant information? Should we just keep one of those columns?
> > 2. Should we make all binary columns binary for consistency (eg `hotspot` and `activating` as well?)
> 
> CNV_loss_counts is the number of copy number losses found for tp53 in the sample and copy_number is the actually value of copy loss for tp53. So it is redundant information but I guess I added `copy_number` as the evidence column for `CNV_loss_counts` . For example BS_ZV21J6YW is the only sample that has copy_number ==0 overlapping P53_tetramer, P53, P53_TAD domains of TP53 which would be lost with just binary column? Maybe I can change the name from `copy_number` to `CNV_loss_evidence`?
> 
> ```r
>     SNV_indel_counts= length(unique(HGVSp_Short[!is.na(HGVSp_Short)])), 
>     CNV_loss_counts = length(unique(copy_number[!is.na(copy_number)])),
>     HGVSp_Short = toString(unique(HGVSp_Short)),
>     copy_number = toString(unique(copy_number))
> ```

Ok, sure, sounds good.

kgaonkar6:
We have these multiple RNA matched with multiple DNA I don't think there are unique values in columns to match, I used "sample_id",
"cell_line_composition",
"cancer_predispositions",
"sample_type",
"Kids_First_Participant_ID"

[multi_rows_RNA.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5914678/duplicated_rows_RNA.txt)

jharenza:
> We have these multiple RNA matched with multiple DNA I don't think there are unique values in columns to match, I used "sample_id",
> "cell_line_composition",
> "cancer_predispositions",
> "sample_type",
> "Kids_First_Participant_ID"
> 
> [multi_rows_RNA.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5914678/duplicated_rows_RNA.txt)

I think having these in the output file should be OK, however, when we go to do any plots for ROC or other, we should make sure we are using unique RNA biospecimens.

kgaonkar6:
@jharenza thanks! I've also updated the readme with descriptions for each column now.

kgaonkar6:
@jharenza thanks! I've also updated the readme with descriptions for each column now.

jharenza (approved):
looks ready!

jaclyn-taroni (approved):
👍  I did not run this myself but the logic seems to do what is described in the comments.

PR #893, opened 2021-01-06
Remove MAF file hyperlink in SNV callers README

### LINKED ISSUE #885: Updated documentation: SNV callers
[snv callers](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers#snv-caller-comparison-analysis)

The consensus maf file link in paragraph 1 is broken because we do not store this file in GitHub, so there should not be a hyperlink.

NA

Unsure

Unsure

cansavvy:
### Background 

There's a broken link in the README for snv-callers. I'm removing it here. 

Closes #885


jaclyn-taroni (approved):
Does what says, looks good to me! 👍

jaclyn-taroni:
Docs only change, merging without CI!

Issue #898, opened 2021-01-11
Make mapping histology groups for plotting -- Part 2 implementing the use of the mapping table


cansavvy:
#### Background:

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/627#issuecomment-754813774: 

- This mapping TSV file (created in #897) needs to be used in all relevant plotting modules (or wherever else smaller "high-level" labels are needed. 

- `histology_color_palette.tsv` will need to be retired after all modules are updated. 

#### Known relevant modules that need this update

- [x] chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd - Updated in #904
- [x] cnv-chrom-plot/cn-status_heatmap.Rmd - Updated in #911
- [x] mutational-signatures/01-known_signatures.Rmd - Updated in #911
- [x] oncoprint-landscape/01-plot-oncoprint.R - Updated in #918
- [x] sample-distribution-analysis/02-multilayer-plots.R - Updated in #918

** We also need to make sure that the figures/scripts equivalents of these figures are up to date: 

- [x] fig1-sample-distribution.R - Updated in #918
- [ ] TelomeraseActivitites.R
- [ ] transcriptomic-overview.R - Updated in #921 

#### What input data are required for this analysis?

`pbta-histologies.tsv`

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
This is dependent on #897

jaclyn-taroni:
As we're going through these, let's be careful to make sure that things don't stay commented out in CI. I believe everything covered in #882, where the CI errors are documented, is covered here now and we are going to retire `histology_color_palette.tsv`.

jaclyn-taroni:
As far as I can tell, this is done.

Issue #897, opened 2021-01-11
Make mapping histology groups for plotting -- Part 1 creating the table


cansavvy:
#### Background:

From https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/627#issuecomment-754813774: 

> I'm going to suggest that we create a **high level histology** list that we can then use for figures. This does not need to be recorded in `pbta_histologies.tsv` but can be a separate table that we keep in the repo directly that serves as a dictionary to map `harmonized_diagnosis` and/or `broad_histology` to some subset of plotable categories. For color reasons, we should strive to have this set of categories be on the order of 10-15 labels.
> 
> At the moment, `broad_histology` has about 24 categories, but many of these only have a few samples, so it would seem likely these could be combined with others for plotting.
> 

To start working on a method to address #627, we need to first create a mapping from harmonized_diagnosis and/or broad_histology to larger groups -- as mentioned in the quote, we ideally want this to narrow down to 10-15 labels that can be used for large summary figures like those of Chromosomal instability breakpoints and CN Status heatmap and Oncoprint plotting. 

This mapping TSV file will be used in all relevant plotting modules and probably end up making `histology_color_palette.tsv` obsolete. 

#### Steps toward creating a mapping TSV table

Step 1) Create a mapping TSV notebook 

I think this mapping should take place in a Rmd that would live in the `/figures` folder and would use whatever the latest data release's pbta-histologies.tsv to create a separate TSV table that contains these columns for mapping. 

- `Kids_First_Biospecimen` 
- `broad_histology` 
-  `summary_group`   - a reasonable name that would be used on the plots themselves as labels 
- `hex_code` - a hex code that can be used as the color for plotting. 

This notebook's main function will be creating this `summary_group` and will ideally send out easy to interpret warnings for when a histology group from `pbta-histologies.tsv` is either missing but expected. 

`summary_group` will probably be created by 
1) Evaluate how many samples there are per `broad_histology` group.
2) Recode larger `summary_group`s based on their `broad_histology` groups. 

Step 2) We'll need someone with domain expert knowledge to check the groupings and make sure they are reasonable. 

Step 3) figures/README.md#color-palette-usage will need updating

Step 4) Implement the resulting mapping TSV file in all relevant modules and probably retiring `histology_color_palette.tsv`. This part is tracked separately here: #898 

#### What input data are required for this analysis?

`pbta-histologies.tsv`

[...]

PR #899, opened 2021-01-12
Mapping high level histology color labels -- Part 1

cansavvy:

This PR addresses Step 1 of 4 discussed in #897. This PR adds that main notebook described where we attempt to group biospecimens into "high level" histology groups that will be more practical for plotting purposes. 

Note that subsequent PRs will deal with the updating of the README and retirement of `histology_color_palette.tsv` in favor for the mapping table, `palettes/histology_label_color_table.tsv`, that is created in this notebook. 

#### What was your approach?

Copied from the notebook added in this PR itself: 
```
# Purpose: 

The histology label variables included in `pbta-histologies.tsv` from data releases are not always useful for visualizing the full set of biospecimens due to the large number of different values.
Having too many different possible values makes the colors harder to distinguish.
In addition, there are some groups that are represented by only a very few samples; giving such groups a distinct color may be counterproductive.

The goal of this notebook is to use the currently existing `broad_histology` groups from `pbta-histologies.tsv`, to form 10-15 "high level histology" group labels that can used for plotting purposes.

## The output table

The output of this notebook is a TSV file: `palettes/histology_label_color_table.tsv` that contains the following fields:

**Copied from `pbta-histologies.tsv`**:    
- `Kids_First_Biospecimen_ID` (from `pbta-histologies.tsv`)  
- All the original histology label variables (`broad_histology`, `short_histology`, etc.)  
  
**Created in this notebook**:  
- `display_group` - the high-level histology labels that should be used for plotting   
- `hex_codes` the direct colors that should be used for plotting  

With this info, `histology_label_color_table.tsv` can be used by all plots and figures that summarize high level data  while displaying histology information. 

# How `display_group` is made:

Here's how `broad-histology` groups are [combined into the higher-level groupings of `display_group`](#declare-new-equivalent-groups).

1) "Neuronal and mixed neuronal-glial tumor" and "Diffuse astrocytic and oligodendroglial tumor" were labeled as LGAT in previous releases so we will group these into `Low-grade astrocytic tumor` in `display_group`. 

2)  `Germ cell tumor`  samples are combined into the `Non-CNS tumor` group. 

3) `Benign tumor` and `Non-tumor` biospecimens are combined into a `Benign` group. 

4) Anything not in the above categories gets its `broad_histology` label carried over. 
```

#897 

- How do we feel about this notebook and its ability to deal with upcoming releases' versions of pbta-histologies.tsv 
- How do we feel about these broad histology groupings? Aka how scientifically wrong are they? (Note that we've always known "high level" groupings will not be more accurate or specific; that is not the goal here, the goal is to be broad and have something easier to visualize in a plot. 

Yes. The ones included can be. 

We have a list of colors and 16 + Normal groups. I have not tried it on an official plot yet. (upcoming PR) 

- [x] This analysis has been added to continuous integration. -- I've added it at the very top because the first few CI tests will need it once #898 is addressed and those plots use the table created here. 

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date. -- No, this will be in a subsequent PR. The README will require quite a bit of updating. 
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date. -- it is not because its not located in the analyses folder.


figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,194 @@
...
+Declare how many colors we need. 
+
+```{r}
+n_colors <- nrow(summary_groups)
+```
+
+Make a named list color key where histologies are the names. 
+
+```{r}
+subset_colors <- sample(color_palette, n_colors)
```
cansavvy, line 165: Even though I set the seed, this `sample()` changes with each re-run? That's not what I expected.

cansavvy: jashapiro helped me resolve this. This was my own misunderstanding on how `sample` + `set.seed` work together.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,237 @@
...
+input_dir <- file.path("..", "data")
+output_dir <- "palettes"
+```
+
+# Read in metadata 
+
+Which variables are we keeping for this table? 
+
+```{r}
+histology_variables <-
```
cansavvy, line 79: For ease of downstream use, I'm wondering if we should drop everything that's not `Kids_First_Biospecimen_ID` `summary_group` and `hex_codes`, because downstream we'll need to drop everything. OR do we save a separate file just like `histology_color_palette.tsv` but that has each biospecimen's `summary_group` and `hex_codes`.

jashapiro (commented):
This looks like a good start. I think my main comment is that we should try to be a bit more intentional about grouping things before resorting to "Other". There may well be reasonable groupings that take some of the small-count histologies and link them to large-count histologies that they are related to rather than just to other small-count groups. It may be helpful to have a table which shows the `integrated_diagnosis` values for each `broad_histology` to assist with such grouping.

My other suggestions are mostly about display. This is a notebook for internal evaluation, so having more easily visible data at each stage in tables is more valuable than some of the summaries or figures. 

I also suggest treating "Other" as a special group, with a defined color that remains constant. 

With regards to your question:
>For ease of downstream use, I'm wondering if we should drop everything that's not Kids_First_Biospecimen_ID summary_group and hex_codes, because downstream we'll need to drop everything. 

I think we should keep as much information as possible for now. It is potentially easier to evaluate groupings using the full table, and the cost of keeping it is essentially zero.

cansavvy:
@jharenza , we could use some more domain knowledge about these histologies at this point. 
In particular, which of these `display_group`s (which are mostly from `broad_histology`) would be most reasonable to combine into other groups? 

```
display_group	n
Lymphoma	2
Melanocytic tumor	2
Other tumor	2
Other astrocytic tumor	6
Metastatic tumors	9
Tumor of pineal region	10
Histiocytic tumor	12
Choroid plexus tumor	21
Pre-cancerous lesion	27
Non-CNS tumor	29
Mesenchymal non-meningothelial tumor	48
Meningioma	59
Tumors of sellar region	73
Benign	78
Tumor of cranial and paraspinal nerves	83
Ependymal tumor	174
Embryonal tumor	339
NA	833
Low-grade astrocytic tumor and adjacent	1034
```

jaclyn-taroni:
Without looking at the code, I wanted to point out that the `NA` should be normal samples. So you may want to include some kind of filtering to tumor samples if you have not already.

cansavvy:
> Without looking at the code, I wanted to point out that the `NA` should be normal samples. So you may want to include some kind of filtering to tumor samples if you have not already.

I did have that previously but with the changes I made most recently with Josh's review we lost that. I'll add it back.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,232 @@
...
+Let's print out the summary. 
+
+```{r}
+broad_summary %>% 
+  knitr::kable()
+```
+
+There's handful of very small groups (many are n = 2). 
+
+## Declare new equivalent groups
```
cansavvy, line 121: @jharenza this is where we set up the equivalent groups which are then recoded in the next section.

jharenza:
Hi @cansavvy - 

>1) "Neuronal and mixed neuronal-glial tumor" and "Diffuse astrocytic and oligodendroglial tumor" were labeled as LGAT in previous releases so we will group these into `Low-grade astrocytic tumor` in `display_group`. 

This is not true - These are three separate entities and really should not be combined. The neuronal and mixed glial tumors comprise of many distinct tumors, so they had a wider array of `short_histology`. The Diffuse astrocytic and oligodendroglial tumors were all `HGAT` (High-grade astrocytic tumors) and should be kept as a separate group. So, I would keep those three groups, especially since they have a bulk of the tumors.

>In particular, which of these display_groups (which are mostly from broad_histology) would be most reasonable to combine into other groups?

I think we can combine:
```
Lymphoma	2
Melanocytic tumor	2
Other tumor	2
Metastatic tumors	9
Non-CNS tumor	29
```

to `Non-CNS or Other tumor ` 

I think you can combine `Other astrocytic tumor	6` into LGAT for now, since it is a low-grade astrocytoma. Possibly, we can update that `broad_histology` in v19. 

I think that will leave us with 15.

cansavvy:
@jharenza , thanks for these comments. I'm going to implement them and then re-request you for a review when its ready.

cansavvy:
> I think that will leave us with 15.

This brings us to 16 + Normal. Are there any other combos that seem reasonable or should we leave it at that?

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,237 @@
...
+```
+
+Make a named list color key where histologies are the names. 
+
+```{r}
+# Set seed so the colors are consistent upon re-run
+set.seed(2021)
+
+# Sample from the 18 colors
+subset_colors <- sample(color_palette, n_colors)
```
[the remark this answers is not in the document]
cansavvy: Alternatively, we could drop this "Other" group from the large summary figures.

jharenza:
> Are there any other combos that seem reasonable or should we leave it at that?

I would be more prone to dropping `Benign tumor` first, then `Other` if we want to drop more.

cansavvy:
> I would be more prone to dropping `Benign tumor` first, then `Other` if we want to drop more.

If we are talking about dropping that completely from the figures, than I think we'd probably keep this data in the table for now, but change that at the figure scripts?

But I was more talking about groups we could combine into other groups.

jharenza:
Ah, yeah I think that is as far as we can combine right now.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
+---
+title: "Mapping histology labels for plots"
+output:   
+  html_notebook: 
+    toc: true
+    toc_float: true
+author: Candace Savonen for ALSF - CCDL
+date: 2020
```
jharenza, line 8: 2021 now :)

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+- `display_group` - the high-level histology labels that should be used for plotting   
+- `hex_codes` the direct colors that should be used for plotting  
+
+With this info, `histology_label_color_table.tsv` can be used by all plots and figures that summarize high level data  while displaying histology information. 
+
+# How `display_group` is made:
+
+Here's how `broad-histology` groups are [combined into the higher-level groupings of `display_group`](#declare-new-equivalent-groups).
+
+1) "Lymphoma", "Melanocytic tumor", "Other tumor", "Metastatic tumors", "Non-CNS tumor" are combined into a `Other non-CNS group` in `display_group`.
```
jharenza, line 37: This should be "Other or non-CNS tumor" rather than just other non-CNS tumor, so that people don't assume they're all non-CNS.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+# How `display_group` is made:
+
+Here's how `broad-histology` groups are [combined into the higher-level groupings of `display_group`](#declare-new-equivalent-groups).
+
+1) "Lymphoma", "Melanocytic tumor", "Other tumor", "Metastatic tumors", "Non-CNS tumor" are combined into a `Other non-CNS group` in `display_group`. 
+
+2) `Benign tumor` and `Non-tumor` biospecimens are combined into a `Benign` group. 
+
+3) `Other astrocytic tumor` samples are combined into the existing `Low-grade astrocytic tumor`.
```
jharenza, line 41: We should add that the other astrocytic tumors were low-grade SEGA tumors, therefore can be lumped (just so it is clear why we did this).

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+metadata <-
+  readr::read_tsv(file.path(input_dir, "pbta-histologies.tsv"), guess_max = 10000)
+```
+
+Now we'll select histology variables we mentioned above and so capitalization differences don't get in the way with this process, we will change everything to lower case for now. 
+
+```{r}
+working_metadata <- metadata %>% 
+  dplyr::select(Kids_First_Biospecimen_ID, sample_type, histology_variables) %>% 
+  dplyr::mutate_at(histology_variables, tolower)
```
jharenza, line 95: Love reading your code and seeing new cool functions of tidyverse! :)

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+```{r}
+histology_table <- working_metadata %>% 
+  dplyr::mutate(
+    # NAs are really Normals
+    display_group = tidyr::replace_na(broad_histology, "normal"),
+    # Now do the group combining
+    display_group = forcats::fct_collapse(display_group,
+      "low-grade astrocytic tumor" = lgat,
+      "other non-cns tumor" = other_tumor,
+      "benign" = begnign
```
jharenza, line 150: Typo "benign"

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+These groups we'll combine into a non-CNS/other tumor group.
+
+```{r}
+other_tumor <- c("lymphoma", "melanocytic tumor", "other tumor", "metastatic tumors", "non-cns tumor")
+```
+
+These groups we'll combine as a benign.
+
+```{r}
+begnign <- c("benign tumor", "non-tumor")
```
jharenza, line 130: Typo "benign"

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,240 @@
...
+
+```{r}
+histology_table <- working_metadata %>% 
+  dplyr::mutate(
+    # NAs are really Normals
+    display_group = tidyr::replace_na(broad_histology, "normal"),
+    # Now do the group combining
+    display_group = forcats::fct_collapse(display_group,
+      "low-grade astrocytic tumor" = lgat,
+      "other non-cns tumor" = other_tumor,
```
jharenza, line 149: As mentioned above - should be "other or non-cns tumor"

figures/mapping-histology-labels.Rmd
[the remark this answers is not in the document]
jharenza: We consider treating "Benign" as the grey group @jashapiro and @cansavvy - because it's more like a control and we would be less interested (maybe) in those genomics.

jharenza (approved):
Hi @cansavvy! Looks good- made a few minor comments, but otherwise it looks good!

jashapiro:
> We consider treating "Benign" as the grey group @jashapiro and @cansavvy - because it's more like a control and we would be less interested (maybe) in those genomics.

This makes sense to me. Maybe make both `Benign` and `Other` "special" so that their colors aren't random?

jharenza:
Sure!

figures/mapping-histology-labels.Rmd
cansavvy: Lol. Good catch.

figures/mapping-histology-labels.Rmd
cansavvy: Okay. We'll go with `Other tumor` then.

figures/mapping-histology-labels.Rmd
cansavvy: My brain kept looking at this when I wrote it because I thought it was wrong but then my brain convinced me it was right. LOL

figures/mapping-histology-labels.Rmd
cansavvy: Added this note.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,243 @@
...
+# Sample from the 18 colors
+subset_colors <- sample(color_palette, n_colors)
+names(subset_colors) <- display_group_df$display_group
+```
+
+We want `Other tumor` and the `Benign` group to both always be gray. 
+
+```{r}
+subset_colors[names(subset_colors) == 'other tumor'] <- "#D3D3D3"
+subset_colors[names(subset_colors) == 'benign'] <- "#808080"
```
jashapiro, line 209: Looking at these, I would probably switch them on the assumption that "benign" is less important than "other", so we we might want "benign" to be lighter in color.

I also feel like "normal" should either not get a color or get a specific color? I don't know if we have any plots where we include the normal samples.

figures/mapping-histology-labels.Rmd
```diff
@@ -0,0 +1,243 @@
...
+metadata <-
+  readr::read_tsv(file.path(input_dir, "pbta-histologies.tsv"), guess_max = 10000)
+```
+
+Now we'll select histology variables we mentioned above and so capitalization differences don't get in the way with this process, we will change everything to lower case for now. 
+
+```{r}
+working_metadata <- metadata %>% 
+  dplyr::select(Kids_First_Biospecimen_ID, sample_type, histology_variables) %>% 
+  dplyr::mutate_at(histology_variables, tolower)
```
jashapiro, line 95: My only concern here is that the output file now has different histology labels from the original file. 
One approach is to name your function, then you can use `broad_histology_lower` etc. for the comparison steps.
```suggestion
  dplyr::mutate_at(histology_variables, lower = tolower)
```

jashapiro (commented):
Looks good. Just a question about what to do with "Normal" colorwise, and some concern about capitalization consistency between files.

figures/mapping-histology-labels.Rmd
cansavvy: Not sure how to get back from `broad_histology_lower` to `broad_histology` labels for display group without doing bad things that would probably break in a future release?

jashapiro: You won't have to. Just use `select(histology_variables)` or similar to drop the `_lower` variants before output.

cansavvy: Okay. Slack message you because I misunderstood and thought you were referring to `display_group`'s borrowed labels from `broad_histology` but really I accidentally was altering ALL the `histology_variables` to be lower and I didn't intend that so we will get rid of that.

figures/mapping-histology-labels.Rmd
cansavvy: I think they almost always get dropped (at least the plots I've made they do) but I'm open to suggestions if we have a color we'd like for the Normals.

cansavvy: Normal biospecimens are never plotted, so I set the hex_code to NA.

cansavvy: Nope. Switched to black.

figures/mapping-histology-labels.Rmd
```diff
@@ -205,8 +205,14 @@ names(subset_colors) <- display_group_df$display_group
 We want `Other tumor` and the `Benign` group to both always be gray. 
 
 ```{r}
-subset_colors[names(subset_colors) == 'other tumor'] <- "#D3D3D3"
-subset_colors[names(subset_colors) == 'benign'] <- "#808080"
+subset_colors[names(subset_colors) == 'other tumor'] <- "#808080" 
+subset_colors[names(subset_colors) == 'benign'] <-  "#D3D3D3"
+```
+
+Normal biospecimen's should not get plotted, so we will put their hex code as black.
```
jashapiro, line 212:
```suggestion
Normal biospecimens should not get plotted, so we will put their hex code as black. 
```

jashapiro (approved):
One typo, but looks good!

PR #904, opened 2021-01-14
Using mapping histology groups for plotting -- Part 2 implemention (PR 1 of 4)

### LINKED ISSUE #897: Make mapping histology groups for plotting -- Part 1 creating the table

cansavvy:

This PR addresses Step 3 and part Step 4 discussed in #897 and #898. This PR uses the  `palettes/histology_label_color_table.tsv` table that is created by the notebook added in #899  

#### What was your approach?

This PR is 1 of a few PRs where I am updating things in response to the table added in #898

Here's what's accomplished in this PR: 

- The figures/README is updated to reflect this new `palettes/histology_label_color_table.tsv`
- figures/mapping-histology-labels.Rmd is now run in generate_figures.sh -- this way when figures are generated we know this table is up to date. 
- I've implemented the use of `palettes/histology_label_color_table.tsv` in one of the fives places (chromosomal-instability heatmaps) mentioned in #898 (needed to try out this some place so I knew how to update the README instructions). 
- I deleted the portion of the figures/scripts/color_palettes.R script that is responsible for creating `palettes/histology_color_palette.tsv` which is now in the process of being retired (as described in #898). 

Note that the `palettes/histology_color_palette.tsv` file itself is not deleted yet, I will do this in the next PR so the other modules that use this file don't break in CI yet. 

Closes #897 and 
begins to address #898 

- How do we feel about how the updated chromosomal-instability heatmap plots look? (see zip file below if you don't want to check out the branch)
- Is the figures/README updated appropriately and with enough detail? 

Yes. Take a look at the chromosomal-instability plots here to see what you think: 

[chromosomal-instability-heatmaps.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5837679/chromsomal-instability-heatmaps.zip)

Results are the same, the labels are changed. 

- [x] This analysis has been added to continuous integration. -- This was added in #899 

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date. --the README in figures is now up to date. 
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date. -- it is not because its not located in the analyses folder.


jashapiro (commented):
Overall, this looks good to me, but I had a thought that will will likely change some of the readme content if implemented:

One thing that I didn't comment on in previous PRs but I am now thinking more heavily about here is that we should not only have the display histology order standardized as well as the color. Arranging alphabetically doesn't really make sense, as we will want `Other` and probably `Benign` to be at the end, and there may be other ordering we want to use (largest to smallest?)

 I'm guessing this should happen when reading in the data, but we should probably make `display_group` a factor. My first thought was to just use `forcats::fct_relevel(display_group, "Other tumor", "Benign", after = Inf) to push those two categories to the end.

On further reflection, it may be worth adding a `display_order` column to the data table to make this easier to standardize across scripts: Then you could use  `forcats::fct_reorder(display_group, display_order)`  to make changes that would easily propagate, and can allow us to specify ordering of all factors just as easily. (Do we want to keep LGAT and HGAT next to each other, for example, or order by # of samples)

cansavvy:
> On further reflection, it may be worth adding a `display_order` column to the data table to make this easier to standardize across scripts: Then you could use  `forcats::fct_reorder(display_group, display_order)`  to make changes that would easily propagate, and can allow us to specify ordering of all factors just as easily.

I like this idea. Will implement. 

> (Do we want to keep LGAT and HGAT next to each other, for example, or order by # of samples)

I'd say small to large makes sense with the exception that Other and Benign should be at the end?

jashapiro:
> I'd say small to large makes sense with the exception that Other and Benign should be at the end?

Large to small, but yeah, I think that is the place to start until somebody expresses a different preference.

cansavvy:
I'm not sure how much this is or isn't an issue, but do different data types have different sample sizes so `display_order` would technically not always be large to small if we set it based on the full set of biospecimens?But my hunch is the consistency is still good so this is still a good plan?

cansavvy:
@jashapiro, I added a `display_order` column to the original mapping notebook and then updated the instructions and chromosomal instability notebook accordingly. See what you think.

figures/mapping-histology-labels.Rmd
```diff
@@ -172,6 +172,37 @@ if (nrow(display_group_df) > 17) {
...
+
+Get ranks in order of big to small and make them into a new column in `display_group_df`. 
+We will always want the "normal", "benign", "other_tumor" groups to come last so we are gonna remove them from this ranking and then add back when we get to `histology_table`. 
+
+```{r}
+control_like_groups <- c("normal", "benign", "other tumor")
+
+display_order_df <- display_group_df %>% 
+  dplyr::filter(!display_group %in% control_like_groups) %>%
+  dplyr::mutate(display_order = as.numeric(order(n, decreasing = TRUE))) %>%
```
jashapiro, line 185: Is `as.numeric()` needed?  It might display things like `1L` but I don't think those will print. 

Also, probably worth including display_group in the order to break ties by alphabet, which means that we have to get a bit tricky and use `-n`: 
```suggestion
  dplyr::mutate(display_order = order(-n, display_group))) %>% 
```

figures/mapping-histology-labels.Rmd
```diff
@@ -172,6 +172,37 @@ if (nrow(display_group_df) > 17) {
...
+  # Join on the display orders
+  dplyr::left_join(display_order_df, by = "display_group") %>% 
+  # Add ranks for the control like groups
+  dplyr::mutate(display_order = dplyr::case_when(
+    display_group == "benign" ~ max_rank + 1, 
+    display_group == "other tumor" ~ max_rank + 2, 
+    display_group == "normal" ~ max_rank + 3, 
+    TRUE ~ display_order
+    ))
+```
```
jashapiro, line 204: This ends up a bit convoluted I think. Why do two separate joins rather than sort out  & add the order at the same time as the display_groups earlier?

Also, I think you can simplify a lot and get rid of the confusing `case_when` and `max_rank` stuff:
My suggestion (not checked):
```
display_group_df %>%
  mutate(display_group = forcats::fct_reorder(display_group, n, .desc = TRUE) %>%
                         forcats::fct_relevel("benign", "other tumor", "normal", after = Inf),
         display_order = as.numeric(display_group))
```

analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd
```diff
@@ -56,9 +56,12 @@ Import color palettes.
 ```{r}
 # Import standard color palettes for project
 histology_label_mapping <- readr::read_tsv(
-  file.path(figures_dir, "palettes", "histology_label_color_table.tsv")
+  file.path(figures_dir, "palettes", "histology_label_color_table.tsv"), 
+  # Read in each column as a specific type - skip extra columns
+  col_types = paste0(c("c", rep("-", 6), "f", "i", "c"), collapse = "")
```
jashapiro, line 61: Do you need this? I guess it is helpful for skipping columns, but the factor will be automatically converted when you call fct_reorder, and the order should get read in correctly as numeric. You are implicitly specifying column order, which is something I try to avoid as it can be brittle.

analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd
```diff
@@ -56,9 +56,12 @@ Import color palettes.
...
   ) %>% 
-  dplyr::select(Kids_First_Biospecimen_ID, display_group, hex_codes)
+  # Reorder factor based on display_order
```
jashapiro, line 63:
```suggestion
  # Reorder display_group based on display_order
```

jashapiro (commented):
I made a suggestion for streamlining, because I was confused by the max_level stuff for longer than I want to admit.

figures/mapping-histology-labels.Rmd
jashapiro: This is subsumed by my later comment.

analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd
cansavvy: I agree that I'm not happy about this thing which is rigged with respect to column order. However, fct_reorder will not as the table here is read in. `display_group` does not get read in as a factor which is what is causing the problem. 

I can instead get rid of this `col_types` line and add back a `mutate` where we force it to be a factor first and then also add back a `select` to get to the four columns we want. (this is what I had previously. It includes more steps but it is less brittle with respect to column order).

figures/mapping-histology-labels.Rmd
cansavvy: Ah. This is much better.

analyses/chromosomal-instability/02a-plot-chr-instability-heatmaps.Rmd
jashapiro: Hmm, I've never had `forcats` complain about being given a character vector. Maybe I haven't used whatever old version we currently have.

cansavvy: I switched back to using a `select` and `as.factor()`. It may be more official "step" but it is more robust.

cansavvy: Okay. I must be imaging things. I dropped the as.factor and it works fine. Never mind everything I said. will push that change.

cansavvy:
Things have been streamlined. Thanks for those suggestions. @jashapiro

figures/mapping-histology-labels.Rmd
```diff
@@ -172,6 +172,28 @@ if (nrow(display_group_df) > 17) {
...
 ```
 
+# Make `display_order`
+
+Get ranks in order of big to small and make them into a new column in `display_group_df`. 
+We will always want the "normal", "benign", "other_tumor" groups to come last so we are gonna remove them from this ranking and then add back when we get to `histology_table`. 
+
+```{r}
+control_like_groups <- c("normal", "benign", "other tumor")
+
```
jashapiro, line 182: I don't think you use this again?
```suggestion
```

figures/mapping-histology-labels.Rmd
```diff
@@ -172,6 +172,28 @@ if (nrow(display_group_df) > 17) {
 }
 ```
 
+# Make `display_order`
+
+Get ranks in order of big to small and make them into a new column in `display_group_df`. 
+We will always want the "normal", "benign", "other_tumor" groups to come last so we are gonna remove them from this ranking and then add back when we get to `histology_table`.
```
jashapiro, line 178: Update to reflect changes.
```suggestion
We will always want the "normal", "benign", "other_tumor" groups to come last so we will push then to the end of the factor order. 
```

figures/mapping-histology-labels.Rmd
```diff
@@ -172,6 +172,28 @@ if (nrow(display_group_df) > 17) {
...
+Get ranks in order of big to small and make them into a new column in `display_group_df`. 
+We will always want the "normal", "benign", "other_tumor" groups to come last so we are gonna remove them from this ranking and then add back when we get to `histology_table`. 
+
+```{r}
+control_like_groups <- c("normal", "benign", "other tumor")
+
+display_order_df <- display_group_df %>% 
+  dplyr::mutate(display_group = forcats::fct_reorder(display_group, n, .desc = TRUE) %>%
+                  forcats::fct_relevel("benign", "other tumor", "normal", after = Inf),
+                display_order = as.numeric(display_group))
```
jashapiro, line 186:
```suggestion
                display_order = as.numeric(display_group)) # save the factor order for text table export
```

jashapiro (approved):
Looks good! Just a few little cleanup suggestions.

figures/mapping-histology-labels.Rmd
cansavvy: Oh yeah, I used it before but don't know.

PR #911, opened 2021-01-15
Using mapping histology groups for plotting -- Part 2 implemention (PR 2 of 4)

cansavvy:

This PR addresses Step 3 and part Step 4 discussed in #897 and #898. This PR is updating modules that should now be using the  `palettes/histology_label_color_table.tsv` table that is created by the notebook added in #899  

#### What was your approach?

This PR is 2 of a few PRs where I am updating things in response to the table added in #898

Here's what's accomplished in this PR and what files have changes that need review (the rest are updated plots): 

- `cnv-chrom-plot/cn_status_heatmap.Rmd` now uses the histology groupings and hex code colors from histology_label_color_table.tsv
- `mutational-signatures/known_signature.Rmd` now uses the histology groupings and hex code colors from histology_label_color_table.tsv
- I also needed to update the custom functions in `mutational-signatures/util` so that they used `display_group` instead of `short_histology` for the plotting.

I deleted the old plots, re-ran both of the notebooks mentioned above, and now am pushing the updated plots and results. That's where the crazy number of files changed comes from, a lot of plot updates. 

*Note that the `palettes/histology_color_palette.tsv` file itself is not deleted yet, I will do this in the last PR of this series so the other modules that use this file don't break in CI yet. 

- How do we feel about how the updated chromosomal-instability heatmap plots look? (see zip file below if you don't want to check out the branch)
- Is the figures/README updated appropriately and with enough detail? 

Yes. Here's updated cn status heatmap from cnv-chrom-plot: 

cn_status_heatmap

And here's some of the updated plots from the known signature analysis in `mutational-signatures` (I did not included the `signature_grouped_barplots` folder of plots; it was too big. 

bubble_matrix_cosmic_mutation_sig
bubble_matrix_nature_mutation_sig

Plots are different and now use the `display_group` histology labels instead of `short_histology`. 

Note that cnv-chrom-plot does not use the histology color palette, but it does still use `short_histology` for its plots -- *I have not updated this notebook here, we should discuss if we should update that in another PR. *

#### Documentation Checklist


analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -380,51 +375,36 @@ This annotation object strategy was originally from [chromosomal-instability](ht
...
 
 ```{r}
 # Make color key specific to these samples
-histologies_color_key_filtered <- unique(histologies$sample_color)
-names(histologies_color_key_filtered) <- unique(histologies$short_histology)
-
-# Drop this column so ComplexHeatmap isn't tempted to plot it
-histologies <- dplyr::select(histologies, -sample_color, -group_n)
+histologies_color_key_filtered <- unique(histologies$hex_codes)
+names(histologies_color_key_filtered) <- unique(histologies$display_group)
```
jashapiro, line 401: I'm a little worried about separate `unique()` calls preserving order... I think it should work fine, but it seems somehow dangerous to me. So I present an alternative below.

```suggestion
histologies_color_key_filtered <- histologies %>%
  dplyr::select(display_group, hex_codes) %>%
  dplyr::distinct() %>%
  dplyr::pull(hex_codes, name = display_group)
```

jashapiro (approved):
Looks good to me. Just one comment because I don't trust the machines.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
cansavvy: Yes. I like this better.

cansavvy: Turns out the name argument for `pull` is not in our version of dplyr for our docker image, so I can't use the last part, but I will add a distinct to this part but in my subsequent PR of this series.

Issue #877, opened 2020-12-22
Updated analysis: Use `harmonized_diagnosis` instead of `integrated_diagnosis` in downstream analysis


jashapiro:
#### What analysis module should be updated and why?

Possibly many, but I have not evaluated the extent yet. Certainly `interaction plots`, but other modules are likely involved.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

As discussed in #870, and illustrated in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/870#issuecomment-749690672, as of v18, the `integrated_diagnosis` column is not longer the most complete column from which to get diagnosis information. Analyses that relied on it should be updated to use the more complete `harmonized_diagnosis`.

Some of these changes will likely be caught by #874 (which found `interaction plots`), but there may remain other modules that need updating.

#### What input data should be used? Which data were used in the version being updated?

v18

The first step will be to identify modules that need updating.

#### When do you expect the revised analysis will be completed?

Uncertain, probably should be broken into multiple issues.

#### Who will complete the updated analysis?

TBD

cbethell:
It appears that the following files currently use `integrated_diagnosis`:

| Script                                           | Location                                |
|-----------------------|---------------------------------------------------------|
| `molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd`       | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/53c1f7c8534d32da98c20f8e9136899e7dfcce92/analyses/molecular-subtyping-HGG/00-HGG-select-pathology-dx.Rmd#L130 <br> We may want to add `harmonized_diagnosis` here rather than replace `integrated_diagnosis` |
| `molecular-subtyping-LGAT/00-LGAT-select-pathology-dx.Rmd`      | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/53c1f7c8534d32da98c20f8e9136899e7dfcce92/analyses/molecular-subtyping-LGAT/00-LGAT-select-pathology-dx.Rmd#L121 Again, we may want to add `harmonized_diagnosis` here rather than replace `integrated_diagnosis` |
[...]

PR #870, opened 2020-12-11
 PBTA Histologies: Integrated molecular subtyping to base histology (7 of N) 

kgaonkar6:
#### What scientific question is your analysis addressing?
Base histology needs to be updated with results from molecular-subtyping-pathology. In this module I'm adding molecular_subtype, integrated_diagnosis, short_histology, broad_histology and Notes.

#### What was your approach?
Read pbta-histologies-base.tsv and add molecular_subtype, integrated_diagnosis, short_histology, broad_histology and Notes. Check for changes, if there are updates we need to re-run summary modules/plots which use this information. Check for duplicates if any, and fix.

#### What GitHub issue does your pull request address?
#858 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
table

#### What is your summary of the results?
Final histology file for v18 release

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

Issue #838, opened 2020-11-09
Proposed Analysis: Add CNS_region mapping in histologies file to OpenPBTA analyses


jharenza:
#### What are the scientific goals of the analysis?
Add CNS_region to pbta-histologies.tsv using logic utilized on CHOP D3b end

#### What methods do you plan to use to accomplish the scientific goals?
Regions: Hemispheric, Midline, Spine, Ventricles, Posterior fossa, Optic pathway, Suprasellar, Other

Utilize this table to perform CNS_region mapping:
| Updated CNS Region | Site              | primary_site                                  |
|--------------------|-------------------|-----------------------------------------------|
| Hemispheric        | Frontal           | Frontal Lobe                                  |
|                    | Temporal          | Temporal Lobe                                 |
|                    | Parietal          | Parietal Lobe                                 |
|                    | Occipital         | Occipital Lobe                                |
|                    | —                 | Right Occipital                               |
|                    | —                 | Left temporo-parietal                         |
|                    | —                 | Right Parietal                                |
|                    | —                 | Right Frontal                                 |
|                    | —                 | Right Temporal                                |
| Midline            | Brainstem         | Brainstem                                     |
|                    | Midbrain          | Brain Stem- Midbrain/Tectum                   |
|                    | Pons              | Brain Stem- Pons                              |
|                    | Medulla           | Brain Stem-Medulla                            |
|                    | Corpus Collosum   | —                                             |
|                    | Thalamus          | Thalamus                                      |
|                    | Hypothalamus      | —                                             |
|                    | Basal Ganglia     | Basal Ganglia                                 |
|                    | Hippocampus       | Hippocampus                                   |
|                    | Pineal            | Pineal Gland                                  |
|                    | Tectum            | Left Thalmus                                  |
|                    | —                 | Right Thalmus                                 |
|                    | —                 | Bilateral Thalmus                             |
|                    | —                 | —                                             |
| Spine              | Cervical          | Spinal Cord- Cervical                         |
|                    | Thoracic          | Spinal Cord- Thoracic                         |
|                    | Lumbar            | Spinal Cord- Lumbar/Thecal Sac                |
|                    | Sacral            | —                                             |
|                    |                   | Spine                                         |
|                    | —                 | Spine NOS                                     |
| Ventricles         | Lateral           | —                                             |
|                    | Third             | —                                             |
|                    | Fourth            | —                                             |
|                    | Cerebral aqueduct | —                                             |
[...]

Issue #749, opened 2020-08-28
Add total, stranded RNA-Seq for 8 samples


jharenza:
#### What data file(s) does this issue pertain to?
pbta-star-log-final.tar.gz
pbta-star-log-manifest.tsv
pbta-mend-qc-manifest.tsv
pbta-mend-qc-results.tar.gz
pbta-isoform-counts-rsem-expected_count.polya.rds
pbta-isoform-counts-rsem-expected_count.stranded.rds
pbta-isoform-expression-rsem-tpm.polya.rds
pbta-isoform-expression-rsem-tpm.stranded.rds
pbta-fusion-arriba.tsv.gz
pbta-fusion-putative-oncogenic.tsv
pbta-fusion-recurrently-fused-genes-byhistology.tsv
pbta-fusion-recurrently-fused-genes-bysample.tsv
pbta-fusion-starfusion.tsv.gz
pbta-gene-counts-rsem-expected_count.polya.rds
pbta-gene-counts-rsem-expected_count.stranded.rds
pbta-gene-expression-kallisto.polya.rds
pbta-gene-expression-kallisto.stranded.rds
pbta-gene-expression-rsem-fpkm-collapsed.polya.rds
pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
pbta-gene-expression-rsem-fpkm.polya.rds
pbta-gene-expression-rsem-fpkm.stranded.rds
pbta-gene-expression-rsem-tpm.polya.rds
pbta-gene-expression-rsem-tpm.stranded.rds

#### What release are you using?

V16

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
For a few samples in which polyA RNA-Seq was performed, new RNA was rRNA-depleted and sequenced as total, stranded and we will process and release these.

jharenza:
For tracking:

**PNOC samples:**
BS_FXJY0MNH
BS_KABQQA0T
BS_D7XRFE0R
**CBTN samples:**
BS_SHJA4MR0
BS_HE0WJRW6
BS_8QB4S4VA
BS_FN07P04C
BS_SB12W1XT

jaclyn-taroni:
Included in v18 release (#857)

PR #912, opened 2021-01-15
Replace `integrated_diagnosis` with `harmonized_diagnosis` in sample distribution analysis module

### LINKED ISSUE #909: Updated analysis: Use `harmonized_diagnosis` in sample distribution analysis

`sample-distribution-analysis` should be updated as part of #877.

As noted in the [comment on 877](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/877#issuecomment-759796037):

The main instances of `integrated_diagnosis` that need to be replaced with `harmonized_diagnosis` are linked below (noting that there are three individual scripts/notebooks that will need to be updated in this module).
Although only the main instances (instances that would affect the output) are linked below, changes should be propagated throughout each of the scripts/notebooks (including comments, file names, etc.) and the notebooks and/or module should be re-run.

| Script                                           | Location                                |
|-----------------------|---------------------------------------------------------|
| `sample-distribution-analysis/01-filter-across-types.R`     | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/01-filter-across-types.R#L32 <br> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/01-filter-across-types.R#L58 |
| `sample-distribution-analysis/02-multilayer-plots.R` | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/02-multilayer-plots.R#L46 <br> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/02-multilayer-plots.R#L48 |
| `sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd` | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd#L218 <br> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd#L219 <br> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/sample-distribution-analysis/03-tumor-descriptor-and-assay-count.Rmd#L231 |

It appears that this module was last re-run with v17 data in merged PR #782.
We will want to re-run this module with v18.

~ 1 day (potential pitfalls are not obvious)

TBD

cbethell:

This PR addresses the following from issue #877
> As discussed in #870, and illustrated in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/870#issuecomment-749690672, as of v18, the `integrated_diagnosis` column is not longer the most complete column from which to get diagnosis information. Analyses that relied on it should be updated to use the more complete `harmonized_diagnosis`.

#### What was your approach?

Using the [comment on 909](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/909#issue-786219688), I was able to locate the main instances where `integrated_diagnosis` needed to be replaced `harmonized_diagnosis`.

I then used search and replace to find any additional instances that needed to be replaced, and re-ran the module `run-sample-distribution.sh` script.

This PR closes #909.

The plots and results files have been updated in this PR, with what I believe are the expected changes with the change from using `integrated_diagnosis` to using the more complete `harmonized_diagnosis`.

These plots/results should receive another close look to ensure that these changes are indeed expected based on the upstream changes to v18.

More specifically, as I note in the summary of results below:
> In `results/disease_expression.tsv`, the total number of samples is now 969 (using `sum(disease_expression$count)`), compared to the total number being 957 in the file last ran with v17 data.

Does this number seem expected?

Similarly,
> In `primary_sites_counts.tsv`, the total `number_of_types` are now 333 with the v18 data, compared to 243 with the v17 data (using `sum(primary_sites_counts$number_of_types)` to check).

Does this seem expected?

Yes.

The following plots have been updated:
- `plots/distribution_across_cancer_types.pdf`
- `distribution_across_cancer_types_treemap.pdf`

The following results files have been updated:
- `results/disease_expression.tsv`
- `results/plots_df.tsv`
- `results/primary_sites_counts.tsv`

As expected the `integrated_diagnosis` column has been replaced with the `harmonized_diagnosis` column in the results files.
This is accompanied with expected changes in naming (due to updates made in v18), as well as a change in the number of total samples.

In `results/disease_expression.tsv`, the total number of samples is now 969 (using `sum(disease_expression$count)` to check), compared to the total number being 957 in the file last ran with v17 data.

In `primary_sites_counts.tsv`, the total `number_of_types` are now 333 with the v18 data, compared to 243 with the v17 data (using `sum(primary_sites_counts$number_of_types)` to check).

#### Documentation Checklist


jaclyn-taroni:
I have not looked at the code. What did the total number of samples used to be? We have 8 new RNA-seq samples, but, depending on how that information is used, we may want to be summarizing on the _participant_ basis.

jaclyn-taroni:
Ah I see - 957 + 8 does not equal 969 🤔

cbethell:
> I have not looked at the code. What did the total number of samples used to be? We have 8 new RNA-seq samples, but, depending on how that information is used, we may want to be summarizing on the _participant_ basis.

Using [the comment tracking the additional 8 stranded RNA-seq samples](https://github.com/alexslemonade/openpbta-analysis/issues/749#issuecomment-721997172) and the [v18 release notes](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/release-notes.md#release-v18-20201123), I was able to determine that differences in numbers of samples for each category seems to indeed be due to the additional 8 samples and the more complete naming include in `harmonized_diagnosis` (which would be different from the previous values in `integrated_diagnosis`. For example, one of the added samples was classified as `Brainstem glioma- Diffuse intrinsic pontine glioma` and this category's count increased from 8 -> 9 `disease_expression.tsv`.

As for 
> In `primary_sites_counts.tsv`, the total `number_of_types` are now 333 with the v18 data, compared to 243 with the v17 data (using `sum(primary_sites_counts$number_of_types)` to check).

I was still unable to pinpoint the source of the difference in numbers here, even after taking a look at AlexsLemonade/OpenPBTA-analysis#838, but that does not rule it out!

jaclyn-taroni (approved):
These code changes LGTM! To respond to your questions about expected changes:

> In `results/disease_expression.tsv`, the total number of samples is now 969 (using `sum(disease_expression$count)` to check), compared to the total number being 957 in the file last ran with v17 data.

The total number of samples here refers to unique `Kids_First_Participant_ID`, `harmonized_diagnosis` pairs. So I think the question here is "Are there more participants with multiple `harmonized_diagnosis` values than we would expect?"

That's beyond the scope of what you're doing here, but I have filed #913 to track that. 

> In `primary_sites_counts.tsv`, the total `number_of_types` are now 333 with the v18 data, compared to 243 with the v17 data (using `sum(primary_sites_counts$number_of_types)` to check).

My understanding of `number_of_types` is that it counts the total number of `harmonized_diagnosis` values with at least one `Kids_First_Participant_ID` with that `primary_site` value. I think a lot of the increase is probably due to the fact that there are a lot more unique values in `harmonized_diagnosis` in v18 than in `integrated_diagnosis` in v17:

```R
> v17_df <- read_tsv("data/release-v17-20200908/pbta-histologies.tsv")
> v18_df <- read_tsv("data/release-v18-20201123/pbta-histologies.tsv")
> v17_df %>% 
+   select(integrated_diagnosis) %>% 
+   distinct() %>% 
+   count()
# A tibble: 1 x 1
      n
  <int>
1    64
> 
> v18_df %>% 
+   select(harmonized_diagnosis) %>% 
+   distinct() %>% 
+   count()
# A tibble: 1 x 1
      n
  <int>
1    92
```

But some of it may stem from the point above, too.

Issue #880, opened 2020-12-23
v18 breaks: Interaction plots failing -- tried using `broad_histology` 


cansavvy:
#### What analysis module should be updated and why?

With v18 data, interaction plots is broken. 
See the discussion on https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/876#issuecomment-749730138

#### What changes need to be made? Please provide enough detail for another participant to make the update.

It was proposed that changing these scripts to plot `broad_histology` instead of using `integrated_diagnosis` as they do now might fix this problem. 

I tried to implement this by replacing instances of `integrated_diagnosis` with `broad_histology` and re-running with the full data set but it gave the same error -- something about the `comb()` function within `coocurrence()` custom function. 

```
Error in combn(genes, m = 2) : n < m
Calls: coocurrence -> t -> combn
Execution halted
```

Here's the circle CI run where this error was initially found if you'd like to see more: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/4722/workflows/27d03a80-9cfe-415d-ad7f-f941c5ac8307/jobs/5260

#### What input data should be used? Which data were used in the version being updated?

v18 data

#### When do you expect the revised analysis will be completed?

Not sure

#### Who will complete the updated analysis?
Someone who is more familiar with `interaction-plots` ideally.

cansavvy:
I think #889 and #627 might help resolve some issues related to this.

cbethell:
The interaction plots script hard codes which categories to look at, so #889 will need to be tackled before this issue can be tackled as we will need the most accurate `broad_histology` categories for hard coding.

jaclyn-taroni:
I'm going to close this in favor of #910 because I think that's the strategy we should pursue. I will note on that issue that the module also needs to be added back to CI (uncommented out).

PR #915, opened 2021-01-19
Replace `integrated_diagnosis` with `broad_histology` in interaction plots module

### LINKED ISSUE #910: Updated analysis: Use `harmonized_diagnosis` in interaction plots module

`interaction-plots` should be updated as part of #877.

As noted in the [comment on 877](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/877#issuecomment-759796037):

The main instances of `integrated_diagnosis` that need to be replaced with `harmonized_diagnosis` are linked below (noting that there are three individual scripts/notebooks that will need to be updated in this module).
Although only the main instances (instances that would affect the output) are linked below, changes should be propagated throughout each of the scripts/notebooks (including comments, file names, etc.) and the notebooks and/or module should be re-run.

| Script                                           | Location                                |
|-----------------------|---------------------------------------------------------|
| `interaction-plots/scripts/01-disease-specimen-lists.R`    | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/interaction-plots/scripts/01-disease-specimen-lists.R#L67 <br> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/interaction-plots/scripts/01-disease-specimen-lists.R#L91 |
| `interaction-plots/scripts/02-process_mutations.R` | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/interaction-plots/scripts/02-process_mutations.R#L336 |
| `interaction-plots/scripts/03-plot_interactions.R`   | https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ab673df68e4f10124f58463dbeb87ea4f800f9d5/analyses/interaction-plots/scripts/03-plot_interactions.R#L133 <br> If/when the other notebooks noted above in the `interaction-plots` module, the comment here should also be updated |

It appears that the results in this module are from a run with v15 data (merged PRs #595 and #639).
`interaction-plots/01-create-interaction-plots.sh` was updated in merged PR #764 but appears that it was not re-ran.

We will still want to remember to re-run this module with the v18 data when making the updates mentioned above.

~ 1 day (potential pitfalls are not obvious)

TBD

cbethell:

This PR addresses the following from issue #877
> As discussed in #870, and illustrated in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/870#issuecomment-749690672, as of v18, the `integrated_diagnosis` column is not longer the most complete column from which to get diagnosis information. Analyses that relied on it should be updated to use the more complete `harmonized_diagnosis`.

#### What was your approach?

Using the [comment on 910](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/910#issue-786236948),as well as the search and replace tool, I was able to locate the instances where `integrated_diagnosis` needed to be replaced `harmonized_diagnosis`.

I then updated the module's `01-create-interaction-plots.sh` script to agree with the values in `harmonized_diagnosis` and re-ran the script.

This PR closes #910.

As noted above, I updated the module's `01-create-interaction-plots.sh` script to agree with the values in `harmonized_diagnosis`.

More specifically, I used the updated `gene_disease_top50.tsv` file to find what would now be the most common disease categories based on `harmonized_diagnosis` to test.
See lines 41-48: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ed3417f3904a317882e08c6d66722e0e840fe5a6/analyses/interaction-plots/01-create-interaction-plots.sh#L41-L48
Note, for some of the broader categories i.e. for LGAT, there are multiple `harmonized_diagnosis` values that can fall under it like `Low-grade glioma/astrocytoma, BRAF wildtype` and `Low-grade glioma/astrocytoma, BRAF V600E`. 

In this case, I chose the first instance seen for testing in this script. Would you agree with this decision?
I tried changing the names within the disease array to be more accurate and to agree with the more complete `harmonized_diagnosis` values but got the following error:
`Error in parse_args(OptionParser(option_list = option_list)) :  
  Error in getopt(spec = spec, opt = args) :  
  "H3" is not a valid option, or does not support an argument Execution halted`

I also replaced `Ganglioglioma` with `Meningioma` to account for the following R error:
`Error: Discrete value supplied to continuous scale` which appears to result from the minimum amount of samples not being met.

However, I believe that this did not successfully solve the CI failure as `Meningioma` still does not meet the minimum number of samples when using the CI files.

> However, I believe that this did not successfully solve the CI failure as `Meningioma` still does not meet the minimum number of samples when using the CI files.

Should we include logic here for running in CI? Perhaps to adjust the minimum number of samples needed when running in CI or perhaps not testing the individual diseases when running in CI? (Note that the latter is the only solution that has worked for me thus far when testing with the CI files).

Yes.

All of this module's output plots and results files have been updated in this PR, with what I believe are the expected changes with the change from using `integrated_diagnosis` to using the more complete v18 `harmonized_diagnosis`.

With the change from using `integrated_diagnosis` to `harmonized_diagnosis` came expected changes in naming (due to updates made in v18), as well as a change in the number of samples for some categories (again due to updates made in v18).

#### Documentation Checklist


jaclyn-taroni:
@cbethell what does the string that contains a particular `harmonized_diagnosis` do? So for 

```
   disease[HGAT]="High-grade glioma/astrocytoma, H3 wildtype" 
```

How is `"High-grade glioma/astrocytoma, H3 wildtype"` used? I'm asking because maybe we could change it to using regex such that one can do `"High-grade glioma/astrocytoma, H3 wildtype|<other harmonized diagnosis>"` and get everything one wants? Or we could do something a little further away from what already exists and use the outcomes of #897.

cbethell:
> @cbethell what does the string that contains a particular `harmonized_diagnosis` do? So for
> 
> ```
>    disease[HGAT]="High-grade glioma/astrocytoma, H3 wildtype" 
> ```
> 
> How is `"High-grade glioma/astrocytoma, H3 wildtype"` used? I'm asking because maybe we could change it to using regex such that one can do `"High-grade glioma/astrocytoma, H3 wildtype|<other harmonized diagnosis>"` and get everything one wants? Or we could do something a little further away from what already exists and use the outcomes of #897.

@jaclyn-taroni, it appears that `"High-grade glioma/astrocytoma, H3 wildtype"` is being used for labeling/grouping purposes. When I did attempt using regex like you suggest above, it appears that the script did not obey the regex expression as I got the following error: `Error in combn(genes, m = 2) : n < m
Calls: coocurrence -> t -> combn` (as was also [noted in the v18 breaks closed issue #880](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/880#issue-773798165)).

That said, perhaps your second suggestion is what may want here?

analyses/interaction-plots/01-create-interaction-plots.sh
```diff
@@ -40,12 +40,12 @@ combined_plot=${plot_dir}/combined_top50
...
-  disease[HGAT]="High-grade glioma"
-  disease[DMG]="Diffuse midline glioma"
-  disease[Ganglioglioma]="Ganglioglioma"
-  disease[Craniopharyngioma]="Craniopharyngioma"
+  disease[Diffuse-astrocytic-and-oligodendroglial-tumor]="Diffuse astrocytic and oligodendroglial tumor"
+  disease[Tumors-of-sellar-region]="Tumors of sellar region"
+  disease[Embryonal-tumor]="Embryonal tumor"
+  disease[Ependymal-tumor]="Ependymal tumor"
+  disease[Meningioma]="Meningioma"
+  disease[Neuronal-mixed-neuronal-glial-tumor]="Neuronal and mixed neuronal-glial tumor"
```
cbethell, line 48: In the most recent commit, I swapped `harmonized_diagnosis` with `broad_histology` and updated the shell script as seen above to use the broad histology values. I am not sure if this was the best way to represent these values in the array (using a `-` to separate words), so feedback is welcome here! Also, is this what we want?

jashapiro: I would have guessed that you couldn't use `-` in array names, but I would have been wrong! It seems to work just fine.

cbethell:
It appears that the fourth plot in the instance in which CI was failing ([this commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/915/commits/764daba56ff4eab6bad7a740f5ff3d60210a1549)) was the individual disease type plot which would have been LGAT in that case. Not completely sure why but it does appear to be specific to disease type implemented (based on tests passing in the [last commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/915/commits/8469fcb2abbc071db6d232d94aeec2f13ad780a2) where the disease type was swapped the only other disease type with a higher n `total_muts` according to `gene_disease_top50.tsv`):
```
gene	disease	                                                mutant_samples total_muts	
TP53	Diffuse astrocytic and oligodendroglial tumor	46	        57
BRAF	Low-grade astrocytic tumor	                       43	        43
```

jaclyn-taroni (approved):
These changes LGTM. I don't think these disease labels/groups are quite right, but the point of this particular pull request is to return things to an unbroken state which it does successfully! I've filed #917 to track identifying the "right" groups.

PR #918, opened 2021-01-20
Using mapping histology groups for plotting implementation (PR 3 of 4)

cansavvy:

This PR addresses Step 3 and part Step 4 discussed in #897 and #898. This PR is updating modules that should now be using the  `palettes/histology_label_color_table.tsv` table that is created by the notebook added in #899  

#### What was your approach?

This PR is 3 of a few PRs where I am updating things in response to the table added in #898

Here's what's accomplished in this PR and what files have changes that need review (the rest are updated plots): 
These scripts have all been updated to use the colors in `palettes/histology_label_color_table.tsv` and `display_group` instead of the old color palette and short_histology: 

- `oncoprint-landscape/01-plot-oncoprint.R` -- I re-ran these plots by calling `run-oncoprint.sh`
- `sample-distribution-analysis/02-multilayer-plots.R` -- I re-ran these plots by running `run-sample-distribution.sh`
- `figures/scripts/fig1-sample-distribution.R` -- I re-ran this by re-running this script itself. 

Oh and one last thing I squeaked in here is a minor change to `cn_status_heatmap.Rmd` that @jashapiro had commented on the previous PR in this series but it got merged before I implemented it. 

*Note that the `palettes/histology_color_palette.tsv` file itself is not deleted yet, I will do this in the last PR of this series so the other modules that use this file don't break in CI yet. 

- You may notice that the fourth oncoprint plot has two of the groups switched in order, though the color key is still correct. It looks like maybe `oncoprint` overrides factor level order based on sample size? How much do I need to go fuss with this?

- There's a couple questions I have about `sample-distrbution-analysis/02-multilayer-plots.R` since I was digging around in there:

1) There's a stray plot `Rplots.pdf`. I see its gitignored, but is there some reason it hangs in this folder? Can we make it go away?
2) The `plots/histology-treemap.html` doesn't work on my end (something that was in the master branch before I touched anything). It just shows a black rectangle. 

Yes See below where I pasted the updated plots. 

### Results

Plots are different and now use the `display_group` histology labels and the respective hex codes instead of `short_histology`. 

### Updated oncoprint plots:
Here's what the correctly updated plots look like (except for the ones I'm having trouble with from sample-distribution-analyses): 
all_participants_primary_only_goi_oncoprint
all_participants_primary_only_oncoprint
all_participants_primary-plus_goi_oncoprint
all_participants_primary-plus_oncoprint

### Non-final version of sample distribution plot: 
[distribution_across_cancer_types_treemap.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/5849739/distribution_across_cancer_types_treemap.pdf)

### Final version of sample distribution plot: 
fig1-openpbta-distribution

#### Documentation Checklist


analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
```diff
@@ -396,12 +396,17 @@ histologies <-
...
 ```{r}
+# Get a distinct version of the color keys
+histologies_color_key_df <- histologies %>%
+  dplyr::select(display_group, hex_codes) %>%
+  dplyr::distinct()
+
 # Make color key specific to these samples
-histologies_color_key_filtered <- unique(histologies$hex_codes)
-names(histologies_color_key_filtered) <- unique(histologies$display_group)
+histologies_color_key <- unique(histologies_color_key_df$hex_codes)
```
jashapiro, line 405: Do we need the `unique()` here still? The idea of using `distinct()` above was to remove this step. If  `unique(histologies_color_key_df$hex_codes) != histologies_color_key_df$hex_codes` we would presumably have a problem...

jaclyn-taroni:
> There's a stray plot Rplots.pdf. I see its gitignored, but is there some reason it hangs in this folder? Can we make it go away?

I think we would have to track down what, of the multiple plotting steps, is opening a graphics device. To me, it doesn't seem like a big enough deal to track this down.

analyses/cnv-chrom-plot/cn_status_heatmap.Rmd
cansavvy: Oh right! Yes. We should be able to delete that. Will delete.

jaclyn-taroni (approved):
These changes look good to me! And it looks like you have resolved @jashapiro's one comment. To address the other questions I have not commented on yet:

> - You may notice that the fourth oncoprint plot has two of the groups switched in order, though the color key is still correct. It looks like maybe `oncoprint` overrides factor level order based on sample size? How much do I need to go fuss with this?

This seems like something we would want to dig into when we are getting figures _polished_, but seems perfectly reasonable as is at the moment!
> 2) The `plots/histology-treemap.html` doesn't work on my end (something that was in the master branch before I touched anything). It just shows a black rectangle. 

🎟️ ? I'm not sure how heavily used that is, so it may be best just to remove it so we don't have to maintain it.

cansavvy:
> This seems like something we would want to dig into when we are getting figures _polished_, but seems perfectly reasonable as is at the moment!

Sounds good. 

> > 1. The `plots/histology-treemap.html` doesn't work on my end (something that was in the master branch before I touched anything). It just shows a black rectangle.
> 
> 🎟️ ? I'm not sure how heavily used that is, so it may be best just to remove it so we don't have to maintain it.

I don't see that its used other places, so I can just remove it while we are here.

PR #922, opened 2021-01-22
re-run tp53 classifier with v18

kgaonkar6:
Just a re-run of the tp53 classifier to add 5 new stranded samples added in v18

#### What was your approach?
Just a re-run:
docker exec -ti frosty_tesla bash -c "cd /home/rstudio/OpenPBTA && bash analyses/tp53_nf1_score/run_classifier.sh"

Related to #837 since we need the updated tp53_score in some of the annotation scripts.

There seem to be a lot of changes in analyses/tp53_nf1_score/results/TP53_NF1_snv_alteration.tsv from the sample order changing I believe since I don't see any actual values changing.

tables

Manual review of the results of classifier with the 5 stranded samples seems to have changed the values a little. Will plot values in #886 to see if drastic changes occured. 

#### Documentation Checklist


jharenza (approved):
Would this benefit from setting a seed or is the classifier set up such that it will always give us new results? cc @gwaygenomics

jaclyn-taroni:
> Would this benefit from setting a seed or is the classifier set up such that it will always give us new results?

One of the steps in the classifier is to scale by gene:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0ed49b68a62462d2c96495ffea1bdadd5b4ff835/analyses/tp53_nf1_score/01-apply-classifier.py#L21

So it makes sense to me that adding a few samples would have a (minor) impact on the values. If you're referring to the scores for the shuffled data, a seed gets set here:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0ed49b68a62462d2c96495ffea1bdadd5b4ff835/analyses/tp53_nf1_score/01-apply-classifier.py#L57

I think we only apply a classifier to _one_ shuffled version of the matrix - it might be worth doing that a few times to get an idea about variability in those scores, but I have not thought deeply about that.

jaclyn-taroni (approved):
:+1:  I think `analyses/tp53_nf1_score/__pycache__/utils.cpython-35.pyc` was probably an unintentional add - can you remove that and stop tracking (probably via a `analyses/tp53_nf1_score/.gitignore`) before this gets merged please?

jashapiro:
> 👍 I think `analyses/tp53_nf1_score/__pycache__/utils.cpython-35.pyc` was probably an unintentional add - can you remove that and stop tracking (probably via a `analyses/tp53_nf1_score/.gitignore`) before this gets merged please?

We can safely add `*.pyc` to the global ignore (but be sure to remove the files FIRST, or annoying trouble will come).

jaclyn-taroni:
I can make those changes and get this merged.

jaclyn-taroni:
Ah, I see that file was already in `master` which is why it is showing up in the diff still. Will merge once CI finishes!

jharenza:
> I think we only apply a classifier to _one_ shuffled version of the matrix - it might be worth doing that a few times to get an idea about variability in those scores, but I have not thought deeply about that.

I was referring to the classifier scores themselves, which change every time (wasn't referring to shuffle scores).

jaclyn-taroni:
Taking a look at the history of the `analyses/tp53_nf1_score/results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores.tsv`, the last couple commits look like changes in precision that affect only a few samples:

https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/d4b5b7a8395738018fdc86238640cfc0c25ffa3a#diff-28518f4529127b6de6e6406ffccca67bfcc2b1b4f9e261058f8e88bdc350228c

https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/3f8baa5686f9f5e771188a60a3bbb01cff3dc495#diff-28518f4529127b6de6e6406ffccca67bfcc2b1b4f9e261058f8e88bdc350228c

That leads me to believe that the more appreciable changes here are new samples + scaling related, which suggests no changes to the code are necessary.

PR #928, opened 2021-01-29
Splitting up #921: GSEA module changes

### LINKED ISSUE #926: Updated analysis: GSEA throws error with stranded collapse-rnaseq v18 data

I was working on updating the histology color palette for the final versions of the figures (#898) and was having trouble getting the gsea part of transcriptomic-overview figure to run after I had re-ran `collapse-rnaseq` that makes the input files for GSEA. 

How to duplicate this error: 1) be on the master branch with v18 data, run `bash collapse-rnaseq/run-collapse-rnaseq.sh` and then try to run `bash analyses/gene-set-enrichment-analysis/run-gsea.sh` (Or just run the stranded step). 

Something is not lining up as indicated by the `dimnames`. This error comes specifically when the stranded data from collapse-rnaseq from v18 data is run (poly-a runs without error) and [when the actual `GSVA::gsea()` step is run](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/f85831fe7e405b1c6b6c43274cabcc796b7bd3b1/analyses/gene-set-enrichment-analysis/01-conduct-gsea-analysis.R#L98)

```
Estimating GSVA scores for 50 gene sets.
Computing observed enrichment scores
Estimating ECDFs with Gaussian kernels
Using parallel with 8 cores
  |======================================================================| 100%
Error in dimnames(x) <- dn :
Calls: <Anonymous> ... <Anonymous> -> .local -> .gsva -> rownames<- -> rownames<-
In addition: Warning message:
In mclapp(gset.idx.list, ks_test_m, gene.density = rank.scores,  :
  scheduled core 5 did not deliver a result, all values of the job will be affected
Execution halted
```

I can dig into this more, but wanted to post this in case someone more familiar with the `collapse-rnaseq` and/or `gene-set-enrichment-analysis` has more insight. 

It does not fail in CI, so this has something to do with the full set as opposed to the subset of data used in CI. 

I haven't totally pinpointed what the problem is. But I can dig into it more. Just wanted to post in case someone knows off the bat. 

v18 data

We can have a better idea about this once we have a better idea about the source of the problem.

cansavvy:

This PR is a split off of #921 where I am updating the transcriptomic overview plot in response to the histology color palette table added in #898.

This module focuses on updating `gene-set-enrichment analysis`. 

This PR updates gene-set-enrichment-analysis module to use the display group and hex codes of the  `palettes/histology_label_color_table.tsv` table that is created by the notebook added in #899  

#### What was your approach?

This PR updates one of the three modules that are ultimately used in the transcriptomic-overview plot that is ultimately updated in #921. 

`gene-set-enrichment analysis` updates: 
- `gene-set-enrichment-analysis/02-model-gsea.Rmd` was updated from using `short_histology` for the gsea model to using `display_group`
- Related to `gene-set-enrichment-analysis`, the source script: `analyses/gene-set-enrichment-analysis/util/hallmark_models.R` needed a `dplyr::` in one spot because it wouldn't run. 
- There was also a slight detour error that need to be fixed (but ended up being a quick fix) in `01-conduct-gsea.R` See #926 
- After these changes I re-ran `run-gsea.sh`

*Note that the `palettes/histology_color_palette.tsv` file itself is not deleted yet, I will do this in the last PR of this series (#927) so the other modules that use this file don't break in CI yet. 

Closes #926 

- Is there anything I missed? 

Yes 

### Results

Results files that were reporting `short_histology` are now replaced with files that report `display_group`. 

#### Documentation Checklist


cbethell (approved):
Based on the upstream changes made in #899, the updates to the notebooks/scripts and the resulting files in this PR make sense!

I double checked this module and it does not seem that you missed anything and I was able to rerun the module's bash script and reproduce your results. That said, LGTM!

PR #929, opened 2021-01-29
Splitting up #921: Immune-deconv changes

### LINKED ISSUE #926: Updated analysis: GSEA throws error with stranded collapse-rnaseq v18 data

cansavvy:

This PR is a split off of #921 where I am updating the transcriptomic overview plot in response to the histology color palette table added in #898

This PR updates immune-deconv module to use the display group and hex codes of the  `palettes/histology_label_color_table.tsv` table that is created by the notebook added in #899  

#### What was your approach?

This PR updates one of the three modules that are ultimately used in the transcriptomic-overview plot that is ultimately updated in #921. 

`immune_deconv` updates:
- `analyses/immune-deconv/01-immune-deconv.R` need to be updated to use `display_group` instead of `short_histology`.
- Also needed to update `analyses/immune-deconv/02-summary-plots.R` in the same way
- After these changes I re-ran `OPENPBTA_DECONV_METHOD="mcp_counter" ./analyses/immune-deconv/run-immune-deconv.sh` (The CIBERSORT.R Doesn't work unless you have that CIBERSORT.R script which we do not). 

*Note that the `palettes/histology_color_palette.tsv` file itself is not deleted yet, I will do this in the last PR of this series (#927) so the other modules that use this file don't break in CI yet. 

Closes #926 

- I ran it with the DECONV_METHOD used in circle CI and in the generate-figures script, but I notice there were originally plots for this method. Does this mean we don't really want this methods plots? 
- I have more questions about why we are keeping CIBERSORT based methods since we can't re-run this (and haven't for almost a year). (We don't have access to the CIBERSORT.R script needed for this). 

Yes?

### Results

The results themselves don't really change here but will change in transcriptomic-overview. 

#### Documentation Checklist


analyses/immune-deconv/02-summary-plots.R
```diff
@@ -29,13 +29,9 @@ deconvout <- opt$input
 output <- opt$output
 load(deconvout) 
 
-# if short histology is Medulloblastoma or ATRT, use them as broad histology
```
cansavvy, line 32: I didn't think this was relevant anymore so I deleted this step completely, but someone should let me know if something like this is still needed.

jharenza: confirmed this is ok to delete with the switch to harmonized dx

cbethell (commented):
The changes made to the scripts in this PR look good to me!

It appears that parts of the `heatmap_MCPcounter.pdf` and `heatmap_xCell.pdf` plots have been cutoff so the sizing will likely need to be adjusted.

As for your point in this PR's original comment:
> I have more questions about why we are keeping CIBERSORT based methods since we can't re-run this (and haven't for almost a year). (We don't have access to the CIBERSORT.R script needed for this)

I also had the same question so I took a look at the `README`... 

Per this module's `README.md`, 
> "We chose CIBERSORT (abs.) over MCP-Counter as it has the greatest number of overlapping immune cell types with that of xCell".

Also stated that, 
> "We recommend using MCP-Counter instead of CIBERSORT (abs.) when these files are not available to the user".

That said, I would argue that if this PR in particular goes in as is, we should update the `README.md` to reflect the change in method used.

cansavvy:
> Per this module's `README.md`,
> > "We chose CIBERSORT (abs.) over MCP-Counter as it has the greatest number of overlapping immune cell types with that of xCell".
> > "We recommend using MCP-Counter instead of CIBERSORT (abs.) when these files are not available to the user".

Yes. I saw these recommendations in the README. This is part of my question. Why are we keeping this around if it isn't something we can run? It might be something for a separate issue. 

On the separate note, I did notice the plots are trimmed badly, I'll look into fixing that.

cansavvy:
@cbethell , I tracked the CIBERSORT method issue mentioned above here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/931 and I fixed the plot margins. Let me know what you think.

cbethell (approved):
With issue #931 filed, this PR LGTM! 👍

Issue #919, opened 2021-01-21
Proposed Analysis:


natemella:
We would like to propose a continuation of the issue [448](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/448).

#### What are the scientific goals of the analysis?

Batch effects are commonly observed in RNA expression data. Such effects are likely less pronounced in RNA-Seq data than microarray data. However, batch effects may bias conclusions of studies that do not account for these effects. We wish to evaluate the level to which batch effects are present in the RNA-Seq data from this study and create alternative versions of the data that have been adjusted for batch effects.

#### What methods do you plan to use to accomplish the scientific goals?

We propose to use the BatchQC tool to evaluate the data for batch effects. BatchQC provides various visualization tools for evaluating batch effects. We will use these and prepare a summary based on our findings. ComBat is a widely used method that uses empirical Bayes methods for correcting batch effects. Currently, we are unsure whether batch is known for these samples. If batch is known, we will adjust for it using ComBat. If not, we will use SVA to identify surrogate variables and plot those against variables that may have a confounding effect.

#### What input data are required for this analysis?

RNA-Seq data. We will start with gene-summarized values but may also work with transcript-level data.

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?

10-20 hours. It will require at least two steps (examining the data using MultiQC and applying ComBat/sva).

#### Who will complete the analysis (please add a GitHub handle here if relevant)?

@nathanmella and @srp33

#### What relevant scientific literature relates to this analysis?

https://www.ncbi.nlm.nih.gov/pubmed/20838408
https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5167063/
https://www.ncbi.nlm.nih.gov/pubmed/22257669
_______________________________________________

As shown in our [previous work](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/448), we found identified the extent to which batches and histologies are confounded. We plan to finish this analysis and complete the following:
1. Transfer R scripts to an R Notebook file that is documented with the corresponding graphs
2. Clean up code with all functions living in one file in the directory
3. Update Docker File
4. Submit a pull request

jaclyn-taroni:
Linking the closed pull request relevant to this analysis as well: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/628

jharenza:
[...]

PR #930, opened 2021-02-01
#837 updated: RNA QC compare stranded vs polya matched sample_ids

kgaonkar6:
Is there a difference between stranded and polya samples and if so should we use 1 over other while annotating by TP53 status by sample_id? 

#### What was your approach?
matched tp53_score_stranded and tp53_score_polya by sample_id ( only for sample_ids that have multiple RNA_library). For 7316-85 since there are multiple standed sample I took a mean of all samples.

#837

This was not part of the original ticket but as part of #837 we are also compiling the tp53 status and @jharenza  found multiples which lead to additional investigation and 1 part of QC requested in [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/886#issuecomment-769489607) was to compare stranded and polya

image

Yes

plots in notebook

On average it seems the polya samples have lower tp53_scores compared to matched stranded samples.

#### Documentation Checklist


jharenza:
Hmm, there definitely looks like a bias there. However, for the four samples with scores below 0.5, maybe this is within the random variability of the classifier. For 7316-1455, both scores are high, but for 7316-161, the stranded sample would have predicted oncogenic TP53 and the polyA would not. Sorry to keep asking for things - can you create a table of TP53 alterations for each of these samples? (ie - do only 1455 and 161 have alterations)?

@gwaygenomics thoughts on this?

kgaonkar6:
Thanks for the review @jharenza! 
I just copied the rows from compiled TP53 alterations [here](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/integrate_tp53_alt/analyses/tp53_nf1_score/results/tp53_altered_status.tsv)  for 7316-1455 and 7316-161 and added the RNA_library info pre RNA bs id, let me know if you need additional info

sample_id | RNA_library | Kids_First_Biospecimen_ID_DNA | Kids_First_Biospecimen_ID_RNA | cancer_predispositions | tp53_score | SNV_indel_counts | CNV_loss_counts | HGVSp_Short | copy_number | hotspot | activating | tp53_altered
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
7316-161 | stranded |BS_W2QCHQ7E | BS_SHJA4MR0 | None documented | 0.7719887631074178 | 2 | 0 | p.R342P, p.P72Lfs*48 | NA | Yes |   | loss
7316-161 | poly-A |BS_W2QCHQ7E | BS_X0XXN9BK | None documented | 0.4490920924301138 | 2 | 0 | p.R342P, p.P72Lfs*48 | NA | Yes |   | loss
7316-1455 | stranded | NA | BS_HE0WJRW6 | Li-Fraumeni syndrome | 0.9526901023341534 | 0 | 0 | NA | NA |   |   | loss
7316-1455 | poly-A | NA | BS_HWGWYCY7 | Li-Fraumeni syndrome | 0.868128930617709 | 0 | 0 | NA | NA |   |   | loss

jharenza:
Hmm. I am wondering if we should utilize batch correction once completed via #919 to rerun the classifier for this dataset. I wonder if the scores would be closer together for polyA and stranded in that case. I don't really want to choose one over the other without some additional investigation, but for the most part, while there is a bias, they agree and for `7316-161`, although the polyA score is just under 0.5, it is still considered a loss via evidence. This also gets back to perhaps a re-thresholding of the scores for what determines inactivation/oncogenic TP53 which was discussed early on. So, I think for now, perhaps we can average those scores instead of discarding one or the other?

jharenza (approved):

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
```diff
@@ -128,3 +130,56 @@ negative correlation estimate -0.139 at p value 0.297
...
+ggplot(sample_id_multi_library_scores_df,
+       aes(y=value,x=as.numeric(factor(sample_id_multi_library_scores_df$variable)))) +
+  geom_point()+
+  geom_line(aes(color=sample_id)) +
+  xlim(c("stranded",
+         "polya")) + 
+  xlab("RNA_library") + ylab("tp53_score")
+```
+
+On avearage matched polya samples have lower tp53 classifier scores, should we only keep stranded scores when we can ?
```
jharenza, line 184:
```suggestion
On average matched polya samples have lower tp53 classifier scores, should we only keep stranded scores when we can?

We could consider utilizing batch correction once completed via [#919](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/919) to rerun the classifier for this dataset. We can then check whether the scores would be more similar for polyA and stranded. Instead of choosing one over the other without some additional investigation, perhaps we take an average. For the most part, while there is a bias, the scores agree and for 7316-161, although the polyA score is just under 0.5, it is still considered a loss via evidence. Once we investigate all of the evidence, we might consider a re-thresholding of the scores for what determines inactivation/oncogenic TP53 (discussed earlier).
```

jharenza (approved):
just added the comments to the notebook

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
```diff
@@ -128,3 +130,57 @@ negative correlation estimate -0.139 at p value 0.297
...
+            suffix = c("_stranded","_polya")) %>%
+  filter(sample_id %in% sample_id_multi_library$sample_id) %>%
+  group_by(sample_id,) %>%
+  # because 7316-85 has multiple stranded I'm taking a mean here 
+  # A tibble: 8 x 2
+  #   Kids_First_Biospecimen_ID tp53_score_stranded
+  #   <chr>                                   <dbl>
+  # 1 BS_59ZJWJTF                            0.299 
+  # 2 BS_QYPHA40N                            0.0665
+  # 3 BS_SB12W1XT                            0.229
```
jaclyn-taroni, line 163: In my opinion, you are better off _showing this output_ in a chunk rather than including it in a comment that could easily get out of date.

jaclyn-taroni (commented):
I had one minor comment about showing some information that's currently in a comment that could get out of date, but other than that this looks good to me!

analyses/tp53_nf1_score/02-qc-rna_expression_score.Rmd
kgaonkar6: thanks for the review! That's a great point, I've now added a chunk to display the output

jaclyn-taroni (approved):
Will merge once CI passes 🚀

Issue #941, opened 2021-02-12
Updated analysis: chordoma subtyping to use pathology diagnosis


jharenza:
#### What analysis module should be updated and why?
chordoma subtyping

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Along with #608, we also need to switch from using `short_histology` to using `pathology_diagnosis`. Thanks to @baileyckelly's eagle 👀 , it was brought to my attention that some of the samples in this module are not actually chordomas, but choroid plexus tumors, and this was my bad when assigning `short_histology` early on. This is fixed in the current release, but will decrease our N here.

| Kids_First_Participant_ID | Kids_First_Biospecimen_ID_DNA | Kids_First_Biospecimen_ID_RNA | pathology_diagnosis      | sample_id |
|---------------------------|-------------------------------|-------------------------------|--------------------------|-----------|
| PT_F1086Z0A               | BS_P4D8Y3S8                   | BS_BWZTMWTM                   | Chordoma                 | 7316-1101 |
| PT_HFQNKP5X               | BS_YB07VF1X                   | BS_9GN1QA3Q                   | Chordoma                 | 7316-2935 |
| PT_HFQNKP5X               | BS_EJ9JKM1C                   | BS_6F49F7WH                   | Chordoma                 | 7316-3295 |
| PT_WTMSD2WB               | BS_67PX06P3                   | BS_JTBM5TSE                   | Chordoma                 | 7316-3632 |
| PT_HFQNKP5X               | BS_DBDXCXT5                   | BS_5B6XZ7YP                   | Chordoma                 | 7316-4062 |
| PT_HFQNKP5X               | BS_GFM6EA61                   | BS_FBJ516WW                   | Chordoma                 | 7316-921  |
| PT_3WA7SBQ6               | BS_0VRSD9V3                   | BS_HN8DE43A                   | Choroid plexus carcinoma | 7316-2248 |
| PT_41BCPB7R               | BS_64HCD9K3                   | NA                            | Choroid plexus carcinoma | 7316-406  |
| PT_7TRGHZBK               | BS_W36RZSFA                   | BS_XEVMEYFS                   | Choroid plexus carcinoma | 7316-431  |
| PT_YZ8A8A36               | BS_GSVCN2XC                   | BS_59FR1NC2                   | Choroid plexus carcinoma | 7316-723  |

#### What input data should be used? Which data were used in the version being updated?
V18

#### When do you expect the revised analysis will be completed?
[...]

PR #944, opened 2021-02-16
Chordoma molecular subtyping update: filter by path_dx and bug fix

kgaonkar6:
molecular-subtyping-chordoma was using short_histology to filter but to be consistent with other subtyping modules we are now using pathology_diagnosis to filter for chordoma samples. I'm also fixing a bug where the RNA and DNA samples were swapped.

#### What was your approach?
- Update the filtering criteria to use pathology_diagnosis == "Chordoma" and 
- Fix assignment of ids in 
https://github.com/kgaonkar6/OpenPBTA-analysis/blob/4a7e268aba969912797a2819eb7b4fba93440032/analyses/molecular-subtyping-chordoma/01-Subtype-chordoma.Rmd#L174-L182

#941 

Just want to verify if we want to keep sample with 2 copies of SMARCB1 as loss or call them diploid? 
Screen Shot 2021-02-16 at 11 08 29 AM

We would need  a column "molecular_subtype" to use to assign an integrated_diagnosis to chordoma subtyping output in  molecular-subtype-pathology module. Do we think we have all criterias in this module to assign a subtype ? ( Some of this would later be updated using the path-review [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/608#issuecomment-778472130))

Yes

tables

7 sample_ids are now filtered as Chordoma samples

#### Documentation Checklist


jharenza (approved):
Copying [this comment ](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/941#issuecomment-780029673 )over. Once this is done, we can merge.


Regarding the comment, shouldn't the path review update be in molecular_subtyping_pathology?

kgaonkar6:
> Copying [this comment ](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/941#issuecomment-780029673)over. Once this is done, we can merge.

Sorry I editted your comment 😅. I had a question regarding the comment, shouldn't the path review update be in molecular_subtyping_pathology?

jharenza:
Oh yes, seems better!

PR #945, opened 2021-02-17
Evaluate TP53 scores using updated TP53 loss annotation

kgaonkar6:
TP53 loss classifier score evaluation with updated TP53 loss annotation ( double hit condition )
```
               TP53 loss status==1
               # IF
               # both SNV/CNV counts are 1 or more than 1 so can be considered as a TP53 double hit
               # SNV == 1 and cancer_predisposition == Li-Fraumeni syndrome
               # SNV == 1 and cancer_predisposition == Li-Fraumeni syndrome
               # SNV count is more than 1 so consider it as a double hit  
               # ELSE
               # TP53 loss status ==0 
```

#### What was your approach?
I updated the evaluation code to fit the new annotation file input and also to add the condition to check for double hit conditions to assign tp53(loss)_status ==1.

#837 

Are all conditions included for TP53 loss ( in the annotation we also used tp53_score >0.5 along with another hit (snv/cnv/germline_caner_predisposition but will not be included here since we want to evaluate the score) 

I've removed NF1 from the evaluation plot since we didn't identify any specific annotation for the gene. Does that sound ok for now and we can update when we have the NF1 loss annotated ?

Yes

plots: 
results/stranded_TP53.png 
results/polya_TP53.png 

classifier score AUC is now 0.92 fro stranded and 0.71 for polya rna data, just a note only 24 stranded and 8 polya are considered as true TP53 loss from SNV/CNV/cancer_predisposition conditions.

#### Documentation Checklist


jharenza (approved):
I think this looks great - the stranded ROC is much improved with the right set of TPs! 

I will let @jashapiro evaluate the code since I am not well-versed in Python. 

The only comment I have is maybe we should output the data file used for this.

analyses/tp53_nf1_score/05-evaluate-classifier.py
```diff
@@ -65,45 +65,33 @@
...
-)
-full_status_df.head(3)
-full_status_df[full_status_df > 1] = 1
-full_status_df = full_status_df.reset_index()
-full_status_df = full_status_df.drop(["No_TP53_NF1_alt"], axis=1)
+full_status_df['tp53_status']= full_status_df.apply(
+    lambda x: (1
+               # IF
+               # both SNV/CNV counts are more than 1 so can be considered as a TP53 double hit
+               if (x['SNV_indel_counts'] !=0 and x['CNV_loss_counts'] != 0 or
```
jashapiro, line 81: I am not sure I understand this line:  the line above says "more than 1" but do you mean "1 or more"? If so, I would probably change these from `!= 0` to `>= 1` just for clarity.

The other minor point is that I tend to find this `ifelse` structure confusing if it spans more than one line. I would use the more conventional:
if( condition1 or
    condition2):
  1 
else:
  0

analyses/tp53_nf1_score/05-evaluate-classifier.py
```diff
@@ -65,45 +65,33 @@
...
+                   # SNV == 1 and cancer_predisposition == Li-Fraumeni syndrome
+                   (x['SNV_indel_counts'] == 1 and (x['cancer_predispositions'] == "Li-Fraumeni syndrome")) or
+                   # SNV == 1 and cancer_predisposition == Li-Fraumeni syndrome
```
jashapiro, line 84:
```suggestion
                   # CNV == 1 and cancer_predisposition == Li-Fraumeni syndrome
```

jashapiro (approved):
This looks good to me. Seems like the classifier is doing much better with these changes! Just some minor style comments.

analyses/tp53_nf1_score/05-evaluate-classifier.py
kgaonkar6: updated the code with this suggestion, it does look more readable. Thanks!

analyses/tp53_nf1_score/05-evaluate-classifier.py
kgaonkar6: I updated the comment . This is not major but added all the conditions in the for loop chunk since I adding the comment per condition in ifelse was causing a syntax error because of indentation I guess ? 🤔

analyses/tp53_nf1_score/Untitled.ipynb
```diff
@@ -0,0 +1,113 @@
+{
```
jashapiro, line 1: Remove this notebook before merging!

analyses/tp53_nf1_score/Untitled1.ipynb
```diff
@@ -0,0 +1,1234 @@
+{
```
jashapiro, line 1: This one too!

jashapiro (commented):
Looks good; that code is much more clear!

But a couple extra notebooks snuck in...

analyses/tp53_nf1_score/05-evaluate-classifier.py
```diff
@@ -65,45 +65,30 @@
...
-full_status_df.head(3)
-full_status_df[full_status_df > 1] = 1
-full_status_df = full_status_df.reset_index()
-full_status_df = full_status_df.drop(["No_TP53_NF1_alt"], axis=1)
+# Obtain a binary status matrix
+for idx, val in enumerate(full_status_df.itertuples()):
+    # if both SNV/CNV counts are 1 or more than 1 this sample can be considered as a TP53 double hit
+    # if SNV or CNV has 1 hit with Li-Fraumeni syndrome cancer predisposition can be considered as double hit
+    # if SNV has more than 1 hit sample can be considered as double hit
+    if full_status_df.loc[idx, 'SNV_indel_counts'] >= 1 and full_status_df.loc[idx,'CNV_loss_counts'] >= 1 or full_status_df.loc[idx,'SNV_indel_counts'] == 1 and (full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome") or full_status_df.loc[idx,'CNV_loss_counts'] == 1 and (full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome") or full_status_df.loc[idx,'SNV_indel_counts'] >1:
```
jashapiro, line 82: Actually, can you split this onto multiple lines as well? I missed how very long this got...

analyses/tp53_nf1_score/Untitled.ipynb
kgaonkar6: Oh I was testing the code up date here as well, removed it now

analyses/tp53_nf1_score/05-evaluate-classifier.py
kgaonkar6: Thanks for the review, I've updated to multi lines and grouped with parenthesis.

analyses/tp53_nf1_score/05-evaluate-classifier.py
```diff
@@ -65,46 +65,38 @@
...
+        full_status_df.loc[idx,'CNV_loss_counts'] >= 1):
+        full_status_df.loc[idx,'tp53_status']=1
+    elif (full_status_df.loc[idx,'SNV_indel_counts'] == 1 and 
+          full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome"):
+        full_status_df.loc[idx,'tp53_status']=1
+    elif (full_status_df.loc[idx,'CNV_loss_counts'] == 1 and 
+          full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome"):
+        full_status_df.loc[idx,'tp53_status']=1
+    elif (full_status_df.loc[idx,'SNV_indel_counts'] >1):
+        full_status_df.loc[idx,'tp53_status']=1
```
jashapiro, line 91: I know I said group with parenthesis, and shorter lines, but if you are switching to `elif` structure, (which I think is a fine choice!) , I would make the opposite recommendations!
```suggestion
    if full_status_df.loc[idx, 'SNV_indel_counts'] >= 1 and full_status_df.loc[idx,'CNV_loss_counts'] >= 1:
        full_status_df.loc[idx,'tp53_status'] = 1
    elif full_status_df.loc[idx,'SNV_indel_counts'] == 1 and full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome"):
        full_status_df.loc[idx,'tp53_status'] = 1
    elif full_status_df.loc[idx,'CNV_loss_counts'] == 1 and full_status_df.loc[idx,'cancer_predispositions'] == "Li-Fraumeni syndrome":
        full_status_df.loc[idx,'tp53_status'] = 1
    elif full_status_df.loc[idx,'SNV_indel_counts'] > 1:
        full_status_df.loc[idx,'tp53_status'] = 1
```

jashapiro (approved):
Looks good, and seems much more clear to me now.

Just a minor formatting suggestion, which is in conflict with my previous ones, unfortunately.

kgaonkar6:
> Looks good, and seems much more clear to me now.
> 
> Just a minor formatting suggestion, which is in conflict with my previous ones, unfortunately.

Thanks for the re-review! The suggested formatting is updated and re-run in the latest commits

PR #938, opened 2021-02-09
#932 Part1 : Recurrence snv/indels in OpenPBTA using all calls from strelka2,mutect2,vardict and lancet

kgaonkar6:
#### What scientific question is your analysis addressing?
In general the question in #932 is to generate/check for hotspots in our data specially to look for sites we don't see them in MSKCC cancer database downloaded [here](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/recurrence-snv/analyses/hotspots-detection/input/hotspots_v2.xls)  . 

In this PR I'm using strelka2,mutect2,vardict and lancet calls 
- Filtering for deleterious 
```
Filter IMPACT ='HIGH|MODERATE'
Filter Variant_Classification = 'Missense_Mutation|Splice_Region|In_Frame_Del|Frame_Shift_Del|Splice_Site|Splice_Region|Nonsense_Mutation|Nonstop_Mutation|In_Frame_Ins|Frameshift_Ins'
```
- Filter for oncogene,transcriptionfactor, kinase, TSGs genes list + brain-goi from @jharenza 

- Find recurrence
   Filter Tumor_Sample_Barcode to keep samples to independent-specimens.wgswxs.primary-plus.tsv to include primary tumors when available, if not select other tumor_descriptor types. Now any genomic site found in 2 or more independent sample and annotated as brain-goi is checked for recurrence. 

#### What was your approach?
I used 01-setup_db.py from snv-callers by @jashapiro and @cansavvy  to set up sql database of all the callers. I just updated `needed_cols` to include IMPACT, HGVSp_Short,VAF and Protein_position in all mafs which were needed for the filtering hotspots detections.

01-reccurence-hotspot-overlap.Rmd filters and combines calls from all callers and summarises the data to save recurrent mutated hotspots and additionally annotates the site as seen in MSKCC cancer hotspot database (file in input folder)

#### What GitHub issue does your pull request address?
#932 

#### Which areas should receive a particularly close look?
The recurrent counts were obtained per unique `Chromosome,Start_Position,End_Position,Amino_Acid_Position,Hugo_Symbol` I kept the chrom start/end mostly  because Amino_Acid_Position is derived from the Protein_position column in maf which we might change depending on transcript used for annotation. But also as in previous slack convos we discussed that there might be multiple different nucleotide changes could lead to the same amino acid change. Let me know if this column selection should be updated.

#### Is there anything that you want to discuss further?
[...]

Issue #932, opened 2021-02-03
Proposed Analysis: Generate hotspot and hot region lists


jharenza:
#### What are the scientific goals of the analysis?
Generate hotspot and hot region file for #819 

#### What methods do you plan to use to accomplish the scientific goals?
After some discussion with David Wheeler (formerly, BCM, now St. Jude), we have decided to derive a cancer hotspots (oncogenes) and hot regions file with which to perform recovery of potentially missed oncogenic DNA alterations. The rationale behind this is that frequently, many oncogenes are mutated recurrently at the same nucleotide(s), whereas tumor suppressor genes (TSG) frequently can have a non-recurrent mutation in a functional domain or a truncated alteration, leading to inactivation of the gene. In addition, kinase genes often may have recurrent or non-recurrent alterations within the kinase domain, leading to activation of the kinase.

We will first search for deleterious recurrent point mutations
```
Filter IMPACT ='HIGH|MODERATE'
Filter Variant_Classification = 'Missense_Mutation|Splice_Region|In_Frame_Del|Frame_Shift_Del|Splice_Site|Splice_Region|Nonsense_Mutation|Nonstop_Mutation|In_Frame_Ins|Frameshift_Ins'
```
within the OpenPBTA cohort (N>=2) within the annotated oncogene and TSG lists from `fusion_filtering` [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/fusion_filtering/references/genelistreference.txt) to determine whether we are missing any pediatric brain-specific hotspots from the cancer hotspots v2 file found [here](https://www.cancerhotspots.org/#/download). We will inspect those and potentially add them to the list of hotspots.

For hot regions, we will create a BED file of all kinase domain coordinates in which to scavenge back deleterious alterations.

For TSGs, we will include the gene as the "region" and scavenge back deleterious alterations that occur within these genes. We may either add these to the region list or just add this as a step in #819. 

#### What input data are required for this analysis?
```
pbta-snv-lancet.vep.maf.gz
pbta-snv-mutect2.vep.maf.gz
pbta-snv-strelka2.vep.maf.gz
pbta-snv-vardict.vep.maf.gz
```

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
1 week, yes

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@kgaonkar6 

#### What relevant scientific literature relates to this analysis?
[...]

PR #946, opened 2021-02-18
Update setup db

kgaonkar6:
Can we update the sql database creation step such that all maf tables contain the following columns ?
```
needed_cols = [
    'Hugo_Symbol',
    'Entrez_Gene_Id',
    'Chromosome',
    'Start_Position',
    'End_Position',
    'Variant_Classification',
    'Variant_Type',
    'Reference_Allele',
    'Tumor_Seq_Allele1',
    'Tumor_Seq_Allele2',
    'dbSNP_RS',
    'Tumor_Sample_Barcode',
    'Transcript_ID',
    't_depth',
    't_ref_count',
    't_alt_count',
    'n_depth',
    'n_ref_count',
    'n_alt_count',
    'Allele',
    'flanking_bps',
    'Protein_position',
    'IMPACT',
    'HGVSp_Short',
    'gnomAD_AF',
    'VAF'
]

``` 

I had made a copy of the script inPR #938 to include some columns needed for check overlap with known hotspots or plotting purposes but if the columns don't affect the consensus calling process I think best would be update the script here, thoughts?

#### What was your approach?
Just add in columns required

#932 

Does including these columns affect consensus calling or should I make a copy of the script to add the columns?

This is just a code update, the sql file gets saved into scratch folder which is ignored in this repo.

-

-

#### Documentation Checklist


jharenza (approved):

jashapiro (approved):
I agree that having a single copy of the script makes sense, and adding extra columns should have no effect on any downstream analysis, so this seems fine to me!

cansavvy (approved):
Seems good to me!

PR #950, opened 2021-03-02
First merge attempt

natemella:
Batch effects are commonly observed in RNA expression data. Such effects are likely less pronounced in RNA-Seq data than microarray data. However, batch effects may bias the conclusions of studies that do not account for these effects. We wish to evaluate the level to which batch effects are present in the RNA-Seq data from this study and create alternative versions of the data that have been adjusted for batch effects.

1. Does the PBTA data have batch effects? If so, will adjusting the batch effects remove the differences between patients with different histologies? 
2. How confounded are histologies and batch effects?

#### What was your approach?
We graphed PCA plots showing color as the histology and batch as the shape. We used three covariates as batch variables in three separate analyses:

1. the scientific cohort
2. the sequencing center
3. the library preparation method (polyA vs ribodepletion)

We filtered the graphs to only include the three most common histology types.

***919***

The reviewers should mostly look at 2 things:

1. Please look at the PCA plots before batch correction. In the comments of my markdown file, I state whether or not there appear to be batch effects. 
2. Please look at the PCA plots after batch correction to see if the adjustment is helpful or harmful.

Please look through the util/functions.R to see the details of how I create my figures.

In a later merge request, I will include a bash script and various R scripts which can all be run in Docker. I want to make this merge request small by only include an R markdown file and an identical jupyter notebook file (which can be run in an R kernal)

The figures are ready for review. The comments that go along with the figure may need some adjustments.

Both tables and figures

We tested for batch effects in three different areas:

1. Batch = Sequencing Center
2. Batch = Cohort
3. Batch = polyA vs ribodepletion

Within each of these analyses I evaluated RNA seq data that had been prepared using either RSEM normalization or Kallisto (tpm) values.
The null hypothesis is that there are no batch effects in the data

###### PolyA before
rsem_polya_sequence_histology
- Conclusion: There are batch effects. On a small scale analysis, you can see that batch 1 groups around the top left corner while batch 2 groups in the bottom left corner.
###### PolyA After
rsem_polya_sequence_batch-adjusted-histology
###### Stranded before
rsem_stranded_sequence_histology
- Conclusion: Cannot determine from this image if there are batch effects. BatchQC is required
###### Stranded After
rsem_stranded_sequence_batch-adjusted-histology
###### PolyA before
kallisto_polya_sequence_histology
- Conclusion: Cannot determine from this image if there are batch effects. BatchQC is required.
###### PolyA After
kallisto_polya_sequence_batch-adjusted-histology
###### Stranded before
kallisto_stranded_sequence_histology
- Conlusion: There appears to be batch effects. Batch 2 tends to group around the bottom of the graph
###### Stranded After
kallisto_stranded_sequence_batch-adjusted-histology
###### PolyA before
rsem_polya_cohort_histology
- Conclusion: There are batch effects. You can see that batch 1 groups around the top left corner while batch two groups in the bottom left corner
- ***NOTE***: For this gene expression file, there are only 59 patients. As such, the batches cohort and sequence center overlap perfectly. Therefore, one cannot remove the batch effects of one without removing the batch effects of the other
###### PolyA After
rsem_polya_cohort_batch-adjusted-histology
###### Stranded before
rsem_stranded_cohort_histology
- Only one batch
###### Stranded After
***NONE***
###### PolyA before
kallisto_polya_cohort_histology
- Conclusion: Cannot determine from this image if there are batch effects. BatchQC is required.
###### PolyA After
kallisto_polya_cohort_batch-adjusted-histology
###### Stranded before
kallisto_stranded_cohort_histology
- Only one Batch
###### Stranded After
***NONE***
###### Before
rsem_method_histology
- Conclusion: There are batch effects. You can see that batch 1 groups around the top left corner while batch two groups in the bottom left corner
###### After
rsem_method_batch-adjusted-histology
###### Before
kallisto_method_histology
- Conclusion: There are batch effects. You can see that batch 1 groups around the top left corner while batch two groups in the bottom left corner.
###### After
kallisto_method_batch-adjusted-histology

***Because I was asked to write an R markdown file, I have not added this to continuous integration. However, after this merge request, I have two bash scripts and 6 R scripts that are designed for continuous integration***

#### Documentation Checklist


analyses/batch-effects/graph-generator.Rmd
```diff
@@ -0,0 +1,426 @@
...
+***This should not be run for a full analysis!***
+
+
+```{R}
+# Shorten Data Files
+dat_rsem_polya = shorten(dat_rsem_polya)
+dat_rsem_stranded = shorten(dat_rsem_stranded)
+dat_kallisto_stranded = shorten(dat_kallisto_stranded)
+dat_kallisto_polya = shorten(dat_kallisto_polya)
+```
```
jaclyn-taroni, line 75: I don't think you want to subset to the first 100 lines, which is what `shorten()` does, by _default_. Instead, I'd recommend making this optional, where you do **not** do it by default. There is a limit to the amount of RAM that the instance running CI has (I _believe_ it's 8 GB), so it could come in handy for that, too.

My first question: Do you have an idea of how much memory is required to run the analysis without the `shorten()` step?

If it is likely to be too costly in terms of memory, we have some options.

If you're going to use R Markdown, the way I'd recommend doing this is with [`params`](https://rmarkdown.rstudio.com/lesson-6.html).

In the YAML header of the file, you'd add:

```yaml
params:
  shorten: FALSE
```

And then in this chunk you could do something like:

```R
if (params$shorten) {
  dat_rsem_polya = shorten(dat_rsem_polya)
  dat_rsem_stranded = shorten(dat_rsem_stranded)
  dat_kallisto_stranded = shorten(dat_kallisto_stranded)
  dat_kallisto_polya = shorten(dat_kallisto_polya)
}
```

If you wanted to run the R Markdown document from the command line and run the shorten step, you can do the following:

```sh
Rscript -e "rmarkdown::render('analyses/batch-effects/graph-generator.Rmd', params = list(shorten = TRUE), clean = TRUE)"
```

These chunks are untested, but are hopefully enough to give you an idea of how to accomplish this!

Another way to "shorten" the data frame that is a little more principled (at least for plotting PCs) would be to subset to genes with high variance. You can use [`matrixStats::rowVars()`](https://www.rdocumentation.org/packages/matrixStats/versions/0.58.0/topics/rowVars) for calculating the variance and if I recall correctly, that calculation probably won't be prohibitive in terms of memory or time.

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
+# function used to cut files
+shorten = function(df) {
+  return(df[1:100,])
```
jaclyn-taroni, line 3: See other comment about using `rowVars()` instead.

analyses/batch-effects/graph-generator.Rmd
```diff
@@ -0,0 +1,426 @@
...
+
+#### Group the Kallisto Data
+
+
+```{R}
+# Summarize the kallisto data on the gene level in order to join with batch
+dat_kallisto_polya = grouper(dat_kallisto_polya)
+dat_kallisto_stranded = grouper(dat_kallisto_stranded)
+```
```
jaclyn-taroni, line 84: This collapses to the gene mean, is that correct? Is this step necessary for batch correction?

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+}
+
+# function used on kallisto data for summarizing transcripts at the gene level
+grouper = function(df){
```
jaclyn-taroni, line 7: I like to start my custom functions with a verb and make the names as descriptive as possible. When I revisit my code months later, this approach is super helpful for getting me back up to speed!

Something like the following tells me a lot about what the function is doing:

```suggestion
collapse_to_gene_mean = function(df){
```

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+    group_by(gene_id) %>%
+    summarise_at(colnames(df)[3:ncol(df)], mean, na.rm = TRUE)
+  return(df)
+}
+
+# transpose data through gather and spread functions, rename the index column
+trans = function(data, Kids_First_Biospecimen_ID, val, starting_col){
+  data = data %>%
+    gather(Kids_First_Biospecimen_ID, val, starting_col:(ncol(data))) %>%
+    spread("gene_id", val)
```
jaclyn-taroni, line 18: It is not clear to me what the advantage of this is over `t()`. Does this transpose only happen when you have a bunch of metadata alongside the expression values?

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+}
+
+# for batchQC to work, the batches must be numeric values
+batch_prep = function(data_t){
+  # verify that batch is indeed a factor and numeric
+  batch = as_factor(data_t$batch)
+  batch = as.numeric(batch)
+  
+  # make so that batch values start at 1
+  levels(batch) <- 1:length(levels(batch))
```
jaclyn-taroni, line 87: Could you use `as.integer()` instead and save some lines (untested)?

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+  
+  if (run_combat == FALSE){
+    print("Combat set to not run on this data set")
+    return()
+  }
+  
+  plot_list = tryCatch({
+    # filter out genes without variance
+    var_zero_genes = poly1matrix[which((apply(poly1matrix, 1, var)==0)),]
+    poly1matrix = poly1matrix[which((apply(poly1matrix, 1, var)>0)),]
```
jaclyn-taroni, line 165: Might look into using `matrixStats` here, too!

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+
+# make two PCA plots before and after applying combat
+make_before_and_after_combat_plots = function(data_t, batch, report_name, run_combat) {
+  
+  data_t$batch = batch_prep(data_t)
+  
+  #make a matrix representing just the gene expression data (this gets plugged into combat)
+  poly1matrix = t(data_t[,4:ncol(data_t)])
+  
+  pca = prcomp(log2(data_t[,4:ncol(data_t)]+ 1))
```
jaclyn-taroni, line 151: It looks like you have to do this transformation twice: once before ComBat and once for this call to `prcomp()`. 

Instead, what if you did your `log2()` step elsewhere (e.g., in one of your cleaning functions), before `data_t` gets to this function?

I'd also consider whether or not you need to transpose before you get to the `prcomp()` steps.

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+  plot_list = tryCatch({
+    # filter out genes without variance
+    var_zero_genes = poly1matrix[which((apply(poly1matrix, 1, var)==0)),]
+    poly1matrix = poly1matrix[which((apply(poly1matrix, 1, var)>0)),]
+    print(sum(var_zero_genes))
+    
+    poly2matrix <- log2(poly1matrix + 1)
+    
+    
+    combat_matrix = ComBat(dat=poly2matrix, batch=batch)
```
jaclyn-taroni, line 171: I have more general ComBat question(s) here: Is `ComBat()` intended to be used on log2-transformed TPM?  Does BatchQC include _ComBat-seq_ ([Zhang _et al._ 2020](https://doi.org/10.1093/nargab/lqaa078))?  I think when we're thinking about these questions we should also consider how we anticipate folks will use corrected values downstream.

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+    message("ComBat failed! It's likely that you are trying to adjust for batches with just 1 batch")
+    # message(cond)
+    return(list(p1, NULL, data_t))
+  })
+  
+  return(plot_list)
+  
+}
+
+# only include three most common histologies
```
jaclyn-taroni, line 189: Is this because the three most common histologies happen to be represented in poly-A and stranded? This seems like a case where we might want to hardcode the `short_histology` values that are in both datasets.

analyses/batch-effects/util/functions.R
```diff
@@ -0,0 +1,361 @@
...
+  #Make before and after PCA plots
+  return(make_before_and_after_combat_plots(data_t, batch, report_name, run_combat))
+
+}
+
+# root functions that calls sub functions to combat and batchQC looking for batch affects
+run_batchQC = function(df, id_batch, gene_id, report_name, file_name, run_combat){
+  
+  # create output directory (if necessary)
+  dir.create(file.path(analysis_dir, "results"), showWarnings = FALSE)
```
jaclyn-taroni, line 226: This `util/functions.R` file could be sourced from anywhere, which means that where it gets sources may or may not have `analysis` directories in the global environment. You could pass `analysis_dir` as an argument here (or maybe plan to have `file_name` contain the full path) or you can include this in a code block in the notebook you're intending to use it.

jaclyn-taroni (commented):
Hi @natemella, thanks for this submission! I have not yet gone through all the code in detail, but I had a few high-level comments that I wanted to return to you in the meantime. I anticipate some of these questions will be starting points for discussion like what kinds of batch corrected values do we think other folks will use in their analyses. 

Please let me know if you have any questions. If a comment is unclear, I'd encourage you to reply to that comment directly; that helps us keep discussion organized.

Thanks again! 🙂

analyses/batch-effects/README.md
```diff
@@ -0,0 +1,75 @@
+# Batch Effect Analysis
```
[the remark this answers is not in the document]
natemella: I'm totally fine with saving the README.md for a later pull. Should I submit a new pull request without the README.md or attempt to edit the current one?

analyses/batch-effects/graph-generator.ipynb
```diff
@@ -0,0 +1,1045 @@
+{
```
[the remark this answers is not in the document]
natemella: I'll submit the new pull request with the .ipynb. The two files should be almost identical (I just have a little bit more documentation on the top of the .ipynb). It's super easy for me to convert .ipynb to .Rmd.

analyses/batch-effects/graph-generator.Rmd
natemella: This is a great idea. I'll work on integrating the changes. I'm not sure the exact amount of memory that this requires. But I can get back to you on that.

Dockerfile
```diff
@@ -90,6 +94,17 @@ RUN ./install_bioc.r \
...
+RUN ./install_bioc.r \
+    fftwtools \
+    heatmaps \
+    BatchQC \
+    tibble \
+    tidyr \
+    ggplot2 \
+    readr \
+    dplyr \
+    forcats
```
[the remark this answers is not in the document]
natemella: Thanks for pointing this out. I'll update my functions so I don't use the newer tidyr functions. Do you know which commands in tidyr aren't supported in the older version?

analyses/batch-effects/util/functions.R
natemella: I'll relook into t() instead of using gathering and spread. I wrote this function a while ago. I might have had to do it this way because of the ID columns, but I'm not sure. I'll send you an update later this week on this.

analyses/batch-effects/util/functions.R
natemella: I can't remember if I tested using as.integer(). I'll explore this option.

analyses/batch-effects/util/functions.R
natemella: Thanks! I'll look into this

analyses/batch-effects/util/functions.R
natemella: I'll add analysis_dir as a parameter.

analyses/batch-effects/graph-generator.Rmd
natemella: I was doing that so I could run this analysis on my machine that has eight GB. It would be great if you could run this on a server that has 32 GB. Then I could just delete the code that does this trimming. Would that work?

jaclyn-taroni: We will still want to run it in continuous integration, which has a cap of 8 GB of RAM as well.

analyses/batch-effects/graph-generator.Rmd
natemella: This is not necessary for batch correction. We could apply combat to the transcript level values. However, that is a less proved approach, and people generally sum transcript level values to gene-level values. I had been using the mean as a way of summarization, but Steve Piccolo said that it would be better to summarize using sum. [Here's ](https://github.com/griffithlab/rnaseq_tutorial_wiki/blob/master/Kallisto.md) a tutorial where this is mentioned as well. So I can change that if you're okay with that?

analyses/batch-effects/util/functions.R
natemella: I'll go ahead and make this change

analyses/batch-effects/util/functions.R
natemella: I only used ComBat, but I'll switch the code to use ComBat-seq instead.

jaclyn-taroni (commented):
Hi @natemella, I wanted to respond to some of your comments all at once, so I am returning a review with those replies.

analyses/batch-effects/graph-generator.Rmd
natemella: I'm resending the graphs from the full dataset. I can generate these graphs when I run them on a server with more memory

natemella: The `matrixStats::rowVars()` method seems to work well until I get to the third "batch" where I'm combining polyA and rsem files together. If I select rows based on which ones have the most variance, I end up with a different subset of genes in the two different gene expression files. As such, I'm not able to join the two files together and do any batch correction.

This isn't a problem if I run my full analysis (without shortening files) on a big server with a lot of memory or if I do a simplified file shortening method for this last batch as I did earlier:
`df=df[1:500,]`

To your point though, we don't want to draw conclusions on a few genes alone. I could try seeing what's the largest amount of genes that I could include without hitting the 8GB memory limit.

jaclyn-taroni:
This pull request has not been updated in a few months, so I am going to go ahead and close it per a discussion with Dr. Piccolo.

natemella:
Jaclyn,

I could try to do a big last effort this weekend to finish up the analysis
if it's not too late.

Best,
Nathan

jaclyn-taroni:
@natemella At this stage in the project, we are looking to either tweak or finish up existing modules (e.g., modules that have already had several PRs go through). This particular analysis is larger in scope than that in my opinion, so unfortunately I don't think we have the spare cycles for review at this point.

natemella:
Makes sense,

Sorry that I wasn't able to finish

Best,
Nathan

Issue #819, opened 2020-10-20
Proposed Analysis: add scavenging of cancer hotspots to consensus SNV calls


jharenza:
#### What analysis are you proposing and why?
Create a new MAF which contains consensus SNV calls from [consensus snv calling](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/snv-callers) and cancer hotspot calls missed by consensus, as noted below. 

We previously noticed that by taking a 3/3 approach for consensus calls, we are inevitably missing some cancer hotspot mutations. We got around that for one specific cancer (DMGs) because we have clinical reports containing histone variant calls that we can add into molecular subtyping pathology module (#735 and #751). However, we are _likely_ still missing some cancer hotspot mutations and I propose that we add a final step in which we scavenge back cancer hotspot mutations using a well-curated and downloadable [list](https://www.cancerhotspots.org/#/home) of these.

#### What changes need to be made? Please provide enough detail for another participant to make the update.
The next step would be to assess if any of these hotspot mutations are being missed using a 3/3 method and then determining a set of rules for adding these mutations back to the consensus SNV file. For example:
1. If the hotspot is present in 2/3, then retain
2. If the hotspot is present in 1/3, _and_ has X reads supporting the tumor allele (5?), then retain

Perhaps the new file can be called `pbta-consensus-snvs-plus-hotspot.maf.gz`

#### What input data should be used? Which data were used in the version being updated?
Cancer hotspots table, downloadable here: https://www.cancerhotspots.org/#/download plus TERT promoter mutations, noted from [this paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4852159/).
>In 2013, two hotspot point mutations were found in the TERT promoter in 71% of melanomas (32,33). The mutations were located 124bp and 146bp upstream of the translation start site and referred to as C228T and C250T, respectively, based on their hg19 genomic coordinates. 

```
pbta-snv-consensus-mutation.maf.tsv.gz
pbta-snv-lancet.vep.maf.gz
pbta-snv-mutect2.vep.maf.gz
pbta-snv-strelka2.vep.maf.gz
pbta-snv-vardict.vep.maf.gz (maybe? TCGA was not run with VarDict, so perhaps we should add a separate step to assess what hotspots VarDict only detects)
```

```
pbta-tcga-snv-lancet.vep.maf.gz
pbta-tcga-snv-mutect2.vep.maf.gz
[...]

PR #954, opened 2021-03-11
Snv database re-run to include all maf columns

kgaonkar6:
Can we update the sql database creation step such that all maf tables contain all maf columns ? We will mainly need it when we want to  scavenge back hotspots called by 1 or more calls as a maf format.

#### What was your approach?
I added mutect2 and vardict to the list of table names that should contain all maf columns

#819 

Adding all columns of a maf file creates a huge 50G file, maybe there is some way to reduce the file size or save the database elsewhere for easy access?

NA

This is just a code update, the sql file gets saved into scratch folder which is ignored in this repo.

NA

NA

#### Documentation Checklist

### DIFF
--- a/analyses/snv-callers/scripts/01-setup_db.py
+++ b/analyses/snv-callers/scripts/01-setup_db.py
@@ -286,7 +286,7 @@ for table_name, maf_file in callers:
     if table_name in ('strelka', 'lancet', 'consensus'):
         table_types = maf_types
     else:
-        table_types = needed_types
+        table_types = maf_types
     if args.overwrite:
         con.execute("DROP TABLE IF EXISTS {}".format(table_name))
     maf_table_def = ", ".join([" ".join(col) for col in table_types])
--- a/analyses/snv-callers/scripts/01-setup_db.py
+++ b/analyses/snv-callers/scripts/01-setup_db.py
@@ -283,10 +283,10 @@ for table_name, maf_file in callers:
         continue
     print("Reading file {} to table '{}'.".format(maf_file, table_name, ))
     # we need 2 full tables for consensus with one missing
-    if table_name in ('strelka', 'lancet', 'consensus'):
+    if table_name in ('strelka', 'lancet', 'consensus', 'mutect', 'vardict'):
         table_types = maf_types
     else:
-        table_types = maf_types
+        table_types = needed_types
     if args.overwrite:
         con.execute("DROP TABLE IF EXISTS {}".format(table_name))
     maf_table_def = ", ".join([" ".join(col) for col in table_types])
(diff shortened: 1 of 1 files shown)


jashapiro:
> Adding all columns of a maf file creates a huge 50G file, maybe there is some way to reduce the file size or save the database elsewhere for easy access?

This is the main reason this was not done in the first place. I am hesitant to add in all of the columns like this for a limited use case of searching for specific hotspots. That would likely be more efficiently accomplished by creating a subset of the original maf file for sites that match known hotspot locations. 

Alternatively, if there are specific columns you need from all tables, you might add those to `needed_types`. The only reason we kept all of them was so that we could write out a "complete" maf file for the consensus. As I understand the analysis goals for this project, I don't think you need that?

kgaonkar6:
The results of #819 would be maf file, so I believe we would need all columns. Are you suggesting instead of the database I filter merged file from each caller for hotspots instead of using the database?

kgaonkar6:
It seems reading vardict in R is using up all RAM in my laptop, so I think I'll switch to python to read and filter the files. 
But since I have the filtering script in R  I was also thinking of splitting vardict merged file via `split -l <line number>` and then give that as input to the R script , thoughts?

jashapiro:
> It seems reading vardict in R is using up all RAM in my laptop, so I think I'll switch to python to read and filter the files.
> But since I have the filtering script in R I was also thinking of splitting vardict merged file via `split -l <line number>` and then give that as input to the R script , thoughts?

Yes, vardict is huge and this is a perennial problem. The reason for the DB was partly to alleviate this problem when we had to deal with the full data set. In this case, when we need only a small subset and have some simple things to filter by, I would expect that `grep` (or `awk` if you want to get fancy) before reading in the file could be most efficient. 

There is a function in `readr` called `read_csv_chunked()` that could also make the read more efficient if you want to stay in R. You could write your filter function and use that as the `callback` argument to filter a set of lines. But my memory is that this isn't the easiest thing to set up.

kgaonkar6:
Thanks @jashapiro I did think about greping for the genes in the terminal and then filter for Amino_Acid_Position and genomic regions, but then 1 of the input file is in excel format from source [here](https://github.com/kgaonkar6/OpenPBTA-analysis/blob/recurrence-snv/analyses/hotspots-detection/input/hotspots_v2.xls) with 2 tabs of sheets😅  

I ended up splitting the vardict file in #956 and filtering for hotspots since I'd already run other callers through that script. Hmm, maybe I should just save the hotspot database calls as tsv file in input  folder then I could grep, I can try it out.

kgaonkar6:
Please close if #954 seems more reasonable for gathering the hotspots. Thanks!

jaclyn-taroni:
Closing in favor of #956

Issue #952, opened 2021-03-09
Docker installing `units` failures


cansavvy:
### Summary

When Docker builds in the data download step, sometimes a package called `units` fails to install: 

https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5116/workflows/ece6709e-1c0d-4cb8-be7b-9051f956d722/jobs/5714
```
configure: error: in `/tmp/RtmpHpJSHz/R.INSTALL326563bcd8f/units':
configure: error: 
--------------------------------------------------------------------------------
  Configuration failed because libudunits2.so was not found. Try installing:
    * deb: libudunits2-dev (Debian, Ubuntu, ...)
    * rpm: udunits2-devel (Fedora, EPEL, ...)
    * brew: udunits (OSX)
  If udunits2 is already installed in a non-standard location, use:
    --configure-args='--with-udunits2-lib=/usr/local/lib'
  if the library was not found, and/or:
    --configure-args='--with-udunits2-include=/usr/include/udunits2'
  if the header was not found, replacing paths with appropriate values.
  You can alternatively set UDUNITS2_INCLUDE and UDUNITS2_LIBS manually.
--------------------------------------------------------------------------------

See `config.log' for more details
ERROR: configuration failed for package ‘units’
* removing ‘/usr/local/lib/R/site-library/units’
Error in install.packages(opts$PACKAGES, repos = c(bioc_repos, getOption("repos"))) : 
  (converted from warning) installation of package ‘units’ had non-zero exit status
```

I haven't looked into this much yet, but it might be solved by installing `libudunits2-dev` before getting to the R installations.

cansavvy:
`libudunits2-dev` is supposedly already installed: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2c1f5faf1f5687fa40b9b30fa6442798bd38557d/Dockerfile#L23 and there was no error or warning about that install failing. Hmm...

jaclyn-taroni:
Here's the status of my investigation:

- When we build the Docker image in CI using the `--no-cache` option (#976), it builds just fine: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5186/workflows/e8e74608-512f-41a1-87bb-6f890a723a57/jobs/5792
- Similarly, there's no problem building locally with `--no-cache`.
[...]

PR #957, opened 2021-03-15
WIP: Testing Docker build 

cansavvy:

Doing some testing to find out more info about #952

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -88,7 +88,8 @@ RUN ./install_bioc.r \
     rpart \
     rprojroot \
     survival \
-    viridis 
+    units \
+    viridis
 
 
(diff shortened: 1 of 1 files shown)


cansavvy:
This did build fine. I don't feel 100% convinced that this solved the problem but maybe it did?

jaclyn-taroni:
I would say that it solves the problem but doesn't necessarily illuminate _why the problem showed up_ if that makes sense.

cansavvy:
^ Yes this.

jaclyn-taroni:
Closing this, see the explanation in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/952#issuecomment-811276228

PR #960, opened 2021-03-16
quick bug fix in mol-sub-path

jharenza:
molecular subtyping pathology resulted in a file which had coded the integrated dx as the molecular subtype, so I have updated the code to fix this

#### What was your approach?
see above

NA

just double check the results 

no

yes

table

updated integrated dx

#### Documentation Checklist


jaclyn-taroni (approved):
Took a quick look at the output and it seems like the change is accomplishing what we want 👍🏻

Issue #953, opened 2021-03-09
Proposed Analysis: investigate putative functionality of TP53 SV


kgaonkar6:
#### What are the scientific goals of the analysis?
We have investigated SNV  and CNV TP53 alterations in detail in `tp53-nf1-score` but there are other samples with high tp53 classifier scores but were not annotated via the 2 hit TP53 loss or activating mutation. 
![image](https://user-images.githubusercontent.com/34580719/110499453-40902300-80c6-11eb-84f2-753830b26046.png)

Thus in this [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/837#issuecomment-790662695) @jharenza proposes checking SV breakpoints in/upstream of TP53 that might support the TP53 loss phenotype

#### What methods do you plan to use to accomplish the scientific goals?
Gather annotated manta calls to identify TP53 loss status :
- If SVs overlap TP53 gene  (should we include chr17p arm level loss ? ) or has breakpoint (bk) is within the gene

```
TAD = trans-activating domain (essential for function): AD1 (residues 1–42) and AD2 (residues 43–92)
DBD = DNA-binding domain (residues 102–292)
TD = tetramerization domain (residues 326–356)
```

Add `TP53 altered - loss` annotation if:
- SV overlaps or bk is within gene + any SNV (not adding CNV because it technically is a structural variation, thoughts?) ~ TP53 altered - loss
- SV overlaps or bk is within gene + cancer_predispositions == "Li-Fraumeni syndrome"  ~ TP53 altered - loss
- SV overlaps or bk is within gene + RNA-Seq > 0.5  ~ TP53 altered - loss

#### What input data are required for this analysis?
pbta-sv-manta.tsv.gz

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
2 days

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@kgaonkar6 

#### What relevant scientific literature relates to this analysis?
@jharenza could you update the ticket if I missed anything and add references if possible?

Quick searchs got me:
http://dx.doi.org.proxy.library.upenn.edu/10.1038/nature25480
>Recurrently deleted regions were predominantly associated with epigenetic or cell cycle regulators, most commonly TP53, PTEN , SETD2 , and CDKN2A or CDKN2B 

https://doi.org/10.1093/jnen/nlaa124
[...]

PR #962, opened 2021-03-18
#953 Part1: Gather TP53 SV breakpoints with gene loss

kgaonkar6:
Are there SV breakpoints in/overlapping TP53 gene locus that might support the TP53 loss phenotype?

#### What was your approach?
I gathered SV breakpoints that overlap or are within TP53 gene locus by searching for TP53 in the `Gene.name` column in the annotSV formatted manta calls in `pbta-sv-manta.tsv.gz`. Then as a secondary support for TP53 loss status I looked for loss of copy_number in `loss_overlap_domains_tp53.tsv` from a previous step and also if the SV breakpoint is associated with a `1<FPKM` expression 

#953 

- I used INV,DEL and BND mainly because I wanted to check if any breakpoint within/overlapping TP53 will also associate with loss in CNV or low expression. Seems like it's not the case, so I can update it to DEL only if that seems like a more clear filtering for this analysis.

BS_3NX3RBCX and BS_EJV0N3BX have a CNV loss and multiple SV breakpoints overlapping TP53 locus, and most of them are quite long 4034301 to 15772588 bp. DO we need a SV.length filter and since we already capture it in CNV do we want to add to these to the classifier evaluation which is the next step ?
The longest ovelaps a large chunk of chr17p arm
Screen Shot 2021-03-18 at 6 44 12 PM

Yes

table

sample_id 7316-114 and 7316-2217 have multiple breakpoints associated CNV loss

#### Documentation Checklist


jharenza:
@kgaonkar6 Thank you for this analysis! 

>I used INV,DEL and BND mainly because I wanted to check if any breakpoint within/overlapping TP53 will also associate with loss in CNV or low expression. Seems like it's not the case, so I can update it to DEL only if that seems like a more clear filtering for this analysis.

I think this is the correct thing to do, especially if a breakpoint were to result in a translocation that removes most of the TP53 gene after the start site such that a functional protein is not translated, so I think we can keep this as is. 

>BS_3NX3RBCX and BS_EJV0N3BX have a CNV loss and multiple SV breakpoints overlapping TP53 locus, and most of them are quite long 4034301 to 15772588 bp. DO we need a SV.length filter and since we already capture it in CNV do we want to add to these to the classifier evaluation which is the next step ?

I think that SV length will vary too much to use a filter at this moment in time and that it is not necessary to implement right now.

>sample_id 7316-114 and 7316-2217 have multiple breakpoints associated CNV loss

At first I thought, GREAT! But, then I was thinking about how we derived consensus CNV calls in the first place, and one of the tools used is MantaSV, so I think this result makes total sense. That being said, I do not think we can necessarily call the SV and CNV events as unique, as I may have originally planned. The SV breakpoint is more likely resulting in CN loss of the same copy of the gene, rather than the SV breakpoint occurring on one copy and the CNV on another copy. So, it seems that for `BS_3NX3RBCX`, we only have evidence of one hit thus far (SV/CNV being considered one hit), which aligns with the lower _TP53_ score of `0.3218202`. For `BS_EJV0N3BX`, that sample was already annotated as `TP53_altered == loss` since it contains a hotspot SNV and a CNV, and its _TP53_ score is high.

I think since we have the information, we could create a new column to the `tp53_altered_status.tsv` file annotating the presence of an SV, but we will need to be sure that if `CNV_loss_counts == 1` & `SV_counts` == 1, that we do not count them as unique events when determining `tp53_altered` status. Once you add this column and account for the calculation, then we can merge this!

jharenza (changes requested):
See my comments!

kgaonkar6:
Should I add another PR for adding the `SV_counts` column  since the code to create tp53_altered_status.tsv is a different script?

jharenza:
I would say it's within the scope of this one, so you can add here.

jharenza (approved):
Great - logic looks good and the expected results occur.

Issue #931, opened 2021-02-01
Updated analysis: Make default method for immune-deconv NOT CIBERSORT


cansavvy:
#### What analysis module should be updated and why?

CIBERSORT method is the default for immune-deconv but it doesn't actually run because we don't have the script for it. We haven't updated the results on Github in almost a year. See the immune-deconv README for more details on why the default method CIBERSORT can't be run. 

I think if we want to encourage this to be ran more often and to be ran more easily, CIBERSORT should not be the default method. Instead the `mcp-counter` method should be made as default (That's what's actually tested in CI as well). 

#### What changes need to be made? Please provide enough detail for another participant to make the update.
`run-immune-deconv.sh`'s logic needs to be udpated as well as the README. 

Perhaps `01-immune-deconv.R` needs reordering as well or at the very least documentation updates. 

#### What input data should be used? Which data were used in the version being updated?
v18 data

#### When do you expect the revised analysis will be completed?
TBD

#### Who will complete the updated analysis?
TBD

jharenza:
@komalsrathi - thoughts? Would you be able to update this?

komalsrathi:
@jharenza yes I'll address this as soon as I can.

komalsrathi:
@cansavvy 

We actually do have the `CIBERSORT.R` and `LM22.txt` files required to run Cibersort - I registered and downloaded those and that's how I generated the outputs corresponding to [CIBERSORT](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/immune-deconv/plots/heatmap_CIBERSORT(abs.).pdf). Those files are added to the .gitignore because we cannot share them with anybody publicly. Because of the aforementioned reason, we are using MCP-Counter for the CI tests. This issue has been fully discussed here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/292 (you might have to go through this entire thread)

Please let me know your thoughts.

cc: @jaclyn-taroni @jharenza

jaclyn-taroni:
@komalsrathi my read of this issue is changing the _default_ method to be a method that everyone, regardless of whether or not they have registered and downloaded those files (and put them in the right place!), can run this module. I believe what gave rise to this issue was difficulty rerunning this using the v18 data. It seems reasonable to have a default that doesn't cause failure.

> Although, it still needs to be updated with v18 so I can do that.

Agree that we should update with v18.

komalsrathi:
[...]

PR #973, opened 2021-03-30
use only xCell to deconvolute

komalsrathi:

As discussed in #931 we should only be using xCell as it is the most comprehensive and widely used deconvolution method. Using other deconvolution methods has not given any valuable insights to the interpretation of this module so we should remove those methods altogether.
 
#### What was your approach?

I removed any data/plots related to methods other than xCell and only kept a single heatmap corresponding to xCell. 

I have also cleaned up the script slightly, created and moved `heatmap_by_histology` function to `utils/` folder. This will be helpful as I will be adding another function `heatmap_by_molecular_subtype` in the next PR.

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/931

N/A

N/A

Yes

N/A

N/A

#### Documentation Checklist


jaclyn-taroni (changes requested):
Looks good! I had a couple minor comments. Can you also update the CI step as part of this PR? 

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b148e1fd7d4c50486157a8b6093dc621a9761323/.circleci/config.yml#L108

I'd remove the `OPENPBTA_DECONV_METHOD="mcp_counter"` from the command that is no longer needed and update the name such that it reflects what `run-immune-deconv.sh` will be doing.

jaclyn-taroni (approved):
👍

Issue #818, opened 2020-10-19
De novo mutational signature extraction: next steps


jaclyn-taroni:
I am filing this issue to replace #636 with more detailed steps required for completing the de novo mutational signatures analysis and to surface some discussion on #799 (or other, already merged PRs). Note that this issue is _too expansive_ in scope to be completed with a single pull request. It likely should be broken up into smaller issues with more detail, but in an effort to reduce the cognitive burden associated with tracking it exclusively in my head and to get some feedback, I am filing this one large issue to start.

#### The current state of mutational signatures

Right now, the de novo part of [the `mutational-signatures` module](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/520133a1a47a5be7e68116139b11922bac28ffee/analyses/mutational-signatures) extracts signatures from the WGS samples only for a range of number of signatures (_k_), using a low number of iterations. There is a script `analyses/mutational-signatures/scripts/de_novo_signature_extraction.R` that has command line options for the value(s) to use for _k_ during extraction and the number of iterations.

#### What needs to happen next

Of the things I am currently aware of 😅 

- **Select the number of signatures to extract (_k_).** This will be a notebook where we include goodness-of-fit plots from `sigfit` and examine the mutational spectra. We will use this thread of discussion as our guide: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/799/files#r502923369

- **Assess the stability/reproducibility of the extracted signatures after selecting _k_**, by running with a higher number of iterations, using multiple seeds, and measuring average silhouette width https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/799/files#r503268952
  - _Potential gotcha:_ How many iterations do we need for `sigfit::extract_signatures()`? Part of how we assess is with reproducibility.
  - _Maybe:_ Add an option for when you are only specifying a single number for `nsignatures` to `analyses/mutational-signatures/scripts/de_novo_signature_extraction.R` per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/806#discussion_r504933953

- Once we have a set of signatures to move forward with, we must do the following: 
  - **Match extracted signatures to known/reference signatures** (`03-match_de_novo` in #799). _Potential gotcha:_ Comparing to known/reference signature can aid in selecting _k_ https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/799/files#r503205836.
[...]

PR #974, opened 2021-03-30
De novo mutational signature analysis, PR 1/n

sjspielman:

This PR begins to address issue #818 by attempting to determine the optimal `k` for a de novo mutational signature extraction (using the `sigfit` package). The ultimate goal is to perform de novo mutational signature extraction from WGS samples. 

Determining optimal parameters for de novo mutational signatures in pediatric brain tumors

#### What was your approach?

Using the `sigfit` package, performed extraction for `k \in [2,8]` with 1000 iterations each. This was performed under two models in `sigfit` (poisson, which is comparable to the Emu method, and multinomial aka nmf which is comparable to Alexandrov 2013 method), and performed five times using seeds \in [1,5].

#818 

Need to discuss file organization for this analysis module. Currently, the benchmarking code is in analysis directory`scripts/extract_signatures_benchmark.R` and is run across seeds and models with `extract_signatures_benchmark.sh`, and output goodness-of-fit plots are in analysis directory `plots/denovo/gof/`. Does this seem reasonable?

+ What's going on with seed 4 plots?
+ If anyone has _other_ ideas for performing analysis besides just running all k 3, 4, 5.
+ Do we agree it is ok if this doesn't go into CI? It is a precursor to the actual extraction and analysis.

Mostly yes. Note that I am aware READMEs for this analysis and overall project need to be updated, but I suspect it's best to wait until the final PR for this analysis rather than during benchmarking.

Goodness of fit plots in `plots/denovo/gof/`, named according to: `gof_seed_<SEED>_model_<MODEL>.pdf`.

seed | model | optimal k
-----|-------|----------
1    | nmf|  3 
1    | poisson| 3
2    | nmf| 4
2    | poisson| 4
3    | nmf| 4 
3    | poisson| 4 
4    | nmf| 5
4    | poisson| 5
5    | nmf| 3 
5    | poisson| 3

+ For a given seed, models always agree on optimal k 
+ Result seems very dependent on initial conditions
+ seed 3 gives a strange curve, we may want to check with someone with additional expertise

Ideally one would run this for more than 1000 iterations, but the RAM is not so forgiving.

#### Documentation Checklist


sjspielman:
Ooof seems like I accidentally borked one of the existing scripts. Restoring it shortly.

jaclyn-taroni:
This code seems fine @sjspielman. It is not clear to me why including `analyses/mutational-signatures/scripts/extract_signatures_benchmark.R` and `analyses/mutational-signatures/scripts/run-extract_signatures_benchmark.sh` is preferable to making the following changes to what's already in this module:

* Altering `analyses/mutational-signatures/scripts/de_novo_signature_extraction.R`
  * Adding `--model` where the acceptable arguments are "multinomial" and "poisson" & updating the call to `sigfit::extract_signatures()` to use this argument
  * Making `--output_file` optional like `--plot_output` is currently
* Altering `analyses/mutational-signatures/03-de_novo_range_of_nsignatures.sh`
  * Set `denovo_plot_dir=plots/denovo/gof` 
  * Don't pass a range of values for number of signatures to `de_novo_signature_extraction.R` (loop through instead! 👇🏻 ) and pass the same value to `--nsignatures_floor` and `--nsignatures_ceiling` 
  *  Loop through the following: model, seed, nsignatures

I could also see an argument for removing the `--nsignatures_floor` and `--nsignatures_ceiling` and replacing with a `--nsignatures`  in `scripts/de_novo_signature_extraction.R`. I don't think leaving it as is hurting anything.

A benefit of this approach is that I've already done a bit of legwork to figure out how to run an abbreviated version of this in CI. I also think it's a bit easier to follow the thread if you come to the code base for the first time that way.

jaclyn-taroni:
Chatted with @sjspielman - in the interest of time, I'm going to file a PR to this branch with the changes outlined in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/974#issuecomment-814974051

jaclyn-taroni:
FYI @sjspielman: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/992 Should hopefully be as straightforward as merging `master` into this branch!

sjspielman:
@jaclyn-taroni Code all debugged and runs!

analyses/mutational-signatures/scripts/extract_signatures_benchmark.R
```diff
@@ -0,0 +1,126 @@
+# Written by Stephanie Spielman and adapted from code by Anna R. Poetsch, Candace L. Savonen, and J. Taroni
```
jaclyn-taroni, line 1: Remove this one now?

analyses/mutational-signatures/scripts/extract_signatures_benchmark.R
```diff
@@ -0,0 +1,126 @@
...
+# Determine optimal k for de novo mutatational signature analysis as well as robustness to random seed.
+#   Analysis is performed with a small number of iterations (1000) and tests k 2:8 (previous small runs suggest this range is appropraite) provided a given model (multinomial or poisson) and random seed.
+#   Needs up to 20 GB RAM but is single-threaded
```
jaclyn-taroni, line 4: We might want to add this to the README for the module, but I think that can wait as your initial comment suggests.

analyses/mutational-signatures/scripts/run-extract_signatures_benchmark.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
jaclyn-taroni, line 1: Also remove this one!

jaclyn-taroni (approved):
If we take out the duplicated stuff (and this passes CI), I think we're good!

analyses/mutational-signatures/scripts/extract_signatures_benchmark.R
sjspielman: Sounds good - will update README in next PR with finalized analysis details.

Issue #969, opened 2021-03-26
Proposed Analysis: Create broad histology-specific gene lists for oncoprint


jharenza:
#### What are the scientific goals of the analysis?
Since the oncoprint-landscape module currently utilizes a set of genes from the interaction-plots module, and we would like to break apart the oncoprint (per #968), the lists will not include canonical fusions/CNVs observed in the histologies, nor include canonical rare mutations, so we would like to manually curate broad histology-specific gene lists for the oncoprint.

#### What methods do you plan to use to accomplish the scientific goals?
literature, pedcbio, previously annotated [gene lists](https://github.com/marislab/create-pptc-pdx-oncoprints/tree/master/data) from the [PPTC manuscript](https://www.cell.com/cell-reports/pdfExtended/S2211-1247(19)31268-9).

#### What input data are required for this analysis?
NA

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
2-3 weeks

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@jharenza 

#### What relevant scientific literature relates to this analysis?
will compile as I complete this

jharenza:
@cbethell I created this file below, which I sent to CBTN PIs for review today, so if there are any additions, we can swap the file out afterward. I wanted to be sure we captured the sources to cite and there are currently duplicate genes due to multiple sources, so please just unique those. Hope this works!

[oncoprint-goi-lists-OpenPBTA.csv](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6327968/oncoprint-goi-lists-OpenPBTA.csv)

jharenza:
close with #1046

Issue #966, opened 2021-03-26
Updated analysis: oncoprint module files not writing/filenames inverted/change bgCol


jharenza:
#### What analysis module should be updated and why?
oncoprint-landscape

#### What changes need to be made? Please provide enough detail for another participant to make the update.
I tried to rerun the oncoprint code because I noticed that the oncoprint plot labels for `goi` and not using `goi` are reversed (even though the code seems to look fine, so maybe the files are out of date), but when I run the code, the files do not print.  I also wanted to update the figures to use the latest data and add `bgCol = "#F5F5F5"` in the  `oncoplot()` function.  There are no errors in the terminal when this happens either, so not sure why as of yet. My commands were:

```
docker run --name rerun-onco -d -e PASSWORD=pass -p 8787:8787 -v "$PWD":/home/rstudio/OpenPBTA ccdlopenpbta/open-pbta:latest
docker exec -ti rerun-onco bash -c "cd  /home/rstudio/OpenPBTA && analyses/oncoprint-landscape/run-oncoprint.sh"
```
and also interactively in docker bash:
```
docker exec -ti rerun-onco bash 
cd  /home/rstudio/OpenPBTA
bash analyses/oncoprint-landscape/run-oncoprint.sh
```

#### What input data should be used? Which data were used in the version being updated?
same input data files (for now), but V18 update

#### When do you expect the revised analysis will be completed?
within a week

#### Who will complete the updated analysis?
@cbethell and/or @jharenza

jharenza:
User mistake of course.. I had updated the script in my local master OpenPBTA-analysis repo clone instead of my forked clone, so the printing now works as anticipated. I will look into the labeling and also assess whether we should update the GOI list at all for the manuscript.

jharenza:
@cbethell I am not sure why the filenames are not printing as expected - they seem to be coded correctly. Can you take a look next week, and potentially update the bgCol :) ?

jharenza:
completed with #975

Issue #964, opened 2021-03-23
Updated analysis: Update CN consensus calls to use ControlFREEC CN as default?


jharenza:
#### What analysis module should be updated and why?
[copy_number_consensus_call](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/copy_number_consensus_call)

Early on, we had noticed many gains/losses in our oncoprint, but had since created a focal CN file module. Still, while the oncoprints look better, there are still a lot of gains/losses within the TMB plots listed above the oncoprints, and worth mentioning/trying the below.

#### What changes need to be made? Please provide enough detail for another participant to make the update.
@zhangb1 noticed that for PNOC003 samples, there were still many losses being called for samples which, in some cases, had known amplifications (from clinical sequencing). He has summarized his findings [here](https://www.notion.so/d3b/Consensus-CN-issue-b10335e6c9474b65bcd1780a71273aaf). His rationale was that ploidy is based on ControlFREEC and thus, using its corresponding CN would probably result in a better estimate of overall CN. 

This piece of code:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2c1f5faf1f5687fa40b9b30fa6442798bd38557d/analyses/copy_number_consensus_call/scripts/bed_to_segfile.R#L119-L128

would change to:

```
copynum = ifelse(is.finite(freec_cn), # use cnvkit if available
                                 freec_cn, cnvkit_cn), #otherwise use freec
```

#### What input data should be used? Which data were used in the version being updated?
data being used and to be updated: 
consensus_seg_annotated_cn_autosomes.tsv.gz
consensus_seg_annotated_cn_x_and_y.tsv.gz
pbta-cnv-consensus.seg.gz

other data to be updated:
pbta-cnv-consensus-gistic.zip

#### When do you expect the revised analysis will be completed?
one week, if QC needed

#### Who will complete the updated analysis?
@kgaonkar6 or @jashapiro or @jharenza ?

Thoughts, @jashapiro and @jaclyn-taroni ?

jaclyn-taroni:
I am not opposed to this change in principle and would welcome a pull request. The evaluation would need to be more systematic (e.g., examine more if not all samples) and should be captured in this repository in a notebook.

jharenza:
> I am not opposed to this change in principle and would welcome a pull request. The evaluation would need to be more systematic (e.g., examine more if not all samples) and should be captured in this repository in a notebook.

[...]

PR #975, opened 2021-03-30
Update `bgCol` in oncoprint module 

cbethell:

The purpose of this PR is to update the `bgCol` of the oncoprint figures.

This is an update to an existing analysis.

#### What was your approach?

As recommended in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/966#issue-841969880, this PR updates the `oncoplot()` function in `oncoprint-landscape/01-plot-oncoprint.R` to use `bgCol = "#F5F5F5"` and re-runs the module script to update the figures in `analyses/oncoprint-landscape/plots`.

I also re-ran the oncoprint section of the `figures/generate-figures.sh` script to update the pub-ready oncoprint.

This PR addresses #966

Does the background color of the updated plots look as expected?

If I interpreted your comment [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/966#issuecomment-808565453) correctly @jharenza:
> I am not sure why the filenames are not printing as expected - they seem to be coded correctly

It appears that the filenames are indeed printing as expected, but the plots themselves may be displayed a bit differently than one would expect.

I came to this conclusion because I cross-checked the genes included in our genes of interest list (compiled of [`analyses/interaction-plots/results/gene_disease_top50.tsv`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/interaction-plots/results/gene_disease_top50.tsv) and [`analyses/focal-cn-file-preparation/results/consensus_seg_focal_cn_recurrent_genes.tsv`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/focal-cn-file-preparation/results/consensus_seg_focal_cn_recurrent_genes.tsv) and determined that the genes represented on the `_goi_oncoprint.png` plots are synonymous with those represented in this list.

With that being said, it appears that a default behavior of the [`oncoplot()`](https://bioconductor.org/packages/devel/bioc/manuals/maftools/man/maftools.pdf) function is to display the top 20 genes with mutations, a functionality that can be over-ridden if supplied a genes of interest list with more than 20 genes. This is what seems to be happening in our case as when I tried supplying the argument `top = 50` rather than the default `top = 20` and re-running the script, the `goi` plots remained unchanged while the plots not using goi were updated with more genes. Would you agree?

See the comparison of the plots when adjusting this argument below:

with `top = 50`
all_participants_primary_only_oncoprint

with the default `top = 20`
all_participants_primary_only_oncoprint

**Question for reviewers:** Do we want to update the `top` argument for the plots that do not use the goi or do we want to wait for the updated goi (mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/969#issue-842262236) before making that decision?

Yes.

No new results, just updated the oncoprint plots.

No new results to summarize.

#### Documentation Checklist


jharenza:
Hi @cbethell ! Thanks for checking in on this. The bgCol looks good! Could we also change the multi-hit color while we have this open, since it is pretty bright, doesn't really fit with the other colors, and we have many multi-hits here?

I actually didn't catch the `top` behavior, but I think top 20 or even 25 would be fine, but 50 would likely be too much for the visual. 

> analyses/focal-cn-file-preparation/results/consensus_seg_focal_cn_recurrent_genes.tsv 

Ahh, I think I missed that this was going into the `goi` list, and had thought the oncoprints without the (what seem to be artifactual) CN losses were not the `goi` lists and that the ones not labeled with `goi` were coming from the `goi` list because those genes are mostly biologically relevant here. We are also re-assessing CN with #964, so hopefully, this will change. It is odd that of all the recurrent CNVs, I am only seeing deletions and the fact that all deletions are occurring in the same samples is perhaps suggesting either these are artifacts or these genes are all on one chr arm which is deleted, so they would not be focal. However, this is way past beyond the scope of your PR :), and I will look closer into that analysis once #964 is completed.

That being said, if we have recurrent SNVs and CNVs in our GOI lists, then it might follow that we add recurrent fusions or fused genes, which can be found in this [results folder](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering/results). Since the recurrent fusion analysis was done per histology, and there are not very many, we could simply add [these genes](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/fusion_filtering/results/pbta-fusion-recurrently-fused-genes-byhistology.tsv) to our `goi` list. I can add a separate ticket for that since this is also beyond the initial ticket.

cbethell:
Thanks for the review @jharenza!

> Could we also change the multi-hit color while we have this open, since it is pretty bright, doesn't really fit with the other colors, and we have many multi-hits here?

I changed the multi-hit color in the last commit (I agree it was pretty bright), so you can let me know what you think of the updated color! I'll be happy to update it if you had something else in mind.

> I actually didn't catch the `top` behavior, but I think top 20 or even 25 would be fine, but 50 would likely be too much for the visual.

I also updated the `top` argument to `top = 25`. 

Overall, let me know what you think of the oncoprints now (as far as the scope of this PR that is)!

jharenza (approved):
looks good! thanks for the update!

Issue #968, opened 2021-03-26
Updated analysis: Add oncoprints by `broad_histology` + update consensus maf to use hotspot consensus maf


jharenza:
#### What analysis module should be updated and why?
oncoprint-landscape

#### What changes need to be made? Please provide enough detail for another participant to make the update.
- Add oncoprints by `broad_histology` in panel form for easier viewing. 

#### What input data should be used? Which data were used in the version being updated?
- Use histology-specific gene lists per #969 
- Update consensus maf to use hotspot consensus maf

#### When do you expect the revised analysis will be completed?
?

#### Who will complete the updated analysis?
someone from CCDL

jharenza:
@cbethell I saw you added a PR as I was writing this! 

In looking at the `independent-specimens.wgswxs.primary.tsv` file being used for the oncoprint, I started thinking of how we want to create panels for this plot since we have such a spread for N per histology (below)

broad_histology | Freq
-- | --
Low-grade astrocytic tumor | 227
Embryonal tumor | 132
Diffuse astrocytic and oligodendroglial   tumor | 80
Ependymal tumor | 60
Benign tumor | 33
Tumors of sellar region | 30
Neuronal and mixed neuronal-glial tumor | 25
Tumor of cranial and paraspinal nerves | 20
Meningioma | 17
Mesenchymal non-meningothelial tumor | 15
Pre-cancerous lesion | 12
Germ cell tumor | 8
Choroid plexus tumor | 7
Histiocytic tumor | 6
Tumor of pineal region | 4
Metastatic tumors | 3
Other astrocytic tumor | 2
Lymphoma | 1
Melanocytic tumor | 1
Non-CNS tumor | 1
Non-tumor | 1
Other tumor | 1

It may look best if we create plots for the following 4 broad histologies and combine all of the others into `Other CNS` while keeping their `broad_histology` labels. 
Low-grade astrocytic tumor
Embryonal tumor
Diffuse astrocytic and oligodendroglial tumor
Ependymal tumor

I am also thinking we may want to remove benign tumor, pre-cancerous lesion, non-CNS tumor, and non-tumor as well, but I will confirm this latter point tonight with @adamcresnick when we meet.

cbethell:
> It may look best if we create plots for the following 4 broad histologies and combine all of the others into `Other CNS` while keeping their `broad_histology` labels.
> Low-grade astrocytic tumor
> Embryonal tumor
[...]

PR #983, opened 2021-04-01
Refactor oncoprint module to include plotting by broad histology

cbethell:

The purpose of this PR is to refactor the oncoprint module to include an option for plotting by broad histology per #968

NA

#### What was your approach?

I included some logic in the `oncoprint-landscape` module bash script to iterate through the given broad histology values (currently supplied in an array in the bash script) and adapted to `01-plot-oncoprint.R` script to filter the metadata file, as well as the additional data files, to include only the data associated with the given broad histology.

I was able to create oncoprints for the categories specified on the associated issue https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/968#issuecomment-812175023
> It may look best if we create plots for the following 4 broad histologies and combine all of the others into Other CNS while keeping their broad_histology labels.
Low-grade astrocytic tumor
Embryonal tumor
Diffuse astrocytic and oligodendroglial tumor
Ependymal tumor

The broad histology goi lists should run in this script without a glitch once provided to the `goi_list` flag in the bash script.

Per another https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/968#issuecomment-812223409 left on the related issue, the `Normal` display group value is associated with `broad_histology` values of `NA` (a decision that was made when mapping the histology color labels https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/899#issuecomment-759504742), but no longer appears on the oncoprints as the "Other CNS" plots are now created by hardcoding the `broad_histology` categories we do want (the hard-coding is mostly due to some weird R behavior and in adhering to the preferred categories specified above and on the related issue).

Part of #968 (this PR does not cover placing the plots in a panel form and the consensus maf file does not yet appear ready for implementation)

The added broad histology plots will need a particularly close look.

I do want to note that on the "Other CNS" oncoprints, the display group value `Low-grade astrocytic tumor` appears as it is associated with the `Other astrocytic tumor` broad histology. Is this intended behavior?
Screen Shot 2021-04-02 at 10 17 34 PM

**Edit:** This behavior indeed results from the intentional decision made when mapping the histology color labels per https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/899#issuecomment-759634809.

> I think you can combine `Other astrocytic tumor` into LGAT for now, since it is a low-grade astrocytoma. Possibly, we can update that `broad_histology` in v19. 

Yes.

Oncoprints for each broad histology.
The added plots include:
`all_participants_primary_only_diffuse-astrocytic-and-oligodendroglial-tumor_goi_oncoprint.png`
`all_participants_primary_only_diffuse-astrocytic-and-oligodendroglial-tumor_oncoprint.png`
`all_participants_primary_only_embryonal-tumor_goi_oncoprint.png`
`all_participants_primary_only_embryonal-tumor_oncoprint.png`
`all_participants_primary_only_ependymal-tumor_goi_oncoprint.png`
`all_participants_primary_only_ependymal-tumor_oncoprint.png`
`all_participants_primary_only_low-grade-astrocytic-tumor_goi_oncoprint.png`
`all_participants_primary_only_low-grade-astrocytic-tumor_oncoprint.png`
`all_participants_primary_only_other-cns_goi_oncoprint.png`
`all_participants_primary_only_other-cns_oncoprint.png`

`all_participants_primary-plus_diffuse-astrocytic-and-oligodendroglial-tumor_goi_oncoprint.png`
`all_participants_primary-plus_diffuse-astrocytic-and-oligodendroglial-tumor_oncoprint.png`
`all_participants_primary-plus_embryonal-tumor_goi_oncoprint.png`
`all_participants_primary-plus_embryonal-tumor_oncoprint.png`
`all_participants_primary-plus_ependymal-tumor_goi_oncoprint.png`
`all_participants_primary-plus_ependymal-tumor_oncoprint.png`
`all_participants_primary-plus_low-grade-astrocytic-tumor_goi_oncoprint.png`
`all_participants_primary-plus_low-grade-astrocytic-tumor_oncoprint.png`
`all_participants_primary-plus_other-cns_goi_oncoprint.png`
`all_participants_primary-plus_other-cns_oncoprint.png`

The plots appear to represent the accurate information and number of samples.

#### Documentation Checklist


analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -215,7 +260,7 @@ maf_object <- prepare_maf_object(
 # Given a maf object, plot an oncoprint of the variants in the
 # dataset and save as a png file.
 png(
-  file.path(plots_dir, opt$png_name),
+  file.path(plots_dir, tolower(gsub(" ", "-", opt$png_name))),
```
jaclyn-taroni, line 263: Good add 👍🏻 🚀

analyses/oncoprint-landscape/run-oncoprint.sh
```diff
@@ -6,6 +6,9 @@
 set -e
 set -o pipefail
 
+# Print oncoprints by broad histology - will not by default
+HISTOLOGY=${RUN_BROAD_HISTOLOGY:-0}
```
jaclyn-taroni, line 10: I think we can take this logic for skipping splitting up by broad histology out. We know we probably want to break the plots up into separate panels. The way you've altered `01-plot-oncoprint.R` means one could just run without `--broad_histology` and get plots for the entire cohort if we wanted!

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -156,18 +162,13 @@ if (!is.null(opt$goi_list)) {
   goi_list <- unique(unlist(goi_list))
 }
 
-# Read in recurrent focal CNVs file
-if (!is.null(opt$focal_file)) {
-  focal_df <- readr::read_tsv(file.path(opt$focal_file))
-}
-
```
jaclyn-taroni, line 163: This is something that should have been removed earlier, correct?

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -177,6 +178,50 @@ metadata <- metadata %>%
...
+  if (!opt$broad_histology == "Other CNS") {
+    metadata <- metadata %>%
+      dplyr::filter(broad_histology == opt$broad_histology)
+  } else {
+    metadata <- metadata %>%
+      dplyr::filter(
+        broad_histology %in% c(
+          "Tumors of sellar region",
+          "Neuronal and mixed neuronal-glial tumor",
+          "Tumor of cranial and paraspinal nerves	",
```
jaclyn-taroni, line 193: This extra spacing is required? 🤔

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -177,6 +178,50 @@ metadata <- metadata %>%
...
+          "Histiocytic tumor",
+          "Tumor of pineal region",
+          "Metastatic tumors",
+          "Other astrocytic tumor",
+          "Lymphoma",
+          "Melanocytic tumor",
+          "Other tumor"
+        )
+      )
+  }
```
jaclyn-taroni, line 207: I think this logic works well for now, but am making a note that this is an area where we might want to take a different approach when we have different genes of interest lists. I don't think it makes sense to make any changes to accommodate #969 just yet!

jaclyn-taroni (commented):
This looks good! I had a few comments and questions, the most substantive of which is moving to separate plots by broad histology as our default.

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -177,6 +178,50 @@ metadata <- metadata %>%
...
+  
+  # There's a duplicated `Tumor_Sample_Barcode` column that we will want to
+  # remove before filtering `maf_df`
+  maf_df <- maf_df[, -16]
```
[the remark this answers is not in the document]
cbethell: > * Is this a new column or did we see this in other releases?

@jaclyn-taroni, this duplicate column was added to the `00-map-to-sample_id.R` script [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6bdd22b504016e09df0d2801d79c34ddd6d83026/analyses/oncoprint-landscape/00-map-to-sample_id.R#L176) in merged PR #243 (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/243#discussion_r345794392). 

My understanding is that the original `Tumor_Sample_Barcode` column in the maf file is needed to filter to tumor samples only as this column holds the `Kids_First_Biospecimen_ID` values https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6bdd22b504016e09df0d2801d79c34ddd6d83026/analyses/oncoprint-landscape/00-map-to-sample_id.R#L127

However, because we use `sample_id` values in the `01-plot-oncoprint.R` script, the `Tumor_Sample_Barcode` column was recoded to hold the `sample_id` values
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6bdd22b504016e09df0d2801d79c34ddd6d83026/analyses/oncoprint-landscape/00-map-to-sample_id.R#L176

which resulted in a duplicated `Tumor_Sample_Barcode` column.

I initially tried to remove the original `Tumor_Sample_Barcode` column with the biospecimen IDs in the `00` script using:

`dplyr::select(-Tumor_Sample_Barcode) %>%` before the `rename` step linked above, but this initially produced an error for me, that I was interestingly unable to reproduce this morning so this may be a better solution here?

Note that it does not tell us it's duplicated, both columns are named `Tumor_Sample_Barcode`.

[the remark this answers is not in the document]
jaclyn-taroni: > `dplyr::select(-Tumor_Sample_Barcode) %>%` before the `rename` step linked above, but this initially produced an error for me, that I was interestingly unable to reproduce this morning so this may be a better solution here?

My understanding is as follows: we want to replace the biospecimen IDs in the `Tumor_Sample_Barcode` column with `sample_id` so that we can match between the experimental strategies, _but_ we need to call the `sample_id` column `Tumor_Sample_Barcode` because that's a requirement of `oncoplot()`. That makes me think that that what you lay out that previously gave you an error is a good move!

analyses/oncoprint-landscape/01-plot-oncoprint.R
cbethell: Correct!

cbethell:
In my [most recent commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/983/commits/27384147e6ce13aa3352d725bd0eceb72536bbb2) @jaclyn-taroni, I updating the logic handling the "duplicate" `Tumor_Sample_Barcode` column as I mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/983#discussion_r607980625, removed the flag added to the bash script as you suggested in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/983#discussion_r607333844 and removed that extra spacing that snuck into the hard-coded broad histology values.

Upon re-running, the "Other CNS" plots were updated due to that update in spacing (good catch!), otherwise the results stayed the same 👍.

All that said, I believe this is ready for another review!

analyses/oncoprint-landscape/run-oncoprint.sh
```diff
@@ -72,18 +84,30 @@ Rscript --vanilla 00-map-to-sample_id.R \
...
+
+  # Genes of interest only version of oncoprint
+  Rscript --vanilla 01-plot-oncoprint.R \
+    --maf_file ${intermediate_directory}/${primaryplus_filename}_maf.tsv \
+    --cnv_file ${intermediate_directory}/${primaryplus_filename}_cnv.tsv \
+    --fusion_file ${intermediate_directory}/${primaryplus_filename}_fusions.tsv \
+    --metadata_file ${histologies_file} \
+    --goi_list ${genes_list} \
+    --png_name ${primaryplus_filename}_"$histology"_goi_oncoprint.png \
+    --broad_histology "$histology"
```
jaclyn-taroni, line 111: Can we put this in the same loop as the primary only steps? I can't think of a reason not to at this stage.

jaclyn-taroni (approved):
This looks good! One comment about putting the primary and the primary plus steps in the same loop, but no need for me to review!

analyses/oncoprint-landscape/run-oncoprint.sh
cbethell: Good point, done!

PR #984, opened 2021-04-05
Fix CNV conensus call workflow

jashapiro:

In https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/964#issuecomment-811164531 @kgaonkar6 ran into some trouble with the CNV consensus workflow, and specifically the merge step. I confirmed the bug, and guessed that it might be due to the very long list of arguments to `cat` in that step of the workflow.

#### What was your approach?

To address the error, I converted the `merge_all` step to use a python loop and append to the output file rather than trying to pass the long file list to the shell. This seemed to solve the problem. It is probably a bit less efficient, but definitely not anything noticeable in the workflow as a whole. 

Results did not change, except for a single small ordering change on one line (not sure why that happened, but the new order is actually in order, so 🤷🏼).

Confirm that this fixes the bug for others, not just me.

If it works, I don't think so!


kgaonkar6 (approved):
Worked for me, thanks!

PR #987, opened 2021-04-06
Part1 Freec as default : Cnv consensus update

kgaonkar6:
According to Bo's rationale since ploidy is based on ControlFREEC using its corresponding CN would probably result in a better estimate of overall CN.

#### What was your approach?

Just a change in 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2c1f5faf1f5687fa40b9b30fa6442798bd38557d/analyses/copy_number_consensus_call/scripts/bed_to_segfile.R#L119-L128

to use controlfreec if available if not use cnvkit:

```
                copynum = ifelse(is.finite(freec_cn), # use freec if available
                                 freec_cn, cnvkit_cn), #otherwise use cnvkit
```

#964 

👀  Downstream analysis to check as first pass can be reviewed here :

1. 🚧  focal-cn-file-preparation will be affected by this change which is re-run with this update https://github.com/kgaonkar6/OpenPBTA-analysis/pull/2(https://github.com/kgaonkar6/OpenPBTA-analysis/pull/2) please note figures in `plots` is not updated. 
2. oncoprint-landscape uses `consensus_seg_most_focal_cn_status.tsv.gz` from `focal-cn-file-preparation` so we have also  re-run https://github.com/kgaonkar6/OpenPBTA-analysis/pull/3 to look at changes across all samples with this update.
3. cnv-chrom-plot was also re-run https://github.com/kgaonkar6/OpenPBTA-analysis/pull/4 with the controlfreec as default CN calls to look for genome wide changes.

Yes

table

By comparing `pbta-cnv-consensus.seg.gz` in this branch with v18 version file I found: 262 BS ids have atleast 1 segment `copy.num` updated . In total there are 2489 rows/regions per bs id that changed CN.
Columns named`_latest` are from the file in the branch and `_previous` is v18 release file 
[changelog.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6274148/changelog.txt)

#### Documentation Checklist


jharenza:
After doing some digging into the `changelog.txt` file above, I found the following and have updated a ticket in our D3b bixu tracker repo to ask for more information from @zhangb1. Copying below what I posted there (even though some links to the private repo can't be opened):

>hi @zhangb1  @yuankunzhu @aadamk - I need to re-open this given the request to update this within OpenPBTA. @kgaonkar6 had been working on this issue and thus far has opened 4 PRs:
>1. [Update consensus CNV calls to use ControlFREEC CN](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/987)
>2. [Update focal CNV files](https://github.com/kgaonkar6/OpenPBTA-analysis/pull/2)
>3. [Update oncoprint](https://github.com/kgaonkar6/OpenPBTA-analysis/pull/3)
>4. [Update chromosome wide plots](https://github.com/kgaonkar6/OpenPBTA-analysis/pull/4) - here, she has created a difference file and we see the following:
  a. 76% of calls have the same CN status after the change (new_status == 52610, same_status == 168895)
  b. There are 144 samples with calls having different CN status, and of these, there are 15 PNOC003 samples
  c. Of the PNOC samples, 0 genes in [this goi list](https://github.com/d3b-center/d3b-pnoc003-HGG-DMG-omics/blob/master/analyses/Oncoplot/input/pnoc003-dx/combined_gene_list) for the oncoprint are affected, thus, the oncoprints look identical using either ControlFREEC CN or CNVkit CN, below:

**ControlFREEC CN**
pnoc_primary_only_goi_oncoprint_freec
**CNVkit CN**
pnoc_primary_only_goi_oncoprint_cnvkit

>I cannot reproduce the above plot with the large number of deletions using either of these codes from OpenPBTA.

>My questions for @zhangb1 are:
>1. Have you tried using OpenPBTA consensus CNV + focal CNV code since May 2020? There may have been updates to these modules between May 2020 and now which may have resolved these deletions. 
>2. When you say 
>> only apply the tumor_ploidy from the histology file 

>in the comment above, what do you mean by that? The ploidy in the CNV module should match what is in the histology file.

>3. Based on [this comment](https://github.com/d3b-center/bixu-tracker/issues/665#issuecomment-637719049) from @aadamk, Was there a time when you were not using ploidy from the histology file, but instead using directly from ControlFREEC's output file, and if so, were there discrepancies between the two?

>General for @yuankunzhu and @zhangb1  - I think we should double check that all controlFreeC ploidy output matches what is in the histology file. I think this should still be the case because if it did not match and @zhangb1 was using info from ControlFREEC output and not the histology file, then I would expect I would see the large number of losses in my oncoprints.

jharenza (approved):
Based on the analysis provided in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1010#issuecomment-824923618), and internal checks performed on the PNOC samples compared to clinical reports generated and used in [this publication](https://pubmed.ncbi.nlm.nih.gov/30861105/) as discussed in [my comment here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1010#issuecomment-825190797), I think we should move forward with another option 1 (converting neutral CN to NA and using ControlFreeC as default CN when generating the consensus seg, which is reflected in this PR. Option 1 confirms the closest to true positives for this analysis. For all options 2,3,4, we still see spurious losses and gains, as evidenced by the gene matrices plotted in #1010.

jaclyn-taroni (approved):
Approving on the basis of the results discussed in #1010 and because the code changes in here address #964 and #1010. I haven't examined all the plots here and am going on trust!

kgaonkar6:
I thought about this a little more and checked for all changes in #1009, one change that I had not pulled from remote cnv-consensus-update (changed in #1009) was change in`figures/palettes/oncoprint_color_palette.tsv` thus the figures used the old palette and thus causing conflicts. 

I did the following : 
```
# gather changes from #1009 added in master
git checkout origin/master oncoprint-landscape
# gather all the changes from master merged in remote cnv-consensus-update
git pull origin cnv-consensus-update
# run oncoprint-landscape/run-oncoprint.sh
docker exec -ti cnv-rerun bash -c "cd /home/rstudio/OpenPBTA/analyses/ && oncoprint-landscape/run-oncoprint.sh
# commit oncoprint-landscape
git push origin cnv-consensus-update
```

Issue #991, opened 2021-04-12
CI failure: Molecular Subtyping - Ependymoma


jaclyn-taroni:
#### Context

The EPN subtyping step run in CI is now failing in CI. [Here's the `command` step for this module in CI](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/52f64f3debc7763520ac64ff685c69a7dd68bb2d/.circleci/config.yml#L61):

```
OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-EPN/run-molecular-subtyping-EPN.sh
```

This happens across multiple, seemingly-unrelated pull requests:

* https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5236/workflows/831388b0-c127-49f9-a939-b96aa127d5a7/jobs/5850
* https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5234/workflows/648ee559-a1d2-4d1b-95d4-7f8554338784/jobs/5848

But it is not occurring for pull requests that haven't been updated in 6+ days:

* https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5175/workflows/cf5140e2-563b-4eee-b4a8-e542c9f7babc/jobs/5780
* https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5212/workflows/7e17b9c2-7709-43a8-be7e-f2c69baaec78/jobs/5823
* https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5224/workflows/12bc726c-94fc-41f0-afef-159fbd0995dd/jobs/5835

#### Text of error

```
Generating analyses/molecular-subtyping-EPN/results/EPN_molecular_subtype.tsv that maps DNA and RNA ID's
Generating analyses/molecular-subtyping-EPN/results/EPN_all_data.tsv  that has all the relevant data needed for subtyping
[NbConvertApp] Converting notebook 03-subgrouping_samples.ipynb to html
Traceback (most recent call last):
  File "/usr/local/lib/python3.5/dist-packages/traitlets/traitlets.py", line 528, in get
    value = obj._trait_values[self.name]
KeyError: 'kernel_manager_class'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #992, opened 2021-04-12
Temporarily skip EPN subtyping in CI

jaclyn-taroni:

Related to #991. We have some unrelated pull requests failing CI (and failing to be tested themselves!) because the EPN subtyping module is failing for what seems to be a new reason. To avoid holding up PRs in the backlog, I think we should comment this step out temporarily. I've noted that this should be undone as part of #991, which also contains details of the error.


jashapiro (approved):
Does what needs to be done.

PR #1003, opened 2021-04-13
Try workaround for kernel issues in nbconvert step of EPN subtyping

jaclyn-taroni:

Related to #991 - we're having a problem with the kernel in the `nbconvert` step of the EPN subtyping module. This has only cropped within the past week, which is suggestive of some python dependency not being fully pinned. Unfortunately, we push `ccdlopenpbta/open-pbta:latest` to Dockerhub every time we merge a pull request so I can't easily go back and identify what versions of packages were installed before this error cropped up. 

Most of the information I've been able to dig up about this error so far is related to `conda`, so I'm not really sure what to make of that.  

Because we can't _execute the notebook_, we're not able to run the code that actually does the subtyping & update the table if the underlying data changes. A workaround for this is to use `nbconvert` to convert to a script and then run that script. This solution ends up being hacky because the notebook itself uses `display` so we can append the converted script such that `display` is imported with `sed`.

I don't like this, but it does run. Some more correct things to do are:

- Get to the bottom of the kernel error (?) - this seems like it could be kind of a rabbit hole for something that _doesn't necessarily need to be a notebook_. 
- Use a script instead in earnest, e.g., drop the notebook and `nbconvert` entirely
- Use `pip freeze` and either hardcode all of the python requirements in the Dockerfile or track the requirements file in git and use it to install Python dependencies when the Docker image is built (we should file an issue and do this anyway)

Tagging @jashapiro for discussion. I'll note that this is not particularly high priority and, again, we should do the `pip freeze` thing either way.


jaclyn-taroni:
@jashapiro can apparently run this module using a copy of the image he has locally 🎉  - here's the output of `pip3 freeze` on the image I have, which _does not_ run this module: [requirements.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6318335/requirements.txt)

jaclyn-taroni:
Closing in favor of #1008

PR #1009, opened 2021-04-15
Recode Oncoprints

### LINKED ISSUE #981: Updated analysis: update Multi-Hit definition
oncoplot-landscape

I noticed many of the `KIAA1549--BRAF` positive samples are annotated as `Multi-Hit`. I am wondering if this is because of the presence of reciprocal fusions `KIAA1549--BRAF` and `BRAF--KIAA1549` or if there are additional hits. I think that we need to separate out `Multi-Hit` from `reciprocal fusions`, and collapse these fusions, as a reciprocal fusion should only be counted once. If, however, there are multiple genes fused to one gene (which I am assuming is not the case here, but can be in other cases), then we might also consider a `Multi-Hit-Fusion` annotation so that we are delineating from `Multi-Hit`, which I usually take to mean distinct events (eg two different SNVs or an SNV+CNV or SNV+fusion).

same, v18

~1 week? 

@cbethell or someone at CCDL?

cbethell:

Per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/981#issuecomment-812181380, this PR recodes the `gain` and `loss` values as `Amp` and `Del`, respectively, for the purpose of taking advantage of default maftools behavior that requires this coding when plotting.

#### What was your approach?

This PR recodes the CNV file's `gain` and `loss` values as `Amp` and `Del` within the `01-plot-oncoprint.R` script of the `oncoprint-landscape` module.

This PR closes #981 as it allows for better visualization of the `Multi_Hit` instances.

The updated oncoprint plots in `oncoprint-landscape/plots` should receive a close look to ensure this is the behavior we expect and want.

I will note that upon recoding, I updated the `oncoprint_color_palette.tsv` file in `figures/palettes` to match the new coding. I also removed the `amplification` value from this file as it did not appear in any of the previous oncoprints and would conflict with the new coding.

Yes.

No new results, just the broad histology oncoprints in `oncoprint-landscape/plots` were updated.

Based on the example left in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/981#issuecomment-812181380, the plots appear to now reflect the behavior we would want.

#### Documentation Checklist


jharenza:
Hi @cbethell - I think this looks great for Amp/Dels. I did notice (especially for LGG `KIAA1549--BRAF` fused tumors) that we are still getting a `Multi-Hit Fusion` for these.  

From [this issue](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/981), I had added:
> I think that we need to separate out Multi-Hit from reciprocal fusions, and collapse these fusions, as a reciprocal fusion should only be counted once. 

To clarify - the `Multi-Hit Fusion` was mainly created within the PPTC project because there were some instances of a gene with promiscuous partners. Since in this dataset, we have so many kinase fusions which often have reciprocals (eg: `KIAA1549--BRAF` and `BRAF--KIAA1549`), I think we should collapse these as one fusion event - so any occurrence of a reciprocal fusion - same genes/opposite direction, we should just note as `Fusion` in the oncoplot. I think that will keep things clearer for those reading the plots.

jharenza (changes requested):
As noted in the comment, if we can update the reciprocal fusions to not be multi-hit, I think that would take care of the rest of this PR.

analyses/oncoprint-landscape/00-map-to-sample_id.R
```diff
@@ -185,10 +185,36 @@ readr::write_tsv(maf_df, maf_output)
...
+  group_by(Sample, FusionName, Fusion_ID) %>%
+  mutate(SortedFusionName = paste(sort(unlist(strsplit(FusionName, "--"))), collapse = "--")) %>%
+  ungroup()
+
+# Now let's perform some filtering on our sorted data frame
+fus_filter <- fus_sort %>%
+  group_by(Sample, SortedFusionName) %>%
+  count() %>%
+  # Filter out redundant instances that are not true multi hit fusions
+  filter(n <= 1)
```
jharenza, line 200: Here, you can select out the columns you need for `fus_sep` to shorten this df.

jharenza (changes requested):
Hi @cbethell! Thanks for updating this!

I think the output looks exactly like we expect now and I did run the code and do some spot checking, and all looks good with your logic. I just made a small suggestion about shortening one of the dataframes.

I also noticed that `all_participants_primary_only_oncoprint.png` and `all_participants_primary-plus_oncoprint.png` are still in the code/plot output. Either we should keep them and update this new fusion code to be rendered in those plots or remove that piece of the code. I think that removing them should be fine.

Otherwise, I think this is almost set!

cbethell:
> Hi @cbethell! Thanks for updating this!
> 
> I think the output looks exactly like we expect now and I did run the code and do some spot checking, and all looks good with your logic. I just made a small suggestion about shortening one of the dataframes.
> 
> I also noticed that `all_participants_primary_only_oncoprint.png` and `all_participants_primary-plus_oncoprint.png` are still in the code/plot output. Either we should keep them and update this new fusion code to be rendered in those plots or remove that piece of the code. I think that removing them should be fine.
> 
> Otherwise, I think this is almost set!

@jharenza, thanks for the re-review! Although the updated code was not quite ready as I noticed instances where there are more than one `multi-hit fusions` per sample but only one would be labeled. That said, it appears that I successfully fixed this issue in the most recent commit and _now_ everything should look as we would expect!

I also could not locate where `all_participants_primary_only_oncoprint.png` and `all_participants_primary-plus_oncoprint.png` are still in the code, but I removed the plots from this modules `plots` directory as you mentioned and renamed the histology plots (in the shell script) to exclude the `all_participants` naming.

analyses/oncoprint-landscape/00-map-to-sample_id.R
cbethell: This df was removed altogether in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009/commits/b640f5d9d133e004b37b84a306cffd92b216423e so I am going to resolve this conversation.

jharenza:
>@jharenza, thanks for the re-review! Although the updated code was not quite ready as I noticed instances where there are more than one multi-hit fusions per sample but only one would be labeled. That said, it appears that I successfully fixed this issue in the most recent commit and now everything should look as we would expect!

I think something happened to the reciprocal fusions in this latest commit in that I think we are missing a lot of them. I have been using LGG as a guide (eg in which we should see upwards of 50% of these samples harboring `KIAA1549--BRAF` fusions). See this [oncoplot](https://pedcbioportal.kidsfirstdrc.org/results?plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D&tab_index=tab_visualize&Action=Submit&session_id=5ff36630080e9a001092b7ce) from the same data in pedcbio. Also pasted below.

Screen Shot 2021-04-26 at 7 07 32 PM

>I also could not locate where all_participants_primary_only_oncoprint.png and all_participants_primary-plus_oncoprint.png are still in the code, but I removed the plots from this modules plots directory as you mentioned and renamed the histology plots (in the shell script) to exclude the all_participants naming.

I see they are now removed - I think the code was still in the bash script, but don't see it now. Thanks!

jharenza (changes requested):
See the comment about the reciprocal fusions gone missing after last update.

cbethell:
> I think something happened to the reciprocal fusions in this latest commit in that I think we are missing a lot of them. I have been using LGG as a guide (eg in which we should see upwards of 50% of these samples harboring `KIAA1549--BRAF` fusions). See this [oncoplot](https://pedcbioportal.kidsfirstdrc.org/results?plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D&tab_index=tab_visualize&Action=Submit&session_id=5ff36630080e9a001092b7ce) from the same data in pedcbio. Also pasted below.

Ahh yes great catch @jharenza, after some combination of the code in the previous commits, I believe the logic _now_ produces what we would expect and we appear to have the reciprocal fusions back while retaining the true multi-hit fusions in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009/commits/8cec80d4d46a94e7e3b79a0b22aaaf189009833d (after using LGG as a guide and some further spot checking, this seems to be the case at least)! Let me know if you agree or if it appears that we may be missing some data -- I've also refactored a bit in the last set of updates so please feel free to comment on the structuring there!

jharenza (approved):
Thanks for updating this @cbethell. I don't know why I didn't think of this before, but @kgaonkar6 had created a [function](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/03e8360e06000efba00850b3d7c7722036dd9b0f/analyses/fusion_filtering/04-project-specific-filtering.Rmd#L171-L195) in `fusion_filtering` to identify reciprocal fusions and in fact,  that populates the `reciprocal_exists` column in the putative oncogenic file. So, if you want, you can leverage that column to possibly shorten that code. Up to you!

Otherwise, the outputs look good and I spot checked the few MET multi-hits from HGG and they look as expected.

cbethell:
> Thanks for updating this @cbethell. I don't know why I didn't think of this before, but @kgaonkar6 had created a function in fusion_filtering to identify reciprocal fusions and in fact, that populates the reciprocal_exists column in the putative oncogenic file. So, if you want, you can leverage that column to possibly shorten that code. Up to you!

> Otherwise, the outputs look good and I spot checked the few MET multi-hits from HGG and they look as expected.

I'll take a look at the code you linked, thanks @jharenza! 
Glad to hear that the output looks as expected otherwise :) 👍

cbethell:
> I don't know why I didn't think of this before, but @kgaonkar6 had created a [function](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/03e8360e06000efba00850b3d7c7722036dd9b0f/analyses/fusion_filtering/04-project-specific-filtering.Rmd#L171-L195) in `fusion_filtering` to identify reciprocal fusions and in fact, that populates the `reciprocal_exists` column in the putative oncogenic file. So, if you want, you can leverage that column to possibly shorten that code. Up to you!

Unfortunately, using the `reciprocal_exists` column did not necessarily shorten the code (perhaps if I went further down the rabbit hole it would have), but in the meantime, I _was_ able to refactor and shorten the code a bit otherwise (with the output looking as expected) in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009/commits/bbfef3657023e64174ea359ab7da7c4b002f1573.

jaclyn-taroni (commented):
#1063 includes my comments about how the fusions are being handled, returning a comment about recoding the CNAs because it might disappear if #1063 gets merged!

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -144,7 +144,10 @@ maf_df <- data.table::fread(opt$maf_file,
 
 # Read in cnv file
 if (!is.null(opt$cnv_file)) {
-  cnv_df <- readr::read_tsv(opt$cnv_file)
+  cnv_df <- readr::read_tsv(opt$cnv_file) %>%
+    dplyr::mutate(Variant_Classification = dplyr::case_when(Variant_Classification == "loss" ~ "Del",
+                                                            Variant_Classification == "gain" ~ "Amp",
```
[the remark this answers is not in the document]
cbethell: Good point, added this in the latest commit https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009/commits/54e5b57c198708c437ab9a4000bb8299af44bd33 👍

cbethell:
> #1063 includes my comments about how the fusions are being handled, returning a comment about recoding the CNAs because it might disappear if #1063 gets merged!

I've addressed your comment about recoding the CNAs in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009/commits/54e5b57c198708c437ab9a4000bb8299af44bd33 @jaclyn-taroni and re-ran this PR with said change and the merged changes from #1063. That said, I believe the plots _now_ look as we expect so this is ready for a re-review.

jaclyn-taroni:
I'm going to merge this now, since there are a number of outstanding oncoprint PRs that will need to be updated!

Issue #747, opened 2020-08-27
Updated analysis: medulloblastoma consensus subtypes


jharenza:
#### What analysis module should be updated and why?
#742 via <s>#738</s> #743

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Will need to be updated once #746 goes in

#### What input data should be used? Which data were used in the version being updated?
RNA-Seq collapsed tables
File to be released in #746 

#### When do you expect the revised analysis will be completed?
unsure

#### Who will complete the updated analysis?
probably @komalsrathi

komalsrathi:
@jharenza I think it should be PR https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/743 instead of https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/738. That's where I am using the expected subtypes to compare the classifiers with the pathology report.

jharenza:
> @jharenza I think it should be PR #743 instead of #738. That's where I am using the expected subtypes to compare the classifiers with the pathology report.

updated, thanks!

jharenza:
@komalsrathi - will you use the new file below as input for the classifier accuracy assessment? Thank you! 

[openPBTA-mb-pathology-subtypes.csv](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6321093/openPBTA-mb-pathology-subtypes.csv)

komalsrathi:
@jharenza can you assign this to me? I never see these tickets unless they are assigned (only get notified and then I have to search for it).

jharenza:
> @jharenza can you assign this to me? I never see these tickets unless they are assigned (only get notified and then I have to search for it).

done!

komalsrathi:
> @komalsrathi - will you use the new file below as input for the classifier accuracy assessment? Thank you!
> 
> [openPBTA-mb-pathology-subtypes.csv](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6321093/openPBTA-mb-pathology-subtypes.csv)

@jharenza This file does not contain two columns that we have in the current input file:
`Kids_First_Biospecimen_ID` and `other_molecular_findings`. These are also added in the output files. If you can update the csv file, it would be easy to incorporate. I don't want to make any local changes to this file on my end as I am seeing this was referenced in other tickets as well. Thanks!

jharenza:
Added BS_ID but removed `other_molecular_findings` to keep this lite for the release.
[...]

PR #1019, opened 2021-04-20
Updated analysis: medulloblastoma consensus subtypes

komalsrathi:

Classifier accuracy was re-evaluated with an updated pathology file. To assign molecular subtypes, a consensus was taken from the predictions of the two classifiers: medulloPackage and MM2S.

For the consensus output using batch-corrected input data, 29/43 (67.44%) and for the consensus output using uncorrected data, 30/43 (69.76%) match with the reported subtype. The difference between the two was that `BS_V96WVE3Z` is correctly predicted as SHH by consensus uncorrected output but not by batch-corrected output. Further digging shows that medulloPackage correctly predicts BS_V96WVE3Z as SHH using both batch-corrected and uncorrected input data but MM2S incorrectly predicts this sample as Group3 using batch-corrected input and hence no consensus was obtained for this sample using batch-corrected data.  

The following table shows the difference between the two outputs:

| Consensus Output | Kids_First_Biospecimen_ID | pathology_subtype | MM2S_best_fit | medulloPackage_best_fit | molecular_subtype | match |
|------------------|---------------------------|-------------------|---------------|-------------------------|-------------------|-------|
| Uncorrected      | BS_V96WVE3Z               | SHH               | SHH           | SHH                     | SHH               | TRUE  |
| Corrected        | BS_V96WVE3Z               | SHH               | Group3        | SHH                     | SHH                | NA    |

#### What was your approach?

1. Updated v17 histologies to v18
2. Used https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6338682/openPBTA-mb-pathology-subtypes.csv provided by @jharenza -> converted to rds and used that instead of `input/expected_class_v17.rds`

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/747

N/A

N/A

Yes

Table

#### Documentation Checklist


jharenza (changes requested):
Hi @komalsrathi - thanks for this update.

I noticed a few things that I think we should update. 

1. For `pathology_subtype == non-WNT`, if there is a subgroup predicted by a classifier as non-WNT (SHH, Group3, or Group4), you are saying this prediction is `false`, but I think we need to update those to `true`. Will you update this and update the accuracies? I am interested to see whether there is a strong difference in classifiers then or if they still remain on par with each other.
2. For cases in which the consensus does not match, but we have a `pathology_subtype`, I think we should add the `pathology_subtype` as the molecular_subtype. 

Just a general comment that I also found the accuracy dropped when using batch corrected consensus, which helps justify the use of uncorrected data here.

komalsrathi:
@jharenza updated both, thanks!

komalsrathi:
@jharenza updated both, thanks!

analyses/molecular-subtyping-MB/README.md
```diff
@@ -143,14 +143,7 @@ For each sample_id with multiple RNA samples, the following logic is implemented
...
 
-For both the consensus corrected output and consensus uncorrected output, 25/32 match with the reported subtype so the % accuracy is the same i.e. 78.125%. Between the two consensus outputs, there is an overlap of 24/25 matched subtypes. The following samples are mismatched between the two consensus outputs:
-
-BS_HB03GSHF is correctly predicted by consensus batch-corrected output but not by uncorrected output:
-
-| Consensus Output | Kids_First_Biospecimen_ID | pathology_subtype | MM2S_best_fit | medulloPackage_best_fit | molecular_subtype | match |
-|------------------|---------------------------|-------------------|---------------|-------------------------|-------------------|-------|
-| Corrected        | BS_HB03GSHF               | WNT               | WNT           | WNT                     | WNT               | TRUE  |
-| Uncorrected      | BS_HB03GSHF               | WNT               | SHH           | WNT                     | NA                | NA    |
+For the consensus corrected output, 29/43 (67.44%) and for the consensus uncorrected output, 30/43 (69.76%) match with the reported subtype. The following table shows the difference between the two outputs:
```
jharenza, line 146: @komalsrathi is this still correct? It seems to be that if each individual classifier has 90 or 95% accuracy alone, that we should not see such a drastic drop when combining them.

jharenza (changes requested):
Hi @komalsrathi thanks for the changes and the classifiers now look to be on par with their respective publications, which is great!

When I mentioned 
>For cases in which the consensus does not match, but we have a pathology_subtype, I think we should add the pathology_subtype as the molecular_subtype.

I neglected to think about the behavior when the pathology reports are not exact and I think that in these cases, we should only keep the pathology subtype when it is confident in one single subtype. Sorry about that - will you make that change?

Also made a comment about combined accuracy - think that may need an update as well?

analyses/molecular-subtyping-MB/README.md
komalsrathi: @jharenza thanks for catching that - I forgot to update the README.

jharenza (approved):
This looks great now, thank you!

komalsrathi:
@jharenza can you please check if this can be merged? Thanks!

jharenza:
> @jharenza can you please check if this can be merged? Thanks! 

I realized that with v19, we have added the mb pathology subtype file. Will you point to that now instead of the input RDS? Then we can merge. Thanks!

jharenza:
> @jharenza can you point me to that file - sorry looked under data/release-v19-20210423 but maybe missing something (probably need more coffee lol)

Ahh, it looks like we missed it in the md5sum file, so it is not being downloaded in this release. I will merge this once checks pass and we will add to v20.

PR #1018, opened 2021-04-20
De novo mutational signature extraction, PR 2/n

sjspielman:
### Purpose/implementation Section

This PR performs the next steps in analysis after PR #974, associated with issue #818. 

#### What was your approach?

Using `sigfit`, a range of 2-5 signatures was extracted with 3000 iterations each, using 5 random seeds and two inference models. Previous analyses from PR #974 suggested that \k in [3,4] would be appropriate, but a range of \k in [2,5] was run since `sigfit` will create a cosine similarity goodness-of-fit plot when >=4 k's are applied (hence, 2,3,4,5).

Extracted signatures were compared across inference conditions.

#### What GitHub issue does your pull request address?
Issue #818  

#### Which areas should receive a particularly close look?

+ Since we run `sigfit` with a range of signatures twice, I have modified the `03-de_novo_range_of_nsignatures.sh` to take an additional argument 0/1 for whether initial benchmarking should be performed or a more (attempted) robust inference. This may not have been ideal since PR #974 has already been merged, but seemed better than writing a duplicate script. 
+ The file `04-analyze_de_novo.nb.html` contains a table with identified signatures across conditions, and I tried to identify and contextualize a couple signatures which kept cropping up. Note that none of the identified signatures correspond to known CNS signatures.
+ Currently I only export signatures to a results file, but could export other information like sample exposures if we think this analysis is worth pursuing further.

#### Is there anything that you want to discuss further?

Further visualization _could_ be performed with some of the results, again if we think this analysis is worth pursuing given the lack of robustness to random seed.

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
 
Yes. Note that the README is not yet filled in with this analysis information; waiting until further discussion to solidify what this analysis fully entails.

#### What types of results are included (e.g., table, figure)?

+ Table is embedded in `04-analyze_de_novo.nb.html` with identified signatures
+ Results file `results/de_novo_signatures.RDS` contains the extracted _signatures_.

#### What is your summary of the results?

Clearly, we have extreme sensitivity to starting conditions, but the repeated observation of signatures associated with "defective DNA mismatch repair and microsatellite instability (MSI)" are intriguing. 

#### Reproducibility Checklist

- [X ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [X ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
[...]

PR #1020, opened 2021-04-20
Install gt from GitHub

jaclyn-taroni:
Related to https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1018 - `gt` is new to CRAN as of March 2020, which predates the MRAN snapshot for our Docker image. However, it only depends on R >= 3.2.0. We can install the most recent version from GitHub.


jaclyn-taroni:
Nope rlang >= 0.4.5 isn't going to happen, closing!

PR #1023, opened 2021-04-23
chordoma subtyping update: add pbta-histologies-base.tsv

kgaonkar6:
molecular subtyping modules would need to use `pbta-histologies-base.tsv` file as input since the final `pbta-histologies.tsv` file is only created after integrating subtyping results in molecular-subtyping-integrate.

#### What was your approach?
update `pbta-histologies.tsv` to `pbta-histologies-base.tsv` in the scripts

NA

NA

NA

No changes in results

#### Documentation Checklist


jharenza (approved):
looks good!

PR #1028, opened 2021-04-26
PBTA Histologies v19 Part 8 of N: Run molecular subtyping 

kgaonkar6:
🚨  🚨  🚨 merge #1039, #1040, #1041, #1042, #1043, #1044, #1045 before this PR.
I checked out each module as a different PR so that  it can be reviewed and then this PR can be specifically for molecular-subtyping-* modules which might have fewer ( then 124) files to review.

#### What scientific question is your analysis addressing?
Run molecular subtyping for v19 release

#### What was your approach?
I re-ran https://github.com/kgaonkar6/OpenPBTA-analysis/blob/v19-release/scripts/run-for-subtyping.sh with an addition in code to run tp53_nf1_score module to add TP53 alteration status for HGG samples and newly Ok-ed chordoma subtyping.

#### What GitHub issue does your pull request address?
#867 

#### Which areas should receive a particularly close look?
Changes in results per subtyping modules would be of most interest, please read my summary of the changes below.

#### Is there anything that you want to discuss further?
Update embryonal molecular subtyping

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What types of results are included (e.g., table, figure)?
tables

#### What is your summary of the results?
 - BS_JXF8A2A6: is removed as per https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/862
 - BS_F8K4VQMF and BS_GG9W3J9Y: is added to HGG subset because the pathology free text is updated in v19 release.

Kids_First_Biospecimen_ID |	pathology_free_text_diagnosis_latest | pathology_free_text_diagnosis_previous 
-- | -- | --
BS_F8K4VQMF  |	glioblastoma, idh-1 negative, who grade iv |	gliomatosis cerebri
BS_GG9W3J9Y | High-grade glioma/astrocytoma (WHO grade III/IV) | Gliomatosis Cerebri

In v18 because https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/972, the sample was removed in previous hgg subset using v18 because of updated exclusion if `Gliomatosis Cerebri` without grad 3or 4 in pathology free text diagnosis 

 - BS_D7XRFE0R and BS_KABQQA0T are now updated to the correct sample_id to match to it's DNA sample so the subtyping results is matched correctly

Kids_First_Biospecimen_ID | sample_id_latest | sample_id_previous
-- | -- | --
BS_D7XRFE0R | A18777 | 7316-5812
[...]

Issue #867, opened 2020-12-07
Planned release: V19


jharenza:
#### What data file(s) does this issue pertain to?
`pbta-histologies.tsv`
all RNA-Seq files

#### What release are you using?

V18

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

Remove BS_JXF8A2A6 from dataset per #862 
- [x] pbta-star-log-final.tar.gz
- [x] pbta-star-log-manifest.tsv
- [x] pbta-mend-qc-manifest.tsv
- [x] pbta-mend-qc-results.tar.gz
- [x] pbta-isoform-counts-rsem-expected_count.stranded.rds
- [x] pbta-isoform-expression-rsem-tpm.stranded.rds
- [x] pbta-fusion-arriba.tsv.gz
- [x] pbta-fusion-putative-oncogenic.tsv
- [x] pbta-fusion-recurrently-fused-genes-byhistology.tsv
- [x] pbta-fusion-recurrently-fused-genes-bysample.tsv
- [x] pbta-fusion-starfusion.tsv.gz
- [x] pbta-gene-counts-rsem-expected_count.stranded.rds
- [x] pbta-gene-expression-kallisto.stranded.rds
- [x] pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds
- [x] pbta-gene-expression-rsem-fpkm.stranded.rds
- [x] pbta-gene-expression-rsem-tpm.stranded.rds
- [x] independent-specimens.rnaseq.primary-plus-stranded.tsv
- [x] fusion_summary_lgat_foi.tsv
- [x] pbta-histologies-base.tsv
- [x] pbta-histologies.tsv
- [ ] md5sum.txt
 
Add new medulloblastoma subtype pathology results file #746

Update all molecular subtypes, broad_histologies, short_histologies, and integrated_diagnoses.

jaclyn-taroni:
I am not sure that this issue is totally up to date, but I did want to make note of something that will need to be included in the v19 release. 

[...]

Issue #862, opened 2020-12-01
Data: BS_JXF8A2A6 likely needs removal from dataset


jharenza:
#### What data file(s) does this issue pertain to?
all RNA-Seq files + histology file

#### What release are you using?

v17

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
`BS_JXF8A2A6` is listed as a craniopharyngioma tumor, but was subtyped as EWS due to the presence of an _EWSR1-FLI1_ fusion. In discussions with CHOP pathology, their opinion is that the pathology report is correct and the tumor sample for this patient is craniopharyngioma, however, there is no precedence for presence of an _EWSR1-FLI1_ fusion in this histology, therefore, this is likely a labeling or sample swap error. There are no DNA-Seq files associated with this patient, `PT_EHE800JJ`. We are awaiting review of slides (which have been requested and will take a few days), but if those confirm craniopharyngioma, we should remove this sample from the dataset.

cc @kgaonkar6

jharenza:
Slide review confirmed this tumor sample was a craniopharyngioma, so this RNA-Seq does not match and should be removed.

jharenza:
removed with #1026

PR #1040, opened 2021-04-27
PBTA Histologies v19 Part 2 of N: Update independent samples

kgaonkar6:
#### What scientific question is your analysis addressing?
Rna summary modules needed for subtyping need to be updated because BS_JXF8A2A6 is removed as per #862. In this PR  I'm re-running the independent-samples module.

| Module |  Reason |Brief Description | output
|--------|----------|--------|------
| [`independent-samples`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples) | BS_JXF8A2A6 is removed as per #862 | Generates independent specimen lists for WGS/WXS samples updated in  #795 and #797 and [comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/750#issuecomment-691526896)| `results/independent-specimens.wgs.primary.tsv` (included in data download) <br> `results/independent-specimens.wgs.primary-plus.tsv` (included in data download) <br> `results/independent-specimens.wgswxs.primary.tsv` (included in data download) <br> `results/independent-specimens.wgswxs.primary-plus.tsv` (included in data download) <br> `results/independent-specimens.rnaseq.primary-plus-stranded.tsv`(included in data download) <br> `results/independent-specimens.rnaseq.primary-plus-polya.tsv` (included in data download)

#### What was your approach?
Just re-run as part of the scripts/run-for-subtyping.sh in kgaonkar6:v19-release in #1028
Then create a new branch for module specific review:
git checkout v19-release analyses/independent-samples/

#### What GitHub issue does your pull request address?
#867 

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
NA

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What types of results are included (e.g., table, figure)?
tables

#### What is your summary of the results?
Some rna-seq sample selections have changed I believe because of the random selection 

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1041, opened 2021-04-27
PBTA Histologies v19 Part 3 of N: Update fusion filtering 

kgaonkar6:
| Module |  Reason |Brief Description | output
|--------|----------|--------|------
| [`fusion_filtering`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering) | BS_JXF8A2A6 is removed as per #862  | Standardizes, filters, and prioritizes fusion calls  | `results/pbta-fusion-putative-oncogenic.tsv`(included in data download) <br> `results/pbta-fusion-recurrent-fusion-byhistology.tsv` (included in data download) <br> `results/pbta-fusion-recurrent-fusion-bysample.tsv` (included in data download) 

#### What was your approach?
Just re-run as part of the scripts/run-for-subtyping.sh in kgaonkar6:v19-release in #1028
Then create a new branch for module specific review:
git checkout v19-release analyses/fusion_filtering/

#867 

Not fully sure why the samples are missing from `pbta-fusion-recurrent-fusion-bysample.tsv `

#### Documentation Checklist


jashapiro (commented):
There seem to be a lot of lines removed here in the recurrent lists, more than I would have expected. I initially thought this might be due to sample changes in the independent samples analysis, but then I would have expected some additions, not just removals. 

`FilteredFusion.tsv` only seems to remove `BS_JXF8A2A6` as expected, but other files have more samples removed, which I was not expecting. 

Is it possible that there was a mismatch between the data files & sample list for some part of this module, making it so the changes in  https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1040 might not be fully reflected here?

One other note: `pbta-fusion-putative-oncogenic.tsv` has more changes than I would have expected. Spot checking it looks like most of this is ordering changes, but I didn't do a careful analysis.

kgaonkar6:
I was not able to figure out the issue before while filing the PR but I think I figured out:

https://github.com/kgaonkar6/OpenPBTA-analysis/blob/5dba0062ae333a74ff6ee14bea6086fc21703f79/analyses/fusion_filtering/06-recurrent-fusions-per-histology.R#L40-L43

This code needs to be change cohort name from `PNOC003` to `PNOC`, That changed the 15 samples other than BS_JXF8A2A6.

kgaonkar6:
Updating the code to use PNOC instead of PNOC003 fixed those unwanted changes.

jashapiro (approved):
This looks good now, but I'm still a bit confused by the large number of changes in `pbta-fusion-putative-oncogenic.tsv` As I said, I don't think it is a problem; it seems like there might be just an ordering change there.

If you want to fix it, it looks like just adding an `arrange(Sample, FusionName)` at line 211 of 
`05-QC_putative_onco_fusion_dustribution.Rmd` would do the trick to prevent this from coming up again in the future, and make it easier to distinguish any changes that do occur after this point.  (You might want to fix the typo in the file name while you are at it. 😉 )

jashapiro:
@kgaonkar6 Thanks for the updates in efe4de5... you will want to make sure to delete the old files as well, so GH recognizes it as a move rather than new files.

kgaonkar6:
Ah yes removed those old files!

PR #1046, opened 2021-04-29
Update oncoprints to use histology-specific goi lists

### LINKED ISSUE #1004: Updated analysis: Update oncoprint-landscape module to use histology-specific goi lists

The `oncoprint-landscape` module should be updated to use the broad histology specific goi lists (as noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/968#issue-842260376) once they are available.

The `run-oncoprint.sh` script will need to be updated to use the appropriate broad histology specific gene of interest list for the respective broad histology oncoprint.

- histology specific gene lists per #969 

< 1 day

Likely @cbethell or someone at the CCDL

cbethell:

This PR incorporates histology-specific genes of interest lists for the appropriate histology oncoprints.

#### What was your approach?

Using the [oncoprint-goi-lists-OpenPBTA.csv](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6327968/oncoprint-goi-lists-OpenPBTA.csv) file, linked in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/969#issuecomment-821575204, I prepared a script (saved in `util`) to make each a column into its own TSV file, named appropriately for the associated histology and stored in `oncoprint-landscape/data` for use with `01-plot-oncoprint.R`

I then adjusted the file paths to the goi lists in the `run-oncoprint.sh` script.

This PR closes #1004

The goi plots have been updated so they should receive a close look to ensure that they were updated as expected.

The file path to the main goi file (in the `run-oncoprint.sh` script) will likely need to be updated.

Yes, although it will need to be re-ran once v19 is merged and the file path to the main goi file will likely need to be updated if it is planned to be included in the data release.

Updated goi oncoprints

The plots seems to be updated as expected

#### Documentation Checklist


analyses/oncoprint-landscape/util/prepare-goi-lists.R
```diff
@@ -0,0 +1,99 @@
...
+
+#### Command line options ------------------------------------------------------
+
+# Declare command line options
+option_list <- list(
+  optparse::make_option(
+    c("-g", "--goi_file"),
+    type = "character",
+    default = NULL,
+    help = "file path to comma-separatedGOI file that contains genes of interest per broad histology",
```
jharenza, line 45:
```suggestion
    help = "file path to comma-separated GOI file that contains genes of interest per broad histology",
```

jharenza (changes requested):
Hi @cbethell! I think this looks good and oncoprints and lists match as expected, but the `top = 25` parameter in the oncoplot function is not working as expected (I am seeing >25 genes in some of the oncoprints). 

We should get #1009 merged before this, but after the v19 release.

cbethell:
> Hi @cbethell! I think this looks good and oncoprints and lists match as expected, but the `top = 25` parameter in the oncoplot function is not working as expected (I am seeing >25 genes in some of the oncoprints).
> 
> We should get #1009 merged before this, but after the v19 release.

Hi @jharenza, is it the goi oncoprints that appear to not adhering to the expected behavior? If so, as noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/975#issue-604396060,

> it appears that a default behavior of the oncoplot() function is to display the top 20 genes with mutations, a functionality that can be over-ridden if supplied a genes of interest list with more than 20 genes.

In other words, the `top = 25` parameter is over-ridden in cases where a goi list has been supplied. That said, if a goi list contains >25 genes, >25 genes will be displayed on the oncoprint.

jharenza:
> In other words, the `top = 25` parameter is over-ridden in cases where a goi list has been supplied. That said, if a goi list contains >25 genes, >25 genes will be displayed on the oncoprint.

Ah, yes. I remember us discussing this, but could not find it at the time. Looking back at the PPTC paper code, it looks like what I had done to create the oncoprints as a max of N genes while using genelists was to get the summary of mutations per gene first, then read just that gene list into the oncoprint function:
https://github.com/marislab/create-pptc-pdx-oncoprints/blob/13e2fd942b33ae7f4ba3f0cec844c0631283b8f5/R/create-complexheat-oncoprint-revision.R#L20-L59

What do you think about implementing this?

cbethell:
> > In other words, the `top = 25` parameter is over-ridden in cases where a goi list has been supplied. That said, if a goi list contains >25 genes, >25 genes will be displayed on the oncoprint.
> 
> Ah, yes. I remember us discussing this, but could not find it at the time. Looking back at the PPTC paper code, it looks like what I had done to create the oncoprints as a max of N genes while using genelists was to get the summary of mutations per gene first, then read just that gene list into the oncoprint function:
> https://github.com/marislab/create-pptc-pdx-oncoprints/blob/13e2fd942b33ae7f4ba3f0cec844c0631283b8f5/R/create-complexheat-oncoprint-revision.R#L20-L59
> 
> What do you think about implementing this?

Ah yes @jharenza, I can refactor the module to take in GOI lists and top `N` as arguments and where both specified, count up the number of mutations in the genes in the GOI list from the MAF file and then subset the GOI list to the top N genes on the basis of the number of mutations, which is what seems to be implemented in the code you linked above!

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -251,6 +239,37 @@ maf_object <- prepare_maf_object(
...
+  if (!is.null(opt$top_n)) {
+    
+    # Select top `n` genes if the argument is provided
+    top_n <- ifelse(nrow(gene_sum) < opt$top_n, nrow(gene_sum), opt$top_n)
+    
+    goi_list <- goi_ordered[1:top_n,]
+    
+  }
+  
+}
```
jaclyn-taroni, line 272: The actual code changes I'm suggesting are untested, but I believe you will only have to take these steps if someone specifies a GOI list _and_ a top n, so you can simplify this a bit:

```suggestion
# Subset `maf_object` for histology-specific goi list
if (!is.null(opt$goi_list) & !is.null(opt$top_n)) {
  
  maf_object <- subsetMaf(
    maf = maf_object,
    tsb = metadata$Tumor_Sample_Barcode,
    genes = goi_list,
    mafObj = TRUE
  )
  
  # Get top mutated genes per this subset object
  gene_sum <- mafSummary(maf_object)$gene.summary
  
  # Sort to get top altered genes rather than mutated only genes
  goi_ordered <-
    gene_sum[order(gene_sum$AlteredSamples, decreasing = T),]
  
  # Select top `n` genes if the argument is provided
  top_n <- ifelse(nrow(gene_sum) < opt$top_n, nrow(gene_sum), opt$top_n)
  goi_list <- goi_ordered[1:top_n,]
  
}
```

analyses/oncoprint-landscape/01-plot-oncoprint.R
```diff
@@ -154,12 +144,10 @@ if (!is.null(opt$fusion_file)) {
...
 if (!is.null(opt$goi_list)) {
-  goi_files <- unlist(stringr::str_split(goi_list, ",| "))
-  # Read in using the `read_genes` custom function and unlist the gene column
-  # data from the genes of interest file paths given
-  goi_list <- lapply(goi_files, read_genes)
-    # Include only the unique genes of interest
-  goi_list <- unique(unlist(goi_list))
+  # Read in using the `read_tsv()` function
+  goi_list <- readr::read_tsv(tolower(gsub(" ", "-",opt$goi_list))) %>%
+    as.matrix()
```
jaclyn-taroni, line 149: I had a few questions about this step:

* Why do you need this to be a matrix? I suspect you want this to be a vector in practice?
* Why are the `tolower()` and `gsub()` steps necessary for reading in a file, where the user will specify the path to the gene list?

It's a bit tricky that the header for each GOI list is different because it encodes information about the disease type the genes are related to and one could accidentally change the file name and lose that info. I guess in the future this would be a good use case for JSON!

analyses/oncoprint-landscape/01-plot-oncoprint.R
cbethell: Upon testing this method does not appear to obey the `top_n` argument.

jaclyn-taroni: Can you expand on that a bit?

cbethell: It does not limit the amount of genes being displayed to `top_n`, it instead shows all of the genes listed in the goi list.

jaclyn-taroni: But the two `if` way worked?

analyses/oncoprint-landscape/01-plot-oncoprint.R
cbethell: > I had a few questions about this step:
> 
> * Why do you need this to be a matrix? I suspect you want this to be a vector in practice?

Good question @jaclyn-taroni, if I am not mistaken, it seems that maftools requires this to be a matrix. I did expect this to be a vector, but upon giving the `oncoplot()` function (or even the `subsetMaf()` function that was recently added) this object as a vector or even a list, I receive an error that looks like the following:

> Error in oncoplot(maf_object, clinicalFeatures = "display_group", genes = goi_list,  : 
> Oncoplot requires at-least two genes for plottng.

This error was fixed by the `as.matrix()` step.

> * Why are the `tolower()` and `gsub()` steps necessary for reading in a file, where the user will specify the path to the gene list?

This decision was made to satisfy the logic in the module shell script which uses a for loop to run `01-plot-oncoprint.R` for each broad histology and create arguments specific to each histology like the following:
> --goi_list ${oncoprint_data_directory}/"${histology}"_goi_list.tsv

analyses/oncoprint-landscape/01-plot-oncoprint.R
cbethell: That's correct.

analyses/oncoprint-landscape/01-plot-oncoprint.R
jaclyn-taroni: > This decision was made to satisfy the logic in the module shell script which uses a for loop to run `01-plot-oncoprint.R` for each broad histology and create arguments specific to each histology like the following:
> 
> > --goi_list ${oncoprint_data_directory}/"${histology}"_goi_list.tsv

This suggests to me that this is a problem to be solved in the shell script, not the R script.

Imagine a situation where you use `01-plot-oncoprint.R` outside the context of the shell script you are looping through and the actual file name includes capital letters - this would break!

analyses/oncoprint-landscape/01-plot-oncoprint.R
jaclyn-taroni: That suggests that the logic _isn't quite right_ because it's using the original list (`goi_list`) - the conditions for the `if()` are not being met. Since you don't want to subset the MAF and do the `mafSummary()` steps unless you have to and you only have to if you have a `top_n` argument I would see if you can get to the bottom of it.

cbethell:
As mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1053#issuecomment-833702530

It appears that in sample `7316-2285`, when looking at the **lgat** plots, the `display_group` value is `Neuronal and mixed neuronal-glial tumor` while the `broad_histology` value is `Low-grade astrocytic tumor`.

See the `display_group` representation in the example of `primary-plus_lgat_goi_oncoprint.png` below:

primary-plus_lgat_goi_oncoprint

I guess the question now is should we leave this info as is or should we recode the `display_group`? (A similar situation is occurring in the **Other CNS** plots where `Chordoma` is the `display_group` value associated with the broad_histology `Choroid plexus tumor` -- the value we do want). Any thoughts here @jharenza?

Otherwise, this PR now appears to do what we expect re incorporating `goi_list` and `top_n`.

jharenza:
hi @cbethell! For `7316-2285`, in v19, this has `broad_histology == Neuronal and mixed neuronal-glial tumor`, so this should be fixed now. For `Chordoma` samples, they now have `broad_histology == Chordoma`. The choroid plexus tumor was a mismatch because the WHO 2016 didn't have a spot for chordomas and we made it its own broad histology, as it is different. So, I think this should help.

cbethell:
> hi @cbethell! For `7316-2285`, in v19, this has `broad_histology == Neuronal and mixed neuronal-glial tumor`, so this should be fixed now. For `Chordoma` samples, they now have `broad_histology == Chordoma`. The choroid plexus tumor was a mismatch because the WHO 2016 didn't have a spot for chordomas and we made it its own broad histology, as it is different. So, I think this should help.

Great, thanks @jharenza! I'll be sure to re-run with v19 files.

jaclyn-taroni:
@cbethell can you run this with v19 again now that #1054 is in please? I'm going to start merging the subtyping PRs that should probably get rerun with v19!

cbethell:
As expected and noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1054#issuecomment-834304155, the ependymal plots were the only ones to be updated in the most recent commit https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/f735ceb22ef4f33d9df8cfa77ec69c31be86efd3 as a result of the updates in #1054 and they look as we would expect. 

The re-run with the v19 files also fixed the `display_group` issue mentioned in the comment above https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046#issuecomment-833874735 👍

jharenza (approved):
hi @cbethell and @jaclyn-taroni. I think this generally looks good as far as goi output. There is one thing I noticed - not sure if we want to fix here or elsewhere, and that is the display of genes when there are no mutations - these we should remove from the plots. I wasn't expecting this, but also did not add that to the ticket specifically. I will approve and let you both decide whether we need a new ticket for that.

cbethell:
Given the findings in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046#issuecomment-836652399, do we want to instead filter the `Intron`, `5'Flank`, and `3'Flank` cases out of the MAF file before subsetting and plotting?

cbethell:
> > Given the findings in [#1046 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046#issuecomment-836652399), do we want to instead filter the `Intron`, `5'Flank`, and `3'Flank` cases out of the MAF file before subsetting and plotting @jharenza ?
> 
> Oh, nice find! We definitely do not want to include intronic variants here, so I would suggest we can either remove them from the MAF to lighten that load or remove them from the altered samples table. We should keep the 3' and 5' flank - _TERT_ promoter mutations from #819 will be included as `5' Flank`. Based on the hotspot maf, we have `Variant_Classification` ==
> 
> ```
> Missense_Mutation 
> Splice_Site    
> Nonsense_Mutation
> Frame_Shift_Del   
> In_Frame_Ins   
> In_Frame_Del      
> Frame_Shift_Ins
> 5'Flank
> ```
> 
> so, everything else is currently captured.a

Thanks for the feedback @jharenza!

In the most recent commit https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/52ac162d3a9062412d23c84e39de1fed746e53a8, you'll find that @jaclyn-taroni and I made the decision to filter to genes with `AlteredSamples > 1`, as to handle any cases with 0 (or near to) mutations being displayed on the oncoprints.

We also added a flag called `include_introns`, which by default is equal to `FALSE` -- meaning that we have removed intronic variants from the oncoprints by default while the option to include them is also there if we need to at some point in the future.

jharenza:
>In the most recent commit 52ac162, you'll find that @jaclyn-taroni and I made the decision to filter to genes with AlteredSamples > 1, as to handle any cases with 0 (or near to) mutations being displayed on the oncoprints. 

Thanks! This looks like it does what it should - filters out introns and removes non-mutated genes, except now I am seeing that the TMB portion of the plot seems to be missing bars.
This can be seen vividly in **LGAT goi** and 
Screen Shot 2021-05-12 at 6 54 26 PM
**CNS other goi**
Screen Shot 2021-05-12 at 6 54 47 PM
Scrolling through above, I see this happened a few other times, and perhaps warrants its own issue. 

The other thing I noticed is that these oncoprints are missing the reciprocal fusion logic in that commit, but I see it in your latest commit [ea6e207](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/ea6e20794b33fe7d7362a39cdcf927a3200659c4). However, in this latest commit, I am seeing the odd behavior of goi and non-goi oncoplot names being reversed, but not even reversed because the goi lists aren't being populated properly. For instance: 

**primary-plus-low-grade-astrocytic-tumor_goi_oncoprint.png:**
Screen Shot 2021-05-12 at 7 03 34 PM
This is not the goi list, but the file refers to it as such

**primary-plus-low-grade-astrocytic-tumor_oncoprint.png:**
Screen Shot 2021-05-12 at 7 02 45 PM
There are genes in this that are not in the goi list, so I am not sure where exactly this list comes from?

The TMB issue seems to be [mostly] resolved in commit [ea6e207](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/ea6e20794b33fe7d7362a39cdcf927a3200659c4). I say mostly because there are just a few samples with no bars, but they seem to be very few compared to before and maybe those TMBs are below the visible log10 threshold which would be fine.

jaclyn-taroni:
The fact that there are no PNG files in files changed when we are changing how we set file names in the shell script here is certainly _suspicious_.

jharenza:
> The fact that there are no PNG files in files changed when we are changing how we set file names in the shell script here is certainly _suspicious_.

yes, I checked multiple times because I thought I was going crazy...

jaclyn-taroni:
Rerunning locally now 🎲

jaclyn-taroni:
New plots added in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/3ad02ded2656eb637d3e6d03802ecf66c0877247 @jharenza. I haven't looked at them in detail yet (that is a tomorrow problem!) but seems like the TMB issue might still exist in the GOI plots?

jharenza:
> New plots added in [3ad02de](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/3ad02ded2656eb637d3e6d03802ecf66c0877247) @jharenza. I haven't looked at them in detail yet (that is a tomorrow problem!) but seems like the TMB issue might still exist in the GOI plots?

Ok great - the oncoprint content looks correct now, but yes, the TMB issue still exists. It looks like the goi subset is resulting in a counting of only those gene alterations for the TMB calculation and in the non-goi plots, you can see that the TMB plots contain _all_ alterations. I may have seen this before when subsetting the maf. I think the only way to perhaps get around this would be to refactor the code in this order:
1. subset maf to goi list
2. gather top N goi from summary
3. plot full maf (not subsetted maf) but only plot genes from 2 (instead of top N from 1). 
If we want to test this in a new PR, I can submit an issue.

cbethell:
> > New plots added in [3ad02de](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/3ad02ded2656eb637d3e6d03802ecf66c0877247) @jharenza. I haven't looked at them in detail yet (that is a tomorrow problem!) but seems like the TMB issue might still exist in the GOI plots?
> 
> Ok great - the oncoprint content looks correct now, but yes, the TMB issue still exists. It looks like the goi subset is resulting in a counting of only those gene alterations for the TMB calculation and in the non-goi plots, you can see that the TMB plots contain _all_ alterations. I may have seen this before when subsetting the maf. I think the only way to perhaps get around this would be to refactor the code in this order:
> 
> 1. subset maf to goi list
> 2. gather top N goi from summary
> 3. plot full maf (not subsetted maf) but only plot genes from 2 (instead of top N from 1).
>    If we want to test this in a new PR, I can submit an issue.

With the changes made and committed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046/commits/d71f76b595bb14adb75c82b26358f52ef82e1841, it looks like the missing bars associated with the TMB portion of the goi plots have been retained, would you agree @jharenza?

jharenza:
>With the changes made and committed in d71f76b, it looks like the missing bars associated with the TMB portion of the goi plots have been retained, would you agree @jharenza?

Yes, this looks great and I think good to go!

jharenza (approved):
LGTM now!

Issue #990, opened 2021-04-09
Updated analysis: SNV consensus calling (for things other than substitutions)


jashapiro:
#### What analysis module should be updated and why?

`snv-callers`, and in particular the generation of the consensus MAF file.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

As noted in another review (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/961#pullrequestreview-632629314), the current SNV consensus script may not handle combining indel variants correctly. Matching of individual calls across callers is currently performed only on the basis of start location, on the erroneous assumption that all calls are a single nucleotide in length. This is true for SNVs (and MNVs, all of which are reduced to SNVs before merging), but does not apply to the small indels that might occur. 

INDEL variants should probably be treated entirely separately from substitutions, but the logic of how to do that is something that requires some investigation, as there are many possible edge cases. 

- The simple case is that start and end positions match across all three callers. Those we are almost certainly capturing, but they are probably being mixed up a bit with cases where...

- Start matches, but callers choose different end positions. (And vice versa). These should be rare, but are certainly possible if one caller decides a deletion is 5 bp and another 6bp, for example. I guess I would go with majority rule among the callers, if possible, in this case?

- Overlapping deletion calls with no shared start and end. I expect this to happen in cases where the first base(s) of a deletion is repeated at the end: For example, a `CATATAT -> CAT` deletion could start at position 2 or position 4 and different callers might make different, equally correct, decisions. The nice thing in this case is that they should match as to the reference and alt alleles. (Both are `ATAT -> _`, but this might not always be the case in situations like this.

- 'Overlapping' insertions could be quite difficult to detect! A change of `CAT -> CAAAT` might put the inserted bases before or after the original `A`, and I don't know that we would be able to easily identify that...

This is not the complete list of edge cases. 

We should start by pulling out all indel calls from strelka, mutect and lancet so we can investigate the properties a bit more carefully, then decide on a set of heuristics to guide the generation of an INDEL consensus.

I will also note that one option is to simply "trust" one caller for INDELs and use its calls exclusively, avoiding many of these potential issues. This has the advantage of being easy, but has obvious downsides as well!

#### What input data should be used? Which data were used in the version being updated?

maf files from strelka, mutect and lancet.

#### When do you expect the revised analysis will be completed?

1-2 weeks? Depends on results of initial investigations, and discussions about methods.

#### Who will complete the updated analysis?

[...]

PR #1050, opened 2021-05-04
update col types

kgaonkar6:
v19 maf files have updated columns https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/990#issuecomment-826879406.

#### What was your approach?
I updated the maf_coltypes.RDS to include the HotSpotAllele and remove column names not part of the maf formats. Since the individual callers also have some unique columns I'm filtering the individual caller files to the columns in maf_coltypes.RDS.

#990 

NA

NA

Yes

No results ( intermediate files are placed in scratch) the output will be created in #956 

NA

#### Documentation Checklist


jashapiro (commented):
I think it would be worth including a very small script in `utils`  that generates the coltypes RDS in addition to the RDS itself. This will make the change here more transparent, and easier for future users to adapt if necessary.

kgaonkar6:
Added a script to create maf_coltypes.RDS. I didn't number this script since we already have a 00-* script. I've added it to the run script to add it as a step in the analysis.

PR #1053, opened 2021-05-05
Oncoprint GOI list revisions

jaclyn-taroni:
Related to #1046 - summarizing the changes we chatted about in our meeting this morning and I've implemented here @cbethell:

- Changing the genes of interest list prep Rscript such that it's run once and the input file is hardcoded (`analyses/oncoprint-landscape/util/prepare-goi-lists.R` -> `analyses/oncoprint-landscape/00-prepare-goi-lists.R`)
- Using associative arrays (SO post: https://stackoverflow.com/questions/28725333/looping-over-pairs-of-values-in-bash) to loop through histology labels (e.g., the value in `broad_histology`) and GOI list file names

I think there are some slight differences in some of the outputs that I have not thought about in any kind of depth, but wanted to put this out there while I can because internet woes today 😂


cbethell (commented):
This looks good and ran fine locally @jaclyn-taroni , thank you for your suggestions/revisions!

I took a look at the plots that have changed and appears that the following plots are exhibiting similar behevior:
- `primary_only_other_oncoprint.png`
- `primary_only_other_goi_oncoprint.png`
- `primary-plus_other_oncoprint.png`
- `primary-plus_other_goi_oncoprint.png`
- `primary-plus_embryonal_oncoprint.png`
- `primary-plus_embryonal_goi_oncoprint.png`

where there appears to be some loss of `loss` and `gain` values, and additional `display_group` values have snuck in.

In the example below (including the `primary-plus_other_goi_oncoprint.png`), you can see where there are less `loss` and `gain` values displayed.

However, this looks to be what we want as it appears that this behavior is due to a loss of samples in the plots (meaning that we have successfully subsetted things I presume)!

Current figure in this PR:
primary-plus_other_goi_oncoprint

Previous figure:
Screen Shot 2021-05-06 at 7 43 00 AM

Also to note is the value `Chordoma` being added as a `display_group`. I have not yet discovered why this is the case.

As for the `primary-plus_lgat_oncoprint.png` plot, it looks like `Neuronal and mixed neuronal-glial tumor` snuck in there (see `display_group`).

jaclyn-taroni:
In the interest of the process of elimination, did you use the v19 data by any chance? Or maybe my focal data was "behind" yours 🤔

cbethell:
> In the interest of the process of elimination, did you use the v19 data by any chance? Or maybe my focal data was "behind" yours 🤔

I used v18 data, but perhaps the focal data is indeed not in sync, I can try deleting the output and re-running everything from your branch (including the focal script) to see if that changes things.

cbethell:
Update re the display group issue @jaclyn-taroni: 

It appears that in sample `7316-2285`, the `display_group` value is `Neuronal and mixed neuronal-glial tumor` while the `broad_histology` value is `Low-grade astrocytic tumor` so the logic appears to be working fine there.

I guess the question now is should we leave this info as is or should we recode the `display_group`? (A similar situation is occurring in the "Other CNS" plots where `Chordoma` is the `display_group` value associated with the broad_histology `Choroid plexus tumor` -- the value we do want)

As for the loss of `gain` and `loss` values in some of the plots, my guess here is that perhaps these files were not subsetted as expected previously and now they are being accurately subsetted, could that be possible?

My reasoning comes from taking another look at the differences between the figures I linked above in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1053#pullrequestreview-652849263

jaclyn-taroni:
> As for the loss of `gain` and `loss` values in some of the plots, my guess here is that perhaps these files were not subsetted as expected previously and now they are being accurately subsetted, could that be possible?
> 
> My reasoning comes from taking another look at the differences between the figures I linked above in [#1053 (review)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1053#pullrequestreview-652849263)

Because you're talking about `primary-plus_other_goi_oncoprint.png`, my guess is that it's because `top_n` + `goi_list` is working as intended = those losses / gains aren't included? Does that make sense?

cbethell:
> > As for the loss of `gain` and `loss` values in some of the plots, my guess here is that perhaps these files were not subsetted as expected previously and now they are being accurately subsetted, could that be possible?
> > My reasoning comes from taking another look at the differences between the figures I linked above in [#1053 (review)](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1053#pullrequestreview-652849263)
> 
> Because you're talking about `primary-plus_other_goi_oncoprint.png`, my guess is that it's because `top_n` + `goi_list` is working as intended = those losses / gains aren't included? Does that make sense?

Ahh YES of course!

jaclyn-taroni:
For the display group questions - I think if we think the code in this PR is _correct_ we should merge this and move the discussion to #1046.

PR #1054, opened 2021-05-06
Update top_n logic 

jaclyn-taroni:
Related to #1046 - updates logic as discussed in the thread https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1046#discussion_r626068485

The way it was written if someone passed `top_n` but not a GOI list, the top argument would not be respected because `25` was hardcoded. Also, if someone passed a GOI file but no `top_n`, I believe this would have been referencing a `Hugo_Symbol` column that may not have existed.

### DIFF
--- a/analyses/oncoprint-landscape/02-plot-oncoprint.R
+++ b/analyses/oncoprint-landscape/02-plot-oncoprint.R
@@ -98,7 +98,7 @@ option_list <- list(
   optparse::make_option(
     c("-n", "--top_n"),
     type = "integer",
-    default = NULL,
+    default = 25,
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
There were a few plot updates in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1054/commits/1e044595e1c0dc4db3c3dda8d5b3e0b3c3ffd76a. However, now the number of samples listed at the top in the ependymal plots with and without GOI lists are the same, which seems _more_ correct to me?

cbethell (approved):
LGTM ✅

Issue #998, opened 2021-04-12
Updated analysis: molecular subtyping pathology - JXG


jharenza:
#### What analysis module should be updated and why?
As outlined in #967, we would like to integrate some of the information in `pathology_free_text` diagnosis to the `integrated_diagnosis` or `harmonized_diagnosis`. For this ticket, we will further subtype `Other` tumors to `Juvenile xanthogranuloma` per the [WHO 2016 CNS subtypes](https://link.springer.com/content/pdf/10.1007/s00401-016-1545-1.pdf) based on information gathered from the pathology reports.

The following samples will be annotated here:
```
> v18 %>%
+   filter(grepl("xanthogranuloma|jxg", pathology_free_text_diagnosis)) %>%
+   select(pathology_free_text_diagnosis) %>%
+   distinct()
# A tibble: 2 x 1
  pathology_free_text_diagnosis           
  <chr>                                   
1 non- langerhans histiocytosis (jxg- cns)
2 xanthogranuloma 
```

#### What changes need to be made? Please provide enough detail for another participant to make the update.
If `pathology_diagnosis == Other` and `pathology_free_text_diagnosis` contains the below terms, then annotate as below:

pathology_diagnosis | subtyping module | pathology_free_text_diagnosis | broad_histology | short_histology | harmonized_diagnosis
-- | -- | -- | -- | -- | --
Other | NA | if contains "xanthogranuloma" or "jxg" | Histiocytic tumor | JXG | Juvenile xanthogranuloma

#### What input data should be used? Which data were used in the version being updated?
`pbta-histologies.tsv`

#### When do you expect the revised analysis will be completed?
this week?

#### Who will complete the updated analysis?
@kgaonkar6 or @jharenza

kgaonkar6:
closed via #1056

PR #1056, opened 2021-05-10
Add pathology-free-text-diagnosis based JXG subtypes

kgaonkar6:
Using WHO 2016 CNS subtypes to improve Juvenile xanthogranuloma harmonized diagnosis

#### What was your approach?
I  updated @jaclyn-taroni code for meningioma code for this PR, my thought process was that this way the path-free-text-dx code updates are easier to read/review. 
I used the following table from [#998](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/998) to recode the `harmonized_diagnosis` values for samples with a `pathology_diagnosis` of Juvenile xanthogranuloma:

pathology_diagnosis | subtyping module | pathology_free_text_diagnosis | broad_histology | short_histology | harmonized_diagnosis
-- | -- | -- | -- | -- | --
Other | NA | if contains "xanthogranuloma" or "jxg" | Histiocytic tumor | JXG | Juvenile xanthogranuloma

#998 

Eventually this code will be numbered and add to compiled subtype file to be used in `molecular-subtype-integrate`. Currently the output is in `analyses/molecular-subtyping-pathology/results/juvenile-xanthogranuloma-subtypes.tsv`

table

Updated 4 samples which contains "xanthogranuloma" or "jxg" in their pathology_free_text_diagnosis.

#### Documentation Checklist


jharenza (approved):
Thanks for this @kgaonkar6! This captures the two patients with this diagnosis and since the code was reviewed previously, we can probably merge this quickly.  cc @jaclyn-taroni

analyses/molecular-subtyping-pathology/pathology_free_text-subtyping-jxg.Rmd
```diff
@@ -0,0 +1,124 @@
...
+                          "xanthogranuloma|jxg") ~ "Histiocytic tumor",
+      TRUE ~ broad_histology
+    ),
+    short_histology = dplyr::case_when(
+      stringr::str_detect(pathology_free_text_dx_lower,
+                          "xanthogranuloma|jxg") ~ "JXG",
+      TRUE ~ short_histology
+    ),
+    Notes = "Updated via OpenPBTA subtyping from pathology_free_text_diagnosis"
+  ) %>%
```
jaclyn-taroni, line 105: I'm going to merge this as is because I believe it to be correct, but for future reference I believe you could simplify this by moving the `str_to_lower()` step earlier, filtering like you do in 75 (now with the added "safety" of everything being lowercase), and then doing the following for the `mutate()` step:
```R
  # String detection in pathology free text per the table
  dplyr::mutate(
    harmonized_diagnosis =  "Juvenile xanthogranuloma"
    broad_histology = "Histiocytic tumor",
    short_histology = "JXG",
    Notes = "Updated via OpenPBTA subtyping from pathology_free_text_diagnosis"
  )
```

The idea being that you would not have any samples that didn't meet the  `stringr::str_detect(pathology_free_text_dx_lower,
                          "xanthogranuloma|jxg")` conditions in your data frame so you don't need the `case_when()`. (All of this is untested but I do something similar elsewhere...)

jaclyn-taroni (commented):
I will merge once CI finishes, but wanted to provide a bit of feedback that doesn't need to be acted upon here!

analyses/molecular-subtyping-pathology/pathology_free_text-subtyping-jxg.Rmd
kgaonkar6: Ah right makes sense, that would have been easier!

PR #1063, opened 2021-05-11
Leverage reciprocal_exists for handling reciprocal fusions in OncoPrints

jaclyn-taroni:
Related to #1009 - We were having an issue where reciprocal fusions were showing up inappropriately as multi-hit fusions in the OncoPrints. During review, using `reciprocal_exists` in the putative oncogenic fusion file was raised https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1009#pullrequestreview-647703181 which I came across during my own review of #1009. It occurred to me that we only need to sort gene partners of fusions in alphabetical orders _when we know a reciprocal fusion exists_, where we then use the sorted results to collapse reciprocal fusions/make sure we're only counting them once. 

So here I'm proposing fusion code updates that do the following:

* Handle `reciprocal_exists == TRUE` and `reciprocal_exists == FALSE` separately, based on the considerations for each (e.g., sorting for `reciprocal_exists == TRUE` only) - I draw from the sorting code in #1009 but use `stringr` functionality because personal preference, mostly
* Binds the reciprocal and no reciprocal fusion data frames
* Handles multi-hit fusions the same way as we do in `master`

@cbethell I spot checked a couple differences between this code and what's in #1009 and it looks like this code correctly IDs some multi-hit fusions missed in #1009.

### DIFF
--- a/analyses/oncoprint-landscape/00-map-to-sample_id.R
+++ b/analyses/oncoprint-landscape/00-map-to-sample_id.R
@@ -22,6 +22,7 @@
 #   --independent_specimens ../../data/independent-specimens.wgswxs.primary.tsv
 
 library(dplyr)
+library(stringr)
 
 #### Command line options ------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
Because I'm not merging into `master`, no review required. Feel free to merge if this looks good, @cbethell !

Issue #1010, opened 2021-04-15
Updated analysis: Neutral region called as losses when compared to ploidy 


kgaonkar6:
#### What analysis module should be updated and why?
The CNV analysis specifically 

While creating the consensus seg file in `copy_number_consensus_call` module we bind a df of  *neutral* region in the code here to fill back neutral calls :
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/84a324790f9cb2848cd9e370832f0f4d0a7610b1/analyses/copy_number_consensus_call/scripts/bed_to_segfile.R#L149-L160

Which adds neutral regions ( that is regions  not called as del/dups in callable region bed) with a copy number = 2 to specify that it is a neutral call. 

However, when we are gathering the status of "loss"," gain" or "neutral" as an intermediate step between the consensus region seg file and gene annotated seg file in `focal-cn-file-preparation` module :
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/84a324790f9cb2848cd9e370832f0f4d0a7610b1/analyses/focal-cn-file-preparation/01-add-ploidy-cnvkit.Rmd#L43-L52

Along with the actual del/dups we are also comparing the copy number in these *neutral* region which were assigned copy number=2 as part of `copy_number_consensus_call` module. These are now being annotated as "loss" in `focal-cn-file-preparation` when they are compared to 3,4 etc tumor ploidy. I believe the neutral status of these regions should be propagated in `focal-cn-file-preparation` while annotating the region with genes. 

#### What changes need to be made? Please provide enough detail for another participant to make the update.

My suggestions are:

1) Add ploidy as copy number to neutral calls:
Depending on that we are still following this [note on ploidy](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-formats.md#a-note-on-ploidy)
Shouldn't the neutral calls that we are back filling in `copy_number_consensus_call` [code](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/84a324790f9cb2848cd9e370832f0f4d0a7610b1/analyses/copy_number_consensus_call/scripts/bed_to_segfile.R#L149-L160) have copy number == tumor_ploidy to be considered as a copy neutral call instead of 2? 
My understanding is that controlfreec also calls regions as neutral if the copy number== tumor_ploidy for samples >2 ploidy so this would be similar.

2) Add NA as copy number to neutral calls:
[...]

PR #1066, opened 2021-05-12
Part1: Freec as default and neutral NA

kgaonkar6:
⚠️ This is a re-do of #987 

According to Bo's rationale since ploidy is based on ControlFREEC using its corresponding CN would probably result in a better estimate of overall CN.

#### What was your approach?

Just a change in 
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2c1f5faf1f5687fa40b9b30fa6442798bd38557d/analyses/copy_number_consensus_call/scripts/bed_to_segfile.R#L119-L128

to use controlfreec if available if not use cnvkit:

```
                copynum = ifelse(is.finite(freec_cn), # use freec if available
                                 freec_cn, cnvkit_cn), #otherwise use cnvkit
```

#964 

NA, it is just a 1 line code update

Note: I have not added any downstream modules  ( like I did in #987) in this PR since those downstream analysis are just re-runs and can come in after this PR is merged to master.

yes

tables

Here I'm comparing ( consensus_seg_annotated_cn_autosomes.tsv.gz + consensus_seg_annotated_cn_x_and_y.tsv.gz ) from the following versions of the files
- controlfreec (default for copy number ) + neutral calls as NA
- cnvkit (default for copy number) + neutral calls as NA
- and v18 release 
image

The changes between using controlfreec and cnvkit are mainly as follows:
- BS_ZSH09N84 and BS_CBMAWSAR are initial samples from the same participant PT_MNSEJCDM, controlfreec calls EGFR and MET gains in both samples, cnvkit only calls in BS_ZSH09N84
- BS_JRFVST47 and BS_3Z40EZHD are initial samples for PT_9GKVQ9QS, controlfreec calls MET gain in both samples , cnvkit only calls for BS_3Z40EZHD
- Other than the above calls being missed, we also see some copy change level changes in KIT,KDR and PDGFRA calls where ,controlfree calls amplification for BS_JRFVST47 and cnvkit shows as gains; for BS_3Z40EZHD cnvkit shows as amplification and controlfreec shows as gains.

My underdstanding is controlfreec helps identifying consistent calls for multiple samples from the same participant ids where as cnvkit seems to be missing the call for 1 of the 2 multiple samples in the participants mentioned above.

 

#### Documentation Checklist


jaclyn-taroni (approved):
👍🏻 approving on the basis of discussion #1010 + the code does what is described in #964 and #1010! (I did not look at the SEG file.)

kgaonkar6:
I have not added any downstream modules ( like I did in #987) in this PR since those downstream analysis are just re-runs and can come in after this PR is merged to master.

Issue #1049, opened 2021-05-01
Updated analysis: Update PFA EPN subtyping for EZHIP over-expression


jharenza:
#### What analysis module should be updated and why?
molecular-subtyping-EPN

Quick update to include over-expression of _EZHIP_ (_CXorf67_) alone, as a biomarker of PFA EPN, according to the following publication:
[EZHIP is a specific diagnostic biomarker for posterior fossa ependymomas, group PFA and diffuse midline gliomas H3-WT with EZHIP overexpression
](https://actaneurocomms.biomedcentral.com/articles/10.1186/s40478-020-01056-8) since 1q gain is not present in _all_ PFA EPNs ([ref](https://actaneurocomms.biomedcentral.com/articles/10.1186/s40478-019-0820-5) and [ref](https://ascopubs.org/doi/10.1200/JCO.18.01765)). 

A quick look at RNA-Seq collapsed data and pedcbio tells me we will have more samples which can fit into this subtype.
<img width="1127" alt="Screen Shot 2021-04-30 at 7 55 35 PM" src="https://user-images.githubusercontent.com/207034/116764151-12580100-a9ee-11eb-9b65-f568a0b47cfb.png">

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Add `row["CXorf67_expr_zscore"]>3) or` to logic for classifying PFA EPN within the below code:

```
def prioritizing_PT_EPN(row, sample_list):
    if((row["CXorf67_expr_zscore"]>3 and row["1q_gain"]>0) or
        (row["TKTL1_expr_zscore"]>3  and row["1q_gain"]>0)):
        sample_list.append(row["sample_id"])
        return("EPN, PF A")
```

#### What input data should be used? Which data were used in the version being updated?
RNA-Seq collapsed matrices

#### When do you expect the revised analysis will be completed?
1-2 weeks

#### Who will complete the updated analysis?
@kgaonkar6 

Also wanted to ask @jashapiro and @jaclyn-taroni why we had settled on a z-score >3 for identifying over-expression? Pedcbio uses >2, but I have a feeling we did this to have higher confidence. Curious.

jaclyn-taroni:
> Also wanted to ask @jashapiro and @jaclyn-taroni why we had settled on a z-score >3 for identifying over-expression? Pedcbio uses >2, but I have a feeling we did this to have higher confidence. Curious.

[...]

PR #1068, opened 2021-05-12
EPN subtyping update EZIP based subtyping

kgaonkar6:
Over-expression of EZHIP (CXorf67) alone should be added as a biomarker of EPN, PF A

#### What was your approach?
Add  EZHIP (CXorf67)  over expression alone for classifying EPN, PF A as:

```
def prioritizing_PT_EPN(row, sample_list):
    if( row["CXorf67_expr_zscore"]>3)  or
        (row["CXorf67_expr_zscore"]>3 and row["1q_gain"]>0) or
        (row["TKTL1_expr_zscore"]>3  and row["1q_gain"]>0)):
        sample_list.append(row["sample_id"])
        return("EPN, PF A") 
```

#1049 

Only one sample (7316-3624) was updated to EPN, PF A did we expect more samples to be subtyped with this condition?  

In the bash script it says if OPENPBTA_SUBSET > 0 then it prints "Subsetting  for CI" but is run by default.
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/9afff4aaf3c8962c778b4743857e6e25afff61f6/analyses/molecular-subtyping-EPN/run-molecular-subtyping-EPN.sh#L39

I changed it to just "Subsetting", let me know if I'm wrong.

Yes

tables

7316-3624 was added as EPN, PF A subtype

#### Documentation Checklist


jharenza (commented):
Hmm, I did a cross-check with the data here and that in pedcbio. I know that here, we use z-score >3 but in pedcbio, they are using z-score >2 to determine a significant threshold for increased expression. However, of the three samples (PT_ZZRBX5JT, PT_ZA95JQEB, PT_2CJ31SBX) using the higher cutoff here, two (PT_ZA95JQEB,  PT_2CJ31SBX) are not in the >2 cutoff in pedcbio, which is suspicious to me and makes me worry about us calling this subtype. Below, I am listing the 20 samples in the [oncoprint below](https://pedcbioportal.kidsfirstdrc.org/results/oncoprint?plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D&data_priority=0&tab_index=tab_visualize&Action=Submit&session_id=609c4179d1301e7a3e47f87c), 18 of which show increased _EZHIP_ expression.
Screen Shot 2021-05-12 at 9 08 17 PM

```
PT_QZN7FZER
PT_2CJ31SBX - mutated, not high expression
PT_3PK66TJY - mutated, not high expression
PT_3R0P995B
PT_BNARR0N6
PT_D3AN5M5W
PT_EB89G70N
PT_F9G665B1
PT_FHDZNN40
PT_K2YDNZDA
PT_N9KYV4WB
PT_NPETR8RY
PT_NYBH025T
PT_R6WWH1QX
PT_T4VKB29K
PT_V3Q78E6F
PT_WKNKNYHH
PT_ZZRBX5JT
PT_ZZWYKZZ8
PT_164RNWTT
```

When using z-score >=3 in pedcbio, we get 13 patients with higher expression of _EZHIP_:
Screen Shot 2021-05-12 at 9 12 03 PM
```
PT_QZN7FZER
PT_2CJ31SBX - mutated, not high expression
PT_3PK66TJY - mutated, not high expression
PT_3R0P995B
PT_BNARR0N6
PT_D3AN5M5W
PT_EB89G70N
PT_N9KYV4WB
PT_NPETR8RY
PT_R6WWH1QX
PT_T4VKB29K
PT_WKNKNYHH
PT_ZZRBX5JT
PT_ZZWYKZZ8
PT_164RNWTT
```

On the pedcbio site,  I see: 
>For mRNA and microRNA expression data, we typically compute the relative expression of an individual gene in a tumor sample to the gene's expression distribution in a reference population of samples. That reference population is all profiled samples (by default for mRNA), or normal samples (when specified), or all samples that are diploid for the gene in question (discontinued). The returned value indicates the number of standard deviations away from the mean of expression in the reference population (Z-score). The normalization method is described here. Please note that the expression results by querying a gene with the default setting (z-score threshold of 2) oftentimes are not meaningful. Since the z-scores were usually calculated compared to other tumor samples, high or low expression does not necessarily mean that the gene is expressed irregularly in tumors. The data is useful for correlation analysis, for example, pick a threshold based on overall expression (using Plots tab) and compare survival data between expression high and low groups.

I just sent a message to one of our developers to ask whether in this case, the reference population is `pbta_all` or if it would be `ependymomas`. I am hypothesizing that this is being done against the whole pbta cohort, which is not what we are doing and hence the differences.

jharenza:
An update on this: the z-score is calculated across the whole study/cohort, so in pedcbio, across all of PBTA. Given that we see differing results between this subtyping module and the pedcbio for samples coming out significant while using a higher z-score threshold here, I advocate we remove PFA subtyping from this module because we can't be confident in that assignment at this point in time.

jaclyn-taroni:
If the goal is to identify overexpression relative to other tumors of the same type, which I believe is the case, I think calculating the z-score within the EPN samples and picking z-score >= 3 would make me more confident in the calls, not less?

jharenza:
> If the goal is to identify overexpression relative to other tumors of the same type, which I believe is the case, I think calculating the z-score within the EPN samples and picking z-score >= 3 would make me more confident in the calls, not less?

It could be - I might want to look at the path reports to see if we have any positive controls we can use to assess if we try to keep this. Making a classifier may be the better way to do this to have the best confidence- I just don't want to mis-classify.

jharenza:
>It could be - I might want to look at the path reports to see if we have any positive controls we can use to assess if we try to keep this. 

@jaclyn-taroni I never responded to this. I did look at the path reports, but there is no classification in any of them. I think I can agree with you that using a more stringent cutoff of z-score >= 3 should be fine, so I think we can keep this in.

jharenza (approved):

PR #1065, opened 2021-05-11
Add atypical choroid plexus papilloma to harmonized diagnosis

jaclyn-taroni:
### Purpose/implementation Section

Adds a notebook for choroid plexus papilloma subtyping to `molecular-subtyping-pathology`

#### What GitHub issue does your pull request address?

Closes #997

#### Is there anything that you want to discuss further?

My read of #997 was if "atypical" was not present in `pathology_free_text_diagnosis`, `harmonized_diagnosis` should be set to `Choroid plexus papilloma` - is this correct?

#### Reproducibility Checklist

- [ ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

PR #1070, opened 2021-05-13
Remove irrelevant file that snuck in with #961

jaclyn-taroni:
Does what it says! This file snuck in with #961. It appears to capture the same information that was added in #1065 (albeit in a slightly different format). As far as I know, there were no accompanying code changes for this file.


jashapiro (approved):
👍🏼

Issue #1074, opened 2021-05-14
Updated analysis: LGAT integrated dx for ganglioglioma


jharenza:
#### What analysis module should be updated and why?
molecular-subtyping-lgat
Ganglioglioma diagnoses are not captured in integrated dx

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Recode `integrated_diagnosis` for samples with `pathology_diagnosis == "Ganglioglioma"` to become `Ganglioglioma, subtype` rather than `Low-grade glioma/astrocytoma, subtype` to be consistent with new pathology free text subtyping.

#### What input data should be used? Which data were used in the version being updated?
v19, histologies file

#### When do you expect the revised analysis will be completed?
tomorrow or monday

#### Who will complete the updated analysis?
@kgaonkar6

kgaonkar6:
Closing with #1078

PR #1078, opened 2021-05-17
add Ganglioglioma int dx

kgaonkar6:
Adding "Ganglioglioma" in integrated_diagnosis for LGG samples in `molecular-subtyping-pathology`.

#### What was your approach?
replace "LGG, <subtype>" for samples where pathology_diagnosi==Ganglioglioma samples to "Ganglioglioma, <subtype>

#1074 

I only updated the integrated_diagnosis in molecular-subtype-pathology do we want to update the molecular_subtype in moleculat-subtyping-LGAT as well? I only updated in molecular-subtype-pathology since the other pathology-free-text-diagnosis based updates are also in this module. 

I reran the bash script to update all the compiled subtype files and it seems an updated medullo subtyping is also adding to the changes.

Yes

table

95 Ganglioglioma samples now have integrated_diagnosis prefix "Ganglioglioma, <subtype>". 

#### Documentation Checklist


jharenza (approved):
Hi @kgaonkar6! Thanks for this PR. The logic looks correct and I think we can move forward with this. 

However, I noticed that there are many samples which have canonical _KIAA1549--BRAF_ fusions that got removed. Looking back, it looks like these are removed because they are glialneuronal tumors.  In talking with Cassie Kline, she mentioned we should keep them out of the LGG group, but it appears we should still subtype them. 

@jaclyn-taroni I propose going back to subtyping them in LGAT, but making their `molecular_subtype`, `GNT, subtype`, along with updating the `integrated_diagnosis` with the appropriate tumor type. This way, we don't duplicate the module for another tumor type. Thoughts? If that sounds OK, then I can make a ticket.

jharenza:
> > I propose going back to subtyping them in LGAT, but making their `molecular_subtype`, `GNT, subtype`, along with updating the `integrated_diagnosis` with the appropriate tumor type. This way, we don't duplicate the module for another tumor type. Thoughts? If that sounds OK, then I can make a ticket.
> 
> 
> 
> Are we only interested in the _KIAA1549--BRAF_ fusions or do we want to examine all the same molecular lesions?

All of the same lesions

jaclyn-taroni:
Okay then not duplicating makes sense to me!

Issue #1061, opened 2021-05-10
Proposed Analysis: Compile molecular-subtype+pathology and clinical feedback + pathology_free_text_diagnosis subtypes to a file


kgaonkar6:
#### What are the scientific goals of the analysis?
Combine pathology_free_text_diagnosis base subtyping with the existing compiled subtype results to be used as input for `molecular-subtype-integrate`. Tagging @jharenza @jaclyn-taroni for inputs to combined the pathology_free_text_diagnosis updates to `compiled_molecular_subtypes_with_clinical_pathology_feedback_and_report_info.tsv`

#### What methods do you plan to use to accomplish the scientific goals?

For v19 we already had created `analyses/molecular-subtyping-pathology/results/compiled_molecular_subtypes_with_clinical_pathology_feedback_and_report_info.tsv` to include pathology-free-text-diagnosis based meningioma  subtypes in input for `molecular-subtype-integrate`

To me looks like there are 2 types of updates

1. The following subtypes can be directly added to 'analyses/molecular-subtyping-pathology/results/compiled_molecular_subtypes_with_clinical_pathology_feedback_and_report_info.tsv` as well since these subtypes don't have molecular-subtyping results:

File | Logic to include as results |  Description ticket 
 -- | -- | -- 
analyses/molecular-subtyping-pathology/results/cns-lymphoma-subtypes.tsv | Append | #1057  
analyses/molecular-subtyping-pathology/results/juvenile-xanthogranuloma-subtypes.tsv | Append | #1056 
analyses/molecular-subtyping-pathology/results/choroid_plexus_papilloma_subtypes.tsv  | Append | #1065

2. The samples in the files below have molecular-subtyping results which are already part of the compile file `analyses/molecular-subtyping-pathology/results/compiled_molecular_subtypes_with_clinical_pathology_feedback_and_report_info.tsv` so we would have to incorporate some logic to merge these samples correctly. 

File | Logic to include as results | Previous subtype | Description ticket 
-- | -- | -- | -- 
analyses/molecular-subtyping-pathology/results/lgat-pathology-free-text-subtypes.tsv #1060  | Replace existing values | LGAT | #1000 
analyses/molecular-subtyping-pathology/results/cranio_adam_subtypes.tsv #823  | Replace existing values | CRANIO | #994
analyses/molecular-subtyping-pathology/results/glialneuronal_tumor_subtypes.tsv  | Replace existing values | LGAT | #1017 

[...]

PR #1060, opened 2021-05-10
Recode integrated-diagnosis for LGAT based on pathology-free-text-diagnosis

kgaonkar6:
#### What scientific question is your analysis addressing?
Using WHO 2016 CNS subtypes to improve LGAT harmonized diagnosis

#### What was your approach?
I followed the same notebook format as #1016 but I added a function `recode_integrated_diagnosis()` to be used to recode a integrated_diagnosis based on the pathology_free_text_diagnosis. The code will add the new integrated_diagnosis if NA or incorporate the new term with the molecular subtype if integrated_diagnosis was not NA.

To obtain the pathology_free_text_diagnosis term to use for subtyping I used the following table from [#1000](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1000) to recode the `harmonized_diagnosis` values for samples with a `pathology_diagnosis` of LGAT:

pathology_diagnosis | subtyping module | pathology_free_text_diagnosis | broad_histology | short_histology | integrated_diagnosis
-- | -- | -- | -- | -- | --
Low-grade glioma/astrocytoma (WHO grade I/II) | LGAT | contains "sega or subependymal giant cell astrocytoma" | Low grade astrocytic tumor | LGAT | Subependymal Giant Cell Astrocytoma, `molecular_subtype`
Low-grade glioma/astrocytoma (WHO grade I/II) | LGAT | contains "fibrillary" | Low grade astrocytic tumor | LGAT | Diffuse fibrillary astrocytoma, `molecular_subtype`
Low-grade glioma/astrocytoma (WHO grade I/II) | LGAT | gliomatosis cerebri, type 1, ia | Low grade astrocytic tumor | LGAT | Gliomatosis cerebri, `molecular_subtype`
Low-grade glioma/astrocytoma (WHO grade I/II) | LGAT | if contains "jpa" or "juvenile astrocytoma" or "pilocytic" or "pilocystic" (typo) or "pilomyxoid" but does not contain "fibrillary" | Low grade astrocytic tumor | LGAT | Pilocytic astrocytoma, `molecular_subtype`
Low-grade glioma/astrocytoma (WHO grade I/II) | LGAT | oligodendroglioma who ii | Diffuse astrocytic and oligodendroglial tumor | Oligodendroglioma | Oligodendroglioma, `molecular_subtype`
[...]

PR #1057, opened 2021-05-10
Add pathology-free-text-diagnosis based  Lymphoma subtypes

kgaonkar6:
#### What scientific question is your analysis addressing?
Using WHO 2016 CNS subtypes to improve CNS lymphoma harmonized diagnosis

#### What was your approach?
I updated @jaclyn-taroni code for meningioma code for this PR, my thought process was that this way the path-free-text-dx code updates are easier to read/review.
I used the following table from [#999](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/999) to recode the `harmonized_diagnosis` values for samples with a `pathology_diagnosis` of CNS lymphoma:

pathology_diagnosis | subtyping module | pathology_free_text_diagnosis | broad_histology | short_histology | harmonized_diagnosis
-- | -- | -- | -- | -- | --
Primary CNS lymphoma | NA | burkitt's lymphoma | Lymphoma | CNS lymphoma | CNS Burkitt's lymphoma

#### What GitHub issue does your pull request address?
#999 

#### Which areas should receive a particularly close look?
Eventually this code will be numbered and add to compiled subtype file to be used in molecular-subtype-integrate. Currently the output is in `analyses/molecular-subtyping-pathology/results/cns-lymphoma-subtypes.tsv`

#### What types of results are included (e.g., table, figure)?
table

#### What is your summary of the results?
2 samples with burkitt's lymphoma in pathology_free_text_diagnosis are now subtypes as "CNS Burkitt's lymphoma"

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1013, opened 2021-04-19
Recode meningioma harmonized diagnosis per #993 

jaclyn-taroni:
#### What scientific question is your analysis addressing?

Here I am recoding the `harmonized_diagnosis` field for samples with a pathology diagnosis of meningioma using string detection on the `pathology_free_text_field`. This is getting added to the `molecular-subtyping-pathology` module because it is subtyping on the basis of clinical information, rather than any kind of molecular data.

In the future, we will need to reorganize and document the `molecular-subtyping-pathology` module but that can wait until after all of our additions have gone in already.

#### What GitHub issue does your pull request address?

Closes #993 

#### Is there anything that you want to discuss further?

The table in #993 says that `harmonized_diagnosis` should be set to `Meningioma` when there is an absence of the specific strings we use for subtyping. I didn't explicitly code that into my `dplyr::case_when()` step, but rather assumed that setting `TRUE ~ "Meningioma"` would be sufficient. When I examined the results table, I thought this seemed fine.

#### What types of results are included (e.g., table, figure)?

Subtyping table

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1084, opened 2021-05-25
Compile pathology_free_text_diagnosis subtypes 

kgaonkar6:
Adding back the individual pathology_free_text_diagnosis subtypes to the `compiled_molecular_subtypes.tsv` 

#### What was your approach?
There are 2 kinds of updates:
The samples in the files below have molecular-subtyping results which are already part of the compile file `analyses/molecular-subtyping-pathology/results/compiled_molecular_subtypes.tsv` so we will be updating the values for these samples using outputs from the pathology-free-text-diagnosis terms.
File | Logic to include as results | Previous subtype | Description ticket 
-- | -- | -- | -- 
analyses/molecular-subtyping-pathology/results/lgat-pathology-free-text-subtypes.tsv #1060  | Update existing values | LGAT | #1000 
analyses/molecular-subtyping-pathology/results/cranio_adam_subtypes.tsv #823  | Update existing values | CRANIO | #994
analyses/molecular-subtyping-pathology/results/glialneuronal_tumor_subtypes.tsv  | Update existing values | LGAT | #996

The following subtypes can be directly added to `analyses/molecular-subtyping-pathology/results/compiled_molecular_subtypes.tsv` as well since these subtypes don't have molecular-subtyping results 
File | Logic to include as results |  Description ticket
  -- | -- | -- 
analyses/molecular-subtyping-pathology/results/cns-lymphoma-subtypes.tsv | Append | #1057  
analyses/molecular-subtyping-pathology/results/juvenile-xanthogranuloma-subtypes.tsv | Append | #1056
analyses/molecular-subtyping-pathology/results/choroid_plexus_papilloma_subtypes.tsv  | Append | #1065
analyses/molecular-subtyping-pathology/results/meningioma_subtypes.tsv | Append | #1013 

#1061 

- I added just 1 script here to compile the pathology_free_text_diagnosis here since even though there are 2 types of updates they are all from pathology_free_text_diagnosis. Let me know if you would prefer it to be divided into 2 scripts.
- 04-incorporate-pathology-feedback.Rmd needs some careful review
     -  like PT_00G007DM was coded to `ETMR, C19MC-altered` in  but is not needed anymore since there are no duplicate subtypes for this patient in v19
     - PT_EHE800JJ is coded in 04-incorporate-pathology-feedback.Rmd but is removed in v19

- There are 2 subtypes for this medulo BS_5BMNK8SY , it seems to be from matching to multiple RNAseq #1085 
- In addition the compiling and adding pathology_free_text_diagnosis subtypes there are other updates to `compiled_molecular_subtypes.tsv` from the hotspots updates / downstream molecular-subtype modules 

Yes

tables

#### Documentation Checklist


analyses/molecular-subtyping-pathology/02-compile-subtyping-pathology-free-text.Rmd
```diff
@@ -0,0 +1,154 @@
...
+  # append
+  bind_rows(choroid_plexus_papilloma_subtypes_df) %>%
+  select(-pathology_diagnosis,-pathology_free_text_diagnosis)
+
+```
+
+
+## Save to file
+```{r}
+compiled_subtypes_df %>% write_tsv(output_file)
```
jharenza, line 151: can you add an `arrange()` here for future change comparisons?

jharenza (changes requested):
I did a quick check on the final output file and saw some things that need to be updated - I haven't gone through the code to know from where this is derived, but a few things:
- Some `integrated_diagnosis` values do not have subtypes (they look to be `LGG, To be classified`) - Pilocytic astrocytoma, Diffuse fibrillary astrocytoma, Subependymal Giant Cell Astrocytoma, Pleomorphic xanthoastrocytoma. This column should always have subtypes and if no subtype, then there should be an `NA` value.
- Some `integrated_diagnosis` values are derived from `pathology_free_text_diagnosis`, but if these samples were not subtyped via module, then they should not be in this field, but rather `harmonized_diagnosis`. Eg: `Central Neurocytoma` - I recall we took care of that before, but maybe that was in molecular-subtype-integrate? Should we put that here?
- It seems like because we are adding `harmonized_diagnosis` due to `pathology_free_text_diagnosis` subtyping here, that we should also add the logic normally in molecular-subtype integrate which pushes `integrated_diagnosis --> harmonized_diagnosis` here. Thoughts on that?

kgaonkar6:
>It seems like because we are adding harmonized_diagnosis due to pathology_free_text_diagnosis subtyping here, that we should also add the logic normally in molecular-subtype integrate which pushes integrated_diagnosis --> harmonized_diagnosis here. Thoughts on that?

I think we can add the logic have integrated_diagnosis as harmonized_diagnosis in molecular_subtyping_pathology but I'll add a different PR.

kgaonkar6:
The following issues were fixed:
- Removed integrated_diagnosis terms in `LGG,To be classified` pathology_free_text_diagnosis updates should only have `harmonized_diagnosis`
- Glial and choroid plexus needed `Notes` to be added 
- Only add harmonized_diagnosis for subtypes from 
   - pathology_diagnosis updates only ( neurocytoma ) 
   - pathology_feedback only (chordoma )
- Duplicate for BS_5BMNK8SY fixed
- PT_EHE800JJ is removed from 04-incorporate-pathology-feedback.Rmd because it is removed as per v19

jharenza (changes requested):
Thanks for the updates @kgaonkar6!

I just see one thing needing an update:
The `meningioma_subtypes.tsv` file is not in the `compiled_subtypes_df` of `02-compile-subtyping-pathology-free-text.Rmd` and should be added.

Otherwise, I think the recode update in #1097 should probably go in first to get the ganglioglioma updates through.

jharenza:
Could you also pop #1108 into this PR since it is in the same module?

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/fad3359c16e7dbcff5169d08f854769c55f57821/analyses/molecular-subtyping-pathology/01-compile-subtyping-results.Rmd#L283-L285

should be 
```
  mutate(integrated_diagnosis = case_when(molecular_subtype == "CRANIO, ADAM" ~"Adamantinomatous craniopharyngioma",
                                          molecular_subtype == "CRANIO, PAP" ~"Papillary craniopharyngioma", 
                                          TRUE ~ NA_character_),
```

jharenza:
I also wanted to check that you plan to update `molecular-subtype-integrate` after this PR with updated `harmonized_diagnosis` since I do not see those filled in here... in which case, I think we may need a ticket for that one?

kgaonkar6:
In the latest commit, I've updated the typo Adamantinomatous craniopharyngioma and added back meningioma. 

About `molecular-subtype-integrate` didn't we decide that we will only run the module per data release so that the pbta-histologies.tsv file matches? Maybe I mis-understood, a ticket will be good.

jharenza:
That's possible, but I think the way it is set up now is via an excel sheet matching, right, and there needs to be some logic added for updating the harm dx based on free text dx, rather than just pulling over path dx, right?

kgaonkar6:
> That's possible, but I think the way it is set up now is via an excel sheet matching, right, and there needs to be some logic added for updating the harm dx based on free text dx, rather than just pulling over path dx, right?

Hmm I don't think the logic needs to be updated in the integrate step.

Any BS_ID already subtyped ( ie. exists in compiled_molecular_subtypes_with_clinical_pathology_feedback_and_report_info.tsv) is not going to be part of the [code](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/fad3359c16e7dbcff5169d08f854769c55f57821/analyses/molecular-subtyping-integrate/01-integrate-subtyping.Rmd#L66-L98) that pulls harm_dx/broad_hist/short_hist from the excel sheet. That part is only for the non-subtyped + pathology_diagnosis=="Other" BS_IDs. We could do have `molecular-subtyping-integrate` rerun PR to see if that's true or I could have missed some key logic.

jharenza:
gotcha! that makes sense.

jharenza (approved):
Thank you for the updates @kgaonkar6 - I think this looks good now!

Issue #1094, opened 2021-05-27
Updated analysis: Add fusion as TP53 loss


kgaonkar6:
#### What analysis module should be updated and why?
tp53_nf1_score module update 04-tp53-sv-loss.Rmd to include fusion breakpoints. @jharenza  noticed that we had missed out adding fusion in TP53 that removes most of the gene and functional domains in BS_NJ4WPQVK as a TP53 loss. 
[annofuse_TP53_bk_BS_NJ4WPQVK](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6554209/annofuse_bpleft.pdf)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
 - Add a code chunk to read in fusion calls and generate granges
```
putative_onco_tp53_fusion_gr <- putative_onco_fusion %>%
  filter(str_detect(FusionName,"^TP53--|--TP53$")) %>%
  left_join(histology,
            by = c("Sample" = "Kids_First_Biospecimen_ID")
  ) %>%
  dplyr::select("Sample",
                "sample_id",
                "FusionName",
                "RightBreakpoint",
                "LeftBreakpoint") %>%
  reshape2::melt(c("Sample","sample_id","FusionName")) %>%
  mutate(
    BreakpointChr = paste0("chr",gsub(":.*$", "", value)),
    Breakpoint = gsub("^.*:", "", value)) %>%
  GenomicRanges::makeGRangesFromDataFrame(seqnames.field = "BreakpointChr",
                                          start.field = "Breakpoint",
                                          end.field = "Breakpoint",
                                          
                                          keep.extra.columns = TRUE,
                                          starts.in.df.are.0based = FALSE)
```

 - And then check if fusion breakpoint is in exon 1 (we can get the exon 1 information from the gtf) or intron 1 (introns are not ranked so will need a region) in the TP53 as 5` gene that will result in TP53 loss OR should I just check for TP53 functional domain loss? 

cc @jharenza for a review of the conditions of calling TP53 loss with a fusion ( this is similar to CNV and SV loss annotation) 
-  if sample_id has SNV+(CNV|SV|Fusion) mutation suggesting both alleles are mutated
-  if CNV|SV mutant|Fusion  and has Li-Fraumeni syndrome suggesting germline TP53 mutant as cancer predisposition
[...]

PR #1073, opened 2021-05-13
#959 Tp53 classifier rerun for subtyping

kgaonkar6:
#### What scientific question is your analysis addressing?
Update tp53-nf1-score with hotspots maf + consensus maf  and latest consensus CNV file.

#### What was your approach?
 - added bind(consensus_maf),hotspot_maf) %>% unique() https://github.com/kgaonkar6/OpenPBTA-analysis/blob/73499652d84eeec13ec85840ba350c66f0c658df/analyses/tp53_nf1_score/05-tp53-altered-annotation.Rmd#L136-L178

- add code in bash script to gather most up-to-date consensus seg file with status 
Rscript -e "rmarkdown::render('../focal-cn-file-preparation/02-add-ploidy-consensus.Rmd', clean = TRUE)"
cp ${scratch_dir}/consensus_seg_with_status.tsv ${analysis_dir}/input/consensus_seg_with_status.tsv

#### What GitHub issue does your pull request address?
#1072

#### Which areas should receive a particularly close look?

Should we update the TP53 CNV loss filtering process?

We had previously  (cnvkit copy_number as default) only retained TP53 CNV losses with 1 **or** 0 copies because we had a lot of copy_number ==2 as loss (from neutral calls being assigned 2 copy).

After the update to use Freec as default for copy_number we see some changes where the copy_number has mostly changed to 2 for samples with >=3 ploidy which is a loss compared to the ploidy but we are missing them out because of our filter to use 1 **or** 0 copy loss calls only.
Here's a snippet of the TP53 loss calls that are missed : 

```
> overlaps[overlaps$Kids_First_Biospecimen_ID %in% c("BS_2J4FG4HV","BS_5JC116NM","BS_7M7JNG00","BS_823V5X6Z","BS_ZV21J6YW"),] 
DataFrame with 15 rows and 13 columns
                   cnv_gr Kids_First_Biospecimen_ID copy_number tumor_ploidy      status               domain_gr hgnc_symbol     pfam_id gene_start  gene_end         NAME
                <GRanges>               <character>   <numeric>    <numeric> <character>               <GRanges> <character> <character>  <integer> <integer>  <character>
1   chr17:114696-21459816               BS_2J4FG4HV           2            3        loss chr17:7670637-7673573:-        TP53     PF07710    7661779   7687550 P53_tetramer
2   chr17:114696-21459816               BS_2J4FG4HV           2            3        loss chr17:7673755-7676387:-        TP53     PF00870    7661779   7687550          P53
[...]

PR #1095, opened 2021-05-27
Add fusion within TP53 loss annotation 

kgaonkar6:
⚠️ merge #1073 before this PR

Add fusions in TP53 as a gene loss annotation.

#### What was your approach?
 - Added a code chunk to read in fusion calls and generate granges
```
putative_onco_tp53_fusion_gr <- putative_onco_fusion %>%
  filter(str_detect(FusionName,"^TP53--|--TP53$")) %>%
  left_join(histology,
            by = c("Sample" = "Kids_First_Biospecimen_ID")
  ) %>%
  dplyr::select("Sample",
                "sample_id",
                "FusionName",
                "RightBreakpoint",
                "LeftBreakpoint") %>%
  reshape2::melt(c("Sample","sample_id","FusionName")) %>%
  mutate(
    BreakpointChr = paste0("chr",gsub(":.*$", "", value)),
    Breakpoint = gsub("^.*:", "", value)) %>%
  GenomicRanges::makeGRangesFromDataFrame(seqnames.field = "BreakpointChr",
                                          start.field = "Breakpoint",
                                          end.field = "Breakpoint",
                                          
                                          keep.extra.columns = TRUE,
                                          starts.in.df.are.0based = FALSE)
```

 - And then checked if fusion breakpoint is in exon 1 which would result in loss of functional domain of TP53 gene

#1094 

I checked for intron overlaps but I didn't see any ranking on introns so can't figure out a filter for intron1  so I was wondering if it would be easier to just check for domain loss of TP53?

In `05-tp53-altered-annotation.Rmd` we aggregate the calls which needs a logic to add the fusion update as a loss , I added fusion to the CNV|SV logic we have already
-  if sample_id has SNV+(CNV|SV|Fusion) mutation suggesting both alleles are mutated
-  if CNV|SV mutant|Fusion  and has Li-Fraumeni syndrome suggesting germline TP53 mutant as cancer predisposition
-  if tp53 inactivating score for RNA_Seq is greater that 0.5 and has 1 or more SNV| (CNV|SV|Fusion) loss in TP53

Yes

table

7316-349 ( DNA:BS_YETTZ1NC, RNA:BS_NJ4WPQVK) is now annotated as TP53 loss since the classifier score is greater than 0.5 and also has a fusion.

#### Documentation Checklist


jharenza (approved):
Thanks for adding this @kgaonkar6 ! As we were talking about intron 1, I got that from the canonical _TP53_ fusions, especially in osteosarcoma, which result in non-functional TP53. It does look like, from a quick lit search, that these have been found in other [metastatic cancers](https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0218618), consistent with this finding. Theoretically, we might be able to extend the breakpoint region between the start site and prior to the first domain, but I think sticking to intron 1 for now works.

kgaonkar6:
I wasn't able to find any ranking information for introns so could not select for intron1 specifically (if we select a canonical TP53 transcript we could select by start and stop) and more importantly the fusion  TP53--PSMG4 overlaps the 1st exon in BS_NJ4WPQVK. Does that sound ok ? 
It would be easiest to just check if breakpoint is anywhere between the start and first domain, should I implement that?

It seems we are missing BS_NJ4WPQVK with the TP53--PSMG4 fusion which is failing the CI checks, should I update the test files ?

jharenza:
No, I think your parameters now are fine- just noting.

@jaclyn-taroni is it worth updating CI files for this one sample?

jaclyn-taroni:
I'll let you all decide if you'd like to file a PR to `create-subset-files` to include that sample, but some considerations:

* There is a precedent for such things (see: https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/c0744576ee80914636443b01012168efc2da22a9/analyses/create-subset-files#special-considerations)
* At this stage in the project, I do not anticipate updating the CI files in this matter to break things but it is a possibility. 
* What would we not be testing if we didn't add that sample and how critical is it?

jharenza:
I don't think it's critical if you are ok merging without the pass since it's just one sample.

jaclyn-taroni:
If this goes in and does not pass, that means all PRs will have to be merged with it failing which I don't want to do because it 1. means modules that come after it will not be tested 2. everything needs to be merged with administrative power. 

We can skip this step in CI if it's not critical (and this may be what you meant above!) or what might be a better/more robust approach would be to add logic to the code to handle if the sample is missing.

jharenza:
Ah, i see- let's add logic to the code to handle if a sample is missing @kgaonkar6

kgaonkar6:
Sounds good, I'll update the code. Thanks for the input @jaclyn-taroni @jharenza !

Issue #1085, opened 2021-05-25
Updated analysis: BS_5BMNK8SY assigned 2 MB subtypes


kgaonkar6:
#### What analysis module should be updated and why?
[molecular_subtyping_MB](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-MB)
BS_5BMNK8SY has 2 subtypes from matching with 2 different RNA-Seq with sample_id==7316-14

Kids_First_Biospecimen_ID | sample_id | aliquot_id | Kids_First_Participant_ID | cohort_participant_id | previous_parent_aliquot_id | parent_aliquot_id | molecular_subtype | pathology_diagnosis | pathology_free_text_diagnosis | integrated_diagnosis | harmonized_diagnosis
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
BS_5BMNK8SY | 7316-14 | 654436 | PT_M0AP3Z15 | C19803 | 7316-14-T-8725.WGS | 7316-14-T-8725.WGS | MB, To be classified | Medulloblastoma | medulloblastoma | NA | Medulloblastoma
BS_5BMNK8SY | 7316-14 | 654436 | PT_M0AP3Z15 | C19803 | 7316-14-T-8725.WGS | 7316-14-T-8725.WGS | MB, Group4 | Medulloblastoma | medulloblastoma | Medulloblastoma, group 4 | Medulloblastoma, group 4
BS_QEYPFYHD | 7316-14 | 654201 | PT_M0AP3Z15 | C19803 | 7316-14-T-8725.RNA-Seq | 7316-14-T-8725.RNA-Seq | MB, To be classified | Medulloblastoma | medulloblastoma | NA | Medulloblastoma
BS_Z7PKVY9J | 7316-14 | 1030649 | PT_M0AP3Z15 | C19803 | 7316-14-T-8711.RNA-Seq | 7316-14-T-8711.RNA-Seq | MB, Group4 | Medulloblastoma | medulloblastoma | Medulloblastoma, group 4 | Medulloblastoma, group 4
BS_ZH666SJW | 7316-14 | 612624 | PT_M0AP3Z15 | C19803 | 7316-14-N-8710.WGS | 7316-14-N-8710.WGS | NA | NA | NA | NA | NA

#### What changes need to be made? Please provide enough detail for another participant to make the update.
[...]

PR #1096, opened 2021-06-02
fix DNA-RNA mapping using parent_aliquot_id

komalsrathi:

Multiple MB subtypes assigned to `BS_5BMNK8SY`

#### What was your approach?

Use `parent_aliquot_id` to map DNA to RNA. This will map `BS_5BMNK8SY` to just `BS_QEYPFYHD`. The other RNA `BS_Z7PKVY9J` will be mapped to `NA`.

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1085

Output tsv files

N/A

Yes

#### Documentation Checklist


komalsrathi:
@kgaonkar6 can you check if this makes sense now? This should not affect any results but only the mapping of DNA samples to RNA samples in the output tsv file.

kgaonkar6 (approved):
This looks right to me, thanks!

jharenza (commented):
Hi @komalsrathi - thanks for adding this. @kgaonkar6 I don't think that it is this simple, but checking with @baileyckelly now. Some samples have matching IDs, and some don't, but the format is not the same across the board and I don't think we can use it globally as a match. TBD.

jharenza (changes requested):
ok after talking with @allisonheath and @baileyckelly, it seems like this field is going to be changing a bit, in the past has not been fully completed and DNA/RNA ids don't always align. For now, I would suggest for only samples which have two RNA-Seq samples for the same event that we assess the classifications. I think we can use the parental ID matching for this sample _only_, but also use a 3/4 consensus here, which is 3 for Group 4 and 1 for SHH. MM2S had a discrepancy between the two, which should not happen for the same tumor and time point, so we could flag this mismatch as being wrong. I probably prefer the latter since these samples are from the same event and should be annotated the same.

komalsrathi:
For obtaining the consensus subtype, we can have the 3/4 cutoff for all duplicated samples (not sure if there are others too) or do you want me to hardcode for this sample only? Also, I think I will have to revert back to what we had before i.e. only use KF Participant ID and sample_id to map DNA and RNA because if I use the parental aliquot ID, `BS_Z7PKVY9J` (RNA) sample won't get mapped to `BS_5BMNK8SY` (DNA) and then we cannot apply the 3/4 rule. 

Is that all or am I missing anything?

jharenza:
> For obtaining the consensus subtype, we can have the 3/4 cutoff for all duplicated samples (not sure if there are others too) or do you want me to hardcode for this sample only? Also, I think I will have to revert back to what we had before i.e. only use KF Participant ID and sample_id to map DNA and RNA because if I use the parental aliquot ID, `BS_Z7PKVY9J` (RNA) sample won't get mapped to `BS_5BMNK8SY` (DNA) and then we cannot apply the 3/4 rule.
> 
> Is that all or am I missing anything?

That's right - for 3/4, it would only be where samples for the same patient have the same `sample_id` and thus same `tumor_descriptor`. In addition, it sounds like the `parental_aliquot_id` will be changing in future releases, so using that will be unstable.

komalsrathi:
@kgaonkar6 okay, I think this should fix the issue.

kgaonkar6:
Perfect! This is ready to merge

jaclyn-taroni:
We need a re-review from @jharenza because there are changes requested (or for me to dismiss the previous review) before this can be merged.

jharenza (approved):
This looks good, thank you @komalsrathi

Issue #1113, opened 2021-07-20
Updated analysis: Subset manta FILTER=="PASS" consensus cnv


kgaonkar6:
#### What analysis module should be updated and why?
Add a FILTER==PASS for manta calls used in `copy_number_consensus_call`. I think this filter is necessary so that we use a subset of high-confidence broad SVs called by manta. https://github.com/Illumina/manta/blob/master/docs/userGuide/README.md#vcf-format-fields

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Add a filter to only keep calls where FILTER=='PASS' 
`$11 == 'PASS'`
[here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d70158c8945e6cdbd96bc3e68e5a114c394c79ff/analyses/copy_number_consensus_call/Snakefile#L101-L106)

#### What input data should be used? Which data were used in the version being updated?
pbta-sv-manta.tsv.gz

#### When do you expect the revised analysis will be completed?
1day

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
closed with #1114 #1115 #1116

Issue #1099, opened 2021-06-17
Updated analysis: CNV deep deletions


kgaonkar6:
#### What analysis module should be updated and why?
update focal-cn-file-preparation
and
rerun ocnoprint-landscape

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Change:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6bf301eea42a7f8812c3b14cee35a961097f8ecf/analyses/focal-cn-file-preparation/04-prepare-cn-file.R#L240-L247

To:
```

# Merge and annotated no X&Y
autosome_annotated_cn <- process_annotate_overlaps(cnv_df = cnv_no_xy,
                                                   txdb_exons = tx_exons) %>%
  # mark possible amplifications and deep loss in autosomes
  dplyr::mutate(status = dplyr::case_when(
    copy_number > (2 * ploidy) ~ "amplification",
    copy_number == 0  ~ "deep loss",
    TRUE ~ status
  ))

```
Change:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6bf301eea42a7f8812c3b14cee35a961097f8ecf/analyses/focal-cn-file-preparation/04-prepare-cn-file.R#L265-L267

To:
```
  sex_chrom_annotated_cn <- process_annotate_overlaps(cnv_df = cnv_sex_chrom,
                                                      txdb_exons = tx_exons) %>%
  # mark possible deep loss in sex chromosome
  dplyr::mutate(status = case_when(
    copy_number == 0  ~ "deep loss",
   TRUE ~ status
  ))
```

#### What input data should be used? Which data were used in the version being updated?
pbta-cnv-consensus.seg.gz v19

#### When do you expect the revised analysis will be completed?
~1 day

#### Who will complete the updated analysis?
@kgaonkar6 

cc @jharenza @jaclyn-taroni  for input

jharenza:
closed with #1101

PR #1088, opened 2021-05-26
Consensus seg update for oncoplot

kgaonkar6:
#### What scientific question is your analysis addressing?
We are adding CNV calls from consensus seg files for autosomes , X and Y chromosomes as inputs for oncoplots. 

#### What was your approach?
Update the input files to : 
```
consensus_seg_autosomes_cnv_file=${focal_directory}/consensus_seg_annotated_cn_autosomes.tsv.gz
consensus_seg_cnv_xy_cnv_file=${focal_directory}/consensus_seg_annotated_cn_x_and_y.tsv.gz
```

#### What GitHub issue does your pull request address?
#1087

#### Which areas should receive a particularly close look?
The figures need careful review , the only update in consensus seg files and we are currently coding loss as "Del" and amplification and gain as "Amp" to match the maftools requirements as in #1009 

#### Is there anything that you want to discuss further?
Na

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
figures

#### What is your summary of the results?
ATRX calls are now added with use of X and Y chromosome calls and all CNV overlapping gene locations are being plotted. 

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1116, opened 2021-07-21
Update manta FILTER=='PASS' Part3 : oncoprint rerun 

kgaonkar6:
Rerun with FILTER==PASS for manta calls used in `copy_number_consensus_call`

#### What was your approach?
Rerun

Add deep deletion status in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1099
Update with manta FILTER==PASS https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1113

I have`deep deletion` and `amplification` in the oncoprint.
There seems to have been an issue with Varian_Classification [update](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1088/commits/e864fe290ca50dedb6cbb5787b292c96a10a727d) not going into master from #1088 so I've added back in this PR.

Yes

figures

Previous primary_only_hgat_goi_oncoprint.png with status=="loss" as Del and "amplification" as Amp
image

Current file primary_only_hgat_goi_oncoprint.png with status=="loss" as Del and "amplification" as Amp along with manta `FILTER==PASS`
image

#### Documentation Checklist


jharenza:
Thanks @kgaonkar6 - can you show the GOI plots? I don't think the top N gene plots are as useful here.

Also, I will say no to a stepwise review - we shouldn't call any non-pass variants.

kgaonkar6:
Is deep deletion filter ok? I reverted back to use "loss" as Del in the last commit to make it comparable to master

jharenza (approved):
This 100% looks much better. I checked the LGAT plots as well and those unexpected amplifications across the fusion genes are also gone (which they should be).

For plotting `Del` - we tend to use deep deletions in these plots, so that is what we should do here and can specify in the manuscript.

jaclyn-taroni:
> For plotting `Del` - we tend to use deep deletions in these plots, so that is what we should do here and can specify in the manuscript.

My interpretation of this comment is that `analyses/oncoprint-landscape/01-map-to-sample_id.R` needs to be revised before this goes in - is that correct?

jharenza:
> My interpretation of this comment is that `analyses/oncoprint-landscape/01-map-to-sample_id.R` needs to be revised before this goes in - is that correct?

yes, I misunderstood some offline discussion with @kgaonkar6 thinking this current version already had that update.

jaclyn-taroni (changes requested):
Okay no problem, I am going to hit this with request changes so I don't accidentally merge this before that happens!

jaclyn-taroni (approved):

kgaonkar6:
Sorry about the confusion! 

In the last commit status `deep deletion` is labeled as `Del`.

@jharenza also pointed out TTYH1 amplifications that needed to be verified. I've spot checked the TTYH1 in the 2 EMTR samples and they are supported by both cnvkit and freec so we consider these as high confidence calls :

```
# A tibble: 2 x 4
  biospecimen_id gene_symbol copy_number_cnvkit status       
  <chr>          <chr>                    <dbl> <chr>        
1 BS_TE8QFF7T    TTYH1                        7 amplification
2 BS_69VS8PS1    TTYH1                        7 amplification
> check_freec %>%  filter(gene_symbol=="TTYH1",status=="amplification") 
# A tibble: 2 x 4
  biospecimen_id gene_symbol copy_number_freec status       
  <chr>          <chr>                   <dbl> <chr>        
1 BS_K07KNTFY    TTYH1                      12 amplification
2 BS_TE8QFF7T    TTYH1                       8 amplification
```

Issue #1110, opened 2021-07-01
Updated analysis: molecular-subtyping-embryonal sample selection


kgaonkar6:
#### What analysis module should be updated and why?
`molecular-subtyping-embryonal`

It seems because the path_dx is slightly updated in v19 to Embryonal Tumor with Multilayered Rosettes (ETMR);Other but the we filter by exact matching to terms here https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-[…]embryonal/subset-files/embryonal_subtyping_path_dx_strings.json

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Use str_detect() to gather the terms used in json file https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-[…]embryonal/subset-files/embryonal_subtyping_path_dx_strings.json

#### What input data should be used? Which data were used in the version being updated?
`pbta-histologies.tsv`

#### When do you expect the revised analysis will be completed?
1 day

#### Who will complete the updated analysis?
@kgaonkar6

jharenza:
closed with #1117

PR #1117, opened 2021-07-22
Adding str_detect to fix issue with EMTR

kgaonkar6:
Updating the string matching logic because the path_dx was updated in v19 to use term "Embryonal Tumor with Multilayered Rosettes (ETMR);Other" but the we filter by exact matching to terms from [json](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-embryonal/subset-files/embryonal_subtyping_path_dx_strings.json)

#### What was your approach?
Adding str_detect() instead of exact term matching 

#1110 

There seem to be a lot of changes but it looks like only decimal changes in zscore values for "LIN28A", "FOXR2" because of the addition of the RNA-Seq samples associated with sample_id 7316-2577

NA

Yes

table

Adds `7316-2577` to embryonal subtyping results which has path_dx == "Embryonal Tumor with Multilayered Rosettes (ETMR);Other"

#### Documentation Checklist


jaclyn-taroni (approved):
LGTM

jharenza (approved):
This looks good to me, but concerning is that `7316-2577` has a diagnosis of ETMR but does not have the hallmark C19MC amplification or TTYH1 fusions, so this is one I may have to check with pathology about. However, `Embryonal tumor, NOS` will be the designation for this release.

PR #1123, opened 2021-08-04
V20 CNV update part1: Update overlap criteria for consensus CNV

kgaonkar6:
We believe criteria for cnvs to have 50% reciprocal overlap might be too stringent. After the #1116 we are missing some subtype defining CNVs. 

After investigation we see the CNV alterations n controlfreec and cnkit . For example chr19 amplification in BS_K07KNTFY 
is seen in both controlfreec and cnvkit but missed out of consensus calls because  cnvkit region is 11% of controlfreec region

```
BS_K07KNTFY.cnvkit.dup.filtered3.bed: chr19	54138551	54427104
BS_K07KNTFY.freec.dup.filtered3.bed:  chr19	53641020	56141391
```

#### What was your approach?
Our approach was to broaden the criteria to include CNV calls in either caller that has any overlap at this step 
master
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2511e8c9fc4a7542f5b709363f866ccddb73be8b/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L148-L152

consensus-cnv-smallCNV-overlap
```
                        ## For list2's CNV
                        ## If any overlap exists,
                        ## then we add in the start, end coordinate, total overlap length, and total len to different lists
                        ## This is done to account for 1 CNV from list1 overlapping with MULTIPLE CNVs from list2
                      if (end - start +1) / (end_list2 - start_list2 + 1) >= 0:
```

And at the following snippet we allow CNV overlaps that completely overlap a smaller CNV in caller X by a larger CNV in callerY
master
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2511e8c9fc4a7542f5b709363f866ccddb73be8b/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L181

consensus-cnv-smallCNV-overlap
```
if (coverage_list1 >= 0.5 and coverage_list2 >= 0.5) or (coverage_list1 >=0.9 and coverage_list2 >0 ) or (coverage_list1>0 and coverage_list2 >=0.9):
```

#1125 1125

Does the criteria for consensus region with less that 50% overlap look right? 

Yes

tables

We now include smaller overlapping regions in CNV which will help identify the focal CNV.

```{r}
# latest with (coverage_list1 >= 0.5 and coverage_list2 >= 0.5) or (coverage_list1 >=0.9 and coverage_list2 >0 ) or (coverage_list1>0 and coverage_list2 >=0.9)
latest <- read_tsv("~/Documents/OpenPBTA-analysis/analyses/copy_number_consensus_call/results/pbta-cnv-consensus.seg.gz")

# previous with 50% reciprocal overlap
prev <- read_tsv("~/Documents/OpenPBTA-analysis/data/pbta-cnv-consensus.seg.gz")

## Overall changes
change_log_in_latest <-setdiff(latest, prev) 
```

```
## Difference in CNV calls latest consensus cnv seg
> change_log_in_latest$copy.num %>% table()
.
   0    1    2    3    4    5    6    7    8   12   20 
  71 1177  342 1852  425  313   92   23    9    3    1 

```

```
> change_log_in_latest %>% mutate(size = abs(loc.start-loc.end)) %>% select(size) %>% summary() 
      size          
 Min.   :        1  
 1st Qu.:    75456  
 Median :   681020  
 Mean   :  6754822  
 3rd Qu.:  5921025  
 Max.   :158482914  
```

#### Documentation Checklist


jharenza (approved):
These changes are minimal and kept stringent, but capture important known CN alterations which were lost with the 50% reciprocal criteria requirement. In examining the difference file, almost all genes are cancer-relevant and important to capture.

jaclyn-taroni:
Hi @kgaonkar6 and @jharenza, thanks for submitting this! Because of the nature of this change and how others may be tasked with writing about it or explaining it, I think we need to make sure the documentation is easy to understand and that extends to the method section of the manuscript, too. Apologies in advance if I misinterpreted something, but all the more reason to make sure the docs are very clear!

- There are still comments that are out of date in `compare_variant_calling_updated.py`:

  - https://github.com/kgaonkar6/OpenPBTA-analysis/blob/10477be2aa23453062d616d22f8b3c99146182d9/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L13
  - https://github.com/kgaonkar6/OpenPBTA-analysis/blob/10477be2aa23453062d616d22f8b3c99146182d9/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L184


- I think when we use the terminology `coverage` in the script, we mean proportion of CNV call that overlaps with calls from the other caller. With that in mind – I'm finding the terminology "complete overlap" to be confusing. What I think we're doing is retaining calls where 90% or more of a CNV from the first caller (CNV1) overlaps with a call in the other caller (CNV2), provided _any_ of CNV2 overlaps with CNV1, and vice versa. I would not call that complete overlap, I would call that 90% overlap that is not required to be reciprocal. Am I missing something and/or did I follow incorrectly?
  
  I also don't think the comments/docs in these two places would necessarily be interpreted mean the same thing, as only one of these mentions the "any overlap" part:

  - https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1123/files#diff-c09bef42ba1c6a1f2a7e2addd4590cd09c31e04bd8e941ca0493d0561a036342R63
  - https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1123/files#diff-cf0dccbd1dd23f21864b536db7c5142ea6648038c4f4d9fbac0ac7f4f2936b2cR181-R182

- In the interest of writing methods, explaining this to others, etc., can you explain why we don't need to change anything that comes _after_ the change in logic at line 183?

- The CNV consensus calling methods need to get updated in the manuscript before it gets merged in my opinion or at the very least there needs to be an issue tracking that update! We don't have the new Manta filtering in there yet either.

Thank you!

kgaonkar6:
Thank for the review, I'll update the comments accordingly. 

> * There are still comments that are out of date in `compare_variant_calling_updated.py`:
>   
>   * https://github.com/kgaonkar6/OpenPBTA-analysis/blob/10477be2aa23453062d616d22f8b3c99146182d9/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L13
>   * https://github.com/kgaonkar6/OpenPBTA-analysis/blob/10477be2aa23453062d616d22f8b3c99146182d9/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L184

I've updated the comments here to incorporate the new criteria


> * I think when we use the terminology `coverage` in the script, we mean proportion of CNV call that overlaps with calls from 
the other caller. With that in mind – I'm finding the terminology "complete overlap" to be confusing. What I think we're doing is retaining calls where 90% or more of a CNV from the first caller (CNV1) overlaps with a call in the other caller (CNV2), provided _any_ of CNV2 overlaps with CNV1, and vice versa. I would not call that complete overlap, I would call that 90% overlap that is not required to be reciprocal. Am I missing something and/or did I follow incorrectly?

Ah yes, it is not a "complete overlap" I can update that terminology. You are right, I have update the comments to correctly specify that we will be using _any_ CNV in a caller overlap 90% or more of CNV in another caller. 


>   I also don't think the comments/docs in these two places would necessarily be interpreted mean the same thing, as only one of these mentions the "any overlap" part:
>   
>   * https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1123/files#diff-c09bef42ba1c6a1f2a7e2addd4590cd09c31e04bd8e941ca0493d0561a036342R63
>   * https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1123/files#diff-cf0dccbd1dd23f21864b536db7c5142ea6648038c4f4d9fbac0ac7f4f2936b2cR181-R182

I reworded the README to match the comments in the code to specify that we will retain any CNV that overlaps 90% or more of a CNV in another caller in the consensus CNV file.

> * In the interest of writing methods, explaining this to others, etc., can you explain why we don't need to change anything that comes _after_ the change in logic at line 183?

The logic after line 183 gather the consensus CNV region as:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/187df0fcf28117b869be534159eda536fa7ce07e/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L187-L189
which satisfies the expected such that the region in consensus CNV overlaps all CNVs in list2 that satisfy the criteria of 50% reciprocal or any CNV in 1 caller overlapping 90% of CNV in another caller. 

For example chr19 amplification in BS_K07KNTFY is seen in both controlfreec and cnvkit 
```
BS_K07KNTFY.cnvkit.dup.filtered3.bed: chr19	54138551	54427104
BS_K07KNTFY.freec.dup.filtered3.bed:  chr19	53641020	56141391
```
So the consensus CNV we want to see  would be `chr19       53641020	56141391`


> * The CNV consensus calling methods need to get updated in the manuscript before it gets merged in my opinion or at the very least there needs to be an issue tracking that update! We don't have the new Manta filtering in there yet either.
Yes, I'll add in a ticket! Thanks for the reminder

jaclyn-taroni:
> The logic after line 183 gather the consensus CNV region as:
> 
> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/187df0fcf28117b869be534159eda536fa7ce07e/analyses/copy_number_consensus_call/scripts/compare_variant_calling_updated.py#L187-L189
> 
> which satisfies the expected such that the region in consensus CNV overlaps all CNVs in list2 that satisfy the criteria of 50% reciprocal or any CNV in 1 caller overlapping 90% of CNV in another caller. 
> 
> 
> 
> For example chr19 amplification in BS_K07KNTFY is seen in both controlfreec and cnvkit 
> 
> ```
> 
> BS_K07KNTFY.cnvkit.dup.filtered3.bed: chr19	54138551	54427104
> 
> BS_K07KNTFY.freec.dup.filtered3.bed:  chr19	53641020	56141391
> 
> ```
> 
> So the consensus CNV we want to see  would be `chr19       53641020	56141391`
> 

Wouldn't the more conservative approach, which I'm going to tend to favor generally speaking, be to take the smaller alteration in that case? I.e., the range for the CNA that has 90% overlap with another call which we might expect to be smaller

kgaonkar6:
>Wouldn't the more conservative approach, which I'm going to tend to favor generally speaking, be to take the smaller alteration in that case? I.e., the range for the CNA that has 90% overlap with another call which we might expect to be smaller 

Thanks for pointing this out! I can update the code to keep the smaller region that is overlapped 90% or more within the larger CNV region. Sounds like a more justifiable code to keep just the focal CNVs.

jaclyn-taroni:
@jashapiro and I chatted about this briefly this afternoon and he thinks that effectively you may end up with the smaller region _anyway_ because some later step only takes the union of the regions. He suggested taking a look at some number of cases (like the one for BS_K07KNTFY & chr 19 amplification) to see if we consistently get the smaller CNV in the consensus.

kgaonkar6:
This is the output , which covers the smaller start site and smaller of the end sites
```r
# A tibble: 2 x 7
  ID          chrom loc.start  loc.end num.mark seg.mean copy.num
  <chr>       <chr>     <dbl>    <dbl> <lgl>       <dbl>    <dbl>
1 BS_K07KNTFY chr19  53641206 54049403 NA           1.27       12
2 BS_K07KNTFY chr19  54138551 54427104 NA           1.36       12
```

jaclyn-taroni:
Okay I think we can get this merged, and then:

1. Update the v20 consensus SEG file in the release (e.g., on S3), if not already done
2. Merge the fix in #1130
3. Start on fixing conflicts / rerunning / merging the other CNV PRs (#1124, #1126, #1127) as needed

kgaonkar6:
v20 s3 bucket was updated when this PR was created so we have the consensus seg file update.

b9284650be04df3538e6c6dba29b8eb0  pbta-cnv-consensus.seg.gz

PR #1124, opened 2021-08-04
V20 CNV update part2: focal cnv and oncoprint rerun

kgaonkar6:
#### What scientific question is your analysis addressing?
Just a rerun for focal-cn and oncoprint after #1123 

#### What was your approach?
I added the conditions to use base histology in `focal-cn-preparation` module to run for subtyping when needed.

#### What GitHub issue does your pull request address?
#1125

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
NA

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What types of results are included (e.g., table, figure)?
tables

#### What is your summary of the results?
Rerun of focal calls and oncoprint from updated overlap #1123 

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1130, opened 2021-08-09
Update focal-cn-preparation to use consensus SEG file in data download

### LINKED ISSUE #1129: Focal CN preparation takes over an hour to run in CI
#### Context

Right now, the focal CN preparation step in CI takes over an hour: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5780/workflows/9c9f689d-5a14-4bfd-8fd8-8b8259f528ef/jobs/6483

This seems to me like a _somewhat_ recent development based on my memory alone, so I randomly picked this run from earlier in the year and it used to take 10 minutes: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/4994/workflows/ad73ca10-0742-4d59-af13-12cf4d890ecb/jobs/5575 Also true in May: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/5453/workflows/3ac8bb24-b7d8-4dcf-a9ae-bedfffd1f27c/jobs/6101

#### Problem

This is a major bottleneck for getting PRs in, especially when updates that touch multiple modules required separate PRs. 

#### Next steps

I'd probably start by looking at the history for this module and seeing what changed within the last few months or so.

jaclyn-taroni:
The `focal-cn-preparation` module was taking over an hour to run in CI. This is because it was using the consensus SEG file that is committed to the repository, which is not for a subset of samples but instead contains all samples. With the release of v20, there's no reason to continue to use the file in `analyses/copy_number_consensus_call/results` so here I am changing the path in the second notebook of `focal-cn-preparation` to use the file in the data download. In the context of CI, this will be the subset file.

Closes #1129.


analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd
```diff
@@ -41,9 +41,7 @@ if(!dir.exists(output_dir)) {
 ```
 
 ```{r}
-# TODO: the consensus SEG file is not currently in the data download -- when it
-# gets included we will have to change the file path here
-consensus_seg_file <- file.path("..", "copy_number_consensus_call", "results", 
+consensus_seg_file <- file.path("..", "..", "data",
```
jaclyn-taroni, line 44: @jharenza and @kgaonkar6 - is there any reason (that you are aware of) that prohibits me from making this change at this point?

jharenza: I do not think so - the latest file is in the download.

jaclyn-taroni: We probably need to wait until all the v20 changes go through though? Or would just getting #1123 through -> updating the release (if not done yet) be sufficient?

jaclyn-taroni:
It is curious, though, that the consensus CN steps happens right before this one and I would expect then that the file in `results` for that module would then be from the subset of samples. There might be something snakemake-y happening that I don't understand.

analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd
kgaonkar6: I've updated v20 release data with the latest consensus seg with #1123 `b9284650be04df3538e6c6dba29b8eb0  pbta-cnv-consensus.seg.gz`

I added the relative path so that we can run the code while running all the preprocessing steps of subtyping, so maybe we can add a logic to change to relative path or not  within the if (params$base_run ==0)

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/a6281c2aebfc69213cf8cc0ff9b2b246b96ea294/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L48-L58

jaclyn-taroni: In which scenario would you use the file in `analyses/copy_number_consensus_call/results`? When you use `pbta-histologies-base.tsv`?

kgaonkar6: yes, we will rerun the consensus seg file module first so the latest consensus seg file can be used as input for the focal-cn module

jaclyn-taroni:
Closing in favor of #1124

PR #1131, opened 2021-08-12
remove previous CNV results to force full snakemake run in CI

jashapiro:

We had some odd CI behaviour where it seems like the CNV consensus step may not have been fully running in CI. The most likely cause of this would be `snakemake` deciding not to run because the output files were newer than the inputs. To prevent that, I added a step to remove any lingering output files before running the pipeline. 

CI completeness

#### What was your approach?

Add a step in the shell script for CNV consensus to remove any previous output files

NA

Do all the steps run in CI?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -44,94 +44,94 @@ jobs:
 
      ### MOLECULAR SUBTYPING ###
 
-    #   - run:
-    #       name: Molecular Subtyping - HGG
-    #       command: OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-HGG/run-molecular-subtyping-HGG.sh
+      - run:
+          name: Molecular Subtyping - HGG
+          command: OPENPBTA_SUBSET=0 ./scripts/run_in_ci.sh bash analyses/molecular-subtyping-HGG/run-molecular-subtyping-HGG.sh
 
-    #   - run:
-    #       name: Molecular subtyping - Non-MB/Non-ATRT Embryonal tumors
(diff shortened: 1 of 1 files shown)


jaclyn-taroni:
I'm not sure that this change is enough to gather the information we're after. In #1124, we changed the focal CN code such that it will use the data download (e.g., the subset files) in CI: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/87d5d4b0afced3410a0d87263c57a8917a111634/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L48-57

So you may need to temporarily change that notebook to use the SEG file in `analyses/copy_number_consensus_call/results` and if the focal CN prep step continues to take a matter of minutes, that's a good indication that this is working as intended in CI. Thoughts?

(I'd probably temporarily comment out everything before the CN consensus steps in the config file, too, in the interest of speed.)

jashapiro:
> I'm not sure that this change is enough to gather the information we're after. In #1124, we changed the focal CN code such that it will use the data download (e.g., the subset files) in CI: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/87d5d4b0afced3410a0d87263c57a8917a111634/analyses/focal-cn-file-preparation/02-add-ploidy-consensus.Rmd#L48-57
> 

Looking at those changes, it seems like we have two options: change the `consensus_seg_file` in the version when `params$base_run == 0` or we could change the invocation of the scripts to test when `params$base_run == 1`. That would use a different histology file as well, but seems like it should be a reasonable test. 

I am assuming that the base histology file is also a subset in CI... is that a safe assumption?

Issue #1144, opened 2021-08-18
Updated analysis: update some figures to include annotations by `cancer_group` for manuscript


jharenza:
#### What analysis module should be updated and why? What changes need to be made? Please provide enough detail for another participant to make the update.

- [x] interaction-plots - update plotting to be by `cancer_group`
- [x] telomerase-activity-prediction - change `broad_histology` and `short_histology` plots to `cancer_group` for the main figure. Could keep all in the analysis module</s>
- [ ] <s> transcriptomic-dimension-reduction - color by `cancer_group` </s>
- [x] sample-distribution-analysis - add second level with `cancer_group`,  3rd/4th/5th with WGS/WXS/RNA, 7th with tumor descriptor, 8th with sex

#### What input data should be used? Which data were used in the version being updated?
v20

#### When do you expect the revised analysis will be completed?
this and next week

#### Who will complete the updated analysis?
@kgaonkar6 

Note, ticket still a WIP

kgaonkar6:
While working in the interaction plots I've identified some points that I'd like to get clarification on:

1) Do we want to plot coocurrence plot per each cancer_group corresponding to the broad_histology here https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/0f5a7263af2b02f80c6949f9d33730fe006eaab2/analyses/interaction-plots/01-create-interaction-plots.sh#L40-L49 ? Asking because dividing some cancer_groups only have a couple of samples 
```
             broad_histology                        cancer_group   n
1 Low-grade astrocytic tumor      Diffuse fibrillary astrocytoma   7
2 Low-grade astrocytic tumor                       Ganglioglioma   1
3 Low-grade astrocytic tumor                 Gliomatosis cerebri   1
4 Low-grade astrocytic tumor                                 GNG  47
5 Low-grade astrocytic tumor                   Oligodendroglioma   1
6 Low-grade astrocytic tumor               Pilocytic astrocytoma 123
7 Low-grade astrocytic tumor       Pleomorphic xanthoastrocytoma  11
8 Low-grade astrocytic tumor Subependymal Giant Cell Astrocytoma   9
```

[...]

PR #1161, opened 2021-08-24
#1144 part2: Sample distribution multi-layer plot

kgaonkar6:
Update sample distribution multi-layer plot centered on cancer_group.

#### What was your approach?

So as per requested multi-layer plot in #1144 I've added 

- 2nd level with cancer_group
- 3rd with WGS Available/Not Available
- 4th with WGS Available/Not Available
- 5th with WGS Available/Not Available
- 6th with tumor descriptor,
- 7th with germline_sex_estimate

#1144 

- Do the 3rd/4th/5th layer for WGS/WXS/RNA-Seq look ok? I kept them as `Available` and `Not Available` so each layer is color coded the same way.  I did try putting in "WGS" when WGS is available in 3rd layer but then the color for "WGS","WXS" and "RNA-Seq" would be different so could be too much color wise? Let me know if we would like some other representation.
 - Does distribution_across_cancer_types_treemap.pdf look as expected? I had to remove the subgroup param because `cancer_group` is the only histologies term we are using in the latest plot.

- Incomplete rows are coded to be removed, so samples that are cancer_group==NA will be removed, does that sound ok?

Yes

figures

multi-layer sample distribution plot, example vis:
Screen Shot 2021-08-25 at 5 26 28 PM

#### Documentation Checklist


kgaonkar6:
In the latest commits I've updated the plot to use sample_id and added figures/palettes/tumor_descriptor_palette.tsv for the tumor_descriptor values as mentioned above.

kgaonkar6:
hmm, seems to error out in CI with `testing` but runs ok locally with testing data. Rerunning CI to see if it fixes it.

analyses/sample-distribution-analysis/02-multilayer-plots.R
```diff
@@ -40,127 +41,174 @@ histology_label_mapping <- readr::read_tsv(
...
+  dplyr::inner_join(histology_label_mapping
+                   , by= c( "level2"="cancer_group")) %>%
+  dplyr::rename(color=cancer_group_hex_codes)%>%
+  dplyr::select(colnames(tm))
+
+level3 <- tm %>%
+  dplyr::filter(level == 3 ) %>%
+  dplyr::mutate(color = dplyr::case_when(level3 == "WGS" ~ "#E5E5E5",
+                                         level3 == "WXS" ~ "#B3B3B3",
+                                         level3 == "Targetted Sequencing" ~ "6b6b6b",
```
jharenza, line 166:
```suggestion
                                         level3 == "Targeted Sequencing" ~ "6b6b6b",
```

jharenza (approved):
small typo, but otherwise, looks great! thank you!

kgaonkar6:
Thanks @jharenza !

However, it seems the `figures` code to plot sample distribution errors out in CI I've removed `Rscript figures/scripts/fig1-sample-distribution.R` from CI . Since we will be coping over the sunburst pie plot from here to figures. 

Does that sound ok @jaclyn-taroni ?

jaclyn-taroni:
If we're not going to use `figures/scripts/fig1-sample-distribution.R`, which is what I understand from your comment, we could also just remove that script from the repository in addition to removing it from CI (that seems less confusing to me).

kgaonkar6:
ok removed `figures/scripts/fig1-sample-distribution.R`

.circleci/config.yml
```diff
@@ -28,9 +28,9 @@ jobs:
           name: Sample Distribution Analyses
           command: ./scripts/run_in_ci.sh bash "analyses/sample-distribution-analysis/run-sample-distribution.sh"
 
-      - run:
-          name: Sample Distribution Figure
-          command: ./scripts/run_in_ci.sh Rscript figures/scripts/fig1-sample-distribution.R
+      # - run:
+      #    name: Sample Distribution Figure
+      #    command: ./scripts/run_in_ci.sh Rscript figures/scripts/fig1-sample-distribution.R
```
jaclyn-taroni, line 33: Script has been removed, so I don't see any reason to keep this in here commented out. We'll have a new Fig 1 script eventually, but it might have a totally different name!

```suggestion
```

PR #1143, opened 2021-08-17
oncoprint update PR 1/n

jharenza:
#### What scientific question is your analysis addressing?
First of N PRs to update oncoprints for manuscript submission

#### What was your approach?
- update gene lists (add a few C19MC key genes for ETMRs to embryonal, add ALK, ROS1, EZHIP to hgat, move ependymal genes to other list, remove ependymal list)
- move ependymal tumor to CNS other oncoprint since there were so few genes in the plot - to enable better panel set up for a main figure
- convert relative paths to point to the data directory
- remove rowbars
- remove tumor sample barcodes (7316 ids)
- updated README

#### What GitHub issue does your pull request address?
#1141 

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
No

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What types of results are included (e.g., table, figure)?
- Updated data folder for goi lists
- Updated oncoprints

#### What is your summary of the results?
4 oncoprint figures instead of 5, with better sample and gene distributions for creating a panel figure

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [x] This analysis module has a `README` and it is up to date.
- [x] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1164, opened 2021-08-26
Update oncoprint panel figure

### LINKED ISSUE #1148: Updated analysis: update oncoprint panel figure code
`figures`: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/oncoprint-landscape.R

Update the panel for 4 figures instead of 5, since #1143 combined ependymal tumors into other

same as current script

next week?

@kgaonkar6 or @cbethell cc @jaclyn-taroni

cbethell:

This PR closes #1148 

It updates [`figures/scripts/oncoprint-landscape.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/oncoprint-landscape.R) to include 4 instead of 5 panels because, as noted in #1148, #1143 combined `ependymal tumors` into `other`.

No specific question is being addressed. Updates are being made to the oncoprint landscape figure generation script as to propagate changes from the updated analysis in #1143.

#### What was your approach?

Pulled the latest docker image and updated the local master of this repo before modifying `figures/scripts/oncoprint-landscape.R` to include only 4 panels in each of the output files (namely, `figures/pngs/primary-plus_oncoprint_landscape.png` and `figures/pngs/primary_only_oncoprint_landscape.png`), removing the `ependymal tumor` panel.

#1148

Do the final figures look reasonable? Are there formatting changes we would like to make?

Yes.

The following plots have been updated:

- `figures/pngs/primary-plus_oncoprint_landscape.png` 
- `figures/pngs/primary_only_oncoprint_landscape.png`

The 5th panel appears to have successfully been removed and the plots incorporated appear to be the latest as updated in `master`.

#### Documentation Checklist


cbethell:
In one of the most [recent commits](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/d78a0a34ae0506c27d9aa4097ed2e1dfdfa4e02b), I believe I was able to successfully implement the following changes:

- remove titles on each plot (`Altered in*`)
- Move broad histologies titles to be centered above each plot
- Label A,B (top), C,D (bottom)
- Increase gene name font size as much as possible
- Increase size of the CNV boxes/Make them rectangles
- Try to make the plots more horizontally condensed for a portrait page -- this may possibly further be improved with the removal of the whitespace on either side of the plots, is this something that seems closer to what we may want here @jharenza?

Many of the changes made were only possible when installing the maftools package from github so I also modified the package installation in the `Dockerfile` -- will request @jaclyn-taroni's review here mainly for this purpose.

jharenza:
Thanks so much for the updates @cbethell!

>Try to make the plots more horizontally condensed for a portrait page -- this may possibly further be improved with the removal of the whitespace on either side of the plots, is this something that seems closer to what we may want here @jharenza?

Yes, if possible!

cbethell:
> > Try to make the plots more horizontally condensed for a portrait page -- this may possibly further be improved with the removal of the whitespace on either side of the plots, is this something that seems closer to what we may want here @jharenza?
> 
> Yes, if possible!

Done! 👍

figures/scripts/oncoprint-landscape.R
```diff
@@ -58,49 +58,41 @@ opt <- optparse::parse_args(opt_parser)
...
-ependymal_png <- file.path(onco_dir, paste0(opt$lead_filename, "_ependymal_goi_oncoprint.png"))
 other_cns_png <- file.path(onco_dir, paste0(opt$lead_filename, "_other_goi_oncoprint.png"))
 
 #### Assemble multipanel figure ------------------------------------------------
 
-oncoprint_figure <- multi_panel_figure(columns = 8,
-                                            rows = 2,
-                                            width = 1200,
-                                            height = 300)
+oncoprint_figure <- multi_panel_figure(columns = 4,
```
jaclyn-taroni, line 65: One thing I noticed is that it almost looks like the titles for panels in the lower row are being cutoff ever so slightly during assembly. However, it kind of looks like that in the original PNGs, too. I am wondering if decreasing the `titleFontSize` or increasing the margin at the top a tiny bit might remedy that (in `02-plot-oncoprint.R`)

If we think the problem is instead assembly with `multi_panel_figure()` such that the white background of the top PNGs is responsible for the cutoff, another idea is to make the background of the PNGs transparent, which you can do with `magick` ([ref](https://stackoverflow.com/questions/50479875/how-to-remove-background-of-images-with-the-magick-package)), e.g., in this script do something like the following for each panel:

```r
img_png <- magick::read_image(<path to image>)
img_png <- magick::image_transparent(img_png, 'white')
```

I feel like it's more likely the first thing though in examining the PNGs by broad histology. I'll also note that if this takes more than ~1 hour, I think we've gone down a not super pressing rabbit hole. I also am aware that this can take awhile to run so I would certainly not run the whole module to find out, I'd start with a single broad histology!

jaclyn-taroni (commented):
Really nice work @cbethell! This looks mostly to be in order and I had two main comments (captured in the line comments but I'll reiterate here):

- What happens when one doesn't use the optional `--broad_histology` argument to `02-plot-oncoprint.R`? Do the changes you've made still work without error?
- The titles of the oncoprint panels are being cutoff a little bit in a way that looks more pronounced for the lower panels. I left a couple thoughts about what you might try to remedy that.

In my opinion, the first point needs to be addressed but if the second one is taking a while to get to the bottom of, we should punt until #1165!

I'm going to let @jharenza approve to ensure that you've addressed all of her comments.

jharenza (approved):
Hi @cbethell! thanks for all of the great updates here! I think the below one was only partially addressed, but happy to have this come later.
>Update germline_sex_estimate label to sex and cancer_group to histology(?) and make sure they are not cut off in the figures

For this, I meant for us to remove all of the underscores and have some shorter terminology (`sex`), although `cancer group` without the underscore is also fine if we want to just keep it consistent with the column name. So, I will approve (these are minor things we can fix whenever) since the major updates are in.

figures/scripts/oncoprint-landscape.R
cbethell: I also believe it is the first thing as I noticed this yesterday and went down a mini rabit hole trying to increase the margins but I was unable to successfully increase the top margins so I decreased the title size in the [most recent commit](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1164/commits/b764e012db44f276a4f0abc644f22f6d77cd42f9) (and also slightly adjusted the height of the plots).

cbethell:
> Hi @cbethell! thanks for all of the great updates here! I think the below one was only partially addressed, but happy to have this come later.
> 
> > Update germline_sex_estimate label to sex and cancer_group to histology(?) and make sure they are not cut off in the figures
> 
> For this, I meant for us to remove all of the underscores and have some shorter terminology (`sex`), although `cancer group` without the underscore is also fine if we want to just keep it consistent with the column name. So, I will approve (these are minor things we can fix whenever) since the major updates are in.

Ah yes @jharenza, when I dove into attempting to update this, it was not as straightforward to address as I thought so I figured that would best be included in a later PR with the remaining items on #1165!

jharenza:
Sounds good!

PR #1169, opened 2021-09-01
Update EPN subtyping readme

jashapiro:

AS noticed by @runjin326 [here]( https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/150), the  readme for the EPN subtyping module does not match the code as implemented in the module, in particular the list of prioritized fusions. 

#### What was your approach?

Updated the table!

NA

This is a documentation-only change, so unless the code should be updated instead of the readme (i.e. if the readme was correct, and the code wrong), this should be straightforward.

The relevant code is here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-EPN/03-subgrouping_samples.ipynb


jaclyn-taroni (approved):
Thanks for the fix! And for catching this @runjin326. I'm going to merge without CI because it's a docs only change.

PR #1188, opened 2021-10-17
add Ns and percents to co-occurrence tables for paper

jharenza:
Added additional stats to tables for ease of manuscript writing.

#### What was your approach?
I added the following calculations to the `cooccur_functions.R` script and reran the code to add this information to the results tables 
- n_mutated_gene1
- n_mutated_gene2
- perc_mutated_gene1
- perc_mutated_gene2
- perc_cooccur_or_mutexcl

NA

Updates in `cooccur_functions.R`

Added comment within the function R script:
```
#' @return A data frame summarizing the co-occurence of pairs of genes in the
#'   gene list with columns `gene1`; `gene2`; counts of each mutations in
#'   each category of sharing (`mut11`: both mutated; `mut10`: mutated in
#'   the first but not second gene, etc.) and summary (`n_mutated_gene1` and 
#'   `n_mutated_gene2`: number of samples with gene1 or gene2 mutated, ) `n_mutated_gene1`); 
#'   `odds_ratio` for co-occurence, `cooccur_sign`: 1 if co-occurence greater than by chance, 
#'   -1 if less frequent than expected; `p` the fisher's exact test p value; and a
#'   `cooccur_score` calculated as `cooccur_sign * -log10(p)`. Overall percent 
#'   of samples mutated denoted as `perc_mutated_gene1` and `perc_mutated_gene2`. 
#'   Percent of gene1-mutated samples in w#' @return A data frame summarizing the co-occurence of pairs of genes in the
#'   gene list with columns `gene1`; `gene2`; counts of each mutations in
#'   each category of sharing (`mut11`: both mutated; `mut10`: mutated in
#'   the first but not second gene, etc.) and summary (`n_mutated_gene1` and 
#'   `n_mutated_gene2`: number of samples with gene1 or gene2 mutated, ) `n_mutated_gene1`); 
#'   `odds_ratio` for co-occurence, `cooccur_sign`: 1 if co-occurence greater than by chance, 
#'   -1 if less frequent than expected; `p` the fisher's exact test p value; and a
#'   `cooccur_score` calculated as `cooccur_sign * -log10(p)`. Overall percent 
#'   of samples mutated denoted as `perc_mutated_gene1` and `perc_mutated_gene2`. 
#'   Percent of gene1-mutated samples in which gene2 mutations are co-occuring or mutually exclusive 
#'   denoted as `perc_cooccur_or_mutexcl`.
```

Is this clear?

Yes

Tables

Appended 5 columns to results tables. All other values and plots remain unchanged.

#### Documentation Checklist


analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -131,14 +135,23 @@ coocurrence <- function(gene_sample_df,
...
-
+      cooccur_score = cooccur_sign * -log10(p)) %>%
+    dplyr::rowwise() %>%
+    dplyr::mutate(    
+      n_mutated_gene1 = sum(mut11, mut10),
+      n_mutated_gene2 = sum(mut11, mut01),
+      perc_mutated_gene1 = sum(mut11, mut10)*100/sum(mut11, mut10, mut01, mut00),
+      perc_mutated_gene2 = sum(mut11, mut01)*100/sum(mut11, mut10, mut01, mut00),
+      perc_cooccur_or_mutexcl = ifelse(cooccur_sign == 1, 
+                                       mut11*100/sum(mut11, mut10),
```
jashapiro, line 152: I'm not sure I understand the logic here. Should this include `mut01`? If not, then it becomes dependent on the order the genes are listed, which seems odd to me. 

I'm also not sure we should combine these two calculations into a single column. It seems likely to result in errors downstream.

analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -81,10 +81,14 @@ filter_mutations <- function(maf_df,
...
-#'   frequent than expected; `p` the fisher's exact test p value; and a
-#'   `cooccur_score` calucated as `cooccur_sign * -log10(p)`.
+#'   the first but not second gene, etc.) and summary (`n_mutated_gene1` and 
+#'   `n_mutated_gene2`: number of samples with gene1 or gene2 mutated, ) `n_mutated_gene1`); 
+#'   `odds_ratio` for co-occurence, `cooccur_sign`: 1 if co-occurence greater than by chance, 
+#'   -1 if less frequent than expected; `p` the fisher's exact test p value; and a
+#'   `cooccur_score` calculated as `cooccur_sign * -log10(p)`. Overall percent 
+#'   of samples mutated denoted as `perc_mutated_gene1` and `perc_mutated_gene2`. 
+#'   Percent of gene1-mutated samples in which gene2 mutations are co-occuring or mutually exclusive 
+#'   denoted as `perc_cooccur_or_mutexcl`.
```
jashapiro, line 91: I think we should reformat this section for easier reading/updating. The current list is not in the order of the output table, so we should also try to make that the case. I have a suggested format below, which also incorporates my suggestion to split cooccurance and mutual exclusivity into separate columns.

```
#' @return A data frame summarizing the co-occurence of pairs of genes in the
#'   gene list with columns:
#'  `gene1`; `gene2` (gene names);
#'  `mut11`; `mut10`; `mut01` (counts of mutations in each category of sharing:
#'     `mut11`: both mutated; `mut10`: mutated in the first but not second gene, etc.);
#'  `odds_ratio` (odds ratio for for co-occurence);
#'  `cooccur_sign` (1 if co-occurence greater than by chance, -1 if less frequent than expected)
#'  `p` (the fisher's exact test p value);
#'  `q` (Benjamini-Hochberg adjusted p value);
#'  `cooccur_score` (calculated as `cooccur_sign * -log10(p)`);
#'  `n_mutated_gene1`; `n_mutated_gene2` (count of samples with gene1 or gene2 mutations); 
#'  `perc_mutated_gene1`; `perc_mutated_gene2` (percent of samples with gene 1 or gene 2 mutations);
#'  `perc_cooccur`;
#'  `perc_mutexl`;
```

analyses/interaction-plots/scripts/cooccur_functions.R
jharenza: >I'm not sure I understand the logic here. Should this include mut01? If not, then it becomes dependent on the order the genes are listed, which seems odd to me.

Are you referring to line 152? 

For co-occurrence, using the example below of TP53 (gene1) and H3F3A (gene2) in HGATs, I wanted the calculation to be: of all samples in which TP53 is mutated (mut10 or mut11), how many have this instance of co-occurrence of gene2, in this case, H3F3A (mut11)? I also realized I do not have a percent co-occurrence/mut-excl in _all_ samples, so I added that as well. So from below, 46% of HGATs have a TP53 mutation, 40% have H3F3A mutated, 28.7% of all HGATs have TP53 and H3F3A mutations co-occurring, and in all HGATs with a TP53 mutation, 54.3% of them have a co-occurring H3F3A mutation. Hopefully that clarifies. Update incoming.

```
> hgat %>%
+   filter(gene1 == "TP53", gene2 == "H3F3A") %>%
+   select(gene1, gene2, mut11, mut10, mut01, mut00, n_mutated_gene1, n_mutated_gene2,
+          perc_mutated_gene1, perc_mutated_gene2, perc_cooccur_all_samples, perc_mutexcl_all_samples, 
+          perc_cooccur_gene1_mutated, perc_mutexcl_gene1_mutated)
# A tibble: 1 x 14
# Rowwise: 
  gene1 gene2 mut11 mut10 mut01 mut00 n_mutated_gene1 n_mutated_gene2 perc_mutated_gen… perc_mutated_ge… perc_cooccur_all… perc_mutexcl_all… perc_cooccur_gen… perc_mutexcl_gen…
  <chr> <chr> <dbl> <dbl> <dbl> <dbl>           <dbl>           <dbl>             <dbl>            <dbl>             <dbl>             <dbl>             <dbl>             <dbl>
1 TP53  H3F3A    25    21    15    26              46              40              52.9             46.0              28.7                NA              54.3                NA
```

 >I'm also not sure we should combine these two calculations into a single column. It seems likely to result in errors downstream.

Yes, will split - went back and forth on that.

jharenza: @jashapiro see the latest commit [d64a173](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1188/commits/d64a1731b56b5baa5d77e0e376ffce6e7fb6b07b)

jharenza: @jashapiro see the latest commit [d64a173](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1188/commits/d64a1731b56b5baa5d77e0e376ffce6e7fb6b07b)

analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -148,9 +166,14 @@ coocurrence <- function(gene_sample_df,
...
-                                       mut11*100/sum(mut11, mut10), 
-                                       sum(mut10,mut01)*100/sum(mut11, mut10, mut01))
+      perc_cooccur_all_samples = ifelse(cooccur_sign == 1, 
+                                        mut11*100/sum(mut11, mut10, mut01, mut00), NA),
+      perc_mutexcl_all_samples = ifelse(cooccur_sign == -1, 
+                                        sum(mut10,mut01)*100/sum(mut11, mut10, mut01, mut00), NA),
+      perc_cooccur_gene1_mutated = ifelse(cooccur_sign == 1, 
+                                          mut11*100/sum(mut11, mut10), NA),
+      perc_mutexcl_gene1_mutated = ifelse(cooccur_sign == -1,
+                                          sum(mut10,mut01)*100/sum(mut11, mut10, mut01), NA)
```
jashapiro, line 176: These statistics are valid whether or not they tend toward co-occurrence or not, so I would remove the `ifelse` completely.

analyses/interaction-plots/scripts/cooccur_functions.R
jashapiro: The case I am concerned about is this: suppose we have 100 samples with mutations in `gene1` and 10 with mutations in `gene2`, all of which have a mutation in gene 1. (`m11 = 10`, `m10 = 90`, `m01 = 0`)

The percent cooccurrence with `gene1` as the denominator is 10% (10/100), but would be 100% with `gene2` as the denominator (10/10). 

Looking back at a bit more code, I can see based on surrounding code `gene1` is always the higher frequency gene the way we are using this function, but that sorting is _not_ part of this function, so somebody using it could end up with different answers depending on the order that gene list is fed in.

By contrast, % mutual exclusivity has a different denominator: it is not relative to gene 1, but to *both* gene 1  and gene 2. This makes sense, because both `mut10` and `mut01` have to be in the numerator, so we need them in the denominator to prevent > 100% values. But I find the difference to be confusing.

jharenza: > The case I am concerned about is this: suppose we have 100 samples with mutations in gene1 and 10 with mutations in gene2, all of which have a mutation in gene 1. (m11 = 10, m10 = 90, m01 = 0)

The percent cooccurrence with gene1 as the denominator is 10% (10/100), but would be 100% with gene2 as the denominator (10/10).

I see what you are saying about this, but am still confused about this:

>Looking back at a bit more code, I can see based on surrounding code gene1 is always the higher frequency gene the way we are using this function, but that sorting is not part of this function, so somebody using it could end up with different answers depending on the order that gene list is fed in.

If `gene1` is always higher freq gene, then no matter if your list is TP53, H3F3A or H3F3A, TP53, then aren't you saying the output would always be TP53 (gene1), H3F3A (gene2)? I had the calculations relative to gene1 based on how I was thinking about this writing. 

>By contrast, % mutual exclusivity has a different denominator: it is not relative to gene 1, but to both gene 1 and gene 2. This makes sense, because both mut10 and mut01 have to be in the numerator, so we need them in the denominator to prevent > 100% values. But I find the difference to be confusing.

I think these are set up correctly?


Based on how I am trying to write this,

jashapiro: > > Looking back at a bit more code, I can see based on surrounding code gene1 is always the higher frequency gene the way we are using this function, but that sorting is not part of this function, so somebody using it could end up with different answers depending on the order that gene list is fed in.
> 
> If `gene1` is always higher freq gene, then no matter if your list is TP53, H3F3A or H3F3A, TP53, then aren't you saying the output would always be TP53 (gene1), H3F3A (gene2)? I had the calculations relative to gene1 based on how I was thinking about this writing.

The way the `coocurrence()` function is written, it takes `gene_sample_df` and a vector `genes` for which to calculate statistics. The order of the `genes` vector is what determines the order of the output table, and hence which of a pair is `gene1` and which is `gene2` (`gene1` is whichever gene is first in the `genes` input vector). It happens that when this function is called, we are giving it a vector of genes sorted by frequency, but this is not enforced at all by the `coocurrence()` function itself. So if the gene list were sorted differently, we would end up with a different table. 

For example, if it were alphabetical we would end up with `H3F3A` as `gene1`, and we would be calculating a different value for the H3F3A/TP53 pair. 

What you could do instead is to make the relevant line:
```
perc_cooccur =  mut11*100/(mut11 + max(mut10, mut01))
```
This would give the same result no matter the order of the genes: always the percentage of the higher frequency mutation. 

But there is also a deeper problem:
We could also end up with a situation where in one disease H3F3A is mutated more than TP53; in that case the order of `gene1` and `gene2` would be reversed in one table relative to the other, and the comparisons of "percent cooccurrence" would be measuring different things: In once case the % of H3F3A mutations with a TP53 mutation, and in the other the % of TP53 mutations that have an H3F3A mutation.

If we instead calculate 
```
perc_cooccur =  mut11*100/(mut11 + mut10 + mut01)
```
This statistic will always have the same meaning: of samples with at least one mutation in `gene1` or `gene2`, what percentage have both mutations. This is also more parallel to the % exclusivity calculation, as the denominators are the same (in fact, the two stats will always add to 100%).

jharenza: Hi @jashapiro - thanks for explaining the `coocurrence()` function - I agree that using 
```
perc_cooccur =  mut11*100/(mut11 + mut10 + mut01)
```
will work just as well, be clearer, and will adapt the wording based on that calculation instead.

analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -79,12 +79,32 @@ filter_mutations <- function(maf_df,
...
+#'     `mut11`: both mutated; `mut10`: mutated in the first but not second gene, etc.);
+#'  `odds_ratio` (odds ratio for for co-occurence);
+#'  `cooccur_sign` (1 if co-occurence greater than by chance, -1 if less frequent than expected)
+#'  `p` (the fisher's exact test p value);
+#'  `q` (Benjamini-Hochberg adjusted p value);
+#'  `cooccur_score` (calculated as `cooccur_sign * -log10(p)`);
+#'  `n_mutated_gene1`; `n_mutated_gene2` (count of samples with gene1 or gene2 mutations); 
+#'  `perc_mutated_gene1`; `perc_mutated_gene2` (percent of samples with gene 1 or gene 2 mutations);
+#'  `perc_cooccur`; (Percent of all surveyed samples with co-occurring gene1/2 mutations when either gene1 or gene2 are mutated);
+#'  `perc_mutexcl`; (Percent of all surveyed samples with gene1/2 mutations being mutually exclusive);
```
jashapiro, line 107: I was confused by the "all surveyed samples" here, so I'm trying a slightly different wording here.
```suggestion
#'  `perc_cooccur`; (Percent of samples with mutations in gene 1 and/or gene 2 with co-occurring gene1/2 mutations);
#'  `perc_mutexcl`; (Percent of samples with mutations in gene 1 and/or gene 2 where gene1/2 mutations are mutually exclusive);
```

analyses/interaction-plots/scripts/cooccur_functions.R
```diff
@@ -131,14 +151,22 @@ coocurrence <- function(gene_sample_df,
...
-    )
-
+      cooccur_score = cooccur_sign * -log10(p)) %>%
+    dplyr::rowwise() %>%
+    dplyr::mutate(    
+      n_mutated_gene1 = sum(mut11, mut10),
+      n_mutated_gene2 = sum(mut11, mut01),
+      perc_mutated_gene1 = sum(mut11, mut10)*100/sum(mut11, mut10, mut01, mut00),
+      perc_mutated_gene2 = sum(mut11, mut01)*100/sum(mut11, mut10, mut01, mut00),
+      perc_cooccur = mut11*100/sum(mut11 + mut10 + mut01),
```
jashapiro, line 167: Just making this parallel with the other calculations here.
```suggestion
      perc_cooccur = mut11*100/sum(mut11, mut10, mut01),
```

Issue #1173, opened 2021-09-02
[Discussion] - Manuscript figure for de novo mutational signatures


jharenza:
We need to sketch out a manuscript figure for the de novo CNS mutational signatures now that #1100 has been merged.

Some things we should have:
1 - Supplemental excel book of DNA BS_ID + signatures with tabs for (CNS signatures `mean` in `fitted_cns_signature_exposures.RDS`), COSMIC, and Alexandrov
2 - plot of CNS signature proportions by `cancer_group`(?) - can be similar to the COSMIC/Alexandrov or stacked bar?

Some things to consider:
- Current plots could be suitable for the supplement (`https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/mutational-signatures/plots/cns/mean_median_exposure_barplot.png`), but not sure what the colors indicate (histologies?)
- We need some sort of mapping of ABC, etc signature to the underlying mechanism of the signature, if known (eg - which signature is due to MMR or HR deficiency, TMZ treatment, etc?) Do we know this information? Will be important to capture this by cancer type so that we can describe the results and also compare to the COSMIC/Alexandrov signatures as to why this is an improvement. Hoping it shows something cool

cc @jaclyn-taroni

jaclyn-taroni:
Link to the signatures we used and the related reference signatures: https://signal.mutationalsignatures.com/explore/studyTissueType/1-6#

jaclyn-taroni:
I'm going to assign myself to this so I can do a bit of poking around in the existing code before making recommendations. For my own reference, the bubble plots we have are an adaptation of this figure: https://www.nature.com/articles/s41586-020-1943-3/figures/3 (from [Alexandrov et al. 2020](https://doi.org/10.1038/s41586-020-1943-3)).

jharenza:
done per #1220

PR #1190, opened 2021-10-21
Save output of fit_signatures(), make it easier just to run CNS fitting steps

jaclyn-taroni:

I am working on #1173. To make that easier, I've altered the CNS mutational signature fitting step to save the entire output of `fit_signatures()`. This file is 2.2 GB in size, so it needs to be ignored. I also want there be a way, using the shell script, to only run the steps required for the fitting CNS signatures.

### DIFF
--- a/analyses/mutational-signatures/run_mutational_signatures.sh
+++ b/analyses/mutational-signatures/run_mutational_signatures.sh
@@ -19,7 +19,7 @@ ABBREVIATED_MUTSIGS=${OPENPBTA_QUICK_MUTSIGS:-0}
 # Split up the consensus MAF files by experimental strategy (writes to scratch)
 Rscript --vanilla 02-split_experimental_strategy.R
 
-if [ "$CNS_FIT_ONLY" -lt "1" ]; then
+if [ "$CNS_FIT_ONLY" == "0" ]; then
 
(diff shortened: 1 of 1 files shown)


analyses/mutational-signatures/run_mutational_signatures.sh
```diff
@@ -9,28 +9,35 @@ set -o pipefail
...
-Rscript -e "rmarkdown::render('01-known_signatures.Rmd', clean = TRUE)"
-
 # Split up the consensus MAF files by experimental strategy (writes to scratch)
 Rscript --vanilla 02-split_experimental_strategy.R
 
-# Run the shell script for determining the number of signatures to use
-# with a low number of iterations if run in CI
-# argument 0 --> run denovo goodness of fit
-QUICK_MUTSIGS=$ABBREVIATED_MUTSIGS ANALYSIS=0 bash 03-de_novo_range_of_nsignatures.sh
+if [ "$CNS_FIT_ONLY" -lt "1" ]; then
```
jashapiro, line 22: I don't know, but I find this easier to follow the logic of. 
```suggestion
if [ "$CNS_FIT_ONLY" == "0" ]; then
```

PR #1193, opened 2021-10-22
Create minimal main display item palette (#1176)

jaclyn-taroni:
### Purpose/implementation Section

Addresses the new palette part of #1176. I haven't added documentation to the `figures` README yet. This is in part by design, because I plan on using this palette to remake figures for a bit to figure out if I've set this up in a way that makes sense. I'm also leaving the old palette in for now.

Key points:

* Putting together the palettes for `broad_histology` and `cancer_group`
* Adding some ordering 
* Columns specifically for inclusion & grouping in the oncoprints

PR #1194, opened 2021-10-22
Update Figure 4 single panel PDFs, add compiled figure

jaclyn-taroni:
⚠️  downstream of #1193

The big idea here is to move to a model where individual panels for figures are saved as PDFs and then assembled via Illustrator, where we could put the `.ai` file somewhere read-only. Then we can track the PDF output and PNG files in the repository. Explained a bit in `figures/pdfs/fig4/README.md`.


cbethell (approved):
I was able to successfully re-run what's in this PR and everything LGTM! ✅

PR #1195, opened 2021-10-26
Add Figure 1A component

jaclyn-taroni:
If we're following the convention put forth in #1194, where we'll track individual panels as PDFs in this repository, then it makes sense to me to include the Fig 1A panel from [the placeholder figure](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/69cf76bd28a84e7587febbd11665d324da22177d/content/images/main/Figure1.png) here.


jharenza (approved):

cbethell (approved):
🚀

PR #1196, opened 2021-10-27
Update chromothripsis panels to use new cancer group palette (Fig 3)

jaclyn-taroni:
⚠️ downstream of #1193

Working towards PDF versions of individual panels in directories `figures/pdfs/<figure number>/panels`  using the new color palette in #1193. 

This is regarding Figure 3 ([placeholder image](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/69cf76bd28a84e7587febbd11665d324da22177d/content/images/main/Figure3.png)). Here I am:

* Adding a shell script to make it easy to run the entirety of the chromothripsis module
* Saving the CNV breaks vs. SV breaks scatter plot panel as a PDF (in addition to a PNG) in the notebook where that plot was originally made
* Adding the chromothripsis figure steps to `figures/generate-figures.sh`
  * Running the module in its entirety via the new shell script for that module 
  * Copying over the PDF of the scatter plot panel to `figures/pdfs/fig3/panels`
  * Adding an Rscript, adapted from `analyses/chromothripsis/04-plot-chromothripsis-and-breakpoint-data.Rmd`, to make the bar plot panel with the new cancer group palette introduced in #1193

The new publication ready plot specific Rscript seemed like a quicker path to getting this panel done (and tweaked as necessary) because I haven't removed the older palette (`figures/palettes/histology_label_color_table.tsv`) yet.

PR #1197, opened 2021-10-27
Update interaction plot to use new cancer group palette

jaclyn-taroni:
⚠️ downstream of #1193 and #1196

Working towards PDF versions of individual panels in directories `figures/pdfs/<figure number>/panels`  using the new color palette in #1193. 

This is regarding Figure 3 ([placeholder image](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/69cf76bd28a84e7587febbd11665d324da22177d/content/images/main/Figure3.png)). Here I am:

* Saving the `interaction-plot` figures for _all disease types_ as PDFs, instead of PNGs
* Removing any labeling of panels by `patchwork`
* Adding patterning to the stacked bar chart because without it it's a little difficult to distinguish between DMG, CRANIO, and HGG astrocytoma
* Copying over the PDF instead of the PNG (see point 1!)


jharenza:
The colors look better to me!

jaclyn-taroni:
The patterning has been removed, as we've updated cancer group colors such that distinguishing between adjacent cancer groups in the stacked bar plots is less of a concern.

jashapiro (approved):
I think this looks good, but I have a couple of questions.

The first is just that I am a little confused when you say you are adding "Adding patterning to the stacked bar chart". Do you just mean the outline, or was there more patterning to add that is not showing up here? I am assuming the outline is all you meant, so I am approving this, but if there was more that you intended to do, you can rerequest me after adding it. 

The other question is if we want to update all the figures to pdf? Probably not, but thought it was worth asking.

Issue #1178, opened 2021-09-12
Create panel of standalone legends for the oncoprint plots


jaclyn-taroni:
⚠️ _Blocked by #1177_

#### Context

Quoting https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1165#issue-980606286

> - Create a unified legend and add to the right side of the plots (I feel like this is a ticket in itself and maybe @kgaonkar6 can tackle - we would maybe have to make color boxes with labels and make them another figure, then put it into this panel - thoughts? (Since there are a lot of samples which are not displayed, some cancer groups in the legend do not have these samples in the plots, so I think we would want to only list those which have samples). It would also be nice to convert these so there are no underscores and the whole words are displayed (eg - `In_Frame_Ins` --> `In-frame insertion`) - honestly, I just made these in excel before (ick) and created a piece of a figure and put it in the final figure, so if this is a huge coding challenge, it can be done like that, as tedious as it is.

#### Next steps

There are at least four steps to getting a unified legend into a final oncoprint figure that I am aware of, which I will detail below.

1. Create a series of [standalone legends using `legend()`](https://stackoverflow.com/questions/48966645/how-can-i-create-a-legend-without-a-plot-in-r/48966924). We need the following:
    - The oncoprint palette legend (using `figures/palettes/oncoprint_color_palette.tsv` but with the color names cleaned up)
    - `broad_histology` panel-specific color palettes for each panel, where the `cancer_group` legend is in the same order as the oncoprint panel itself (https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1174#issuecomment-915351111) and only includes `cancer_group` that actually get displayed
      - A. Low-grade astrocytic tumor
      - B. Embryonal tumor
      - C. Diffuse astrocytic and oligodendroglial tumor
      - D. Other CNS
2. Compile the series of legends into a single panel with something like `multipanelfigure`. I think doing _steps 1 AND 2_ within `oncoprint-landscape` with it's own notebook or script might be the best way to go about this.
3. Alter `analyses/oncoprint-landscape/02-plot-oncoprint.R` such that individual oncoprints don't get legends. (Maybe, this will depend on what decisions are made for #1177.)
4. Update `figures/scripts/oncoprint-landscape.R` to include the legends panel from step 2. 

**I am planning on implementing this myself and if not, some of these steps might need their own more detailed issues.**

#### Sketch putting it all together 

[...]

Issue #1177, opened 2021-09-12
Updated analysis: Oncoprints to use new minimal display palettes, remove cancer group with n < 10


jaclyn-taroni:
#### What analysis module should be updated and why?

`oncoprint-landscape`

#### What changes need to be made? Please provide enough detail for another participant to make the update.

With the changes coming in #1176, we will need to remove any `cancer_group` with n < 10 from the oncoprints for main display. I know there is logic in this module that is pretty tightly tied to `broad_histology` values, so there may be a non-trivial amount of work to get this done. 

We also might consider creating _individual_ oncoprint plots for `cancer_group` with n < 10.

#### Who will complete the updated analysis?

I expect I will complete the analysis and if not, I will need to update this with more specific instructions.

jharenza:
I think I misunderstood when we talked about this - I was thinking we would have a separate set of colors for the oncoprint to ensure we are still plotting all cancer groups with mutations. This will also ensure that our plots match pedcbio for the genes listed so that people don't get confused if they check our figures vs pedcbio. Thoughts?

jaclyn-taroni:
My take in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1174#issuecomment-915298938 was that we could include `cancer_group`-specific oncoprints for groups that didn't meet the sample size criteria:

> > I think I have to check on which cancer_group are >=10, but I think for all figures other than the oncoprint, this may be fine. For the oncoprint, we should still annotate as the specific cancer.
> 
> If the specific cancer does not meet this criteria, we could include the oncoprint in the supplemental material instead and possibly even split up by `cancer_group`, rather than `broad_histology`, which would allow us to avoid having to worry about colors for specific cancer groups with N < 10.

And more generally in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1174#issuecomment-915666312

> And for `cancer_group` or `broad_histology` labels that don't make the cutoff based on sample size, we should devise ways to break those plots out individually as needed.

This what I meant by

> We also might consider creating _individual_ oncoprint plots for `cancer_group` with n < 10.

above, as well. We could then assemble the individual, `cancer_group`-specific plots into a multipanel supplemental figure. 

I _think_ that this will solve for this point:

> This will also ensure that our plots match pedcbio for the genes listed so that people don't get confused if they check our figures vs pedcbio.

[...]

PR #1200, opened 2021-10-30
OncoPrint figures

jaclyn-taroni:
⚠️  Stacked on #1193 

Addressing #1177 and #1178, but made slightly different choices than what is detailed in those issues. Importantly, I've only removed cancer groups with less than 10 samples for the other CNS plot. In the current version of #1193, cancer groups with n < 10 in the other 3 broad histologies are randomly assigned a hex code from the `Greys` color brewer palette. There are over 20 cancer groups with n < 10 that'd be included in the Other CNS plot (fewer than that when you consider which groups actually show up in the plots using the other CNS genes of interest list), so it's not feasible to randomly assign greys to them in the same way. Instead, the Other CNS cancer groups with n < 10 are in a supplemental PDF where they are assigned colors based on the `Paired` color brewer palette.

Similar to #1194, I've created an Rscript (`figures/scripts/fig2-oncoprint-landscape.R`) that creates individual panels & legends as PDFs (`figures/pdfs/fig2/panels`) and then I've assembled a draft figure in Illustrator. You can see that draft as `figures/pngs/figure2.png`. I've removed the old script that used `multipanelfigure` to assemble the figure from the PNGs in the module (`figures/scripts/oncoprint-landscape.R`).

I needed to rerun the `oncoprint-landscape` module because we had been manually updating the OncoPrint palette, rather than updating the script that has all the palettes, so some colors got dropped in #1161 when the tumor descriptor palette was created and those colors were missing when things were run as part of #1189. I've also deprecated the former figure 2 script to hopefully avoid confusion.


jaclyn-taroni:
This is a large number of changed files, but everything in `analyses/oncoprint-landscape/plots` is just a rerun now that I've fixed the oncoprint palette and I'm removing a good number of figures that are deprecated.

figures/palettes/oncoprint_color_palette.tsv
```diff
@@ -11,11 +11,13 @@ In_Frame_Ins	#FFE4E1
 Stop_Codon_Ins	#CC79A7
 Start_Codon_Del	#56B4E9
 Fusion	#7B68EE
-Multi_Hit	#00F021
+Multi_Hit	#CCCCCC
 Hom_Deletion	#313695
 Hem_Deletion	#abd9e9
```
jharenza, line 16:
```suggestion
```
We are not actually using these

figures/palettes/oncoprint_color_palette.tsv
```diff
@@ -11,11 +11,13 @@ In_Frame_Ins	#FFE4E1
...
-amplification	#c51b7d
-loss	#0072B2
-gain	#D55E00
+Del	#0072B2
+Amp	#D55E00
```
jharenza, line 18:
```suggestion
Del	#386db0
Amp	#c51b7d
```
new color suggestions

figures/palettes/oncoprint_color_palette.tsv
```diff
@@ -11,11 +11,13 @@ In_Frame_Ins	#FFE4E1
...
 High_Level_Gain	#FF0000
 Multi_Hit_Fusion	#CD96CD
+Intron	#5ea4fe
+5'Flank	#386db0
+3'Flank	#c51b7d
```
jharenza, line 23:
```suggestion
5'Flank	#7CCBAC
3'Flank	#FD8344
```
new color suggestions - moving these two to Amp/Del, where I had meant them to be previously, but it looks like things got shuffled. Let's see what these look like 😬

figures/palettes/oncoprint_color_palette.tsv
```diff
@@ -11,11 +11,13 @@ In_Frame_Ins	#FFE4E1
...
-Multi_Hit	#00F021
+Multi_Hit	#CCCCCC
 Hom_Deletion	#313695
 Hem_Deletion	#abd9e9
-amplification	#c51b7d
-loss	#0072B2
-gain	#D55E00
+Del	#0072B2
+Amp	#D55E00
 High_Level_Gain	#FF0000
```
jharenza, line 19:
```suggestion
```
We are not using this one either

jharenza (commented):
>Instead, the Other CNS cancer groups with n < 10 are in a supplemental PDF where they are assigned colors based on the Paired color brewer palette.

I like this! I will modify Supplemental Figure 3 in the manuscript repo to add this panel.

For the `oncoprint_alteration_legend.pdf`, it looks like we have a few alterations/colors which we do not use due to the renaming of CNV alterations, so I updated the code to remove those. I also made a few suggestions for swapping some colors.

The only other minor thing I saw was that the oncoprint panel pdfs do not contain `goi` in the filename while the analysis module does, so maybe you can add this to be consistent (I double checked you are using goi here and you are 😄 ).

figures/palettes/oncoprint_color_palette.tsv
jaclyn-taroni: We need to edit the script that generates them (`figures/scripts/color_palettes.R`), not this file directly. Editing this file directly caused problems (see the mentions of #1161 and #1189 in the original PR comment).

jharenza: gotcha!

jaclyn-taroni:
@jharenza I've updated the color palette and rerun the figures. Can you take a look before I invest any more effort, i.e., attempt to put together the multipanel version?

jharenza:
> @jharenza I've updated the color palette and rerun the figures. Can you take a look before I invest any more effort, i.e., attempt to put together the multipanel version?

The colors I mentioned look fine, but what is "Complex_event"? I hadn't seen that before and it is black, along with nonsense mutations, so that would need a new color if we use it. I would think it would fit into other categories, unless it is an SV of some sort?

jaclyn-taroni:
> but what is "Complex_event"? I hadn't seen that before and it is black, along with nonsense mutations, so that would need a new color if we use it. I would think it would fit into other categories, unless it is an SV of some sort?

A couple notes:

- Complex event is not new as of my most recent update - we can see it here in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1200/commits/f6bcf253ac3ceb242e2557e82ad259ea3eb0f729: https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/f6bcf253ac3ceb242e2557e82ad259ea3eb0f729/analyses/oncoprint-landscape/plots/primary-plus_embryonal_oncoprint.png
- Complex event isn't documented as far as I can tell outside of: https://github.com/PoisonAlien/maftools/blob/db66c013049e04999eca8e6aaa440cfe1eb08703/R/oncoplot.R#L644 which I take to mean that we expect to see it when a gene is impacted by both by a mutation and a copy number alteration of some kind
- It's the same navy (or at least close enough) as Translation Start Site. I'll add it to the palette with another color. I'll probably start with a darker green.

jaclyn-taroni:
@jharenza ready for another look

jharenza:
> Complex event isn't documented as far as I can tell outside of: https://github.com/PoisonAlien/maftools/blob/db66c013049e04999eca8e6aaa440cfe1eb08703/R/oncoplot.R#L644 which I take to mean that we expect to see it when a gene is impacted by both by a mutation and a copy number alteration of some kind

Interesting, I had not noticed that before and I wonder if that was earlier code before the CNVs were added as smaller rectangles in which we are able to see a mutation/fusion + a CNV with two different colors. That begs the question why is it showing as complex rather than the two colors for mut/cnv, unless perhaps a multi-hit SNV + CNV? I can't look into that right now, but probably can later tonight. If you get to it before me, let me know.

jaclyn-taroni:
From a cursory look at the code, I think it might arise when there are multiple CNVs. If we set `writeMatrix = TRUE` when we call `oncoplot()`, that might help confirm/reject that idea.

jharenza:
>From a cursory look at the code, I think it might arise when there are multiple CNVs. If we set writeMatrix = TRUE when we call oncoplot(), that might help confirm/reject that idea.

Just noting that we should explain this in the figure legend.

I can't distinguish between the green for complex event and missense, but otherwise, the flank/amp colors are ok for me.

jaclyn-taroni:
Going to request a review from @cbethell! The main file to take a look at is: `figures/scripts/fig2-oncoprint-landscape.R`. We can add the compiled figures later.

figures/generate-figures.sh
```diff
@@ -77,14 +82,8 @@ bash ${analyses_dir}/oncoprint-landscape/run-oncoprint.sh
...
-for filename in "${filenames[@]}"; do
-
-  ## Run the `oncoprint-landscape` figure assembly script
-  Rscript --vanilla scripts/oncoprint-landscape.R \
-    --lead_filename ${filename} \
-    --png_name pngs/${filename}_oncoprint_landscape.png
-
-done
+# Create single panel PDFs and legends
+Rscript --vanilla figures/figures/scripts/fig2-oncoprint-landscape.R
```
cbethell, line 86:
```suggestion
Rscript --vanilla scripts/fig2-oncoprint-landscape.R
```

cbethell (approved):
This LGTM! Just left a comment re: fixing the file path to the figure script in `generate-figures.sh` (tested it to make sure it works successfully and it does) 👍

figures/generate-figures.sh
jaclyn-taroni: Oops! That's super weird, thanks for catching it!

Issue #1203, opened 2021-11-12
Updated analysis: Add CI to interaction-plots module


runjin326:
#### What analysis module should be updated and why?
Add CI to `interaction-plots` module

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Add CIs for the odds ratio calculations.

#### What input data should be used? Which data were used in the version being updated?
N/A

#### When do you expect the revised analysis will be completed?
1-2 days

#### Who will complete the updated analysis?
@runjin326

jaclyn-taroni:
Closed via #1204

PR #1204, opened 2021-11-15
CI calculation added for odds ratio

runjin326:
Add confidence interval calculation for odds ratio in `interaction-plots` module

#### What was your approach?
Three columns - `standard_error_or`, `or_ci_lower_bound` and `or_ci_upper_bound` were added to the results column with the following calculation - 
```
      standard_error_or = sqrt(1/mut11 + 1/mut00 + 1/mut10 + 1/mut01),
      or_ci_lower_bound = exp(log(odds_ratio) - 1.96 * standard_error_or),
      or_ci_upper_bound = exp(log(odds_ratio) + 1.96 * standard_error_or)
```

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1203

Please check and see the addition of CI related columns are as wanted. 

No.

Yes.

All the *.tsv files in the results folder are updated with addition of CI related columns. 

Spot checked a couple entries and the numbers make sense. 

#### Documentation Checklist


jashapiro (commented):
Hi @runjin326-
Thank you for implementing this! 

Unfortunately, the simple OR CI calculation that you implemented here is probably not quite what we will want, as it results in Inf/NA any time we have a `0` count. There are two potential solutions that I can think of:

One is to implement Haldane's Correction when there are zero counts: This would mean adding `0.5` to _all_ counts in the table if _any_ counts are zero, then performing the same standard error calculation as before. This is probably the easiest to implement, but as it starts to get a bit messier, I would pull it out as a separate function, similar to the `row_fisher()` function. One note here is that the CI calculated this way starts from the "corrected" OR, which means that we often end up with an OR that does not include 0, so I would probably set the lower bound value to `min(or_uncorrected, or_lowerbound)`

Another option is to use the `PropCIs::orscoreci()` function, which is probably more accurate (it uses a method from Agretsi), but would involve a more backend work, as it requires adding a new package to the docker image.

runjin326:
@jashapiro, thanks so much for your prompt review -  I would probably take the first approach and ping you for another review when completed.

runjin326:
@jashapiro, I pulled out function for CI and now the results only have one `ci` column with the calculation that you specified. Could you please check to see whether this is what you expect? Thanks!

jashapiro (commented):
Hi @runjin326  I saw that some of the intervals were still not making sense, which I think has to do with when the function is being applied. I also modified the code to more accurately reflect the full adjustment as it has to be applied (if `odd_ratio` is zero, you can't take the log, which is part of the reason for this procedure).

I have not tested the code I wrote, but hopefully it works, or does so mostly!

runjin326:
@jashapiro , thanks so much for the detailed review and instructions. I fixed the scripts based on your suggestion - please check to see whether this now looks good.

jashapiro (approved):
LGTM! I spot checked some of the confidence intervals, and they seem correct to me.

Issue #1183, opened 2021-10-09
Updated analysis: medullo subtyping misses WGS only samples as To be Classified


jharenza:
#### What analysis module should be updated and why?
medulloblastoma-subtyping - missed WGS only samples

```
v21 %>%
  filter(cancer_group == "Medulloblastoma" & sample_type == "Tumor" &
           is.na(molecular_subtype)) %>%
  select(Kids_First_Biospecimen_ID, experimental_strategy, molecular_subtype)

# A tibble: 7 x 3
  Kids_First_Biospecimen_ID experimental_strategy molecular_subtype
  <chr>                     <chr>                 <chr>            
1 BS_3J4T2YYW               WGS                   NA               
2 BS_7F07M7JG               WGS                   NA               
3 BS_AG0BZN6F               WGS                   NA               
4 BS_HZNKSQ17               WGS                   NA               
5 BS_V6HKXX1Z               WGS                   NA               
6 BS_WP9J88EB               WGS                   NA               
7 BS_YYAPSA5P               WGS                   NA        
```

#### What changes need to be made? Please provide enough detail for another participant to make the update.
annotate these samples as to be classified

#### What input data should be used? Which data were used in the version being updated?
v21

#### When do you expect the revised analysis will be completed?
next week, if time?

#### Who will complete the updated analysis?
@komalsrathi 

This is also not a huge deal if not added since they are To be classified, but wanted to track as an issue here

komalsrathi:
Sure, will do. Can you add me as an assignee?

jharenza:
Closing with #1206

PR #1206, opened 2021-11-17
add molecular subtype to wgs-only samples

komalsrathi:

WGS only samples are missing medulloblastoma molecular subtype info. They need to be annotated as "To be Classified"

#### What was your approach?

Added a separate script to append wgs-only samples to the output tsv files

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1183

04-no-RNA-samples.R and output tsv files

N/A

Yes

tsv files

N/A

#### Documentation Checklist


jharenza (approved):
Thanks @komalsrathi - this looks good!

jaclyn-taroni:
Just to note, these changes won't take effect unless we cut another release, correct?

jharenza:
> Just to note, these changes won't take effect unless we cut another release, correct?

I wasn't thinking of another release, but just adding them to our histologies file in the supplement. Do you want us to rerun subtyping and do another release for that to happen?

jaclyn-taroni:
I don't think we want to be in a situation where the release and the supplemental data are out of sync, but perhaps the new release can be the very last thing we do before we cut a release of the codebase.

jharenza:
> I don't think we want to be in a situation where the release and the supplemental data are out of sync, but perhaps the new release can be the very last thing we do before we cut a release of the codebase.

Ok, sounds good - put in #1207 for this

PR #1208, opened 2021-12-02
Add survival curves by TP53 and telomerase scores

jharenza:
Plotting survival curves for all PBTA, HGAT only, and non-HGAT samples for various categorizations of TP53 and telomerase scores for publication

#### What was your approach?
Important points:
- utilized primary stranded RNA samples only (primary for the survival curve to be performed on one phase of therapy and stranded since the TP53 classifier did not perform well on the polyA dataset
- Binned TP53 and telomerase scores in the following ways:
    1. **tp53_0.5** = TP53 high or low using a 0.5 cutoff
    2. **tel_0.5** = telomerase high or low using a 0.5 cutoff
    3. **tp53_strict** = TP53 stricter cutoff using <= 25th and >= 75th quantiles as low/high and in between == mid
    4. **tel_strict** = telomerase stricter cutoff using 25th and 75th quantiles as low/high and in between == mid
    5. **pheno_0.5** = TP53 & telomerase scores binned together using 0.5 as cutoff in both to be high or low
    6. **pheno_strict** = TP53 & telomerase scores binned together using 25th and 75th quantiles to determine high/low/mid
    7. **pheno_extremes** = TP53 & telomerase scores binned together using only 25th and 75th quantiles to determine high/low (left mid out of plot)

NA

No

KM plots and results tables

As expected high telomerase scores/activity corresponds to a lower survival across all tumor types. High TP53 scores also correlate with lower survival compared to lower scores in all PBTA and non-HGAT samples. However, all HGAT have poor survival, so we cannot see any trends with TP53/telomerase, so I took HGAT out of the full analysis.
In both PBTA all and non-HGAT PBTA, we observe higher TP53 scores result in a survival advantage. This has previously been reported in gliomas in "Association of Mutant TP53 with Alternative Lengthening of Telomeres and Favorable Prognosis in Glioma" https://cancerres.aacrjournals.org/content/66/13/6473. In this study, telomerase was measured using a PCR assay and TP53 mutations via sequencing. Here, we can confirm these previous results using classification of RNA-Seq.

#### Documentation Checklist


analyses/survival-analysis/util/quantile_calc.R
```diff
@@ -0,0 +1,52 @@
+# Functions for calculating quantiles of variables used for survival analysis
+
+# using 0.5 as a cutoff
+calc_groups <- function(meta_indep, tp53_quantiles, telomerase_quantiles){
+  meta_indep$tp53_0.5 <- ifelse(meta_indep$tp53_score > 0.5, "tp53 high",
+                                ifelse(meta_indep$tp53_score <= 0.5, "tp53 low",NA))
+  
+  # using 0.5 as a cutoff
+  meta_indep$tel_0.5 <- ifelse(meta_indep$telomerase_score > 0.5, "telomerase high",
+                               ifelse(meta_indep$telomerase_score <= 0.5, "telomerase low",NA))
```
jashapiro, line 10: The nested `ifelse`s in this function are difficult to follow, but can be simplified and made easier to follow with `dplyr` and specifically `case_when`. I am providing a partial example here, but the same changes could be made throughout this function. I think that all of the added columns could be included in a single call to `mutate`

```suggestion
 meta_indep <- meta_indep %>%
  dplyr::mutate(
    tp53_0.5 = dplyr::case_when(
      tp53_score > 0.5  ~ "tp53 high",
      tp53_score <= 0.5 ~ "tp53 low"),
    tel_0.5 = dplyr::case_when(
      telomerase_score > 0.5  ~ "telomerase high",
      telomerase_score <= 0.5 ~ "telomerase low")
    )
```

Note that the `NA` case is handled automatically.

analyses/survival-analysis/run_survival.sh
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash
 
-# K S Gaonkar
+# K S Gaonkar, J Rokita
 
-# Run fusion_filtering
+# Run survival-analysis
 
 set -e
 set -o pipefail
```
jashapiro, line 9: Adding this will allow the script to run with paths relative to its location (this directory).
```suggestion
# Set the working directory to the directory of this file
cd "$(dirname "${BASH_SOURCE[0]}")"
```

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+    p <- cowplot::plot_grid(surv_plot[[1]], surv_plot[[2]], nrow = 2,
+                         rel_heights = c(4, 2))
+    
+    # Print it out here
+    print(p)
+    
+    #Save the plot to a file. 
+    ggsave(filename = file.path(plots_dir, paste(prefix, each, "KM.pdf", sep = "-")), plot = p)
+    # Let's save the model for CI and median survival as well
+    readr::write_tsv(surv_median(kap_fit$model), file.path(results_dir, paste(prefix, each, "model.tsv", sep = "-")))
```
jashapiro, line 205: Looking at the output from this line, I am not sure what I am looking at, as there are a _lot_ of `NA` values. Is that because some categories just have few individuals? I suppose if survival is high, it may not be possible to calculate the upper bound?

jashapiro (commented):
I left a number of code comments below, but I want to also focus a bit on the overall goals of this analysis. The code here generates a large number of plots and statistics, but the reasons for all of the various divisions are not explained in the notebook, which may make it hard to interpret the results.

One thing to start with would be to add the following explanation in the notebook (from the PR):

> Binned TP53 and telomerase scores in the following ways:
> - tp53_0.5 = TP53 high or low using a 0.5 cutoff
> - tel_0.5 = telomerase high or low using a 0.5 cutoff
> - tp53_strict = TP53 stricter cutoff using <= 25th and >= 75th quantiles as low/high and in between == mid
> - tel_strict = telomerase stricter cutoff using 25th and 75th quantiles as low/high and in between == mid
> - pheno_0.5 = TP53 & telomerase scores binned together using 0.5 as cutoff in both to be high or low
> - pheno_strict = TP53 & telomerase scores binned together using 25th and 75th quantiles to determine high/low/mid
> - pheno_extremes = TP53 & telomerase scores binned together using only 25th and 75th quantiles to determine high/low (left mid out of plot)

With that explanation added, the question becomes which of these subdivisions is the "correct" one, or more generally, which should we be the most interested in? With so many different possible tests, we are definitely running into the potential for false positive results, so I think it is worth spending some words in the notebook justifying the analysis that we are running. It will be much easier to find that later than to track through the GitHub comments!

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+  output_stats(metadata=df_of_interest, 
+               model=model,
+               prefix=prefix)
+}
+
+```
+
+
+
+
```
jashapiro, line 234: Adding a session info chunk to aid reproducibility.

~~~suggestion
```{r}
sessionInfo()
```
~~~

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+                             test = "kap.meier",
+                             metadata_sample_col = "Kids_First_Biospecimen_ID_RNA"
+    )
+    
+    # median and 0.95CI survival days
+    kap_fit$model
+    
+    #`survminer` package comes with a convenient plotting function which uses `ggplot2` arguments. 
+    #It requires the original Kaplain-Meier `survfit` object, so we will extract that with `$model`.
+    surv_plot <- survminer::ggsurvplot(kap_fit$model,
```
jashapiro, line 176: Something in this chunk is failing CI with the following error: `Aesthetics can not vary with a ribbon` (and a bunch of anonymous function names, which are unhelpful).

I suspect it is related to the plotting here or below, but I haven't yet had a chance to check. I wonder if it is the `palette = each` line or something else? Running this notebook in the current docker image should allow us to figure out which specific part is failing.

analyses/survival-analysis/util/quantile_calc.R
jharenza: @runjin326 can you work on this piece

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+
+histologies <- readr::read_tsv(hist_file, guess_max = 10000) %>%
+  dplyr::filter(composition=="Solid Tissue" & tumor_descriptor == "Initial CNS Tumor") %>%
+  select(sample_id, broad_histology, short_histology, cancer_group, molecular_subtype, OS_status, OS_days, PFS_days) %>%
+  distinct()
+
+# read in tp53 scores
+tp53_scores <- readr::read_tsv(tp53_file)%>%
+  mutate(Tumor_Sample_Barcode = Kids_First_Biospecimen_ID_DNA) %>%
+  select(Kids_First_Biospecimen_ID_DNA, Kids_First_Biospecimen_ID_RNA, sample_id, tp53_score)
```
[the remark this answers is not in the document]
jharenza: >One thing to start with would be to add the following explanation in the notebook (from the PR):

Yes, I had not gotten to adding any of this to a readme yet - would you prefer this in the module README or within this notebook?  @runjin326 are you able to work on this?

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+```
+
+#### output statistics for paper writing 
+
+```{r}
+# first format the metadata to have appropriate OS_status column
+df_list <- lapply(df_list, format_meta)
+
+# define the formula
+model <- as.formula("survival::Surv(OS_days, OS_status) ~ pheno_0.5")
```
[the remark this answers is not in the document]
jharenza: yeah good point - @runjin326 can you adjust this to print all model stats? I think I had this one in the notebook because I was thinking we would use that as the main figure (to be discussed), but we should add all so we can also describe supplemental figures

analyses/survival-analysis/survival-analysis_histologies.Rmd
jharenza: This was taken from the original module, and yes you are correct. See [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/32a81d2b5b307c5473c5dbb4408c9a64de40e7d5/analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd#L129). I will reference this in the notebook.

analyses/survival-analysis/survival-analysis_histologies.Rmd
jharenza: I did not originally have this using 
```
palette = each
```
but rather had this as 
```
palette = "Paired", color = phenotype 
```
however, when looping, this change was required based on the documentation. This also made the labels pretty ugly, as you noted elsewhere. Open to how to improve the aesthetics for sure.

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+                                       xlim = c(0, 8000),
+                                       break.time.by = 1000,
+                                       ggtheme = theme_pubr(),
+                                       risk.table.y.text.col = TRUE,
+                                       risk.table.y.text = FALSE,
+                                       palette = each
+                                       #color = phenotype,
+                                       #palette = "Paired"
+    
+    )
```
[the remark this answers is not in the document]
jharenza: >I'm not sure how we intend to use these plots, but looking at them it seems to me we could do better than the default labels, which result in strings like pheno_0.5=tp53_high telomerase_high, which is hard to interpret.

yes, this was still a WIP :), but I agree. the labels are based on the `palette = each` - for some reason, I couldn't use the `each` as `color = each` as I did when doing a single plot and then using it in `palette` made it even uglier, but gave the desired printout.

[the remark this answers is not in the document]
jharenza: ok, I figured out updating to `color = pheno`

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+    
+    surv_plot$plot <- surv_plot$plot +
+      #ggtitle(paste0(ptitle)) +
+      theme(legend.position = "right")
+    
+    # Make this plot a meta_indep plot
+    p <- cowplot::plot_grid(surv_plot[[1]], surv_plot[[2]], nrow = 2,
+                         rel_heights = c(4, 2))
```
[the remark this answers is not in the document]
jharenza: Removed cowplot for grids

analyses/survival-analysis/survival-analysis_histologies.Rmd
jharenza: I updated this. Will see if CI passes

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+                                       risk.table.y.text = FALSE,
+                                       palette = each
+                                       #color = phenotype,
+                                       #palette = "Paired"
+    
+    )
+    
+    surv_plot$plot <- surv_plot$plot +
+      #ggtitle(paste0(ptitle)) +
+      theme(legend.position = "right")
```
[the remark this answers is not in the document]
jharenza: I changed this to default top, but may need some help with getting this to not cut off of the pages.

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,234 @@
...
+```
+
+
+Create groupings for scores
+```{r}
+meta_indep <- calc_groups(meta_indep, tp53_quantiles, telomerase_quantiles)
+meta_indep_no_hgat <- calc_groups(meta_indep_no_hgat, tp53_quantiles_no_hgat, telomerase_quantiles_no_hgat)
+```
+
+Set phenotypes to loop through
```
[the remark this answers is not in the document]
jharenza: agreed.

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,221 @@
...
+                                       risk.table.y.text = FALSE,
+                                       color = pheno,
+                                       palette = "Paired")
+    
+    #Save the plot to a file. 
+    pdf(file.path(plots_dir, paste(prefix, pheno, "KM.pdf", sep = "-")))
+    print(surv_plot, newpage = FALSE)
+    dev.off()
+    # Let's save the model for CI and median survival as well. Note: per this notebook, https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/32a81d2b5b307c5473c5dbb4408c9a64de40e7d5/analyses/survival-analysis/survival-analysis_HGG_DMG.Rmd#L129, NAs in the output are derived from skew in the underlying data.
+    readr::write_tsv(surv_median(kap_fit$model), file.path(results_dir, paste(prefix, each, "model.tsv", sep = "-")))
```
jharenza, line 190:
```suggestion
    readr::write_tsv(surv_median(kap_fit$model), file.path(results_dir, paste(prefix, pheno, "model.tsv", sep = "-")))
```

jharenza: Forgot to update this one

jharenza (commented):
Forgot this uodate

analyses/survival-analysis/survival-analysis_histologies.Rmd
jashapiro: Not yet... new error is more helpful though:
```
Error in grDevices::col2rgb(colour, TRUE) : 
  invalid color name 'pheno_strict'
``` 

I think removing `color = phenotype` should fix that.

runjin326: I am starting to work on this - I will make the change and push to see whether CI passes.

jashapiro: @runjin326 Great! For speed of testing, you will probably want to make sure that everything runs in the project Docker container before pushing: If the notebook works there it _should_ work in CI. Most times that isn't the case are usually related to file paths, which should not be an issue with this notebook.

runjin326: @jashapiro , thanks! Sure - then I will address all the comments, run in docker and then push to check CI. And if that does not work, we can look at the errors then and figure out a resolution. I will ping in once all the comments are addressed and pushed.

analyses/survival-analysis/survival-analysis_histologies.Rmd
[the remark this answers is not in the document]
jashapiro: Maybe both? The readme can cover overall approache, but details of what variable names mean, etc. should be in the notebook itself.

analyses/survival-analysis/survival-analysis_histologies.Rmd
jaclyn-taroni: Just to chime in here for a second – if this runs locally in Docker without issue and fails in CI, another source of failure could be the files that we use for testing. To help with debugging, you can snag the testing files: https://github.com/AlexsLemonade/OpenPBTA-analysis#working-with-the-subset-files-used-in-ci-locally

jharenza: > Just to chime in here for a second – if this runs locally in Docker without issue and fails in CI, another source of failure could be the files that we use for testing. To help with debugging, you can snag the testing files: https://github.com/AlexsLemonade/OpenPBTA-analysis#working-with-the-subset-files-used-in-ci-locally

Good point- my runs last night worked in docker so it may be the files!

runjin326:
@jashapiro , I have made the following updates:
1) Print out statistic results for all models 
2) Add a `README.md` for the module and added more description/comments in the notebook
3) Fixed the function in `quantile_calc.R` to be more succinct - please check to see whether it makes sense.

There are still some issues left:
1) Some legends still get cutoff in the figures, any suggestion as to how to fix that? 
2) When running `pairwise_survdiff` with HGAT only sample on `pheno_strict` and `pheno_extremes`, I got:
```
Error in survdiff.fit(y, groups, strata.keep, rho) : 
There is only 1 group
```
- However, I checked and there are always >1 groups in the df and I also tried to set the pheno column as factor before calling the function, none of them works. So I just used a `try` to wrap around the function and those two have error message output instead of real statistics results. Any idea what else I can try?

3) Another question is just clarification - @jharenza: previously for `pheno_strict`, other than 4 extreme calls, only `tp53_low telomerase_mid` and `tp53_mid telomerase_low` were defined and everything else was called `tp53 tel mid`. This would count `NA`, `tp53_high telomerase_mid` and `tp53_mid telomerase_high` as mid as well? Is this intentional? Currently, I modified it to be a combination of `tp53_strict` and `telomerase_strict` but feel free to let me know if that is not what you want and I can change it back.

I will now look into the files used in CI and try to figure out what the issue might be.

analyses/survival-analysis/util/quantile_calc.R
```diff
@@ -2,50 +2,59 @@
...
+      tel_0.5 = dplyr::case_when(
+        telomerase_score > 0.5  ~ "telomerase_high",
+        telomerase_score <= 0.5 ~ "telomerase_low",
+        TRUE ~ NA_character_),
+      
+      # for strict ones, 25th and 75th quantile is used
+      tp53_strict = dplyr::case_when(
+        tp53_score >= tp53_upper ~ "tp53_high",
+        tp53_score <= tp53_lower ~ "tp53_low",
+        tp53_score > tp53_lower & tp53_score < tp53_upper ~ "tp53_mid",
```
jashapiro, line 28: Pretty minor, but for consistency we might want to keep where the "or equal to" in a consistent direction (here and in `tel_strict` below).
```suggestion
        tp53_score >  tp53_upper ~ "tp53_high",
        tp53_score <= tp53_lower ~ "tp53_low",
        tp53_score > tp53_lower & tp53_score <= tp53_upper ~ "tp53_mid",
```

runjin326: Thanks! Updated and reran now.

jharenza:
> other than 4 extreme calls, only tp53_low telomerase_mid and tp53_mid telomerase_low were defined and everything else was called tp53 tel mid. This would count NA, tp53_high telomerase_mid and tp53_mid telomerase_high as mid as well? Is this intentional? 

Oh no this was not intentional- thank you for catching this!

jharenza:
@jashapiro i think this is ready for another look

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,235 @@
...
+```{r Set input file paths}
+hist_file <- file.path(data_dir, "pbta-histologies.tsv")
+indep_file <- file.path(data_dir,
+                        "independent-specimens.rnaseq.primary-plus-stranded.tsv")
+tp53_file <- file.path(analysis_dir, 
+                       "tp53_nf1_score/results/tp53_altered_status.tsv")
+
+# only use stranded since tp53 scores for polyA did not have good accuracy
+tel_stranded_file <- file.path(analysis_dir,
+                        "telomerase-activity-prediction/results/TelomeraseScores_PTBAStranded_FPKM.txt")
```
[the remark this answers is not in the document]
jashapiro: I preliminarily confirmed this with https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1208/commits/89a9f25cd30932e504f443226fef036388636ba0: 
When commenting out previous steps, this step passes CI.

jashapiro (commented):
I am working through a few tests, but I have a suspicion that the failing CI is due to previous steps changing the input files (either tp53 classifier or telomerase activity) required for this analysis. There are a few ways we could fix this, but I think the most productive may be to eliminate some of the KM tests in CI. We can do this by setting an environment variable in `run_survival.sh` and `.circleci/config.yml`, then using that to pass in a param to the notebook that might alter line 141 (`phenotypes <- c(...`) so that we only run a subset of the calculations.

Alternatively, we could modify the tp53 & telomerase steps to not overwrite the github versions of the files, but that may be less preferred. 

But I also want to return to some a fundamental questions that I think remain:

Why are we doing all of these alternative divisions of the data? 

I feel like we should perhaps reduce the number of KM tests to a smaller set with clear justifications. 

In particular, I might eliminate the `pheno_extremes` classificiation, as that is a subset of the `pheno_strict` classification. It be arguable that dropping the intermediates for plotting makes sense, but I worry that the statistics would no longer be appropriate. All of the pairwise comparisons in this group could be derived from the `pheno_strict` classification. (Since it looked in at least one case like this was the group where CI failed, this might have a secondary benefit; I am testing this as well).

I was also wondering if the `_0.5` classifications should be changed to split at the median, rather than 0.5, as that could result in more balanced groups, and might account for cases where the score is not well calibrated. Or perhaps this is not needed in the presence of the 3 way splits in the `_strict` classifications?

More generally, I would like to have some sense in this notebook of what the _primary_ grouping is. To my mind, it may be the `pheno_strict`, with everything else being a kind of subdivision of that. In that case I might want to see the statistics & KM model for that analysis presented first, on their own, with other plots maybe included later?

(One other thing that is nagging me is the generation of the `pheno` groups: I am not sure I understand why we are combining the tp53 and telomerase groupings into a single variable, rather than running the analysis as with those as two separate independent variables (aside from the fact that it doesn't look like the `survival_analysis()` function may not be able to handle that). I'll say I don't really know quite enough about survival analysis to say whether this is a problem, but I wanted to note it.)

analyses/survival-analysis/survival-analysis_histologies.Rmd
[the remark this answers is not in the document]
jashapiro: This is also working (less a timeout in a later step) with removing the `pheno_extreme` from the options tested. I don't know how durable that solution is, so we might want to implement something where we just test one of the phenotypes in the KM loop when in CI.

analyses/survival-analysis/survival-analysis_histologies.Rmd
```diff
@@ -0,0 +1,236 @@
...
+
+```{r}
+meta_indep <- calc_groups(meta_indep, tp53_quantiles, telomerase_quantiles)
+meta_indep_no_hgat <- calc_groups(meta_indep_no_hgat, tp53_quantiles_no_hgat, telomerase_quantiles_no_hgat)
+```
+
+Set phenotypes to loop through. These are columns created by the `calc_groups` function. These columns, as described above, contain information about whether the sample has `high`, `mid` or `low` TP53/telomerase scores.
+```{r}
+# phenotypes <- c("tp53_0.5", "tel_0.5", "tp53_strict", "tel_strict", "pheno_0.5", "pheno_strict", "pheno_extremes")
+phenotypes <- c("tp53_0.5", "tel_0.5", "tp53_strict", "tel_strict", "pheno_0.5", "pheno_strict")
```
jashapiro, line 142: I removed `pheno_extremes` here to (hopefully) pass CI, but I wonder if we really want it at all, as I discussed in my earlier comment.

runjin326:
@jashapiro - thanks for reviewing this! I think removing `pheno_extreme` is fine and I can implement using `median` instead of 0.5 for splitting. 
I think the idea behind all these pheno groups are to see which ones can get statistically significant survival results and I agree, `pheno_strict` can be the main one that we look at. 
And yes we can use multiple covariates for survival analysis (tp53 strict + telomerase strict) instead of using combined pheno strict column but I will need to modify the function a little bit. 
Maybe questions for @jharenza before I make any changes:
1) Should we use median instead of 0.5?
2) It is ok to drop `pheno_extremes`?
3) Should we re-write the notebook to center around `pheno_strict` and then add analysis for other pheno later on in the notebook?
4) Do we want to modify the function to take two separate covariates (`tp53_strict` and `telomerase_strict`) instead of one `pheno_strict`
I will start implementing these once we have a consensus as to what to do next.

jharenza:
> Should we use median instead of 0.5?

I'm not sure if we should do this because the median will be biased higher or lower based on the cancer type in some cases and not necessarily reflect a predicted classification phenotype, for example, in HGG, where most classifier scores are high and TP53 is altered. We would not expect balanced groups there. (This was also seen previously in Osteosarcomas, where TP53 is the major driver, hence almost all tumors had scores >0.7). This is why I had looked at HGG separately from the rest of the samples and had planned to have the main figure of all cancers but a supplement showing the HGG only and one with HGG removed confirming that with or without removal, the overall results are the same. 

My question here is more: should we use the top and bottom extremes and classify the rest as "mid" because we don't have the exact cutoff (0.5 was used previously in the original paper describing TP53 classifier and there wasn't one used for telomerase high/low), or just split at 0.5? 

Additionally, as you mentioned @jashapiro we lose a lot of data using only extremes but the middle samples may also be harder to interpret using that terminology. 


>It is ok to drop pheno_extremes?

Sure - I think we can subset out points for the plots using pheno_strict, which was essentially what was done currently

>Should we re-write the notebook to center around pheno_strict and then add analysis for other pheno later on in the notebook?

This may be ok if you also mean you'd keep the individual tel and TP53 analyses as well. 


>Do we want to modify the function to take two separate covariates (tp53_strict and telomerase_strict) instead of one pheno_strict
I will start implementing these once we have a consensus as to what to do next.

I'm not an expert in survival analyses, either, but I'm not sure these variables are independent as biologically, they are linked, which is why I tried to bin the samples into high for both, low for both, mid for both, or a mix or high/low/mid. Can you do some research on this @runjin326 - what constitutes covariates and how to analyze if they may be dependent on one another?

jashapiro:
> Maybe we also have to add the Cox proportional hazards function within this notebook for the covariates? https://www.datacamp.com/community/tutorials/survival-analysis-R

I think this may be the correct solution, as it would allow us to use the scores directly, rather than by turning them into a discrete variable. You don't get pretty KM plots out of it, but the stats are more appropriate to the data.

runjin326:
Thanks both for the input - I will do the following: 
1) keep 0.5 as a cutoff for now
2) drop pheno_extreme category and first plot out survival using `pheno_strict` and then the rest (e.g., tel 0.5 and tp53 0.5)
3) Research on what to do when two covariates might be inter-dependent when defining the model 
4) Implement CoxPH survival analysis and see how the results look (I can also work on making better figures from CoxPH models).
Will ping you once everything is pushed and ready for another look.

runjin326:
@jashapiro , sure thanks! I will hold off until a more solid plan is in place.

jaclyn-taroni:
Okay, I considered waiting and making this more detailed but I have a busy day tomorrow and better here, less detailed, where others can see it than only in my head/notebook!

From what's here and chatting with @jharenza, here's my understanding of the hypotheses we want to test:

- High TP53 classifier scores mean poorer OS 
- High telomerase activity scores mean poorer OS

We are also interested in whether _TP53_ alterations (where we will use the classifier score as our proxy) in the presence of high telomerase activity improves survival based on some literature.

We _also know_ that disease types have different OS.

We don't necessarily need to try to identify cut points to discretize the scores; we can instead use Cox regression.

As @jashapiro mentioned, we chatted about this a bit today and we came up with the following general plan.

We understand that it is common practice to do univariate analyses before you perform the multivariate analysis. We propose we do the following univariate analyses:

- _TP53_ classifier score (as a continuous variable)
- EXTEND score (as a continuous variable)
- Cancer group, removing cancer groups with less than 10 samples
- HGG vs. non-HGG

I understand from @jharenza that we are interested in demonstrating if a relationship exists between the _TP53_ scores and telomerase activity and OS in brain cancers. That can be accomplished with these univariate analyses (but we also may decide/find out that's not a good idea because we need to include disease label as a covariate!).

Finally, we can build a multivariate Cox regression analysis (_TP53_ classifier score, EXTEND score and whatever disease label we determine is appropriate). @envest has generously agreed to help us out with the specifics of setting up that analysis to make sure we're doing what we want to be doing! One of the ways I think about all of this, as a non-expert, is that we want to control for disease type. Then we can plan to include something like [a forest plot](https://cran.r-project.org/web/packages/survivalAnalysis/vignettes/multivariate.html#a-forest-plot) to display this information.

To summarize the biggest departures from what we're currently doing: 1. we won't "slice" by histology, we'll include it in the model instead 2. we won't try to discretize the scores

As far as moving forward with this plan, should folks agree: I think it might be easier to start a new branch and notebook (that would be my preference as a coder/analyst), but we should keep this branch around. 

I also want to mention that we don't need to worry about publication ready figures just yet. We have this notion of the separate `figures` directory because sometimes it's helpful to keep initial analysis and figure polishing separate. I think it's better to focus on getting the analysis right with the notebook I'm proposing and then we can polish the viz later!

jharenza:
Thank you @jaclyn-taroni! This all sounds good!

>As far as moving forward with this plan, should folks agree: I think it might be easier to start a new branch and notebook (that would be my preference as a coder/analyst), but we should keep this branch around.

@runjin326 when you start on the above, will you create a new branch? Thanks!

runjin326:
Yes - I will start working on this from a new branch.

jaclyn-taroni:
I'm going to close this to avoid confusion – we should be taking a look at #1212 now!

Issue #1210, opened 2021-12-03
Updated analysis: survival analysis generalization


runjin326:
#### What analysis module should be updated and why?
survival analysis that is currently in [PR 1208](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1208)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
1) Currently the script has non-HGAT and PBTA all commented out - need to add back as a loop function to run the analyses on all data frames
2) When looping across different data frames, the quantiles used for HGAT/PBTA and non-HGAT need to be taken into control
3) Add this function as well for output for table:
```
pairwise_survdiff(formula = Surv(OS_days, OS_status) ~ pheno_0.5, 
                  data = meta_no_hgat, 
                  p.adjust.method = "BH", rho = 0)
```

#### What input data should be used? Which data were used in the version being updated?
```
pbta-histologies.tsv
independent-specimens.rnaseq.primary-plus-stranded.tsv
tp53_altered_status.tsv
TelomeraseScores_PTBAStranded_FPKM.txt
```

#### When do you expect the revised analysis will be completed?
1-2 days.

#### Who will complete the updated analysis?
@runjin326

jaclyn-taroni:
I'm going to close this in favor of what's outlined in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1208#issuecomment-996356663

PR #1211, opened 2021-12-06
Modification of survival analysis WIP PR

runjin326:
This 

#### What was your approach?
1) Make the quantile calculation into a function so that we can use matching quantiles
2) Adding a for loop to iterate through all 3 data frames with their individual quantiles 
3) Add 2 more functions for generating the statistics for paper writing 

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1210

1) Note that since I cannot push to upstream `surv` branch - this PR is against the `surv` branch
2) Also, I think the additional changes we see is due to the fact that the current `surv` is not up to date with the master - should go away when upstream `surv` is in sync.

Please let me know if you want me to start a PR against the master instead etc.

Yes.

Tables and figures - the modifications of the code generated the same results as original PR. 
It is just a code change that loop all 3 dfs and their percentiles into the survival model.

N/A - see original PR.

#### Documentation Checklist


jaclyn-taroni:
I’m just going to merge this into the other branch and we can review at the same time.

PR #1212, opened 2021-12-21
Prelim results added for survival analysis 

runjin326:
This PR addresses the discussion we have in [this PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1208)

#### What was your approach?
Currently, the notebook only does univariate analysis for the following:
- TP53 classifier score (as a continuous variable)
- EXTEND score (as a continuous variable)
- Cancer group, removing cancer groups with less than 10 samples
- HGG vs. non-HGG

The cox regression cannot be plotted but the pvals were output.
And for categorical variables (point 3 and 4), the survival plots were generated. 

NA

I did not use the `survival_analysis` function in the notebook to make things easier - please check to see whether function to fit survival model and generate plots make sense.

1. Based on the results - it looks like disease label definitely have an effect on survival (as well as tp53 score and telomerase activity). What would we use as disease label for the multivariate analysis?
2. Any additional analysis that would be of interest? Should we set up the forest plot?

Yes.

Plots 
```
coxph_survival_per_hgg_group.png
coxph_survival_per_short_histology.png
```
	
Results
```
cox_reg_results_per_telomerase_score.tsv
cox_reg_results_per_tp53_score.tsv
log_rank_survival_per_short_histology.RDS
log_rank_survival_per_hgg_group.RDS
```

Looks like all 4 models that we looked at generated significant results. 

#### Documentation Checklist


jaclyn-taroni:
I'm going to take the first look at this and then I'll loop in @envest for thoughts on next steps!

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+  meta_data <- meta_data_list[[i]]
+  
+  # define model
+  model <- paste0("survival::Surv(OS_days, OS_status_fixed) ~ ", ind_var)
+  
+  # run survival analysis
+  fit <- survival::survdiff(formula(model),
+                            data = meta_data)
+  # Obtain p value for Chi-Squared stat
+  fit$p.value <- pchisq(fit$chisq, df = 1, lower = FALSE)
```
jaclyn-taroni, line 164: Seems like this might be a more general solution we can use in the future analyses ([ref](https://stackoverflow.com/questions/36368652/how-to-get-survdiff-returned-p-value/36369255)) ?

```suggestion
  fit$p.value <- pchisq(fit$chisq, df = length(diff$n) - 1, lower = FALSE)
```

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+
+```
+ 
+## Cox Regression 
+
+### TP53 scores & Telomerase score - results
+```{r Run Cox regression model for tp53 and telomerase}
+# run the model with different covariates
+for(ind_var in c("tp53_score", "telomerase_score")){
+  model <- paste0("survival::Surv(OS_days, OS_status_fixed) ~ ", ind_var)
```
jaclyn-taroni, line 112: Based on the `survival` docs, the ordering for the arguments is not super clear to me so that's a case for making this more explicit IMO

```suggestion
  model <- paste0("survival::Surv(time = OS_days, event = OS_status_fixed) ~ ", ind_var)
```

That's your intention, correct?

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+  # Make this plot a combined plot
+  surv_plot_coxph <- cowplot::plot_grid(plot_coxph[[1]], 
+                                        plot_coxph[[2]], 
+                                        nrow = 2, 
+                                        rel_heights = c(2.5, 2.5))
+  
+  print(surv_plot_coxph)
+  
+  # Save the plot
+  cowplot::save_plot(filename = file.path(plots_dir, paste0("coxph_survival_per_", ind_var, ".png")), plot = surv_plot_coxph)
```
jaclyn-taroni, line 193: Although I don't think we should worry much about polishing these plots, I think we should aim to get these to a point where they can be interpreted by (pull request) reviewers. They are a little smushed at the moment – I think changing the width and height could go a long way, but I might also try not making this a combined figure.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+  # Obtain p value for Chi-Squared stat
+  fit$p.value <- pchisq(fit$chisq, df = 1, lower = FALSE)
+  
+  print(fit)
+  # save the output
+  saveRDS(fit, file.path(results_dir, paste0("log_rank_survival_per_", ind_var, ".RDS")))
+  
+  # generate plots fit
+  fit_plot <- survfit(formula(model), data = meta_data)
+  plot_coxph <- survminer::ggsurvplot(fit_plot,
```
jaclyn-taroni, line 172: I think the name of this variable is somewhat misleading -- I would expect `coxph()` followed by `survfit()` rather than `survdiff()` and `survfit()`

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+
+## Import the metadata
+
+```{r Read in metadata}
+# read in histologies file
+indep_rna <- readr::read_tsv(indep_file)
+
+# remove cell lines by selecting only solid tissue
+histologies <- readr::read_tsv(metadata_file, guess_max = 10000) %>%
+ dplyr::filter(composition=="Solid Tissue" & tumor_descriptor == "Initial CNS Tumor") %>%
```
jaclyn-taroni, line 78: Can you add a comment about why we're filtering by `tumor_description == "Initial CNS Tumor"` here please? I kind of expect that this is equivalent to the primary list of RNA-seq specimens?

jaclyn-taroni (commented):
Thank you @runjin326! I've taken a first look and made some comments that are based on my knowledge of the project (e.g., what disease label to use) and what I found by reading the `survival` documentation. At this point, I will request a review from @envest. 

@envest, please take a look at the survival analysis to verify correctness. I will also direct your attention to this comment, where we explain a bit behind what we're hoping to accomplish with these analyses overall: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1208#issuecomment-996356663

Given what we're interested in and these results, can you advise us on how we should set up the multivariate analyses? Let us know if you have any questions and thanks very much for your help! 😄

analyses/survival-analysis/run_survival.sh
```diff
@@ -8,5 +8,6 @@ set -e
 set -o pipefail
```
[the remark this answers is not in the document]
runjin326: sure - fixed both.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+
+## Log Rank analysis 
+
+### Categorical variable cancer group and HGG or not 
+```{r}
+# separate into HGG or non-HGG
+meta_indep <- meta_indep %>% 
+  dplyr::mutate(hgg_group = case_when(
+    short_histology == "HGAT" ~ "HGAT",
+    short_histology != "HGAT" ~ "non-HGAT"
```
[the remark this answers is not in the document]
runjin326: I modified this to use cancer group - basically HGAT matches HGG, DIPG and DMG.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+  ))
+
+# filter to contain only samples in short histology that that have >=10 samples
+meta_indep_sh_filtered <- meta_indep %>% 
+  dplyr::group_by(short_histology) %>% 
+  dplyr::mutate(n_cg = n()) %>% 
+  dplyr::filter(n_cg >=10)
```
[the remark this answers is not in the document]
runjin326: We don't have to use a separate data frame - I modified to have a column showing only cancer group when the n_cg >=10. I did not use and subset though - I think NA would automatically be filtered and not used? But it needs to be confirmed.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+```{r}
+ind_var_list <- c("hgg_group", "short_histology")
+meta_data_list <- list(meta_indep, meta_indep_sh_filtered)
+
+for(i in 1:length(ind_var_list)){
+  ind_var <- ind_var_list[i]
+  meta_data <- meta_data_list[[i]]
+  
+  # define model
+  model <- paste0("survival::Surv(OS_days, OS_status_fixed) ~ ", ind_var)
```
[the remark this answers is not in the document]
runjin326: yes - corrected.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: I modified to `length(fit$n)-1` - should be correct now.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: yes - corrected.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+
+This notebook will do the following survival analysis
+1. Univariant analysis 
+- TP53 classifier score (as a continuous variable)
+- EXTEND score (as a continuous variable)
+- Cancer group, removing cancer groups with less than 10 samples
+- HGG vs. non-HGG
+
+2. Multivariant analysis
+- TP53 classifier score, EXTEND score and whatever disease label we determine is appropriate
```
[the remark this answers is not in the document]
runjin326: corrected.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Yes - this is equivalent to primary list of RNA-seq specimens. I think this filter is as a safe guard? Need input from @jharenza.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: This is actually just for plotting purposes - if I use `survdiff()` fit as the input for `ggsurvplot`, it is not taking it. So the results from coxph are stored as RDS but the plotting is largely Kaplan Meier. This might not be the right way to do it and feedback is appreciated!

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Yes - I have now split them into two plots, move the legends to the right for the survival plots and remove color legend for the risk table. Still not perfect but now we can see all the legends.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,202 @@
...
+# separate into HGG or non-HGG
+meta_indep <- meta_indep %>% 
+  dplyr::mutate(hgg_group = case_when(
+    short_histology == "HGAT" ~ "HGAT",
+    short_histology != "HGAT" ~ "non-HGAT"
+  ))
+
+# filter to contain only samples in short histology that that have >=10 samples
+meta_indep_sh_filtered <- meta_indep %>% 
+  dplyr::group_by(short_histology) %>%
```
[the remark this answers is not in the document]
runjin326: Yes - modified to cancer group now.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Oh sorry - I missed your point. Yes this is actually not coxph and it is log.rank. I renamed the plots now to match.
Additionally, I also add `hgg_group` and `cancer_group` to the coxph analysis and plotted out the results (although there are no groups to compare and just one survival line). For some reason, I was under the impression that coxph is best for continuous and log rank for categorical but now I included categorical variable in the for loop as well.

jaclyn-taroni: I'm not sure which is considered "better/best" for categorical data. I'm fairly convinced we will not be showing any univariate results based on what you have here – we'll have to go with multivariate which will necessarily be Cox.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
...
+  print(ggpar(plot_logrank[[2]], font.legend=6))
+
+  dev.off()
+
+  print(ggpar(plot_logrank[[1]], font.legend=6, legend="right"))
+  print(ggpar(plot_logrank[[2]], font.legend=6))
+}
+
+```
+
```
envest, line 214: It will be nice to keep the univariate analysis above, with the the potential modification of also testing broad histology as an additional univariate predictor. Here is where we can dive into the multivariate analysis, specifically with `tp53_score + telomerase_score + hgg_group` as the model parameters, or maybe tp53_score + telomerase_score + broad_histology` as an additional thing to try.

Once that model is fit, I like `survminer::ggforest()` as a solid visualization tool for Cox PH models.

As a reader, I would also be interested to know what the bivariate distribution of tp53_score and telomerase_score looks like, stratified by HGAT status. A 2D density plot with different colors for HGAT status comes to mind. 🤔

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
...
+  dev.off()
+
+  print(ggpar(plot_coxph[[1]], font.legend=6, legend="right"))
+  print(ggpar(plot_coxph[[2]], font.legend=6))
+}
+
+```
+
+## Log Rank analysis
```
envest, line 168: This is great to keep around for the exploratory evidence stage. The interpretation of logrank can become challenging with more than two groups, like in `cancer_group_over10`.

envest (commented):
I have read though this analysis and examined the various outputs. The progress on univariate analysis looks useful to me, as this exploratory analysis has revealed some potential issues around model convergence.

I think based on this work, I am confident that adding multivariate analysis with HGAT status will be the next big step tackle. I also left some comments about trying with broad histology instead of cancer group. I'm excited to stay involved and contribute to the conversation!

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
+---
+title: "Survival analysis by TP53 and telomerase activity"
+output: html_notebook
+---
+
+This notebook will do the following survival analysis
+1. Univariate analysis 
+- TP53 classifier score (as a continuous variable)
+- EXTEND score (as a continuous variable)
+- Cancer group, removing cancer groups with less than 10 samples
```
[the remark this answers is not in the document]
jaclyn-taroni: Thanks @envest! I was thinking we can use the groups that we use for broad histology display. More specifically, we can use the `broad_histology_display` in `figures/palettes/broad_histology_cancer_group_palette.tsv`, which lumps together broad histologies that do not have at least one cancer group with n >= 10 nested within them into `Other`, for a total of 12 different categories.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: @envest, I have now ran multivariate analysis for `tp53_score + telomerase_score + hgg_group` as well as `tp53_score + telomerase_score + broad_histology` and add `ggforest` for display. It looks like that second analysis had some issues with converging and the figure is not showing - advice on how to solve that will be appreciated.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
...
+### TP53 scores & Telomerase score - results
+```{r Run Cox regression model for tp53 and telomerase for all covariates}
+# run the model with different covariates
+for(ind_var in c("tp53_score", "telomerase_score", "cancer_group_over10", "hgg_group")){
+  model <- paste0("survival::Surv(time = OS_days, event = OS_status_fixed) ~ ", ind_var)
+  
+  fit <- survival::coxph(
+        formula(model),
+        data = meta_indep
+      )
```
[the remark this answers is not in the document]
runjin326: I kept the univiariate and the multivariate is added in the later part of the notebook.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
...
+## Import the metadata
+
+```{r Read in metadata}
+# read in histologies file
+indep_rna <- readr::read_tsv(indep_file)
+
+# remove cell lines by selecting only solid tissue
+histologies <- readr::read_tsv(metadata_file, guess_max = 10000) %>%
+ dplyr::filter(composition=="Solid Tissue" & tumor_descriptor == "Initial CNS Tumor") %>%
+ select(sample_id, broad_histology, cancer_group, molecular_subtype, OS_status, OS_days, PFS_days) %>%
```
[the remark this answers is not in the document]
runjin326: I am using years 1-14 in the survival plots.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,219 @@
...
+# remove samples without RNA-Seq since both classifiers' results use RNA-Seq
+# independent use primary plus rna-seq samples
+meta_indep <- tp53_scores %>%
+  left_join(tel_scores) %>%
+  left_join(histologies) %>%
+  dplyr::filter(Kids_First_Biospecimen_ID_RNA %in% indep_rna$Kids_First_Biospecimen_ID) %>%
+  # recode the categories
+  dplyr::mutate(OS_status_fixed = case_when(
+    OS_status == "LIVING" ~ 0,
+    OS_status == "DECEASED" ~1
```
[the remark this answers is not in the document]
runjin326: Sure - added the comment line.

analyses/survival-analysis/survival-analysis_histology.Rmd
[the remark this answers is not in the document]
runjin326: Thanks @jaclyn-taroni for the suggestion - I have now modified the code to use `broad_histology_display` as covariate.

runjin326:
@envest - thanks so much for the detailed review! I have made the suggested changes and commented/resolved some of your comments above. I think this is ready for another look! @jaclyn-taroni

jaclyn-taroni:
I notice that the step testing these changes was after the one that timed out, so in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1212/commits/4ad379100f07710967c24a517c9b1feaeab19f9f I moved it up so, if we still have the time out problem, we can at least understand if everything for these changes looks okay!

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,257 @@
...
+ select(Kids_First_Biospecimen_ID_DNA, Kids_First_Biospecimen_ID_RNA, sample_id, tp53_score)
+
+# read in telomerase scores
+tel_scores <- readr::read_tsv(tel_stranded_file) %>%
+ mutate(Kids_First_Biospecimen_ID_RNA = SampleID,
+        telomerase_score = NormEXTENDScores) %>%
+ select(Kids_First_Biospecimen_ID_RNA, telomerase_score)
+
+# remove samples without RNA-Seq since both classifiers' results use RNA-Seq
+# independent use primary plus rna-seq samples
```
jaclyn-taroni, line 100: Stating to check my understanding: so the `left_join(histologies)` here means that we're joining the matrices of scores/predictions (which contain all RNA-seq) and keeping all RNA-seq samples until we get to the filter step that uses `indep_rna`.

jaclyn-taroni: Are we going to be missing metadata (i.e., `OS_status` and `OS_years`) for the non-primary samples with this approach?

jaclyn-taroni (commented):
With the exception of some questions I still have around sample selection, all of my comments are addressed. I'll let @envest take it from here!

analyses/survival-analysis/survival-analysis_histology.Rmd
envest: Nice! The `hgg_group` model looks good to me. For `broad_histology`, the idea was to try to get fewer, larger groups in the model than we had with `cancer_group`. The model not converging means the groups are still too numerous and/or small such that we just don't have sufficient data in all parts of the model space. This could be a decision point -- either keep dropping small groups from the model until the model converges, or just focus on the HGAT/nonHGAT analysis.

envest: One thought I had looking at the forest plot... my preference would be to recode non-HGAT as the reference group to be able to see the change in hazard ratio for HGAT.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,257 @@
...
+  print(table)
+  
+  # output the plot
+  forest_coxph <- survminer::ggforest(fit, data = meta_indep)
+  print(forest_coxph)
+}
+
+```
+
+## Bivariate density plot of tp53 score and telomerase activity score
```
envest, line 241: Awesome -- I think this is really helpful to show why the multivariate model is necessary in addition to the univariate.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,257 @@
...
+  left_join(tel_scores) %>%
+  left_join(histologies) %>%
+  dplyr::filter(Kids_First_Biospecimen_ID_RNA %in% indep_rna$Kids_First_Biospecimen_ID) %>%
+  # recode the categories - 
+  # DECEASED maps to a survival event status of 1, LIVING maps to a censored observation with value 0
+  dplyr::mutate(OS_status_fixed = case_when(
+    OS_status == "LIVING" ~ 0,
+    OS_status == "DECEASED" ~1
+ )) %>% 
+  dplyr::mutate(OS_years = OS_days / 365.25)
```
envest, line 111: I think this helps a lot -- thank you!

envest (commented):
Great work on these changes -- it's especially nice to see the multivariate analysis come into form. The big decision after seeing these results is: does the team want to keep trying to model with cancer_group/broad_histology, or just focus on HGAT/nonHGAT? See my comments :) nice additions!

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Thanks for the input - I have now changed non-HGAT as the reference group and re-ran the analysis. 
Additionally, since the broad histology is not converting - I tried using another column called `oncoprint_group` - this categorizes all cancers into 4 groups and the results are included. Seemed to be informative as well! Also added a density plots stratified by these groups.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Since as you pointed out previously, for `histologies`, we filtered to `Initial CNS Tumor` and hence I would think the non-primary participants that only has non primary samples would be missed by this method. Also even if we do not filter to `Initial CNS Tumor` - using the primary independent samples would also filter to samples that are primary. The question is: do we want to keep non-primary?

runjin326:
A summary of the development:
1) Using `broad_histology_display` still does not converge - hence I added the `oncoprint_group` which seemed to be converging and generated informative results 
2) Density plot stratified by `oncoprint_group` is also included 

Things to discuss for next steps:
1) Do we want to only keep the `Initial CNS Tumor` during the histology file filtering step?
2) Does `oncoprint_group` make sense? If so, I can then remove `broad_histology_display` part. Alternatively, other suggestions for grouping them are welcome.

envest:
> A summary of the development:
> 
> 1. Using `broad_histology_display` still does not converge - hence I added the `oncoprint_group` which seemed to be converging and generated informative results
> 2. Density plot stratified by `oncoprint_group` is also included

The addition of `oncoprint_group` looks interesting -- according to the model, within the same `oncoprint_group`, `tp53` and `extend` are less important as predictors.

> Things to discuss for next steps:
> 
> 1. Do we want to only keep the `Initial CNS Tumor` during the histology file filtering step?
> 2. Does `oncoprint_group` make sense? If so, I can then remove `broad_histology_display` part. Alternatively, other suggestions for grouping them are welcome.

My feeling on `oncoprint_group` is: would this analysis make sense _a priori_ before looking at the data? If yes, then that's something to consider including. Unfortunately I am not up to speed on the biological implications of `oncoprint_group` for this project.

With the multivariate models and visualizations in place 👍 , I think it's best I leave it to @jaclyn-taroni to help wrap up / summarize next steps.

jaclyn-taroni:
To weigh in on the `oncoprint_group` discussion, I'm not sure that makes sense here or anywhere outside of the specific purpose it is used for – we expect it to only be used in display for Oncoprints, where individual cancer groups are _also displayed_ and when we have essentially curated lists of genes to display.

runjin326:
@jaclyn-taroni , thanks for the feedback! Yes - I tap into this column since the `broad_histology_display` groups are too granular for the multivariate analysis so I am trying to see whether there are even broader terms to use. Since this is not desired, should we just drop it and broad histology and only keep HGAT vs. non-HGAT for our final analysis?

jaclyn-taroni:
> Since this is not desired, should we just drop it and broad histology and only keep HGAT vs. non-HGAT for our final analysis?

Yea I think that sounds good @runjin326, thank you! Those comparisons both seem well-justified to me but only one of them (HGAT vs. non-HGAT) appears to have sufficient data.

runjin326:
@jaclyn-taroni, changes pushed - now the only question would be the sample selection portion.

jaclyn-taroni:
> now the only question would be the sample selection portion.

I can definitely see an argument for sticking with Initial CNS tumor only. In that case, I don't know why we need to use the primary plus list of independent specimens _unless_ there's no other way to consistently pick an Initial CNS Tumor specimen (if there are multiple) across different analyses without using that list.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: We want to use only primary - but instead of using independent RNA primary-plus sample list, I used `distinct` to directly take only one primary RNA if there are multiple there.

analyses/survival-analysis/survival-analysis_histology.Rmd
runjin326: Using a `distinct` call now.

runjin326:
@jaclyn-taroni, I see your point now! So I modified to not use independent RNA primary-plus list and instead, called `distinct(sample_id, .keep_all=TRUE` on the `meta-indep` after combining TP53 and telomerase scores. Please check to  see whether it looks good now.

jaclyn-taroni (approved):
Thanks for the changes in sample selection @runjin326! I had one comment about making the sample selection step more reproducible in the event that the ordering of samples in files changes. Once you make that change, please ping me and I'll get this merged. Thanks!

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+# read in tp53 scores
+tp53_scores <- readr::read_tsv(tp53_file) %>%
+ select(Kids_First_Biospecimen_ID_DNA, Kids_First_Biospecimen_ID_RNA, sample_id, tp53_score)
+
+# read in telomerase scores
+tel_scores <- readr::read_tsv(tel_stranded_file) %>%
+ mutate(Kids_First_Biospecimen_ID_RNA = SampleID,
+        telomerase_score = NormEXTENDScores) %>%
+ select(Kids_First_Biospecimen_ID_RNA, telomerase_score)
```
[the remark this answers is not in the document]
runjin326: Thanks @jaclyn-taroni for reviewing this - I have now sorted all df on `Kids_First_Biospecimen_ID_RNA` for the scores and on `sample_id` for histology.

analyses/survival-analysis/survival-analysis_histology.Rmd
```diff
@@ -0,0 +1,252 @@
...
+
+# combine tp53 scores and telomerase score to the dataframe
+meta_indep <- tp53_scores %>%
+  left_join(tel_scores) %>%
+  left_join(histologies) %>%
+  # remove samples without RNA-Seq since both classifiers' results use RNA-Seq
+  dplyr::filter(!is.na(Kids_First_Biospecimen_ID_RNA)) %>%
+  dplyr::distinct(sample_id, .keep_all=TRUE) %>%
```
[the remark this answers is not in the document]
runjin326: I also add an additional `arrange` on `Kids_First_Biospecimen_ID_RNA` here to arrange the df before running `distinct`.

runjin326:
@jaclyn-taroni - I have made corresponding changes. Please review!

jaclyn-taroni:
Thanks @runjin326 - looks good! I'll merge once CI passes.

PR #1224, opened 2022-02-16
Switch CircleCI image to Ubuntu 20.04

jaclyn-taroni:
Since we don't specify an image, we were using a Ubuntu 14.04-based image that's about to be deprecated. This should use a Ubuntu 20.04-based image. I don't foresee any issues, but we shall find out!

More context:

* https://circleci.com/blog/ubuntu-14-16-image-deprecation/
* https://circleci.com/docs/2.0/images/linux-vm/14.04-to-20.04-migration/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,13 +3,11 @@ jobs:
   run_analyses:
     machine:
       image: ubuntu-2004:202201-02
+      docker_layer_caching: true
     resource_class: large
     working_directory: ~/OpenPBTA-analysis
     steps:
       - checkout
-      
-      - setup_remote_docker:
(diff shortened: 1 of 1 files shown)


sjspielman:
The config file syntax and updated ubuntu image name look right! 

Let's see if this builds...

sjspielman (approved):

sjspielman:
Welp, time-out fail, as described in #1225

jaclyn-taroni:
Yes, but now when I hit re-run Docker layer caching should work.

sjspielman:
>Yes, but now when I hit re-run Docker layer caching should work.

🤞

Issue #1230, opened 2022-02-28
Updated analysis: Manuscript Figure 4 immune deconvolution panels


sjspielman:
The manuscript Figure 4 panels C/D currently contain temporary heatmaps showing immune deconvolution results. These figures need to be formalized and integrated into the manuscript figures.

#### Who will complete the updated analysis?

@sjspielman

sjspielman:
While going through the history of this module, I am not finding clear methods for _normalization_ as it relates to heatmap visualization.

Previously, heatmaps were created but no normalization was performed. The only calculation was to take the mean of each group going into the heatmap. XCell itself does not contain any normalization of scores after they are calculated.

The [XCell paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5688663/) itself barely describes this for their heatmaps, and does not provide any code to reproduce. In their MS, "Average scores for nine cell types across 24 cancer types from TCGA (The Cancger Genome Atlas). Scores were normalized across rows." (sic)

Therefore, we should actually implement normalization when updating figures, either based on mean or median per row grouping in the heatmap.

sjspielman:
Another note: 0's were removed from heatmaps before plotting, but AFTER means were calculated. This step will be removed as part of incorporating actual row-based normalization.

komalsrathi:
Hi @sjspielman I am _scaling by columns_ here https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/immune-deconv/util/heatmap_by_histology.R#L48-L51 and https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/immune-deconv/util/heatmap_by_molecular_subtype.R#L36-L39.  

@jharenza My expertise on normalizing these scores is limited - hence used the scaling function in pheatmap assuming that should be sufficient. I would let @sjspielman take this on and improve the module.

sjspielman:
Hi @komalsrathi, awesome I see the `scale` arg now! Fwiw, the method provided very little (no?) information provided on normalization schemes for this data, so the current idea (per chats with @jaclyn-taroni is we'll likely go with z-scores. I'll keep moving it along! :)

sjspielman:
Partially addressed with #1251. Closing in favor of #1271 which further updates figure 4.

Issue #965, opened 2021-03-25
Updated analysis: Add additional subtype figures to immune deconv


jharenza:
#### What analysis module should be updated and why?
immune deconv

The overall figures are nice, but by lumping histologies, we are not able to detect subtype differences within histology. We know this because of a recent analysis on medulloblastoma subtypes enabling us to see large differences between subtypes. I suggest we keep the figures we have, but add one heatmap figure per broad histology with rows by molecular subtype so that we can try to detect differences within histologies.

#### What changes need to be made? Please provide enough detail for another participant to make the update.
Add one heatmap figure per broad histology with rows being molecular subtype.

#### What input data should be used? Which data were used in the version being updated?
update to use v18 data

#### When do you expect the revised analysis will be completed?
one week

#### Who will complete the updated analysis?
@komalsrathi

komalsrathi:
@jharenza Can you confirm that there are only 7 broad histologies for which molecular_subtype info is available? Also can you check the attached plot? 

[heatmap_xCell_by_molecular_subtype.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/6224639/heatmap_xCell_by_molecular_subtype.pdf)

komalsrathi:
@jharenza I was initially using the `molecular_subtype` field from the `data/pbta-histologies.tsv` file, which worked fine as it  has a 1-1 mapping for `Kids_First_Biospecimen_ID` and `molecular_subtype`. When I instead switched to using the file under `molecular-subtyping-pathology/results/compiled_molecular_subtypes.tsv`, it seems to have duplicates (not sure if this is a known issue and how to tackle it):

```
   Kids_First_Biospecimen_ID        molecular_subtype
1                BS_5P1TN10Z       CNS Embryonal, NOS
2                BS_5P1TN10Z   HGG, H3 G35, TP53 loss
3                BS_BFDEZK1C       CNS Embryonal, NOS
4                BS_BFDEZK1C              DMG, H3 K28
5                BS_C6ZZ3FY5       CNS Embryonal, NOS
6                BS_C6ZZ3FY5   HGG, H3 G35, TP53 loss
7                BS_G16NM5WJ              DMG, H3 K28
8                BS_G16NM5WJ    EPN, To be classified
9                BS_J8VX4D17              DMG, H3 K28
10               BS_J8VX4D17    EPN, To be classified
11               BS_JXF8A2A6                      EWS
[...]

PR #1232, opened 2022-03-03
Immune deconvolution updated heatmap

sjspielman:

From Issue #1230, this PR uses `ComplexHeatmap` to visualize immune deconvolution scores, with a matrix of cell types across _samples_ (not groupings). The plot is therefore highly similar to that made for GSVA scores.

Figure is [here](https://github.com/sjspielman/OpenPBTA-analysis/blob/immune-deconv-figures/figures/pdfs/fig4/panels/immune_deconv_panel.pdf)

+ Do we want to use this plotting strategy for two plots? 
   + Some pros to this strategy - 
       + Uses the same legend as other panels
       + By plotting samples, we avoid [this situation](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/965) where there were highly distinct scores within histologies but across molecular subtypes. Not summarizing by histology group means we aren't pooling distinct molecular subtypes
       + We may be able to add another panel(s) to "zoom in" on some molecular subtypes or similar?
    + Some cons - 
        + Repetitive panels in the same figure (not sure this is a real problem)
+ The color palette
   + Scores should be "normalized" within cell type groupings (here, rows), but there is no obvious procedure for doing this. Currently, I have implemented z-scores. The problem is the color palette. Z-score distributions are _highly skewed_ with this five number summary - ` -1.902852956 -0.375620332 -0.139271522  0.002413132   47.912638453`. We can't have a color ramp from `[-2, 48]`....

#### Documentation Checklist


sjspielman:
Noting this failed because of CI time out.

Per discussions with @jaclyn-taroni, we may not want two heathap panels in figure 4. An alternative option that will still have the benefit of showing samples directly could be as follows - 
+ A faceted grid of cell types _that show high amounts of variation_ across cancer groups
+ X-axis is molecular subtypes, Y-axis is fraction (explore with(out) z-score normalization)
+ Jitter or (or box/violin with jitter overlay) geom(s)

sjspielman:
To explore this other plotting option, I have added [this figure](https://github.com/sjspielman/OpenPBTA-analysis/blob/e3af66fbed79f52128a36ebeed4e43da6a33a5ca/figures/pdfs/fig4/panels/temporary_box-jitter.pdf) which shows T cell CD8+ _and_ overall immune scores for EPN and HGG cancer groups, with molecular subtypes highlighted.

This plot does not seem very informative to me, except potentially to look at overall immune scores that xCell estimates. Feedback or thoughts on how to move this approach forward (or not?) would be great!

jaclyn-taroni:
I'd facet by `cancer_group` as well. But I think we should go about selecting what gets displayed a different way. My knee jerk reaction would be to subset to features with higher variance but it seems like we might expect scores to be on a different scale?

sjspielman:
I did make a grand viz to get a sense of overall variability within cell types (attached and aptly titled). There are a handful of cell types without much variance, but most have quite a bit of variance across some, but not all, subtypes. The molecular subtypes are proving tricky to integrate here, especially since sample size across them varies so widely. This is just colored by cancer groups, to get an overall sense of the variance in cell type scores.
[woah.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/8200072/woah.pdf)

jharenza:
Hi @sjspielman! Coming back to this. Now that #1243 is merged, can you recreate the tmp and "woah" figures with those cell types + the XCell immune score? 

Also, will you remove any subtypes which are "To be classified" and add MB subtypes to the EPN/HGG tmp figure? I can take a look at the patterning then and see if it makes sense with the literature.

Thanks!

sjspielman:
Yes, but..I think we should actually shut down this PR though and start with a fresh one focusing on xCell etc. The original heatmap I made for this PR was headed in the wrong direction and a cleaner PR will be easier for review I think. I'll link back to this discussion in a new one!

jharenza:
Oh yes, I abbreviated temporary as tmp here! Sorry, was writing it fast, and sounds good re new PR!

Issue #1237, opened 2022-03-08
Add scripts/modules for generating correlation plots for TP53 scores


runjin326:
#### What are the scientific goals of the analysis?
Add two correlation plots:
1) TP53 scores vs. NormEXTEND score
2) TP53 scores vs. breakpoint density 

#### What methods do you plan to use to accomplish the scientific goals?
Grab data from analysis modules or tables and plot the results out

#### What input data are required for this analysis?
NormEXTEND scores
TP53 scores
Breakpoint density of samples

#### How long do you expect is needed to complete the analysis? Will it be a multi-step analysis?
1 day

#### Who will complete the analysis (please add a GitHub handle here if relevant)?
@runjin326 

#### What relevant scientific literature relates to this analysis?
N/A

PR #1239, opened 2022-03-09
figure 5 correlation added

runjin326:
Adding tp53 vs. normEXTEND and tp53 vs. breaks density plots

#### What was your approach?
For tp53 scores vs. normEXTEND - 
1) table S3 from the supplementary tables is used 
2) results are joined using `Kids_First_Biospecimen_ID_RNA`

For tp53 scores vs. breaks density - 
1) breakpoint density data from chromosome instability module is used 
2) results are joined using `Kids_First_Biospecimen_ID_DNA`

https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1237

Please check to the see whether the legends, font, color etc for the figures look good.

No.

Yes.

2 figures in pdf format.

tp53 vs. NormEXTEND - R = 0.3, p <2.2e-16
tp53 vs. breaks density - R = 0.25, p = 1.2e-13

#### Documentation Checklist


sjspielman:
Hi @runjin326, thanks for getting started on this! The immediate thing I notice is, I'm concerned that lumping together cancer groups may not be the right way to go here. This is because, for example, there may be different relationships when comparing values cancer groups. I've seen some previous plots made where there are _negative correlations_ between This means the overall positive relationship is a sort of "paradox" resulting from combining different groupings, essentially [this](https://en.wikipedia.org/wiki/Simpson's_paradox) phenomenon.

Can you make these scatterplots faceted by cancer group as well? This will give us a sense of how pervasive this issue might be and we can go from there! 

In addition, for final plots, we'll want to make sure to use `ggpubr::theme_pubr()` as described [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/figures#overall-figure-theme).

runjin326:
@sjspielman thanks so much for the feedback. I tried running facet by cancer group but it looks like there are way too many cancer groups for this to be meaningful. Should we try either broad/short histology instead?
image

And for the theme part: do you mean the following? I.e., instead of theme_bw() just add `ggtheme = theme_pubr()` in the ggplot? Should I also remove the `theme(axis.text.x ...)` part?

```
ggplot(data = tp53_extend , 
       aes(x = NormEXTENDScores_fpkm, y = tp53_score),
       ggtheme = theme_pubr()) + 
  geom_point() + 
  xlab("NormEXTEND Scores") + 
  ylab("TP53 Scores") + 
  geom_smooth(method=lm, se=TRUE) + 
  stat_cor(method = "pearson", label.x = 0.05, label.y = 1, size = 6) + 
  theme(axis.text.x = element_text(size = 14), axis.title.x = element_text(size = 16),
        axis.text.y = element_text(size = 14), axis.title.y = element_text(size = 16))
```
cc: @jharenza

sjspielman:
Hi @runjin326, for faceting across cancer groups we want to focus on the names in the column `cancer_group_display`, not the `cancer_group` column. This information is [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/palettes/broad_histology_cancer_group_palette.tsv), so you will need to do some joining in there.

The other theme additions are fine - text size always needs to be tweaked a bit on a per site basis! The idea is just that the overall "look" is a pubr theme :)

runjin326:
@sjspielman - thanks so much for the feedback. The figures are updated now. Please check to see whether that is what you expected.

sjspielman:
This looks good faceted, thanks! One more quick change to the figures would be great - can we remove the `NA` cancer groups so there are no `NA` panels?

I think we might have an issue with the correlations themselves, though. Because we are now showing these separately, we may need to correct P-values for multiple testing. The `stat_cor()` function doesn't perform any corrections , so the P-values it returns are not right for this scenario. The correlation coefficients are right, just not the significance. 

One way to deal with this in the code is to calculate correlation coefficients and associated P-values separately (before plotting!), and correct those P-values using Bonferroni correction simply by multiplying P-values by the number of tests performed. Then, this information can be added to the plots with `geom_text()` from a separate data frame, _without_ using `stat_cor()`.

Let me know if this makes sense and/or how I can clarify anything!

runjin326:
@sjspielman - thanks so much for the feedback. The figures are modified accordingly. Please check the code to see whether the calculation was performed as what you expected, particularly the Bonferroni correction. I looked up online and seemed like the correction should be decided by the number of tests instead of multiply? I used division but please double check to see which one is correct.

sjspielman:
> I looked up online and seemed like the correction should be decided by the number of tests instead of multiply? I used division but please double check to see which one is correct.

It depends on how you are assessing significance. With multiple tests, we expect more false positives so we want to be _more stringent_ with our assessments. Therefore, we want p-values to _increase_ (because we expect fewer things are significant!) when you perform multiple comparisons, so you multiply P-values by the number of tests performed. Another perspective is to not change P-values but instead reconsider the _alpha significance threshold_ and lower it by _dividing_ by the number of tests, so that fewer P-values pass the threshold. In this case, we want to modify the P-values with multiplication, and I'll have a look at what you did!

runjin326:
@sjspielman - thanks! The problem with multiplying is that we then would have p value that is >1, which does not make much sense. Please let me know whether there is an alternative method for doing the correction.

sjspielman:
I've updated some of your code to ensure accuracy of the P-value calculation as well as specify _size_ of the output plots, so we can ensure they will always be the same size (7x7 inches for now) reproducibly!

These figures are mostly done, but now that we have corrected the P-values, there is very little that is still statistically significant. I'm not necessarily sure this should still be a main text figure or we should move to SI. Let's see what thoughts @jaclyn-taroni and @jharenza might have now that we don't see any salient trends here.

jaclyn-taroni:
> These figures are mostly done, but now that we have corrected the P-values, there is very little that is still statistically significant. I'm not necessarily sure this should still be a main text figure or we should move to SI.

Hm yea, let's change the output and script name to indicate that this is a supplemental figure for now. It is easy enough to "convert" to a main text figure if that's what we decide to do later.

jharenza:
> These figures are mostly done, but now that we have corrected the P-values, there is very little that is still statistically significant. I'm not necessarily sure this should still be a main text figure or we should move to SI. Let's see what thoughts @jaclyn-taroni and @jharenza might have now that we don't see any salient trends here.

I agree with @jaclyn-taroni on changing the output / script. I am not sure if we should use in SI either, but we can save as supplemental for now @runjin326. Thanks!

runjin326:
@jaclyn-taroni, @sjspielman, @jharenza - thanks so much for the feedbacks and updates! I have now moved the script and figures to the `supp` folder.

sjspielman:
Noting this wasn't merged because MRAN was down at the time. Coming back now to re-run CI and merge.

Issue #1231, opened 2022-02-28
Updated analysis: Manuscript Figure 5


sjspielman:
The manuscript Figure 5 needs to be finalized. This will entail, at a minimum:

+ Using a consistent color scheme for panels A-D 
+ Re-envisioning F-G with updated color scheme and current cancer groups and/or broad histologies
+ Removing panel H
+ Re-envisioning panel I to match F-G updates
+ Creating a new figure for survival analysis

#### Who will complete the updated analysis?

@sjspielman

sjspielman:
@jharenza, I'm wondering if I can get some feedback from you as I start this! In our current figure paneling outline (the google slides), I'm looking at figure 4D which is [this figure](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/plots/tp53_scores_vs_tp53_altered_status_Li-Fraumeni%20syndrome.png) in the repo.

I don't think we can/should use this figure for the manuscript, because one of the groups ("other") is only N=2, which means the statistical comparison is not reliable. Is there another biological grouping (this is Li-Fraumeni) you might recommend to put here for specific tp53 score comparisons? I can investigate and ensure appropriate sample sizes from there. Otherwise, I suggest we remove D altogether.

sjspielman:
@jharenza, I'm wondering if I can get some feedback from you as I start this! In our current figure paneling outline (the google slides), I'm looking at figure **5D** which is [this figure](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/plots/tp53_scores_vs_tp53_altered_status_Li-Fraumeni%20syndrome.png) in the repo.

I don't think we can/should use this figure for the manuscript, because one of the groups ("other") is only N=2, which means the statistical comparison is not reliable. Is there another biological grouping (this is Li-Fraumeni) you might recommend to put here for specific tp53 score comparisons? I can investigate and ensure appropriate sample sizes from there. Otherwise, I suggest we remove D altogether.

sjspielman:
Another question that has come up is why, in panels B/C, different groupings were used for visualization and statistical comparison. Specifically, B which plots expression does not consider "other" (the _vast_ majority of points in the data), but C does. When the "other" grouping is included for the expression comparison, p-values change substantially. For just a 2-group activated/loss, P=0.0035. With bonferroni for comparing 3 groups, the P-value jumps to 0.011: 


[...]

PR #1242, opened 2022-03-09
Re-run notebooks in `tp53_nf1_score` module

sjspielman:
As [discussed](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1231#issuecomment-1063351744), the [`tp53_altered_status.tsv`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/results/tp53_altered_status.tsv) file created in the `tp53_nf1_score` module has some slight numerical inconsistencies when re-run. It is likely that these minor discrepancies were caused by initially running the notebook producing this data file outside of Docker, because running within Docker is consistent. 

As none of the notebooks in this module contained a `sessionInfo()`, this PR has re-run all notebooks in this module to include that information. Along with this, the `TSV` file is updated with the Docker-run version, and some of the PDFs were updated with current aspect ratios (they may have also been out of sync).


jaclyn-taroni (approved):
I don't know that I expect all the quotes around `NA` values in that file, but the changes you're introducing to the Rmd files look fine!

sjspielman:
> I don't know that I expect all the quotes around NA values in that file
Oh, I hadn't noticed that, interesting. As long as they are still NA's!

Issue #1238, opened 2022-03-09
Updated analysis: Update immune deconvolution module to include xCell and quanTIseq


sjspielman:
#### What analysis module should be updated and why?

Currently, `immune-deconvolution` uses only the `xCell` method. This method reports relative scores without units and only allows for comparisons between samples, but _not_ between cell types (see #1234). 

We may want to have a complementary method, such as `quanTIseq`, which provides absolute scores that can be interpreted as fractions of the sample containing that cell type. Further, these scores can be compared _both_ across samples and cell types. 

#### What changes need to be made? Please provide enough detail for another participant to make the update.

Integrate `quanTIseq` into the `immune-deconvolution` in addition to `xCell`, using the same `immunedeconv` package. 

#### What input data should be used? Which data were used in the version being updated?
`v21`

#### When do you expect the revised analysis will be completed?

1-2 days

#### Who will complete the updated analysis?

@sjspielman

jharenza:
updated in #1243

PR #1243, opened 2022-03-09
Incorporate quanTIseq into immune-deconv module

sjspielman:

As discussed in issue #1238, this PR adds the `quanTIseq` method to the `immune-deconv` module.

#### What was your approach?

Slight modification of existing script to run deconvolution to export `RDS` files instead of `.RData` (for tracking different output names per method). We now call the deconvolution script twice from the analysis shell script, once per method. We also add a quick check into the R script to convert any provided method to all lowercase, which `immunedeconv` requires.

#1238 

Updated result files as RDS:

```
results/xcell_deconv-output.rds
results/quantiseq_deconv-output.rds
```

#### Documentation Checklist


analyses/immune-deconv/01-immune-deconv.R
```diff
@@ -87,13 +91,11 @@ deconv <- function(expr_input, clin_file, method) {
...
 combo <- expand.grid(expr_input, deconv_method, stringsAsFactors = F) 
 deconv_output <- apply(combo, 1, FUN = function(x) deconv(expr_input = x[1], clin_file = clin_file, method = x[2]))
 deconv_output <- do.call(rbind.data.frame, deconv_output)
 
-# save output to RData object
-print("Writing output to file..")
-save(deconv_output, file = output_file)
-print("Done!")
+# save output to RDS file
+write_rds(deconv_output, output_file)
```
jaclyn-taroni, line 101: To check my understanding - the `deconvolute()` steps are run for each dataset separately, but then we're putting them into the same data frame. For xCell, the scores don't change depending on whether the data are together or not (see discussion on #292). I don't know that the same is true of `quantiseq`, but what I'm actually getting at is that our result almost implies that we ran them together when that's not the case. I think this (rambling) comment could be addressed by adding a column in the resulting data frame that keeps track of which input dataset the sample comes from.

analyses/immune-deconv/README.md
```diff
@@ -18,6 +21,9 @@ xCell outputs immune scores as arbitrary scores that represent cell type abundan
 Importantly, these scores may be compared between samples (inter-sample comparisons), but _may not_ be compared across cell types or cancer types, as described in the [`immunedeconv` documentation](https://icbi-lab.github.io/immunedeconv/articles/immunedeconv.html#interpretation-of-scores). This is in part because xCell is actually a signature-based method and not a deconvolution method, as is described in the [xCell Publication](https://doi.org/10.1186/s13059-017-1349-1):
 > Unlike signature-based methods, which output independent enrichment scores per cell type, the output from deconvolution-based methods is the inferred proportions of the cell types in the mixture.
 
+Therefore, we also use `quanTIseq` as a complementary method. Although `quanTIseq` looks at fewer cell types, the scores can be interpreted as absolute fractions, thereby allowing comparison _both_ across samples and cell types.
```
jaclyn-taroni, line 24: I'd link the documentation on interpretation of scores again for good measure.

jaclyn-taroni (commented):
The changes you're adding are very reasonable! The one comment that is preventing me from approving yet is about whether we should go a touch furthe.

analyses/immune-deconv/01-immune-deconv.R
sjspielman: :100:

jaclyn-taroni (approved):
👍🏻

PR #1244, opened 2022-03-11
Workflow panel for figure 1

namansri468:

This pull request is for the manuscript issue https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/83 to include a panel in figure 1 showing the workflow process. The icons used are all CCBY.

Some places to look:
+ Is the overall workflow correctly represented?
+ Is there a different color scheme we might want to use?
+ Is anything critical missing from the figure?


jharenza:
Hi @sjspielman and @namansri468! Thanks for working on this. My one minor comment is that perhaps we should stay consistent with the repository and manuscript with text capitalization for "OpenPBTA" rather than "OPENPBTA". Thoughts?

sjspielman:
Closing this PR after some in-house discussions about the direction for this figure. A new PR will be opened later.

Issue #1253, opened 2022-03-21
Create updated panels for Fig 3


jharenza:
Now that Figure 3 is finalized (see [here](https://docs.google.com/presentation/d/1bgaDcMaGgwNEGJl46Lg03FoF4EnS4XLjdp5Ji_ufGsI/edit#slide=id.ge9d4ddd4cc_0_15)), we should save the CNS mutational signatures figure as a panel for Figure 3C in https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/figures/pdfs/fig3/panels.

Do we also need a script to create this figure? It isn't immediately clear to me how this is now being done.

**Who will complete this?**

@sjspielman ?

sjspielman:
In this case it's appropriate to just have a line in [this script](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/generate-figures.sh) copying the figure over, since the figure in the module is pub-ready. This will be quick, can do it!

sjspielman:
Closed with #1256

PR #1256, opened 2022-03-22
Copy figure 3c into final directory

sjspielman:
Per Issue #1253, this PR copies the mutational signatures sina/IQR plot into the final location in the `figures/` directory. No plotting code itself is changed, only the `generate_figures.sh` script.


figures/generate-figures.sh
```diff
@@ -143,3 +143,11 @@ fi
 Rscript -e "rmarkdown::render('${analyses_dir}/cnv-chrom-plot/cn_status_heatmap.Rmd',
                               clean = TRUE, params = list(final_figure=TRUE))"
 
+
+
+######## Mutational signatures
+# Copy the figure to final directory
+cp ${analyses_dir}/mutational-signatures/plots/cns/exposures_sina_IQR.pdf  pdfs/fig3/panels/mutational_signatures_exposuures.pdf
```
jharenza, line 150:
```suggestion
cp ${analyses_dir}/mutational-signatures/plots/cns/exposures_sina_IQR.pdf  pdfs/fig3/panels/mutational_signatures_exposures.pdf
```

jharenza (approved):
just a quick typo fix and looks good!

jaclyn-taroni (approved):
My only comment, and I think it's a future thought, is maybe we should reorganize this shell script a bit so everything in the same figure is grouped. 🤷🏻

sjspielman:
> My only comment, and I think it's a future thought, is maybe we should reorganize this shell script a bit so everything in the same figure is grouped.

I also thought this. I'll put in an issue.

Issue #1255, opened 2022-03-21
Create additional panels for Figure S3


jharenza:
Only the `primary_only_other_supplemental_oncoprint.pdf` panel  is represented [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/figures/pdfs/supp), so the two CN landscape and chromothripsis panels should be added per the slide deck [here](https://docs.google.com/presentation/d/1bgaDcMaGgwNEGJl46Lg03FoF4EnS4XLjdp5Ji_ufGsI/edit#slide=id.ge9d4ddd4cc_0_30).

Figures used for panels can be found:
- B: @jaclyn-taroni  and @cansavvy I cannot find this one at the moment - can you help?
- C: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/chromothripsis/plots/04-breakpoint-data/count_chromothripsis_cnv_breaks_stripchart.png
- D: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/chromothripsis/plots/04-breakpoint-data/count_chromothripsis_sv_breaks_stripchart.png

Who will complete this?
@sjspielman ?

sjspielman:
Yup, I got it!

sjspielman:
I've tracked down panel B here. It was last run with v20, so we might want to update. - https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/cnv-chrom-plot/plots/cn_status_heatmap.png

sjspielman:
Closed by #1259

PR #1259, opened 2022-03-22
Added S3 figure panels B, C, D

sjspielman:
This PR address issue #1255 and creates three PDF panels B, C, and D to use in Figure S3.

Figure S3B is created by copying over the PDF from the relevant module. Therefore that module was also updated to export a PDF image, and its notebook was rerun to ensure v21 data.

Figures S3C-D were created directly in the script added here, where I used the simpsons color scale to color the jitter points for each # chromothripsis regions.


sjspielman:
Note - I am now seeing that the intended FigS3B had some png changes when I re-ran the notebook related to the _palette_. Investigating this now.

jaclyn-taroni:
Yea, I was just headed to chime in about the palette when I saw your edited comment @sjspielman. I think you'd want to use `broad_histology_display`, but perhaps drop the samples where `broad_histology_display == "Other"` because then we'd be lumping disparate diagnoses together and I'm not sure that's what we want to do here.

jaclyn-taroni:
> So, actually, are we sure about cancer group? S2 apparently relates back to Figs 2-3, which focus on cancer groups?

No, I'm not sure. I just thought `broad_histology_display` was more analogous to what was used previously. But if we're going "less module the better," then we probably will have a much easier time making that call/making tweaks. I'm fine with `cancer_group` being what we try first.

sjspielman:
@jaclyn-taroni @jharenza This is ready for a re-review. The heatmap (Fig S3B) code is all in `figures/scripts/` without touching the original module, and heatmap uses cancer groups, where "Other" is on the bottom. 

The other chromothripsis S3C-D plots are still as they were.

jharenza (changes requested):
Thanks for this @sjspielman! I like the new cancer group figure for B, and I agree with @jaclyn-taroni on dropping the "Other" samples from this plot.

For C-D, the color changes look like they match the groupings on the x-axis, which is what the original figure had, but the original figure was set up to be colored as a gradient with the highest number of chromothripsis regions being the darkest. Otherwise, I don't think colors are needed in the plot.

sjspielman:
> I like the new cancer group figure for B, and I agree with @jaclyn-taroni on dropping the "Other" samples from this plot.

👍 

>  but the original figure was set up to be colored as a gradient with the highest number of chromothripsis regions being the darkest.

I wasn't sure about this palette, actually. [Here](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/figures#figure-guidelines) it describes we should only use the gradient palette for numeric data, not for groupings like this. @jaclyn-taroni , any thoughts?

jaclyn-taroni:
Coloring the points in those plots seems a little like duplicating information to me? I suppose the color is somewhat helpful matching between the two panels, but there are not a ton of groups to begin with.

jharenza (approved):
LGTM!

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+
+
+# Read in metadata and merge with colors
+metadata <- read_tsv(file.path(data_dir, "pbta-histologies.tsv"), guess_max = 10000) %>%
+  # Join on the colors 
+  left_join(histology_label_mapping, by = "cancer_group") %>%
+  # Select what is needed
+  select(contains("cancer_group"), 
+                Kids_First_Biospecimen_ID, 
+                tumor_ploidy) %>%
```
jaclyn-taroni, line 56: Very minor point, but this indentation is off.

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+  filter(cancer_group != "Other")
+
+
+### Set up chromosomal sizes for making bins. 
+chr_sizes <- read_tsv(file.path(data_dir, "WGS.hg38.strelka2.unpadded.bed"),
+                             col_names = c("chrom", "start", "end")
+) %>%
+  # Reformat the chromosome variable to drop the "chr"
+  mutate(chrom = factor(gsub("chr", "", chrom),
+                               levels = c(1:22, "X", "Y", "M")
```
jaclyn-taroni, line 67: And so is this for that matter

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+  )) %>%
+  # Remove sex chromosomes
+  filter(!(chrom %in% c("X", "Y", "M")))
```
jaclyn-taroni, line 70: This all looks consistent with https://github.com/sjspielman/OpenPBTA-analysis/blob/3655669f6363494a6281288919c4db20a65b1aa2/analyses/cnv-chrom-plot/cn_status_heatmap.Rmd#L205-L213

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+
+
+### Make row annotation object
+# Get the histologies for the samples in this set and order them by histology
+
+# Figure out cancer group counts, where NA display groups are _excluded_
+cancer_group_counts <- metadata %>%
+  drop_na(cancer_group_display) %>%
+  # No other ? TODO!
+  #filter(cancer_group_display != "Other") %>%
```
jaclyn-taroni, line 145: Can probably take this out now, since you've filtered `metadata` to remove `"Other"` already.

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+cancer_color_key <- cancer_color_key_df$cancer_group_hex
+names(cancer_color_key) <- cancer_color_key_df$cancer_group_display
+
+# Get coordinate start positions
+hist_start <- match(names(cancer_color_key), samples_for_heatmap$cancer_group_display)
+
+# Get coordinate end positions for each histology group
+hist_end <- hist_start + summary(samples_for_heatmap$cancer_group_display)
+
+# Get mid points of
```
jaclyn-taroni, line 199: Looks like this was incomplete in the original notebook, but taking my cue from the chromosome midpoints I think this is right...
```suggestion
# Get mid points of histology group for labels
```

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
...
+names(cancer_color_key) <- cancer_color_key_df$cancer_group_display
+
+# Get coordinate start positions
+hist_start <- match(names(cancer_color_key), samples_for_heatmap$cancer_group_display)
+
+# Get coordinate end positions for each histology group
+hist_end <- hist_start + summary(samples_for_heatmap$cancer_group_display)
+
+# Get mid points of 
+mid_points <- floor((hist_start + hist_end) /2)
```
jaclyn-taroni, line 200: I also assume that it's fine to overwrite `mid_points` from above since that `anno_mark()` step has already taken place.

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -0,0 +1,321 @@
+# Stephanie J. Spielman for CCDL, 2022
```
jaclyn-taroni, line 1: Should we start adding more structure to the `figures/pdfs/supp` directory you think, e.g., adding a subdirectory for each supplemental figure?

jaclyn-taroni: I think filing an issue about this would address this concern.

jaclyn-taroni (approved):
I had only minor points that should probably get addressed before merging and a future thought. No need for me to take another look!

figures/scripts/supp-S3-panels-BCD.R
sjspielman: I am _here for_ this kind of minor point!

figures/scripts/supp-S3-panels-BCD.R
sjspielman: Yes, they were named the same in the original code so kept it here.

figures/scripts/supp-S3-panels-BCD.R
sjspielman: Yes, I agree. I'll file an issue.

figures/scripts/supp-S3-panels-BCD.R
sjspielman: Noting also `tumor_ploidy` doesn't actually need to be kept, so removed from `select()` statement.

Issue #1252, opened 2022-03-21
Updated analysis: oncoprint counts for manuscript


sjspielman:
#### What analysis module should be updated and why?

Related to [MS PR #199](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/199), we need to create a notebook in `oncoprint-landscape` module tabulating certain counts that are slated to appear in the manuscript. The results in this notebook can then be used to populate the respective manuscript sections.

jharenza:
Thanks for opening this @sjspielman! 

I think this addition might be a bit complicated. 

Since we highlight specific cancer group statistics in the manuscript, the goal would be to generate cancer_group specific tables for `*primary_only_*_oncoprint_summary_n.tsv` files. This would entail creating maf subset files and output tables for every cancer group that is represented in the oncoprints currently, including those in the supplemental oncoprint.

Would we want to make a new shell script and notebook for this since it will add a lot to the current workflow?

jaclyn-taroni:
I will pick this up and make the determination about how to implement it along the way

sjspielman:
Closed by #1266

PR #1266, opened 2022-03-24
Count alterations represented in oncoprint on a per cancer group basis

jaclyn-taroni:

We're reporting alterations per cancer group in https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/199, but we didn't yet have a tabular version of that data. What I'm adding here creates the equivalent tables [`analyses/oncoprint-landscape/tables`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/f68714c7a6884a08566bf8f0f17bea2239f3ecef/analyses/oncoprint-landscape/tables)  for an individual cancer group.

#### What was your approach?

Here I'm making the following changes:

* Adding a step to `analyses/oncoprint-landscape/00-prepare-goi-lists.R` where I create a table (`analyses/oncoprint-landscape/data/cancer_group_goi_list_mapping.tsv`) that contains the mapping between `cancer_group` and what genes of interest (GOI) list to use. We only create the tables for _all samples in the `broad_histology`_ when a GOI list is used, so I mirrored that earlier design decision.
* Adding a new script `analyses/oncoprint-landscape/04-alteration-counts-by-cancer-group.R` which does the following:
  * Reads in MAF, CNV, and fusion files prepped earlier -- in practice, this will be a primary only or primary plus set of files and that's how it works in the shell script right now
  * Uses the table described above to read in the GOI list and create a summary of the MAF that is subset to only the samples in that cancer group and the genes in the GOI list
  * Saves that summary to `analyses/oncoprint-landscape/tables/cancer_group_counts` where the subdirectory is determined by whether this is for primary only or primary plus (dictated by the shell script)
* Reworks `analyses/oncoprint-landscape/03-oncoprint-n-count-table.R` to reduce redundant code for primary only vs. primary plus + removes the need to pass broad histology info to that script
* Updates the bash script for the module to reflect these changes, had to make the broad histologies the inner loop.
* Also change to one of the Embryonal oncoprints? To my eye, it only looks like a change in the `cancer_group` palette. If true, this doesn't matter. The pub-ready plots are made in `figures/`.

#1252

* For now, I've only committed the resulting files for DMG otherwise the diff would be 🍌. Please take a close look at those. Or, tell me which cancer group I should use an example instead!
* Does that PNG change look okay?

Yes, but note that I am not adding all of the resulting tables yet.

#### Documentation Checklist


jharenza:
Thanks @jaclyn-taroni - can you print out either ependymoma or craniopharyngiomas? It looks like I did not separate out DMG within the all HGG oncoprint, but did for these other two.

jaclyn-taroni:
Yup, added EPN in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1266/commits/f714412a41b9edd248fbb37c48e5fa8d9a3217ea

analyses/oncoprint-landscape/tables/primary-plus_sample_n_in_oncoprint.tsv
```diff
@@ -0,0 +1,5 @@
+broad_histology	n
```
jaclyn-taroni, line 1: Noting here that these numbers are the same as in `analyses/oncoprint-landscape/tables/sample_n_in_oncoprint.tsv`, which is being deleted here.

analyses/oncoprint-landscape/tables/primary_only_sample_n_in_oncoprint.tsv
```diff
@@ -0,0 +1,5 @@
+broad_histology	n
```
jaclyn-taroni, line 1: Noting here that these numbers are the same as in `analyses/oncoprint-landscape/tables/sample_n_in_oncoprint.tsv`, which is being deleted here.

jaclyn-taroni (commented):
Returning some comments for reviewers

jaclyn-taroni:
I'm going to mark this ready for review now that I've updated the documentation. I am not going to add all the resultant tables until the code has been reviewed. We're using a `maftools` function to do the summary, which is to say that I don't really expect (🤞🏻) any edge cases to pop up that wouldn't also pop up when we made the plots.

jharenza (approved):
>Also change to one of the Embryonal oncoprints? To my eye, it only looks like a change in the cancer_group palette. If true, this doesn't matter. The pub-ready plots are made in figures/.

Yep, looks like palette only to me as well. Those are the ETMR samples.

@jaclyn-taroni this looks good and as expected to me! Thanks!

analyses/oncoprint-landscape/03-oncoprint-n-count-table.R
```diff
@@ -101,136 +67,88 @@ option_list <- list(
...
-}
-
-# write out n number entering oncoprint
-bind_rows(oncoprint_n_table_pp, oncoprint_n_table_po) %>% 
-  readr::write_tsv(file.path(results_dir, "sample_n_in_oncoprint.tsv"))
+fusion_df <- readr::read_tsv(opt$fusion_file)
+
+#### Set up oncoprint counts ---------------------------------------------------
+
+other_cns_histologies <- c(
```
sjspielman, line 94: Agrees with https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/oncoprint-landscape/03-oncoprint-n-count-table.R#L170

analyses/oncoprint-landscape/03-oncoprint-n-count-table.R
```diff
@@ -101,136 +67,88 @@ option_list <- list(
...
+    other_cns_histologies,
+    "Low-grade astrocytic tumor",
+    "Embryonal tumor",
+    "Diffuse astrocytic and oligodendroglial tumor"
+  )) %>%
+  pull(Tumor_Sample_Barcode)
+
+# Filter data files to only include the relevant samples & then pull the IDs
+# We're only counting IDs that are present in the relevant file using this
+# approach.
```
sjspielman, line 123: In an ideal world we'd probably want a function to handle the next three lines of code, but lines are so small this fine.

analyses/oncoprint-landscape/03-oncoprint-n-count-table.R
```diff
@@ -101,136 +67,88 @@ option_list <- list(
...
+  "Other astrocytic tumor",
+  "Lymphoma",
+  "Melanocytic tumor",
+  "Other tumor"
+)
 
+tumor_barcodes_to_include <- metadata %>%
+  filter(broad_histology %in% c(
+    other_cns_histologies,
+    "Low-grade astrocytic tumor",
```
sjspielman, line 115: Might be worth coding this other array up earlier when `other_cns_histologies` is defined

analyses/oncoprint-landscape/04-alteration-counts-by-cancer-group.R
```diff
@@ -0,0 +1,209 @@
...
+
+# This module
+module_dir <- file.path(root_dir, "analyses", "oncoprint-landscape")
+
+# Source the custom functions script
+source(
+  file.path(
+    root_dir,
+    "analyses",
+    "oncoprint-landscape",
```
sjspielman, line 42:
```suggestion
```

Issue #1268, opened 2022-03-25
Update mutational signatures supplementary tables


sjspielman:
Related to PR https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1265/ 
The path to the fitted mutational signatures file currently used to build the supplementary tables needs to be updated from ` fitted_exposures_signal-cns_sigfit.rds` to  `fitted_exposures_signal-cns_deconstructSigs.rds` to match the analysis reported in the main text.

### Who will do it?
@sjspielman 

### How long to complete? 

1 day

PR #1269, opened 2022-03-25
Update table generation notebook and excel files

sjspielman:

From issue #1268 , supplementary tables for mutational signatures had been using the wrong file. The correct file is now used and results are re-generated.
Several other changes were implemented in this notebook:
- Updated other formatting aspects to increase code modularity
- Updated incorrect file paths to TMB files in `data/`
- Knitting the notebook _in the docker_

NA

NA

yes 

Updated output tables to use in manuscript

#### Documentation Checklist


jaclyn-taroni:
`generate_figures.sh` would need to be modified to be able to run on a machine with 8GB to work on CI and, presumably, many of the modules that get re-run in that script are already tested. To me, this scenario seems different because I expect it will run on a machine that size without issue (but I live to be wrong about such matters).

jaclyn-taroni:
@sjspielman we need to use a param to be able to have this work in CI. I think we can make it more general than in CI y/n – we could instead make it the release where the default is the current release. In CI, perhaps you'd pass `"testing"` and if the param is `"testing"`, construct the path to be `data/<file>` instead of `data/<release param>/<file>`

tables/output_tables.Rmd
```diff
@@ -13,132 +12,168 @@ library(openxlsx)
...
+  rename(Kids_First_Biospecimen_ID_RNA = SampleID,
                 NormEXTENDScores_counts = NormEXTENDScores)
 
-telomerase_scores_stranded_fpkm <- readr::read_tsv("../analyses/telomerase-activity-prediction/results/TelomeraseScores_PTBAStranded_FPKM.txt") %>%
-  dplyr::select(SampleID, NormEXTENDScores) %>% 
-  dplyr::rename(Kids_First_Biospecimen_ID_RNA = SampleID,
+telomerase_scores_stranded_fpkm <-
+  read_tsv("../analyses/telomerase-activity-prediction/results/TelomeraseScores_PTBAStranded_FPKM.txt") %>%
+  select(SampleID, NormEXTENDScores) %>% 
+  rename(Kids_First_Biospecimen_ID_RNA = SampleID,
```
jaclyn-taroni, line 176: Can't easily comment on this in this diff - do we want to add a column that indicates which data set these samples come from (`stranded`, `polya`)?

tables/output_tables.Rmd
```diff
@@ -13,132 +12,168 @@ library(openxlsx)
...
 }
+
+
+# Define input files
+tmb_all_file <- file.path(data_dir, "pbta-snv-consensus-mutation-tmb-all.tsv")
+tmb_coding_file <- file.path(data_dir, "pbta-snv-consensus-mutation-tmb-coding.tsv")
+mutsigs_file <- file.path(analyses_dir, "mutational-signatures", "results", "deconstructsigs_exposures_merged.tsv")
+chromo_file <- file.path(analyses_dir, "chromothripsis", "results", "chromothripsis_summary_per_sample.txt")
+tp53_file <- file.path(analyses_dir, "tp53_nf1_score", "results", "tp53_altered_status.tsv")
+quantiseq_file <- file.path(analyses_dir, "immune-deconv", "results", "quantiseq_deconv-output.rds")
```
jaclyn-taroni, line 36: EXTEND score files should go here too IMO

tables/output_tables.Rmd
```diff
@@ -13,132 +12,168 @@ library(openxlsx)
...
+list_s2_table <- list(tumor_mutation_burden = tmb_combined,
+                      fitted_cns_mutational_signatures = cns_signatures,
+                      chromothripsis_events = chromothripsis_region_df
                      )
-openxlsx::write.xlsx(list_s2_table, 
-                     file.path(results_dir, "TableS2-DNA-results-table.xlsx"),
-                     overwrite=TRUE)
+write.xlsx(list_s2_table, 
+           table_s2_file,
+           overwrite=TRUE)
```
jaclyn-taroni, line 137: So _sometimes_ we get errors with the cells `Number Stored as Text`... if I open this file locally in Excel, I see that the `Tmb_all` and `Tmb_coding` columns have these errors in the `tumor_mutation_burden` sheet. The `fitted_cns_mutational_signature` sheet has no such errors, despite the fact that the sheets both have IDs (expected to be text) and then floats. I wonder if `Tmb_all` and `Tmb_coding` are of class `character` for some reason and that's what's up?

jaclyn-taroni: Oh: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1269/files#diff-96afe8a2465c3d1e8fe81bcb0bc6e065771299f64c624c1dcb0f6fa23ef3f7c6R99 Hm

tables/output_tables.Rmd
```diff
@@ -154,22 +189,22 @@ telomerase_scores_combined <- bind_rows(telomerase_scores_polya_combined,
...
 list_s3_table <- list(tp53_scores = tp53_scores,
                       telomerase_scores = telomerase_scores_combined,
-                      quantiseq_immune = quantiseq_immune
+                      quantiseq_immune_deconvolution_fractions = quantiseq_immune
                       )
-openxlsx::write.xlsx(list_s3_table, 
-                     file.path(results_dir, "TableS3-RNA-results-table.xlsx"), 
-                     overwrite=TRUE)
+write.xlsx(list_s3_table, 
+           table_s3_file,
```
jaclyn-taroni, line 206: The `tp53_scores` and `telomerase_scores` sheets have the same problems as mentioned above.

tables/output_tables.Rmd
```diff
@@ -13,132 +12,168 @@ library(openxlsx)
...
+           overwrite=TRUE)
 ```
 
 ## Table S3: RNA results table
 ### TP53 scores
 ```{r s3 RNA results table p1}
 # get tp53 scores 
-tp53_scores <- readr::read_tsv("../analyses/tp53_nf1_score/results/tp53_altered_status.tsv") %>%
+tp53_scores <- read_tsv(tp53_file) %>%
   mutate_all(as.character) %>%
```
jaclyn-taroni, line 145: I don't know why this is here?

jaclyn-taroni (commented):
Something is going on with all the `mutate_all(as.character)` that is happening. 🧐

tables/output_tables.Rmd
sjspielman: > Something is going on with all the `mutate_all(as.character)` that is happening. 🧐

Yeah, I really didn't know why either. I left them in to avoid changing too much code, but will explore more deeply now.

tables/output_tables.Rmd
sjspielman: Realizing I missed all of that telomerase jazz on Friday afternoon. Library column has been added!

tables/output_tables.Rmd
sjspielman: I've removed the `as.character()` business (which I think, but don't know for sure, was trying to work with NAs in Excel?) all around and added argument `keepNA=TRUE` to excel exports where we want NAs to show up instead of being changed to blank cells.

jaclyn-taroni (approved):
Code looks good. I can confirm that the Excel files no longer contain that error, but boy are Excel NA values ugly (`#N/A` really?).

We should address my comment in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1269#issuecomment-1079433524 about making this more general. If we need to regenerate these tables as part of revision, I think we might be glad we took that step! 

If we wanted to take it a _step further_ for revisions sake, you might imagine that you sometimes want to get the TMB from the _analysis directory_ and add a `param` to do that. For instance, if we wanted to put running this notebook as the last step in the generate figures script (which might be handy/helpful), we might want to use the TMB in the analysis directory that we expect to be updated as part of the figure script. 

No need for me to re-review unless you'd like me to.

.circleci/config.yml
```diff
@@ -277,6 +277,12 @@ jobs:
          name: Scavenge back hotspots
          command: ./scripts/run_in_ci.sh bash analyses/hotspots-detection/run_overlaps_hotspot.sh
 
+
+      - run:
+         name: Compile manuscript tables
+         command: ./scripts/run_in_ci.sh Rscript -e "rmarkdown::render('tables/output_tables.Rmd', params = list(testing = 1), clean = TRUE)"
```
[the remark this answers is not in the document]
sjspielman: 🤦

Issue #1273, opened 2022-03-29
Remove sigfit vignette install from Dockerfile


sjspielman:
As noted in #1225 there are occasional CI timeouts. The docker image may benefit from some cleanup to reduce changes of timeouts when there is no cache available.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

In the Dockerfile, `R -e "remotes::install_github('kgori/sigfit', ref = '209776ee1d2193ad4b682b2e2472f848bd7c67a6', build_vignettes = TRUE, build_opts = c('--no-resave-data', '--no-manual'), dependencies = TRUE)"` takes a _long time_ to build the vignettes but the vignettes aren't needed int he docker. We should remove `build_vignettes = TRUE` so the default of `FALSE` is used.

#### Who will complete the updated analysis?

@sjspielman, <1 day

sjspielman:
No longer happening because PR was experiencing CI problems (probably unrelated to vignette removal), but those problems are not a compelling argument for this proceed. In addition, it doesn't really make a dent in runtime overall.

PR #1274, opened 2022-03-29
Removed vignette argument from sigfit installation

sjspielman:
Addresses issue #1273 and removes vignette installation to save a little time building the Docker.


sjspielman:
Ok, this keeps failing CI, which is not a compelling argument for this PR addressing CI problems. It was a bit overkill and isn't needed. Going to close it.

Issue #1248, opened 2022-03-18
Proposed Analysis: investigate whether signatures 3 and 18 also 


jharenza:
#### What analysis module should be updated and why?
mutational signatures

Based on the literature below, we hypothesize we would see a similar phenomenon in our cohort. That is, samples with hypermutator phenotypes have signatures associated with high genomic instability (in our CNS signature analysis, those would be RefSigs 3 andd 18) _and_ _TP53_ dysregulation. 

From doi:10.1038/s43018-020-0027-5:
>Signatures with a high level of genomic instability are dependent on TP53 dysregulation.

>Last, driver mutations in TP53 appear to be strongly associated with multiple signatures (Fig. 6b). Unlike other driver–signature relationships, target plots recurrently present a dark right semicircle and light left semicircle across multiple tissues. Thus, most tumors carrying these signatures are TP53-mutated, but the reverse is not true—not all TP53-mutated tumors carry the individual signatures. Many of the associated signatures can create hypermutator phenotypes resulting in noticable genomic instability (RefSig 17, 18, R1, R3 and R5). We therefore postulate that most mutational signatures with hypermutator phenotypes have a dependency on TP53 dysregulation. 
<img width="708" alt="Screen Shot 2022-03-18 at 1 50 45 PM" src="https://user-images.githubusercontent.com/207034/159056772-6f1daf0e-764f-4b1c-ad97-2221bfcaecb2.png">

#### What changes need to be made? Please provide enough detail for another participant to make the update.
1. We previously looked for a relationship between hypermutator samples and the MMR signature, but we should also look at whether there is a relationship with Sig 3 and/or 18. 
2. Do these hypermutator samples also show dysregulated _TP53_? Since there are only 3 hypermutant samples in this cohort (> 10 Mut/Mb) and 4 ultrahyper mutant samples (>100 Mut/Mb) defined in pediatric samples by https://www.cell.com/cell/pdf/S0092-8674(17)31142-X.pdf, possibly, we can explore these separately or maybe in a table?
[...]

PR #1285, opened 2022-04-05
Figure 4E: Add hypermutator mut sigs heatmap

jharenza:
Stemming from #1248, we decided a visualization + a table would be better than a table alone. 

#### What was your approach?
This PR creates a heatmap of the mutational signatures for the patients who have tumors which are hypermutant or ultra-hypermutant.

Stemmed from the analysis for #1248 

- Annotation/legend labels may need to be cleaned up; I couldn't immediately figure out how to do this within the pheatmap function, so for now, they are being renamed within the dataframe.

- Do we like the color being used for the heatmap? I picked `divergent_high_5` from [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/palettes/divergent_color_palette.tsv
) because the binary palette red was too bright for my liking and the green gradient palette wasn't super aesthetically pleasing no matter how many greens I used. The blue from binary is nice (see below for both versions) but may blend too much with the color for "Initial CNS Tumor", so what color should we use here? Since these go from zero to 1, I did not think coloring the 0 made much sense as it would for a set of values which go from [-X,0,+X].

[hypermutator_sigs_heatmap_red.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/8413848/hypermutator_sigs_heatmap_red.pdf)
[hypermutator_sigs_heatmap_blue.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/8413850/hypermutator_sigs_heatmap_blue.pdf)

- How do we want to color patients? Do we want to list them in the legend? I tried using the divergent low palette, but I thought it would blend too much with the blue heatmap.

- FYI, one sample is different from the pre-PR heatmap I showed and that was because I had taken a cell line instead of tumor for the plot. oops.
- Maybe we should make "Mutator" --> "Mutation status"? @sjspielman 
  
yes

#### Documentation Checklist


analyses/mutational-signatures/08-explore_hypermutators.Rmd
```diff
@@ -14,18 +14,24 @@ editor_options:
 
 
 ```{r setup}
+library(tidyverse)
+library(pheatmap)
+library(ggplot2)
```
sjspielman, line 19: If we load `tidyverse`, no need to load `ggplot2`
```suggestion
```

analyses/mutational-signatures/08-explore_hypermutators.Rmd
```diff
@@ -143,6 +150,78 @@ A couple interesting observations in the table above:
...
+anno$Kids_First_Biospecimen_ID <- NULL
+
+# Specify colors
+ann_colors = list(
+  # mutator color from [this PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1280/files#diff-bc8134008706f38e7795b59696928058d123e44b4172df68105a26ad8aa0a398R298-R301) 
+  Mutator = c(Normal = "grey40",
+              Hypermutant = "orange",
+              `Ultra-hypermutant` = "red"),
+  # tumor descriptor colors from [palette](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/palettes/tumor_descriptor_palette.tsv) 
+  `Phase of Therapy` = c(`Initial CNS Tumor` = "#709AE1",
```
sjspielman, line 210: Rather than setting these colors manually, we should probably read in the file `figures/palettes/tumor_descriptor_palette.tsv` and extract colors from that data frame.

sjspielman:
This looks good! I think this figure is a really concise way to communicate this information. A couple comments to start - 

+ Is it possible to add a label for the color gradient legend for signature exposure?
  + Also, we probably want to use `gradient_col_palette.tsv` to set this scale, as described here: https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/figures#color-palettes
+ The Patient colors are not easily distinguished in colorblind situations (I checked with https://www.color-blindness.com/coblis-color-blindness-simulator/). Maybe we can use Okabe Ito here to be sure, or something from `RColorBrewer`
  + Similarly, can we order the Patient legend based on the order the patients appear in the heatmap? I would guess this is a `factor()` thing
+ Is it possible to add a little more spacing between the legends and the heatmap itself? Doesn't need to be much, but it's a little bunched together now
+ This is failing CI almost certainly because of the CI subsetted data. We can get around this with a param to _not run_ the heatmap if we're in CI. I did this in the `mutational-signatures/07-plot_cns_fit.Rmd` notebook for reference, to deal with an analogous situation

Since this pub-ready figure is being added to the module, we also need to a line in `figures/generate_figures.sh` copying this final PDF figure into `figures/pdfs/fig4/panels`.

analyses/mutational-signatures/08-explore_hypermutators.Rmd
jharenza: updated!

analyses/mutational-signatures/08-explore_hypermutators.Rmd
jharenza: updated

jharenza:
>Is it possible to add a label for the color gradient legend for signature exposure?
I see you tried to use gradient_col_palette.tsv to set this scale, but if it didn't look great then let's stick with this.

I can try to look into this tomorrow

>The Patient colors are not easily distinguished in colorblind situations (I checked with https://www.color-blindness.com/coblis-color-blindness-simulator/). Maybe we can use Okabe Ito here to be sure, or something from RColorBrewer

I updated these

> Similarly, can we order the Patient legend based on the order the patients appear in the heatmap? I would guess this is a factor() thing

Will look into this tomorrow

> Is it possible to add a little more spacing between the legends and the heatmap itself? Doesn't need to be much, but it's a little bunched together now

Yeah, I don't like this, either, but am not sure if this is possible - did some searching, but can try to search some more tomorrow

> This is failing CI almost certainly because of the CI subsetted data. We can get around this with a param to not run the heatmap if we're in CI. I did this in the mutational-signatures/07-plot_cns_fit.Rmd notebook for reference, to deal with an analogous situation

thanks, updated this - let's see

jharenza:
> The legend spacing indeed looks like it may not be possible. I wonder if it's possible to export them separately and then we can position them later with Illustrator? I know `ComplexHeatmap` can export separately but don't know enough about `pheatmap`.

Yeah was thinking of a separate export. I don't know much about either, 😆, but I can check it out.

jharenza:
Hi @sjspielman. I updated to complexheatmap, but couldn't easily figure out how to cut the rows between patients. There is a newer version of complexheatmp (2.5.2) which has pheatmap-like commands which could probably easily handle this, but I didn't want to install and update a lot of dependencies in case that messed up other modules. Also, it looks like the legend is centered, but would probably look better positioned at the top. If you are able to take a look, that'd be great.

sjspielman:
@jharenza I gave this a go, and ended up with a new strategy (we can always revert!) where I've extract the legends as one horizontal row separately from the heatmap, _and_ I made each of these the same width for stacking the heatmap on top of the legends in Illustrator. What do you think of this approach? It can also be modified to stack the legends vertically, but the `packLegend()` function is what allows us to space them out. Of course we could export them each separately, but that's a lot more for Illustrator.

jharenza:
> @jharenza I gave this a go, and ended up with a new strategy (we can always revert!) where I've extract the legends as one horizontal row separately from the heatmap, _and_ I made each of these the same width for stacking the heatmap on top of the legends in Illustrator. What do you think of this approach? It can also be modified to stack the legends vertically, but the `packLegend()` function is what allows us to space them out. Of course we could export them each separately, but that's a lot more for Illustrator.

Thanks! I am ok with this method!

I still would like to create splits between patients so one can more easily look at initial/progressive tumors within patients. Do you think it is OK not having BS_ids in the plot? I suppose I can add these via legend or not mention (we don't in oncoprint). I moved annotations to the right so patients are the first to show up, making it a bit easier to visualize by patient. I will look into cutting the plot.

sjspielman:
> Do you think it is OK not having BS_ids in the plot?

I think it's fine either way, but I don't think we need _both_ the BS_ids and the colors for them.

jharenza:
woohooo! got the row split!

sjspielman:
>Wasn't sure if you wanted me to run the generate figures script

Definitely _do not_ run this script. It's a behemoth in process of being updated.. We'll do a whole run of it on AWS when all the figure scripts are finalized since it will take a long time to run (possibly over a day with all the analyses that get re-run).

sjspielman (approved):
Figure looks great!

Issue #1284, opened 2022-04-04
Remove fig4-telomerase-activities.R script


sjspielman:
PRs #1280 and #1283 complete telomerase figures for the manuscript. Therefore `fig4-telomerase-activities.R` should be _removed_ once they are merged.

jharenza:
completed in #1291

PR #1291, opened 2022-04-07
Remove old figures script

sjspielman:
This PR removes the now-deprecated `fig4-telomerase-activities.R` script and updates README accordingly. This addresses Issue #1284.

A few other small changes were made:
+ Document `supp-tp53-correlation.R` in the figures README. This figure is currently pending inclusion in SI.
+ Removed an old extra PDF from mutational signatures that no longer applies


jaclyn-taroni (approved):
👍🏻

Issue #1295, opened 2022-04-11
Deprecate tmb-compare module


sjspielman:
#### What analysis module should be updated and why?

The `tmb-compare` module is deprecated. Documentation should be updated to reflect this status, including:
+ `analyses/README.md`
+ `analyses/tmb-compare/README.md`
+ Remove `tmb-compare` module from CI

#### When do you expect the revised analysis will be completed?

1-3 days

#### Who will complete the updated analysis?

@sjspielman

sjspielman:
Closed with #1298

PR #1298, opened 2022-04-11
Deprecate tmb-compare module

sjspielman:
This PR from Issue #1295 deprecates the `tmb-compare` module, which is very out of date. A figure in the manuscript is inspired by this module, but all associated code is in `figures/scripts` so this module does not need to be run.

Changes made include:
+ Indicate **deprecated** in READMEs. 
+ Comment out `tmb-compare` lines from CI 
+ On the way, I noticed another recently-deprecated module was missing a clear "deprecated" notice in its README so I went ahead and added it here. 

Question for reviewers: I left the `analyses/README.md` to still say "Yes" that this module is used in the MS, but with a parenthetical "(figure only)" added for context. Given that the figure is exclusively made in `figures/scripts/`, should this actually now say "No"?


analyses/README.md
```diff
@@ -54,7 +54,7 @@ Note that _nearly all_ modules use the harmonized clinical data file (`pbta-hist
...
-| [`tmb-compare`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tmb-compare) | `pbta-snv-consensus-mutation-tmb-coding.tsv` | Compares PBTA tumor mutation burden to adult TCGA data. The D3B TMB calculations (`TMB_d3b_code`) and its comparison notebook (`compare-tmb-calculations.Rmd`) are *deprecated*. | N/A | Yes | 
+| [`tmb-compare`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tmb-compare) | `pbta-snv-consensus-mutation-tmb-coding.tsv` | *Deprecated.* Compares PBTA tumor mutation burden to adult TCGA data. | N/A | Yes (figure only) |
```
jaclyn-taroni, line 57: I feel like this more accurately captures what's going on:
```suggestion
| [`tmb-compare`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/tmb-compare) | `pbta-snv-consensus-mutation-tmb-coding.tsv` | *Deprecated.* Compares PBTA tumor mutation burden to adult TCGA data. | N/A | Not directly, similar figure generated in `figures/` | 
```

jaclyn-taroni (approved):
👍🏻  one comment that is a hill I would not die on

analyses/README.md
sjspielman: 👍

jaclyn-taroni:
Anything related to CI is being commented out. A previous failure was due to a timeout. If commenting things out were to cause an issue, it would likely be with the formatting of the YAML. If the YAML was screwed up in some way, I expect this would fail immediately, which is not the case.

Therefore, merging without CI!

PR #1302, opened 2022-04-12
Alter Figure 1 tumor descriptor distribution plots to have more rows

jaclyn-taroni:

Changing the number of rows in the panel in Figure 1 that shows the distribution for cancer types. You'll see why this was necessary soon when I fill a PR with compiled Figure 1!


sjspielman:
Changes do not require CI, so we can merge before checks are finished.

sjspielman:
Changes do not require CI, so we can merge before checks are finished.

PR #1304, opened 2022-04-12
Figure S1 as PNG and PDF

sjspielman:
This PR adds the compiled Figure S1 as both PDF and PNG. (This figure is a single panel, so there is not much actual "compilation.")

Reviewers: Please check if these files are in the expected place and have the expected name. We may also want to update `figures/README.md` to clearly state naming scheme for these panels, which can be added in this PR during review.


jharenza (commented):
Do we want to have both png and pdf, or just pdf, since we are using pdf for nearly everything else now? If both, they probably should be in the same folder?

jaclyn-taroni:
We need PNG for things to render in the Manubot version and PDF for submission. Putting compiled figures in `figures/pngs` and `figures/pdfs` is documented in `figures/README.md`. If we're going to do a reorganization, it's probably best to do it in the end after we've finished the compiled figures.

jharenza:
> We need PNG for things to render in the Manubot version and PDF for submission. Putting compiled figures in `figures/pngs` and `figures/pdfs` is documented in `figures/README.md`. If we're going to do a reorganization, it's probably best to do it in the end after we've finished the compiled figures.

oh gotcha, this is because it is the full figure! So then, should `Figure_S1.png` go in `figures/pngs/` with a new folder called `supp`, etc and the current PNGs in that folder be deleted and folders added per figure such as with the pdf directory? having one in a folder but one not was what threw me off.

jaclyn-taroni:
These additions are in different places than I would expect. I would expect:

`figures/pdfs/supp/figs1/Figure_S1.png` -> `figures/pngs/figureS1.png`
`figures/pdfs/supp/figs1/panels/Figure_S1.pdf` -> `figures/pdfs/supp/figs1/figureS1.pdf`

The file name style (i.e., `Figure_S1.*` vs. `figureS1.*`) is not well-defined in the docs, but this is closer to what I have in #1305 🤷🏻

sjspielman:
> These additions are in different places than I would expect

Ah, I see I misread something in the README! Will update.

sjspielman:
@jaclyn-taroni this has been updated. I also made a small change to the README about convention for supplementary compiled PDFs.

jaclyn-taroni (approved):
👍🏻

jaclyn-taroni:
Addition of non-programmatically generated figure and docs, merging without CI

PR #1309, opened 2022-04-13
Add PNG and PDF for compiled Figure S2

sjspielman:
This PR adds the compiled Figure S2:

+ `figures/pngs/figureS2.png`
+ `figures/pdfs/supp/figs2/figureS2.pdf`

The orientation is slightly different from the previous design because this is how it all fit together in action!


jharenza (approved):
looks good!

PR #1311, opened 2022-04-13
Update Figures 4B/C with only stranded data

### LINKED ISSUE #1307: Updated analysis: check tp53 Figure 4B,C Ns
scripts for plots in figure 4B, C

First, I think both panels probably should have the same Ns?
Second, panel B Ns add up to more RNA-Seq specimens (1166) than we have in this cohort, so that one is likely wrong. 

check Ns, plots may need to be remade - should be using stranded only 

this week

@sjspielman ?

sjspielman:
Closes #1307 

As noted in #1307, there are different sample sizes in Figures 4B and C, but they should have the same samples.

This PR fixes the bug and ensures _only stranded data_ is used for both of these plots. Samples sizes now match. The P-value for the tp53 scores panel has changed since this panel was previously included polya data. It was P=0.23 and now P=0.9, so the overall conclusions are the same. The P-value for the expression panel is the same (P=0.0035) because it was correctly using stranded data only before.


jharenza (approved):
👍  let's go to master!

sjspielman:
Changes do not require CI check. Merging before checks are complete.

PR #1312, opened 2022-04-13
Add Figure S4 panels and compiled versions

sjspielman:
This PR adds Figure S4. 

+ Panels were apparently not yet in `figures/pdfs/supp/figs4/panels`, so panels were added there by:
   + Adding a step in `generate_figures.sh` to run the mutational-signatures module. This step was previously missing, and was added above the first line in this file that is copying a figure from that module
   + Adding lines to copy the two panels into `figures/pdfs/fig4/panels`
+ A compiled PDF was added as `figures/pdfs/supp/figs4/figureS4.pdf`
+ A compiled PNG was added as `figures/pngs/figureS4.png`


jharenza (approved):
👍

sjspielman:
Code changes do not depend on CI so we can merge before checks are complete.

PR #1313, opened 2022-04-13
Add compiled figure S5

sjspielman:
This PR adds compiled PDF and PNG Figure S5. 

+ The exported size of the two scatterplots in this figure was updated for panels to fit in line
+ A compiled PDF was added as `figures/pdfs/supp/figs5/figureS5.pdf`
+ A compiled PNG was added as `figures/pngs/figureS5.png`

Two questions for reviewers:

+ The UMAP legend for LGG is labeled `colored_subtypes`. I think I should update this to match the other legends
+ We should generally firm up terminology: HGG and LGG, or HGAT and LGAT. I can update legend titles accordingly as needed


jharenza:
>The UMAP legend for LGG is labeled colored_subtypes. I think I should update this to match the other legends

Agree - `subtype` is probably ideal for all legends

>We should generally firm up terminology: HGG and LGG, or HGAT and LGAT. I can update legend titles accordingly as needed

We have been interchanging these, but I suppose since the community knows this better as HGG and LGG, _and_ our subtyping table/histology file uses `HGG` and `LGG`, we should stick with these. We can make sure to go through the paper and update as well.

sjspielman:
This has been updated - 

+ `LGG` and not `LGAT` in legend key
+ Use `subtype` for all subtype legends, and add code to ensure this legend is second when there are two legends
+ Tweak of figure sizes to fit in compiled version more cleanly
+ Compiled PNG/PDF updated with these new plots

jharenza (approved):
👍

Issue #1315, opened 2022-04-14
Update Figure 2 oncoprint panels


sjspielman:
The figure 2 oncoprint primary-only panels have a mismatch between legends and panel: Cancer groups seen in legends are not always seen in the oncoprint figures directly. Specifically:

+ Embryonal legend contains Ganglioneuroblastoma, but this cancer group is not in the landscape plot
+ Diffuse astrocytic legend contains Oligodenroglioma, but this cancer group is not in the landscape plot

Other miscellaneous changes to be made to these panels:
+ Update code to address Issue #1314 
+ Remove underscores from `cancer_group` and `germline_sex_estimate` row labels
+ Make plot titles a bit larger
+ If possible, match the cancer group legend font sizes

jaclyn-taroni:
This might be another good reason to address #1314, yea. I suspect this is the root of the issue!

jaclyn-taroni:
> * Remove underscores from `cancer_group` and `germline_sex_estimate` row labels

For what it's worth, I drop these from my version of the compiled figure, which you can see an older version of [here](https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/f6bcf253ac3ceb242e2557e82ad259ea3eb0f729/figures/pngs/figure2.png). It would not shock me if this was not technically possible.

sjspielman:
> It would not shock me if this was not technically possible.

Indeed, technically not possible -  see https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1316

sjspielman:
Closed with #1316

PR #1316, opened 2022-04-14
Update oncoprint figures

### LINKED ISSUE #1314: Deprecate primary-plus oncoprint plots in `figures/` ?
Currently, the script `figures/scripts/fig2-oncoprint-landscape.R` creates plots for _both_ primary-plus and primary-only. However, at this time, it does not appear that any primary-plus plots will be incorporated in the manuscript directly.

Do we want to deprecate these plots? The script has a for-loop over only vs. plus, so the code is very adaptable.

sjspielman:
Closes #1314 and #1315

Several updates were made to address some minor issues with the oncoprint landscape panels - 

+ Cancer group legends no longer include cancer groups that are not present in the figure
+ Cancer group legends are now in the same order as cancer groups in the figure
+ Removed primary-plus panels (main and supp), and commented out associated code in the script. 
+ Directly use the binary palette file for sex colors instead of hardcoding the hex codes
+ Updated supplementary panel (for Figure S3) since half of a legend key and the full sex legend was not in the plot. Needed a size tweak. _Note that Figure S3 will need to be re-compiled with this updated version._
+ Current titles are too small to be legible, but when their font size is increased the titled run off the plot margins. I have therefore removed titles in favor of (not ideal, but not sure what else to do..) labeling in Illustrator
+ I as unable to remove underscores from annotation labels. This ~feature~ bug turns out to be hardcoded into the `maftools` package, so no can do.


sjspielman:
Converted to draft since something is wonky with 2 PDFs. Will revert back once this bug is addressed.

figures/scripts/fig2-oncoprint-landscape.R
```diff
@@ -398,14 +407,14 @@ for (type_iter in seq_along(data_input_list)) {
...
+                               germline_sex_estimate = germline_sex_estimate_colors),
         fontSize = 1,
         colors = oncoprint_palette,
         bgCol = "#F5F5F5",
         drawRowBar = FALSE,
-        titleText = histology,
-        titleFontSize = 1.3,
+        showTitle = FALSE, # a legible font size is not possible to fit into margins
+        #titleText = histology,
+        #titleFontSize = 1.3,
```
jaclyn-taroni, line 417: Leave it in anyway. It will be less error prone to increase the font size in a vector image editor over adding it in the compiled figure!

figures/scripts/fig2-oncoprint-landscape.R
```diff
@@ -180,10 +182,18 @@ cancer_group_palette <- group_palette_df %>%
...
 names(cancer_group_colors) <- cancer_group_palette$cancer_group
 
+# For coloring male and female:
+binary_palette_df <- readr::read_tsv(
+  file.path(palette_dir, "binary_color_palette.tsv")
+)
+
+# this variable is used again to make the "Other CNS" legend
+germline_sex_estimate_colors <-  c("Male" = binary_palette_df$hex_codes[binary_palette_df$color_names == "binary_1"],
+                                   "Female" = binary_palette_df$hex_codes[binary_palette_df$color_names == "binary_2"])
```
jaclyn-taroni, line 192: This is not an improvement if we wanted to change the germline sex colors in the OncoPrints independently of the binary palette, which I could very well see happening.

Having a vector of `germline_sex_estimate_colors` that gets used both places certainly would be an improvement.

jaclyn-taroni (commented):
I had a couple comments (one of which I've returned already) where I think some of these changes could bite us down the line, e.g., we swap the OncoPrint labels accidentally.

figures/scripts/fig2-oncoprint-landscape.R
sjspielman: Titles are now restored

figures/scripts/fig2-oncoprint-landscape.R
sjspielman: no longer read in binary palette; now we just define the vector and use it

jharenza (approved):
>I as unable to remove underscores from annotation labels. This feature bug turns out to be hardcoded into the maftools package, so no can do.

I think there may have been some non-intuitive way to do this, for example if the column was mutated from `germline_sex_estimate` to `germline sex estimate` then the color dictionary would either have to have underscores (?)... or be have the single back quotes? It uses complexheatmap underneath, so I think it should be doable, but also not worth a ton of trouble!

I trust yours and @jaclyn-taroni for any updates here, so I will approve and you can use your discretion!

sjspielman:
@jharenza `maftools` sneakily, and without permission, removes spaces from annotation label names internally. When I tried this approach and it failed, I went digging through the package's source code and found that disappointing "feature."

sjspielman:
These code changes do not affect CI, so we can merge before checks complete.

PR #1320, opened 2022-04-15
Add cancer group plot to immune deconv notebook

jharenza:
Given that we are not highlighting any subtype-based immune cell type effects on survival, I think we should move the immune cell subtype plot to the supplement and create a cancer group plot as the main figure. This can go along with the GSVA immune scores, as I am discussing in the incoming PR for figure 5 updates in the manuscript repo. 

Update - manuscript repo ticket draft [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/243/files).

#### What was your approach?
- I added DMG as a subtype of interest, so that plot is updated
- I added a cancer group plot. This isn't as pretty as the mutational signatures plot, so needs some tweaking - please help @sjspielman :)

NA

I filtered cancer group > 15 to get the same subset as in mut sigs. I do slightly worry about two very similar plots, but they are showing completely different data, so I think we can go with it, unless you have any other ideas @sjspielman or @jaclyn-taroni ?

- I think we may still add an HR plot or two for immune cells within cancer group to Fig 5, but I will need some more time  after #1264 is completed to dig into the literature to determine what makes the most sense to show.
- I put this on the same y-scale because I thought it would be easier to talk pan-tumor, but I suppose that isn't entirely necessary and we can update this to be free-y. Thoughts?

yes

new figures - updated subtype figure containing DMG and new cancer group figure

I am showing this by cancer group rather than immune cell type because I am talking about immune cells relative to others within a cancer group.

#### Documentation Checklist


jaclyn-taroni:
I'll review this!

jaclyn-taroni (approved):
Looks good. I had a couple minor comments, but mostly an FYI type of thing.

jharenza:
thanks @jaclyn-taroni - took the suggestions!

Issue #1323, opened 2022-04-15
Updated analysis: Create HR figures for HGGs


jharenza:
#### What analysis module should be updated and why?
survival analysis: tp53/telomerase and immune cells (PR #1264)

#### What changes need to be made? Please provide enough detail for another participant to make the update.
1. From tp53/telomerase analysis, create HR plot for Figure 4 panel F from data used below:
<img width="502" alt="Screen Shot 2022-04-15 at 1 14 07 AM" src="https://user-images.githubusercontent.com/207034/163523875-c5845af1-9eb9-4101-b2ff-338f6285302b.png">
2. From upcoming edits to immune cell PR #1264 , add HR plot from univariate analysis of molecular subtypes for HGG group (HGG including DMGs and DIPG - same grouping as from tp53/telomerase analysis).
This will probably also go into figure 4 as 4G.

#### What input data should be used? Which data were used in the version being updated?
same as previous

#### When do you expect the revised analysis will be completed?
within a week

#### Who will complete the updated analysis?
@sjspielman

sjspielman:
I'm not sure that it will be possible to get a panel G here as we'd discussed above, given updated results in #1264. But, 4F is done with #1325!

jharenza:
completed with #1332 and #1346

PR #1325, opened 2022-04-15
Survival panel for figure4f

sjspielman:
This PR partly addresses #1323 by adding a Figure4F panel forest plot showing results from the additive survival model with three terms: tp53 score, telomerase score, and hgg grouping (binary variable).

Points that are significant are filled black, and otherwise filled white. I tried to indicate that the HGG group is referring to HGG (as opposed to reference level of non-HGG), but not sure my labeling conveys this well. No matter what, I think this will have to be a textual explanation in the caption.


figures/scripts/fig4-tp53-telomerase-panels.R
```diff
@@ -414,4 +415,88 @@ cowplot::ggdraw(legend)
...
+  labs(
+    x = "Hazard Ratio ± 95% CI",
+    y = "Coefficient"
+  ) + 
+  # Update term labels for pub-ready
+  scale_y_discrete(
+    labels = c(
+      "HGG group (HGG)",
+      "Telomerase score",
+      "TP53 score"
```
jharenza, line 470:
```suggestion
    labels = c(
      "HGG group: HGG",
      "Telomerase score",
      "TP53 score"
```
I think this will help in showing this refers to HGG as the group shown on the plot, since it seemed to be convention.

jharenza (commented):
@sjspielman - this looks nice!

Can we either add a table to this plot (probably a panel which would need assembling) or add another y-axis with the HR values, CIs, and p-values? 

Something that includes the information such as the plot examples below, and the one in the ticket?
https://www.researchgate.net/figure/A-Overall-survival-OS-hazard-ratio-HR-of-death-with-chemotherapy-versus-control_fig1_5344959
https://www.researchgate.net/figure/Forest-plot-portraying-the-hazard-ratio-HR-and-95-CI-of-the-association-between_fig3_269770487
https://www.researchgate.net/figure/Hazard-Ratios-of-the-Nine-Marker-Signature-learned-by-the-FDR-selection-procedure_fig1_252063792

I think having this information in the plot will make it more intuitive and easier to describe.

sjspielman:
I can make something like this but it will take longer to make, likely by end of the week.

If preferred, we can probably use the `survivalAnalysis` plotting function you found to make this plot and not run into any dependency issues.

jharenza:
I don't have a strong preference except that we have the info in the plot since that's the standard view. One other idea is printing a tidy table as a panel and assembling in illustrator, but that may have been your thought also.

sjspielman:
I think a separate _table_ will be tricky to precisely align to the plot, so I'm going to try out a pure ggplot2 strategy today!

jharenza:
Looks great!!

sjspielman:
@jharenza I was pleasantly surprised! Ready for approval?

sjspielman:
I made a small code update to determine p-value labels with `ifelse()` instead of hardcoding, after separate conversations with @jaclyn-taroni.

sjspielman:
Code changes do not require CI; merging before checks.

Issue #1322, opened 2022-04-15
Copy panels from Figures 5/S5 to panel folders


jharenza:
What needs to be done?
For Figure 5, replace main panel with cancer group plot in #1320 
For Figure S5, add subtyping plot from #1320 to supp panels

Who will complete this analysis?
@sjspielman

Issue #1321, opened 2022-04-15
Update compiled figure 5 and S5


jharenza:
### What needs to be done?
For Figure 5, replace with cancer group plot in #1320 
For Figure S5, make all current plots smaller (UMAP can be on one row?), then add subtype plot from #1320 

### Who will complete this analysis?
@sjspielman

jharenza:
just note figure 5 is merged with #1330, but S5 will need an update

jharenza:
closed with #1363  and #1364

PR #1330, opened 2022-04-20
Compiled figure 5

sjspielman:
Partially addresses issues #1321 and #1322 by updating **Figure 5** panels and adding the compiled PDF and PNG.

This PR copies the updated Figure 5 panels into figures, adds a line in `generate_figures.sh` to perform the copying, and adds compiled Figure 5.

One question: should we retain the `broad_histology` label on the GSVA heatmap?


jharenza:
> One question: should we retain the broad_histology label on the GSVA heatmap?

We did this to match UMAP, but I did find it harder to write about in the manuscript. However, the analysis is also done on broad histology. I think we can either retain broad histology labels and analysis, and add another annotation row for cancer group OR retain broad labels, run another analysis on cancer group and add the annotation for cancer group. I'll just have to update the text for the latter making sure the cancer groups' scores are significant, but it may be useful/easier to write to have the latter results and both labels (that'll require another legend).

sjspielman:
This PR is now out of date given #1336. I'm going to convert this to a draft and block with the 5D panel.

sjspielman:
@jharenza at long last this is ready again!

Note that as part of this I had to update the overall sizing for the quantiseq panel.

jharenza (approved):

sjspielman:
The `immune-deconv` module of interest has passed CI so this can be merged in now.

PR #1332, opened 2022-04-20
Survival analysis of HGG tumor subtypes

jharenza:
❗ staggered on #1264 but I cannot change the base to that branch

This PR addresses bullet 3 in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1264#pullrequestreview-942730959), in which we would like to perform a univariate analysis of subtype within the HGG tumors. This was previously done in v17, but for only H3 WT and DMG K28, so was quite outdated.

#### What was your approach?
I renamed the old notebook which did HGG subtype analysis and performed both logrank and cox regression (univariate) for molecular subtype.

NA: bullet 3 in [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1264#pullrequestreview-942730959)

double check everything

I am not sure why the subtypes which are `DMG, H3 K28` are not showing up in the results.  I ran a double check of the formula manually, and get the same results. There are more samples in this group than in some other groups, so I am baffled. Any ideas @sjspielman ?

yes

- KM plot and logrank results table for hgg subtypes
- cox results table for hgg subtypes

Here I go screen-shotting again...
Screen Shot 2022-04-20 at 2 24 01 PM
HGG tumor subtypes with K28, TP53 loss have a HR of 1.15 (0.41-1.88), p = 0.002, supporting current knowledge that co-occurrence of TP53 with K28 results in worse prognosis. 

#### Documentation Checklist


sjspielman:
@jharenza thanks for working on this!!

Had a look about the missing level. It's "there"! For the cox regression, it's the _reference category_. So, every coefficient can be interpreted as relative to the reference.

```
# Whatever is first is, by default, the reference! Or, you can change the factor levels
levels(factor(metadata$molecular_subtype))
[1] "DMG, H3 K28"                 "DMG, H3 K28, TP53 activated" "DMG, H3 K28, TP53 loss"     
[4] "HGG, H3 wildtype"            "HGG, H3 wildtype, TP53 loss"
```

To get pairwise comparisons among these, you can use [pairwise_survdiff](https://rdrr.io/cran/survminer/man/pairwise_survdiff.html). I went ahead and added a chunk in there so you can pull and see.

jharenza:
> @jharenza thanks for working on this!!
> 
> Had a look about the missing level. It's "there"! For the cox regression, it's the _reference category_. So, every coefficient can be interpreted as relative to the reference.
> 
> ```
> # Whatever is first is, by default, the reference! Or, you can change the factor levels
> levels(factor(metadata$molecular_subtype))
> [1] "DMG, H3 K28"                 "DMG, H3 K28, TP53 activated" "DMG, H3 K28, TP53 loss"     
> [4] "HGG, H3 wildtype"            "HGG, H3 wildtype, TP53 loss"
> ```
> 
> To get pairwise comparisons among these, you can use [pairwise_survdiff](https://rdrr.io/cran/survminer/man/pairwise_survdiff.html). I went ahead and added a chunk in there so you can pull and see.

Oh, this is perfect. Thank you!

jharenza:
@sjspielman this is ready now, and this would be the panel H forest plot I was envisioning in number 2 [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1323)

jharenza:
Just another two notes.
1. I added some code to check for rna-only samples, since they were subtyped. But, they were all non- Initial CNS tumor, so they were not added.
2. I didn't keep it in, but I did check making the reference `HGG, H3 wildtype`, and while almost everything is significant in the hazard plot (all K28 and TP53 subtypes), a weird thing was happening in which there was no reference plotted and `HGG, H3 wildtype`, even though it was the ref, it did not align with HR 1, and showed significance. No idea why, so I just removed this because the more useful comparisons are using H3 K28 as reference.

analyses/survival-analysis/survival-analysis_subtypes.Rmd
```diff
@@ -99,40 +110,44 @@ This way we subset the metadata to 1 single primary sample per patient ID
...
+                  independent_primary$Kids_First_Biospecimen_ID) 
+
+# check whether any RNA pts not in DNA subset
+metadata_rna <- metadata %>%
+  filter(experimental_strategy == "RNA-Seq" & 
+           tumor_descriptor == "Initial CNS Tumor" & 
+           !is.na(OS_years)) %>%
+  unique()
+
+rna_pts_to_add <- setdiff(unique(metadata_rna$Kids_First_Participant_ID), metadata_ind_dna$Kids_First_Participant_ID)
```
sjspielman, line 124: This line is a confusing - There is a `#nope` comment, but when I print this variable out (and it should be printed in the notebook if we need to see the results to get to that "nope" conclusion), it has output consisting of 5 patient ids:

```
> rna_pts_to_add <- setdiff(unique(metadata_rna$Kids_First_Participant_ID), metadata_ind_dna$Kids_First_Participant_ID)
> rna_pts_to_add
[1] "PT_63HDZHCN" "PT_9BZETM0M" "PT_0SPKM4S8" "PT_0MXPTTM3" "PT_S0Q27J13"
```

Based on this seems to me there _are_ RNA patients not in the DNA subset?

analyses/survival-analysis/survival-analysis_subtypes.Rmd
```diff
@@ -99,40 +110,44 @@ This way we subset the metadata to 1 single primary sample per patient ID
...
 # find molecular subtypes which have more than n_participants (3) participants
-filter_molecular_subtype <- metadata %>% group_by(molecular_subtype) %>% dplyr::tally() %>% dplyr::filter(n>n_participants)
+filter_molecular_subtype <- metadata_ind_dna %>% group_by(molecular_subtype) %>% dplyr::tally() %>% dplyr::filter(n>n_participants)
 
-# select only molecular_subtypes that have more than n_participants (3) participants
-metadata <- metadata %>%
+# select only  that have more than n_participants (3) participants
+metadata_ind_dna <- metadata_ind_dna %>%
   dplyr::filter(molecular_subtype %in%
                   filter_molecular_subtype$molecular_subtype)
```
sjspielman, line 134: This can be simplified to only 1 line of code:

```
metadata_ind_dna <- metadata_ind_dna %>%
  group_by(molecular_subtype) %>% 
  # directly add a column for the counts instead of tallying down into a summarized tibble
  mutate(n_molecular_subtype = n()) %>%
  # keep only sufficient n
  filter(n_molecular_subtype>n_participants) %>%
  # rm the counting column and it's all cleaned up
  select(-n_molecular_subtype)
```

analyses/survival-analysis/survival-analysis_subtypes.Rmd
```diff
@@ -167,11 +182,43 @@ Save the plot to a file.
...
+# Check Ns
+as.data.frame(table(metadata_ind_dna$OS_status, metadata_ind_dna$molecular_subtype))
+
+# Run model
+fit_save_model(metadata_ind_dna, 
+               "molecular_subtype",
+               file.path(results_dir, "cox_hgg_subtype.RDS"))
+```
+
+Why are we not seeing H3 K28 subtypes?
```
sjspielman, line 203: Since this subtype is in the model, this should be rephrased or chunk updated accordingly.

analyses/survival-analysis/survival-analysis_subtypes.Rmd
```diff
@@ -167,11 +182,43 @@ Save the plot to a file.
...
+```{r cox manual check}
+# Double check the function isn't missing them - get the same result below
+metadata_recode <- metadata_ind_dna %>%
```
sjspielman, line 206: this line in the chunk is needed for `pairwise_survdiff`, so I'd recommend just 1 chunk here with recoding, survdiff, and the plot.

analyses/survival-analysis/survival-analysis_subtypes.Rmd
sjspielman: We still have two groups going into modeling that only have 1 sample each: `HGG, IDH, TP53 activated ` and `HGG, H3 G35`. That means there is no variation in the groups to model. Hence, we see them coming up with _both_ CI's as `NA` and an unreliable estimate.

On one hand I would not want to include groupings that have insufficient data, but on the other hand if the goal is to compare among groups we cannot cherry-pick them. My personal assessment here is that there is not sufficient data to rely on the model output.

jharenza: I was wondering why we didn't see the G35 in the final table and it is because there aren't enough samples per the n=3 per subtype rule set in place in this notebook prior to my edits. I think some are also not primary tumors, so I'll check why they're even in the model..?

analyses/survival-analysis/survival-analysis_subtypes.Rmd
jharenza: I thought these were not primary tumors or maybe not RNA-only samples, but I can check again and should update the code for this.

jharenza: hmm, I still get `character(0)` - printing and will commit

sjspielman: Hm, let me see again..

analyses/survival-analysis/survival-analysis_subtypes.Rmd
jharenza: I am not seeing these in the model either

jharenza:
@jaclyn-taroni thank you for helping with `ungroup()`!! @sjspielman this is ready for re-review

analyses/survival-analysis/survival-analysis_subtypes.Rmd
```diff
@@ -167,12 +194,29 @@ Save the plot to a file.
 
 ```{r Save survival plot}
 # We can save the plot like a normal ggplot
-cowplot::save_plot(filename = kap_meier_plot_file, plot = surv_plot_all_subtype)
+cowplot::save_plot(filename = kap_meier_plot_file, plot = surv_plot_all_subtype, base_height = 5, base_width = 8)
 # Let's save the model for CI and median survial as well
 readr::write_tsv(surv_median(kap_fit$model), kap_meier_model_file)
```
sjspielman, line 199: We should update this to and RDS (which will include changing the `kap_meier_model_file` variable name!) so that _all_ model information is saved. 
```suggestion
readr::write_rds(kap_fit$model, kap_meier_model_file)
```

jharenza: yes, good idea

sjspielman (commented):
Looks good! Noting to update based on [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1348#issuecomment-1110085619) - set up a separate directory for the result files.

jharenza:
> Looks good! Noting to update based on [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1348#issuecomment-1110085619) - set up a separate directory for the result files.

done!

sjspielman (commented):
Looks like some files went a bit awry at some point here. The files exported by the template Rmd are either deleted in this PR or got deleted elsewhere.

I'm going to add some commits *here* to restore those files. As part of this, I'm updating the template Rmd to export result files to `results/template/` for consistency with what other notebooks are now going to do.

sjspielman:
I had approved this, but realized one additional change needs to be made - we also need to save the data that went into the KM model to make the figure for the paper. So, instead of exporting `kap_fit$model` to RDS, it now exports `kat_fit` to RDS.

This previously passed CI right before this update, and this additional line of code I added here was confirmed to run in Docker locally with a fresh environment. Therefore this is being merged before checks are complete.

PR #1333, opened 2022-04-20
Update compiled figure S5

sjspielman:
Partially addresses issues #1321 and #1322 by updating **Figure S5** panels and adding the compiled PDF and PNG.

This PR copies the updated Figure S5 panels into figures, adds a line in `generate_figures.sh` to perform the associated copying, and adds compiled Figure S5.

A few other changes:
+ The script that makes the UMAP figures was also slightly modified to get a different figure aspect ratio for compiling purposes. 
+ The TERT and TERC scatterplots had some funky encoding. _Any_ type of circular point was rendering as a greek lambda. I found the easiest way to deal with this is just to use square points, which look fine to my eye.

Notes:
+ I also experimented with making the molecular subtypes plot horizontal, but the x-axis text is too overwhelming that way given how many subtypes there are. So, I decided to keep the vertical layout for panel G.
+ Currently panel G is left-justified but can be centered.


jharenza (approved):
looks fine!

jaclyn-taroni:
Changes aren't subject to CI, merging without it!

PR #1338, opened 2022-04-22
Update GSVA analysis and figure

### LINKED ISSUE #1337: Update GSVA analysis
From [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1330#issuecomment-1103995168), we should update the GSVA analysis to consider `cancer_group` instead of `broad_histology` as the sample group variable. This requires both updating the module to update results, and update the code in `figures/scripts/fig5-panels-gsva-umap.R` that makes the associated manuscript figure.

sjspielman:
Closes #1337 

This PR updates the GSVA analysis and associated manuscript figure.

The analysis itself is updated to consider `cancer_group` instead of `display_group`, and associated results as well as README are updated accordingly. Re-running this module also updated some of the retained results because the module hadn't been run with the most recent data. There's a teeny bit of unneeded updates - 
+ Only do a notebook, not notebook + knitted HTML
+ Removed very unneeded lines [here]([analyses/gene-set-enrichment-analysis/util/hallmark_models.R](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1338/files#diff-9323996de14da32079dabd3b0596f2850129fb410191faf6d9723afc628facf8))
+ Added more code chunk labels since the notebook was rendered with repeated chunks unless all chunks were labeled. No idea why, but it's rendered at least.

The figure script is updated to output a new GSVA heatmap with a `cancer_group` annotation as well as an associated legend. The final plot will therefore have separate legends for A and B.

Question: Do we want to remove "Other" from the heatmap, or keep it at the end (as is currently there)?


jharenza:
> Question: Do we want to remove "Other" from the heatmap, or keep it at the end (as is currently there)?

We can remove other - those results won't be super meaningful!

jharenza (approved):
Approving once other is removed. Will go through these results and update the text! thanks @sjspielman :)

jaclyn-taroni (approved):
👍🏻

jharenza:
going to merge and update manuscript results based on this

PR #1339, opened 2022-04-22
Forest plot for panel 5D

### LINKED ISSUE #1336: Create Panel 5D - forest plot
### What needs to be done?
From #1264, add forest plot panel for figure 5: quantiseq cell types + CD274 expression for medulloblastoma tumors 

### Who will complete this?
@sjspielman

sjspielman:
Closes #1336 

This PR adds a forest plot for 5D, but there is an important caveats if we're going to include this plot:

The `Macrophage M1` immune cell category HR has an infinite estimate _and_ upper CI. It cannot therefore be plotted, but it is significant. It's _definitely_ >1, but we have no precision about the estimate other than "it's big." Therefore, this needs to be discussed in the caption and in the text, because it's not possible to graphically represent the full model.

Otherwise, CD274 expression and monocyte fractions are significant, though the estimate for the latter is probably imprecise due to very wide CI.


jharenza (approved):
👍

jaclyn-taroni (approved):
👍🏻

sjspielman:
Does not rely on CI so merging before checks are complete.

PR #1342, opened 2022-04-24
add extent tumor resection/lgg to tp53/telomerase survival analysis

jharenza:
#### What scientific question is your analysis addressing?
After discussion with @adamcresnick, I am adding to the survival analysis two additional terms: extent of tumor resection and LGG group. This is to make sure our results are not strictly due to tumors having been fully resected or belonging to LGG group.

#### What was your approach?
I add in extent of tumor resection and LGG group for univariate analysis. I also add these terms to the multivariate analyses.

#### What GitHub issue does your pull request address?
NA

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
We will want to create a new forest plot for this result.
Update: done in #1343 

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What is your summary of the results?
- Extent of tumor resection: gross/total and LGG group: LGG are not surprisingly very strong positive predictors of overall survival.
- TP53 and telomerase scores alone are each significant with or without LGG samples or samples with tumors fully resected. Note: the p-values did get larger when removing these groups, suggesting it is good we are adding in these terms! TP53 score p went from e-13 to e-8 and telomerase score p went from e-17 to e-8.
- HGG group is a very strong negative prognostic indicator of survival (p=5e-35).
- LGG group is a strong positive predictor of survival (p=1e-4).

- Interaction models are still not significant/only marginally significant, thus additive are the way to go.

**tp53_score+telomerase_score+extent_of_tumor_resection+lgg_group**
- The strongest predictor here is extent of tumor resection with gross/near total resection having a survival benefit and partial resection nearing significance for a survival benefit. High TP53 scores, telomerase scores, and non-LGG are all significant risk factors for poorer overall survival.

**tp53_score+telomerase_score+extent_of_tumor_resection+hgg_group**
- HGG group:HGG and high telomerase scores are significant risk factors for poorer overall survival. Extent of tumor resection: gross/near total is a survival advantage. High TP53 scores trend toward a significant risk for poorer overall survival.

**Additive models: tp53_score+telomerase_score+extent_of_tumor_resection for each cancer group**
- Same results as before

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
[...]

PR #1343, opened 2022-04-24
update tp53/telomerase forest plot(s) 

jharenza:
Here, I update the tp53/telomerase forest plots to include extent of tumor resection. In addition, I am adding a second forest plot which has a term for LGG group. This is staggered on #1342 

#### What was your approach?
Just updated the plots based on analysis in #1342 

NA

NA

I think that we should make the LGG plot now 4F, given the control for LGG. I do think the HGG plot is meaningful since the TP53 scores were not as predictive for HGG (except in DMGs, in which there is a significant hazard - goes along with what we are now learning via the PNOC paper/the subtyping forest plot we will add in 4G), and we can possibly add it to S4 as panel D (may have to landscape that figure). Thoughts on putting the hgg forest plot in S4?

yes

#### Documentation Checklist


sjspielman:
I added a few small commits here to fix the cut off x-axis and remove some hardcoding for ref levels. 

This looks ready to me, but we should decide about the panel location before merging (since the code may need to be in a different script). I'm not sure what exactly is meant by "LGG plot" vs "HGG plot" since the model contains both LGG and HGG? There is only 1 plot here.

jharenza:
closing in favor of #1354

PR #1344, opened 2022-04-24
immune cell survival update to include extent tumor resect + lgg group

jharenza:
#### What scientific question is your analysis addressing?
Adding extent of tumor resection and lgg grouping per @adamcresnick

#### What was your approach?
Same as in #1342 

#### What GitHub issue does your pull request address?
NA

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
Removed subtype analysis from here because of #1332 and that this PR only included RNA samples.

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
yes

#### What is your summary of the results?
pretty much the same as before, except PDL1 expression is not significant across the entire cohort

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [x] The analytical code is documented and contains comments.

PR #1345, opened 2022-04-24
update MB immune cell FP - 5d

jharenza:
update immune cell FP for medulloblastoma per #1344 (staggered on #1344)

#### What was your approach?
added extent of tumor resection terms, reran code

NA

NA

No

yes

same as before-  results are still significant even accounting for extent of tumor resection

#### Documentation Checklist


sjspielman (commented):
It would be good to add the reference level for resection in here too, similar to #1346. Here it seems the reference level is `Biopsy only`. This will require a few more changes to the forest plot script, and we'll want to color or shape as discussed over in #1346.

jharenza:
@sjspielman something weird is happening - I added a grey color for the reference (we may want darker), and it is showing up, for ex, if I print `forest_plot` in line 142, but it is not coming up in the final PDF. Not sure why. Also, do you think grey is sufficient or do you want to also change shape? Would you go about the latter by adding a shape column in the survival_df?

sjspielman (approved):

Issue #1032, opened 2021-04-26
7316-2686: path dx == PNET; diagnosis free text == Ewing sarcoma


jharenza:
#### What data file(s) does this issue pertain to?
`pbta-histologies.tsv`

#### What release are you using?

v19 within this embryonal subtyping update [PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1031)

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.
NA

#### Put your question or report your issue here.
Flagged this sample for our clinical team. The pathology report does not mention EWS, so this may be a data entry error from the site from which this patient came. They have reached out and this may be updated in v20.

jharenza:
Update on this: site coordinator updated the path dx to EWS Abs free test to PNET(EWS), so this will be reflected in v20.

jharenza:
closed with #1118

PR #1346, opened 2022-04-24
Add forest plot for panel 4G

jharenza:
Add HGG subtypes forest plot, staggered on #1332 

#### What was your approach?
Add forest plot, including reference subtype

Ahh, I guess I never made an issue for this..

Upate, found it #1323

NA

From #1032 
>Pairwise comparisons show a signficant difference between “DMG, H3 K28, TP53 loss” and “DMG, H3 K28”, as well as between “HGG, H3 wildtype” and “DMG, H3 K28”

Since DMG H3 K28 is the reference, why isn't the p value for HGG, H3 wildtype significant in this additive model? It would actually be nice if we can show this significance value on the FP - maybe we can recode that one grey and add the pairwise pvalue? 😬 

#### Documentation Checklist


figures/scripts/fig4-hgg-subtype-forest-plot.R
```diff
@@ -0,0 +1,171 @@
...
+
+
+# Get n and event info from glance output
+survival_n <- broom::glance(survival_result) %>%
+  select(n, nevent)
+
+# Convert survival model result to data frame, and exponentiate estimates/CIs to get HRs
+survival_df <- broom::tidy(survival_result) %>%
+  # Add DMG, H3 K28 as reference
+  add_row(term = "molecular_subtypeDMG, H3 K28",
```
sjspielman, line 57: 👍

sjspielman (commented):
Looks good! One other quick thought - I think it's great that you added the reference level here into the plot. I wonder if we should change it's placement and/or color?
Specifically, I think it makes sense for the reference to be the bottom row of the plot (though, still next to the other HGG levels). Also, the shape is white but it's not actually "not significant" - it's just not anything. What do you think about either making it gray, or maybe changing the shape to not be a diamond? Basically, some kind of visual cue that it's a different kind of point.

jharenza:
>Looks good! One other quick thought - I think it's great that you added the reference level here into the plot. I wonder if we should change it's placement and/or color?
Specifically, I think it makes sense for the reference to be the bottom row of the plot (though, still next to the other HGG levels). Also, the shape is white but it's not actually "not significant" - it's just not anything. What do you think about either making it gray, or maybe changing the shape to not be a diamond? Basically, some kind of visual cue that it's a different kind of point.

YES! I also had the thought about grey or another shape, and bottom is fine with me. I have multiple refs in other forest plots because of the groupings, but we can add those to the bottom or bottom of the group - what do you think?

sjspielman:
Ha, I just edited that comment because I realize I missed you had already suggested the gray color!

I think at the bottom of each _grouping_ makes the most sense.

jharenza:
> Ha, I just edited that comment because I realize I missed you had already suggested the gray color!
> 
> I think at the bottom of each _grouping_ makes the most sense.

updated

sjspielman (approved):
I made two quick changes here: 
1) plot export size, and 
2) added a variable `ref_term` since that string is reused.

This does not require CI, so I am going to go ahead and approve and merge this into `surv-hgg-subtype` now in the interest of time.

PR #1347, opened 2022-04-25
medullo cd274 expression/assess immune hot (figure 5E, S5J)

jharenza:
Adding a panel after the medullo forest plot to show the cd274 expression distributions across medulloblastoma subtypes.

#### What was your approach?
plot cd274 expression by mb subtype

NA

When plotting the cd8/cd4 ratios (advice from clinician Jess Foster), I opted to remove all Inf/NaN from the dataframe because all EPN had these values. Is that ok? We can put this in the supplement (S5-J) since I do mention in the text that many of these tumors are immune cold and this could be a good metric, even if rough.

- Should we keep as boxplots or make violins? 
- We need to add stats, but since nearly every pairwise stat is significant, I need some advice on this @sjspielman - should I just export the stat dataframe and write about them in the legend?

yes

two new figures

- CD274 expression, even though low, is significantly different across MB subgroups. This has been previously reported with SHH and WNT having higher expression. We see the highest in WNT.
- We only see MB group 3 & 4 might be considered "immune hot" with a greater number of CD8+ vs CD4+ T cells. Most brain tumors are "immune cold" (also known, but we can show that we can use the quanTIseq fractions to estimate "hot" tumors which might be responsive to immune checkpoint blockade.

#### Documentation Checklist


analyses/immune-deconv/02-visualize_quantiseq.Rmd
```diff
@@ -260,30 +270,112 @@ ggplot(pdl1_cd8) +
...
+Visualize MB only, since that was the only significant result from the survival analyses
+
+```{r pdl1_mb, fig.width = 5, fig.height = 5}
+# select only mb samples
+pdl1_cd8_mb <- pdl1_cd8 %>%
+  filter(cancer_group_display == "Medulloblastoma")
+
+
+# Count group sizes to use in x-axis labels. Use this data frame going forward
+  df_counts <- pdl1_cd8_mb %>%
```
sjspielman, line 306: Undo indent

analyses/immune-deconv/02-visualize_quantiseq.Rmd
```diff
@@ -260,30 +270,112 @@ ggplot(pdl1_cd8) +
...
+    group_by(molecular_subtype) %>%
+    mutate(subtype_count = n()) %>%
+    ungroup() %>%
+    mutate(molecular_subtype = glue::glue("{molecular_subtype}\n(N = {subtype_count})"))
+  
+  wilcox_df <- ggpubr::compare_means(log2_expression ~ molecular_subtype,
```
sjspielman, line 312: Undo indent

analyses/immune-deconv/02-visualize_quantiseq.Rmd
```diff
@@ -260,30 +270,112 @@ ggplot(pdl1_cd8) +
...
+  # remove outliers
+  geom_boxplot(outlier.shape = NA, color = "grey40", size = 0.2) + 
+  geom_jitter(width = 0.15, size = 0.75, alpha = 0.8, aes(color = "black")) + 
+  scale_color_identity() +
+  labs(x = "Molecular subtype of sample",
+       y = "Log2 expression (FPKM) of CD274") +
+  ggpubr::theme_pubr() + 
+  theme(axis.text.x = element_text(hjust = 1, 
+                                   size = rel(0.6), 
+                                   angle = 75))
```
sjspielman, line 329: For getting the P-values in the plot, see these docs and in particular the example: https://rpkgs.datanovia.com/ggpubr/reference/stat_pvalue_manual.html

You need to add a column to `wilcox_df` called `y.position` that contains the (manually determined!) y-axis coordinates for where each label will be placed.

analyses/immune-deconv/02-visualize_quantiseq.Rmd
```diff
@@ -260,30 +270,112 @@ ggplot(pdl1_cd8) +
...
+                          function(x) replace(x, is.infinite(x), NA)))
+
+# drop NaN
+quantiseq_cd8 <- do.call(data.frame,    # Replace Inf in data by NA
+                   lapply(quantiseq_cd8,
+                          function(x) replace(x, is.nan(x), NA)))
+
+# remove NA rows from plot
+quantiseq_cd8 <- quantiseq_cd8 %>%
+  filter(!is.na(score))
```
sjspielman, line 356: Instead of updating the same variable and re-saving it over and over (can cause lots of sneaky bugs), there's a tidyverse way to do this. As far as I can tell, all this replacement only should be done for the `scores` column.

```suggestion
quantiseq_cd8 %>%
   filter(!(is.nan(score)), 
            !(is.infinite(score))
```

analyses/immune-deconv/02-visualize_quantiseq.Rmd
```diff
@@ -260,30 +270,112 @@ ggplot(pdl1_cd8) +
...
+
+# distributions of PDL1 expression
+ratio_plot <- ggplot(quantiseq_cd8) +
```
sjspielman, line 359: I would recommend putting this on a log-scale; [this reference](https://clauswilke.com/dataviz/coordinate-systems-axes.html#nonlinear-axes) describes why ratios are more suitable on a log scale:

> Because multiplication on a log scale looks like addition on a linear scale, log scales are the natural choice for any data that have been obtained by multiplication or division. In particular, ratios should generally be shown on a log scale. 
> On a log scale, the value 1 is the natural midpoint, similar to the value 0 on a linear scale. We can think of values greater than 1 as representing multiplications and values less than 1 divisions. For example, we can write 10=1×10 and 0.1=1/10. The value 0, on the other hand, can never appear on a log scale. It lies infinitely far from 1. One way to see this is to consider that log(0)=−∞. Or, alternatively, consider that to go from 1 to 0, it takes either an infinite number of divisions by a finite value (e.g., 1/10/10/10/10/10/10⋯=0) or alternatively one division by infinity (i.e., 1/∞=0).


We can indicate in the text that 0's are necessarily not shown and potentially write up how many there were in a supp table. If we decide to add this to a table, code should be included here to calculate those numbers.

sjspielman (commented):
For the MB plot, I'd recommend adding the P-values in so we can see how it looks. I've added a comment about how to go about this. I don't see much difference between violin and boxplot for it, though. Boxplot seems fine to me.

sjspielman (commented):
Another quick change I just caught - the cd8/cd4 ratio score should be filtered out of the data frame for the R chunk labeled `subtypes_celltypes_plot`.

analyses/immune-deconv/02-visualize_quantiseq.Rmd
jharenza: This doesn't look super on a log scale, and it also seems to remove 109 more rows (not an error I get with the linear scale after NaN/Inf removal). I kept on linear because it seemed easier for people to think about what the ratios mean in terms of cell populations.. thoughts? It's not a stellar figure to begin with, and only signal in MB, hence the supplement.

jharenza:
@sjspielman I think this is ready for re-review, but see my comment about the log scale plot

analyses/immune-deconv/02-visualize_quantiseq.Rmd
jharenza: Current text for context:
>Finally, we asked whether any molecular subtypes might show an immunologically hot phenotype, as roughly defined by a greater proportion of CD8+ to CD4+ T cells and while we do see the highest ratios in Group 3 and Group 4 medulloblastomas (**Figure {@fig:Fig5}I**), very few tumors had ratios > 1, highlighting an urgent need to identify novel therapeutics for these immunologically cold pediatric brain tumors with poor prognosis.

sjspielman: Given that so few are < 1, and the comparison between <1 and >1 ratios is the main need for log scale, keeping it on linear scale w/ this textual explanation is good with me!

jharenza:
@sjspielman changed the ratio plot back to linear scale, but this was not passing CI before - it holds up at line 608, but even if I add to not run that chunk in CI, still errors out. not sure why...

sjspielman:
@jharenza I'm looking into this CI failure.

figures/generate-figures.sh
```diff
@@ -177,6 +177,9 @@ cp ${analyses_dir}/immune-deconv/plots/cell_types-molecular_subtypes.pdf  pdfs/f
 ###### Forest plot for 5D
 Rscript --vanilla scripts/fig5-forest-plot.R
 
+###### Box plot for 5E
+cp ${analyses_dir}/immune-deconv/plots/cd274_expression_mb.subtypes.pdf  pdfs/fig5/panels/cd274_expression_mb.subtypes.pdf
```
sjspielman, line 181:
```suggestion
cp ${analyses_dir}/immune-deconv/plots/cd274_expression_mb_subtypes.pdf  pdfs/fig5/panels/cd274_expression_mb_subtypes.pdf
```

sjspielman (commented):
There is one typo in the `generate_figures.sh`, but otherwise this looks good now as long as this now passes CI. I'll wait to approve until we are sure about the CI bug.

sjspielman (approved):
Noting that commit [7974920](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1347/commits/7974920dc01c13289b3b592868aeca35e6576f85) passed the immune deconvolution step before CI was killed by another commit. Therefore this passes CI.

This is ready to go! The one other comment I had was font sizing in the plots. But, these aesthetics are is likely to be somewhat tweaked anyways during figure compilation, and current font sizing doesn't preclude interpretation at all, so this is totally fine.

sjspielman:
This has passed CI so can be merged before the next round of checks are complete.

PR #1350, opened 2022-04-26
Remove extra Table S3

sjspielman:
Somehow in #1264, a supp table file got copied into `analyses/survival-analysis/results/TableS3-RNA-results-table.xlsx`. It needs to be removed. This PR removes it.


jharenza (approved):
ha, i was just commentingg about this in another PR, thanks

PR #1351, opened 2022-04-27
Remove duplicated CI step

sjspielman:
CI had two lines for survival analysis:

+ One line _just_ to render the template notebook
+ One line to run the full module

However, the template notebook is also included when running the full module. Therefore it does not need to be run on its own (aka twice). I removed CI line that renders the template notebook.


jaclyn-taroni (approved):
👍🏻

sjspielman:
This only removes from CI, so it should not require re-running checks. Merging now.

PR #1352, opened 2022-04-27
KM HGG panel

### LINKED ISSUE #1335: Panel 4H - KM curve
### What needs to be done?

From #1332, create panel 4H for KM curve of HGG subtypes

### Who will complete this?
@sjspielman

sjspielman:
Closes #1335 

This PR creates the Kaplan-Meier plot from the HGG molecular subtypes survival analysis.

But, based on previous discussion (having some trouble finding the comment where this was mentioned, @jharenza?), given the additional survival plots that have been made, it's not clear whether this should still go into Figure 4. The output directory in the code therefore should be determined as part of review discussion here. 

For now, the main reviews I'm looking for are overall plot aesthetics and any feedback on the code itself.


figures/scripts/hgg-kaplan-meier.R
```diff
@@ -0,0 +1,83 @@
...
+# Makes a pdf panel for the HGG Kaplan-Meier survival analysis for Figure TBD
+
+library(tidyverse)
+library(survival)
+library(patchwork)
+
+# Establish base dir
+root_dir <- rprojroot::find_root(rprojroot::has_dir(".git"))
+
+# TBD!!!!!!!!!!!!!!!!!!
```
jaclyn-taroni, line 12: ❗

figures/scripts/hgg-kaplan-meier.R
```diff
@@ -0,0 +1,83 @@
...
+  data = df
+)
+
+# Establish color-blind friendly colors
+okabe_palette <- colorblindr::palette_OkabeIto[1:5]
+
+km_plot <- survminer::ggsurvplot(fit = fit_factored,
+                       data = df,
+                       palette = okabe_palette,
+                       pval = "P = 0.0002", # P=2e-4, which will look better I think spelled out?
```
jaclyn-taroni, line 54: Can we make use of `format(<the p-value from the object>, scientific = FALSE)` here instead? ([ref](https://stackoverflow.com/questions/21229997/formatting-p-value-in-scientific-notation-to-conventional)) This is the stuff of the test changed but the p-value on the figure didn't nightmares.

figures/scripts/hgg-kaplan-meier.R
```diff
@@ -0,0 +1,83 @@
...
+
+
+# re-factor the km data so legend order matches line order
+df <- km_result$original_data
+df$molecular_subtype <- factor(df$molecular_subtype, 
+                                 levels = c("HGG, H3 wildtype",
+                                            "HGG, H3 wildtype, TP53 loss",
+                                            "DMG, H3 K28",
+                                            "DMG, H3 K28, TP53 activated",
+                                            "DMG, H3 K28, TP53 loss"))
```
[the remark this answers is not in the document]
sjspielman: As in preference for - 
```
df <- df %>%
mutate(molecular_subtype = factor(...))
``` 
?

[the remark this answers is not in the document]
jaclyn-taroni: Preference? No. Am I surprised you went this route? A little.

jharenza:
@sjspielman thanks for working on this! Can you double check your legend because it looks like the colors aren't assigned to the correct subgroups based on survival. Otherwise, aesthetics look nice and yes I think adding to figure 4 still makes sense.

sjspielman:
>Can you double check your legend because it looks like the colors aren't assigned to the correct subgroups based on survival

Good catch! I had tried to change the levels so the legend order matched the line order, but I guess that didn't work.. 

I'll try to make it work, and if I can't the legend order will somewhat differ from the line order (which isn't a big deal I think), and I'll send it into Fig 4.

figures/scripts/hgg-kaplan-meier.R
sjspielman: Noting this is going to Figure 4.

jharenza:
>the legend order will somewhat differ from the line order (which isn't a big deal I think)

yeah nbd! :P

sjspielman:
@jharenza ready for another look!

@jaclyn-taroni I included the `format()` function but had to do some calcs to get there. The model itself doesn't contain the P-value (`ggsurvplot()` calculates it under the hood).

sjspielman:
Does not require CI so merging before checks complete.

PR #1353, opened 2022-04-28
Removed depcrated result file

### LINKED ISSUE #1349: Remove deprecated survival analysis result files
From [this comment](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1348#pullrequestreview-953762566), the `survival-analysis` module which has been heavily reworked recently seems to have a lot of now-deprecated result files in `results/`. 

We should carefully comb through files to identify which result files should and should not exist. A quick approach here would be - once all survival PRs are merged in, we can:
+ temporarily remove all result files
+ re-run the module
+ only keep what comes out of a fresh run, and purge any files that shouldn't be there.

@sjspielman will do this.

sjspielman:
Closes #1349 

Now that all updated survival analysis PRs have been merged in, I did did a pass over the survival module to remove any deprecated results. There was only 1 file that needed to be removed, and that's all this PR does. I confirmed this was accurate by re-running the whole survival module. No other diffs emerged!


jharenza (approved):
👍

sjspielman:
Does not require CI.

PR #1356, opened 2022-04-29
Removed unknown/unavailable resections from relevant survival models

sjspielman:
This PR revisits two of the survival notebooks, `immune` and `tp53_telomerase`, which construct models using `extent_of_tumor_resection`. The models included unknown resections (denoted "Unavailable" or "Not reported" in the data), but these categories are wholly uninterpretable so they should not be included.

This PR updates the specific survival models and associated results where this predictor was included. 

There is one particularly notable result change that I observed - the model with `quantiseq cell type + fractions CD274 + extent of tumor resection` now only converges for HGG, not MB.

PR #1357, opened 2022-04-30
re-create 5D forest plot

jharenza:
With #1356 in, we had to re-create 5D forest plot

#### What was your approach?
rerun script

NA

na

no

yes

new FP

#### Documentation Checklist


sjspielman:
This PR only has the figure. Code?

sjspielman:
The code here should also be updated, not just rerun. For example, see the `term_order` and `term_labels` variables in the script - Unknown/Unavailable should be removed from there. 

In addition, I want to note that in this updated model, Macrophage 1 now has an estimate, but it's lower bound and upper bounds are -Inf and Inf respectively. I think we should continue to exclude this point from the figure, but describe in figure caption that it's coefficient is `-8.951217e+04` with infinite CIs and P=0.99.

jharenza:
Ahhh thanks for catching this. Will update as soon as I can!

jharenza:
@sjspielman code/plot is updated

sjspielman (commented):
One more change here and then I think we're good to go. The _comments_ on lines 35-38 should be updated, since Macrophage M1 estimates/CIs have changed from the original model. We want to indicate in these comments it is NOT significant but has infinite CIs.

jharenza:
> One more change here and then I think we're good to go. The _comments_ on lines 35-38 should be updated, since Macrophage M1 estimates/CIs have changed from the original model. We want to indicate in these comments it is NOT significant but has infinite CIs.

whoops, updated

sjspielman (approved):
👍

sjspielman:
This code does not run through CI so we can merge before checks complete.

PR #1358, opened 2022-05-02
Compiled figure4

sjspielman:
This PR compiles Figure 4 as a PNG and PDF, and includes some code changes (namely sizing and `useDingbats = FALSE`) that are needed to support compilation.

Note the top row panels A-C are a bit tricky to center, because panel A is smaller than B and C are, but I don't think this looks too jarring.


jharenza (commented):
My only comment is that the overall width can probably be decreased because there is a bit of whitespace in every row which can be reduced.

sjspielman:
@jharenza I was able to drop the width by just shy of an inch. How does this look?

jharenza (approved):
👍

sjspielman:
Code changes do not require CI; merging before checks complete.

PR #1364, opened 2022-05-04
Compiled Figure S6

sjspielman:
This PR adds compiled Figure S6, which includes some code changes to resize the immune deconvolution panels. Note that the immune/subtypes panel is also now _wide_ to ensure the two rows of panels are the same widths. As a consequence the T cell ratio plot is a bit more narrow.


jharenza (approved):
Looks fine to me!

PR #1367, opened 2022-05-05
Rename Key Resources Table

### LINKED ISSUE #1366: key resources table
According to the [Cell guidelines](https://www.cell.com/star-supplemental-information), 

> Please note that the key resources table and STAR Methods text are not considered part of the SI.

The key resources table should be renamed from `TableS5-Key-Resources-table.xlsx` to just `Key-Resources-table.xlsx`.

sjspielman:
Closes #1366 

This PR renames the key resources table and updates the tables Rmd with associated context.

Note that I'm planning to implement associated MS changes in a larger copy edit PR today.


jharenza:
@sjspielman can you check my comment on #1366 because I _think_ we need to keep this as a Supp Table.

sjspielman:
@jharenza is correct we need this as a supp table because there are > 10 items. Closing this PR since it's not the right move.

PR #1379, opened 2022-05-06
rerun chordoma subtyping (4 of N)

jharenza:
rerun chordoma subtyping

no notable changes

#### Documentation Checklist


jaclyn-taroni:
Hm, so I think this should get rerun to reflect the expression value changes in the module but I don't think it needs to be done here in this stack. I will close.

jaclyn-taroni:
Okay, I've rerun this in a separate branch and it seems like the differences in expression that I perceived were a matter of aspect ratio of the figure.

PR #1395, opened 2022-05-10
Update deploy step to use a specified Ubuntu image

jaclyn-taroni:
When I added the updates in #1224, I missed the deploy step.


sjspielman (commented):
Seems fine, unless broader changes might actually be needed down the line? 

https://circleci.com/docs/2.0/configuration-reference/#deploy-deprecated

jaclyn-taroni:
The unspecified image will be a problem by the end of the month, so we'll need to get this in before then. We can deal with `deploy` separately.

sjspielman (approved):

jaclyn-taroni:
The only way to test this is to merge it, so I'm going to do that without CI finishing. If the YAML was incorrectly formatted, I think we'd know by now.

PR #1286, opened 2022-04-06
Update generate figures

sjspielman:
This PR addresses Issue #1261 and re-organizes `figures/generate_figures.sh` in order of figures as they will appear in the manuscript.

Deprecated items were removed from the generation script, and figures missing from the script were added in.

**Not ready for review.**

Issue #716, opened 2020-06-10
Updated analysis: Investigate handling of t-SNE in `transcriptomic-dimension-reduction` module


cbethell:
#### What analysis module should be updated and why?

As discovered in PR #704, the t-SNE results in `analyses/transcriptomic-dimension-reduction/results` do not appear to be reproducible as they change across operating systems (when ran with a seed set in Docker) but remain the same for each specific individual OS.

#### What changes need to be made? Please provide enough detail for another participant to make the update.

As noted on PR #704, the following has been determined:
- The seed function must be working as intended because the PCA and t-SNE results stay the same for each individual OS 
- The input files have not changed since the dimension reduction script was last ran and merged into master
- There only appears to be issues reproducing the PCA and t-SNE result files (and not the UMAP result files)
- There has been no evidence of updates to the Rtsne package

That being said, when looking into this issue, one might look into:
- Whether or not various versions of Docker can cause differentiating results across operating systems
- The `prcomp` function as the `Rtsne` function performs an initial reduction using `prcomp`

#### What input data should be used? Which data were used in the version being updated?

No additional input data should be needed.

#### When do you expect the revised analysis will be completed?

~ 2 days

#### Who will complete the updated analysis?

Likely someone at the CCDL

sjspielman:
I don't think the Docker version itself should make a difference, but for what it's worth I'm rebuilding my image now on this docker version - 
<img width="300" alt="Screen Shot 2020-06-11 at 9 28 09 AM" src="https://user-images.githubusercontent.com/4701111/84390766-0a208880-abc6-11ea-9912-a65e6d43b19e.png">

sjspielman:
**Appoach** : two docker using the aforementioned Desktop Docker version, builds as:
```
docker build -t pbtarocker-cache --pull
docker build -t pbtarocker-nocache --pull --no-cache
```
They give the exact same values as one another, but their PCA and tSNE scores differ from master. UMAP is the same as master. 

**Conclusion:** Likely unrelated to docker caching. This is expected and good. Next up: related to docker _version_? Will go download a slightly older Desktop docker and rebuild (from cache!) and check it out there. **Goal**: this also isn't the culprit.

PR #1397, opened 2022-05-11
Rerun transcriptomic dimension reduction module

jaclyn-taroni:
Breaking out something that's different in #1286 in the interest of better examining those changes.


jaclyn-taroni:
I will take a closer look at this tomorrow, but I think what's going on here is just changes in the metadata because this hasn't been run in a year. This module joins the metadata to the scores in a step prior to plotting (and those tables are committed to the repository). It would be better if the scores without metadata were saved and then the metadata was joined for plotting, but I think we'd be getting into recreational revision territory if we were change that at this point.

jaclyn-taroni:
I'll also note that t-SNE values changing is not new (and still a mystery): https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/716

The t-SNE values don't _go anywhere_ (e.g., into a plot), so I think it is sufficient to keep that issue open.

sjspielman:
> I will take a closer look at this tomorrow, but I think what's going on here is just changes in the metadata because this hasn't been run in a year. This module joins the metadata to the scores in a step prior to plotting (and those tables are committed to the repository)

I'll compare to how running the module looks w/ the `v19` data which, according to versioning dates, would have been the data most recently used to generate these files.

sjspielman (dismissed):
I have now re-run this with v21 data to see if a local run produces diffs compared to files in this PR. `X1` and `X2` are changing for t-sne and pca results, which is "expected" (#716), and there are no diffs for UMAP files. The _columns_ are all matching as expected, so it seems we're good to go here!

One small change before merging - a cryptic `Rplots.pdf` showed up in the directory and it should be removed. I'm pretty sure this is a long-standing bug in `cowplot` that has never really been tracked down.

sjspielman:
Noting this will satisfy a checkbox here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1455

sjspielman (approved):
Looks as expected! From skimming the tsv diffs, I see expected (🤷‍♀️) t-SNE number changes , but I don't see PCA/UMAP changes, other than the column rearrangements throughout which look good.

Issue #1368, opened 2022-05-05
Updates required post v22 histologies file release


jharenza:
Per my comments on #1365, some LGG `cancer_group` updates were implemented, but never transferred to `pbta-histologies.tsv`. This resulted in separate `cancer_group` in plots for `Low-grade glioma astrocytoma` (LGGA) and ganglioglioma but not PXA, pilocytic, and SEGA because they remained within LGGA:

```r
# v21 rna
v21 %>%
  filter((short_histology == "LGAT" | short_histology == "Ganglioglioma") & experimental_strategy == "RNA-Seq") %>%
  select(cancer_group, tumor_descriptor) %>%
  table()
                                     tumor_descriptor
cancer_group                          Initial CNS Tumor Progressive Recurrence Second Malignancy
  Ganglioglioma                                      36           8          3                 0
  Low-grade glioma astrocytoma                      200          31         12                 1
  Pilocytic astrocytoma                               0           1          0                 0
  Pleomorphic xanthoastrocytoma                       2           0          0                 0
  Subependymal Giant Cell Astrocytoma                 3           1          0                 0

#v22 rna
histology %>%
  filter((short_histology == "LGAT" | short_histology == "Ganglioglioma") & experimental_strategy == "RNA-Seq") %>%
  select(cancer_group, tumor_descriptor) %>%
  table()
                                     tumor_descriptor
cancer_group                          Initial CNS Tumor Progressive Recurrence Second Malignancy
  Diffuse fibrillary astrocytoma                      5           0          1                 0
  Ganglioglioma                                      36           8          3                 0
  Gliomatosis cerebri                                 1           0          0                 0
  Low-grade glioma astrocytoma                       72          16          6                 0
  Oligodendroglioma                                   1           0          0                 0
  Pilocytic astrocytoma                             106          15          4                 1
  Pleomorphic xanthoastrocytoma                       9           1          1                 0
  Subependymal Giant Cell Astrocytoma                11           1          0                 0

#v21 dna 
v21 %>%
  filter((short_histology == "LGAT" | short_histology == "Ganglioglioma") & experimental_strategy != "RNA-Seq") %>%
  select(cancer_group, tumor_descriptor) %>%
  table()
                                     tumor_descriptor
cancer_group                          Initial CNS Tumor Progressive Recurrence Second Malignancy
  Diffuse fibrillary astrocytoma                      1           0          0                 0
  Ganglioglioma                                      35          10          3                 0
[...]

PR #1401, opened 2022-05-11
Update broad histology and cancer group palettes, etc. to reflect v22 release

jaclyn-taroni:

We need to accommodate additional cancer groups PAST, PXA, and SEGA (#1368). Here's what I came up with:

⚠️  needs to get updated to use the v22 release


sjspielman:
I think this is about as good as we're going to get for separating out the colors!

figures/mapping-histology-labels.Rmd
```diff
@@ -192,6 +195,9 @@ cancer_group_df <- data.frame(
                                 "EPN",
                                 "Teratoma",
                                 "GNG",
+                                "PAST",
```
sjspielman, line 198: This might be more appropriately `PA` - https://www.webmd.com/cancer/what-is-pilocytic-astrocytoma

jharenza: I was using this http://oncotree.mskcc.org/#/home (type "pilocytic").

jharenza: From Peter Madsen: 
>I usually call it a JPA but i think that might be a bit outdated. I think PA is more appropriate

sjspielman (commented):
I wasn't expecting changes to oncoprint colors for groups that were not modified with LGG changes, eg. Oligodendroglioma and embryonal groups. They are all still different shades of grey, but I can't figure out why this would have changed other than the palette file was not previously up to date.

jharenza (dismissed):
I think this looks as expected!

sjspielman (commented):
This final output TSV looks like where we want to be. However, "example" legends printed out in the notebook aren't using the same finalized terminology, since the changes (like "other high-grade gliomas") are implemented _after_ the example legends are shown. This means the examples of how to make the legends won't match what ends up in the manuscript. 
I think it wouldn't be too much work to move the "here's how we can use `legend()`" bits to after some of the wrangling, and update legend code to use the `*_display` columns for text.

jharenza (approved):
Ok, I thought we were missing Other LGG/HGG colors when I updated the interaction plots, but I missed one code update, so this all looks good for the few plots I have generated so far!

jharenza:
Note, as of [294b408](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1401/commits/294b408c868dd9e2eda6b79aeb3961ebfe55a077), this passes CI since #1452 is stacked on this and passes.

jharenza (changes requested):
Hi @jaclyn-taroni. I started a branch for the immune deconvolution figure update and noticed that a lot of subtyped samples from LGG are actually in the `Other` cancer group display column, but it also seems like this may have been the case previously.

For example:
| cancer_group                                         | cancer_group_display | molecular_subtype           |
|------------------------------------------------------|----------------------|-----------------------------|
| Embryonal tumor with multilayer rosettes             | Other                | ETMR, NOS                   |
| Gliomatosis cerebri                                  | Other                | LGG, IDH                    |
| Diffuse fibrillary astrocytoma                       | Other                | LGG, BRAF V600E             |
| Diffuse leptomeningeal glioneuronal tumor            | Other                | GNT, KIAA1549-BRAF          |
| Embryonal tumor with multilayer rosettes             | Other                | ETMR, C19MC-altered         |
| Chordoma                                             | Other                | CHDM, conventional          |
| Neurocytoma                                          | Other                | EVN                         |
| Desmoplastic infantile astrocytoma and ganglioglioma | Other                | GNT, RTK                    |
| Chordoma                                             | Other                | CHDM, poorly differentiated |
| Neurocytoma                                          | Other                | CNC                         |
| Diffuse fibrillary astrocytoma                       | Other                | LGG, IDH                    |
| Glial-neuronal tumor NOS                             | Other                | GNT, other MAPK             |
| Diffuse fibrillary astrocytoma                       | Other                | LGG, RTK                    |
| Desmoplastic infantile astrocytoma and ganglioglioma | Other                | GNT, BRAF V600E             |
| Embryonal tumor with multilayer rosettes             | Other                | CNS Embryonal, NOS          |
| Diffuse fibrillary astrocytoma                       | Other                | LGG, wildtype               |
| CNS neuroblastoma                                    | Other                | CNS NB-FOXR2                |
| Diffuse fibrillary astrocytoma                       | Other                | LGG, KIAA1549-BRAF          |
| Oligodendroglioma                                    | Other                | LGG, wildtype               |
| Rosette-forming glioneuronal tumor                   | Other                | GNT, other MAPK, FGFR       |
| Glial-neuronal tumor NOS                             | Other                | GNT, KIAA1549-BRAF          |

So it looks like maybe we missed these LGG samples previously (maybe because of N when forming the original cancer group hex codes) and they should now go into Other low-grade gliomas so we have a complete group for displays. Likewise, the `Embryonal tumor with multilayer rosettes` above would go into `CNS Embryonal tumor` cancer group display. I think you can use `broad_histology_display` to work in this logic. Can you update `cancer_group_display` when `broad_histology_display`  == HGG, LGG, or embryonal to move those samples into the general display groups?

jaclyn-taroni:
> So it looks like maybe we missed these LGG samples previously (maybe because of N when forming the original cancer group hex codes) and they should now go into Other low-grade gliomas so we have a complete group for displays. Likewise, the `Embryonal tumor with multilayer rosettes` above would go into `CNS Embryonal tumor` cancer group display. I think you can use `broad_histology_display` to work in this logic. Can you update `cancer_group_display` when `broad_histology_display` == HGG, LGG, or embryonal to move those samples into the general display groups?


We can do this, but functionally I think this means that `cancer_group_display` will be somewhere between `broad_histology` and `cancer_group` in terms of specificity, instead of a representation of `cancer_group` where any cancer group that doesn't meet the sample size threshold is set to `Other` (as originally intended). 

We're probably going to need a table in `figures/README` that explains the mapping:

| `cancer_group_display` | Included cancer groups |
|-------------------------|-------------------------|

figures/palettes/broad_histology_cancer_group_palette.tsv
```diff
@@ -6,30 +6,32 @@ Benign tumor	Adenoma	#590024	NA	#b5b5b5	Benign tumor	Other	11	NA	NA	FALSE
...
-Diffuse astrocytic and oligodendroglial tumor	Oligodendroglioma	#ff80e5	NA	#b5b5b5	Diffuse astrocytic and oligodendroglial tumor	Other	2	Diffuse astrocytic and oligodendroglial tumor	#525252	TRUE
+Diffuse astrocytic and oligodendroglial tumor	High-grade glioma astrocytoma	#ff80e5	HGG	#ffccf5	High-grade glioma	Other high-grade gliomas	2	Diffuse astrocytic and oligodendroglial tumor	#ffccf5	TRUE
+Diffuse astrocytic and oligodendroglial tumor	Diffuse midline glioma	#ff80e5	DMG	#ff40d9	High-grade glioma	Diffuse midline glioma	2	Diffuse astrocytic and oligodendroglial tumor	#ff40d9	TRUE
+Diffuse astrocytic and oligodendroglial tumor	Diffuse intrinsic pontine glioma	#ff80e5	DIPG	#bf0099	High-grade glioma	Diffuse intrinsic pontine glioma	2	Diffuse astrocytic and oligodendroglial tumor	#bf0099	TRUE
+Diffuse astrocytic and oligodendroglial tumor	Oligodendroglioma	#ff80e5	NA	#b5b5b5	High-grade glioma	Other	2	Diffuse astrocytic and oligodendroglial tumor	#525252	TRUE
```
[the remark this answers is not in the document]
jaclyn-taroni: I think you and @jharenza are saying the same thing: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1401#pullrequestreview-1014389107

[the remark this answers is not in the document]
jashapiro: Yes, it does look that way! Missed that earlier comment.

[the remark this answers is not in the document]
sjspielman: Following up here a bit - I actually think we need to say _Other high-grade glioma_ (not plural!) and similarly for low-grade, for consistency with other groups which are written as singular and not plural.

jaclyn-taroni:
I'm working on this now, using `broad_histology` values to guide the collapsing into `Other` categories. `Oligodendroglioma` might complicate our plans a bit, as there are cases where samples with `Oligodendroglioma` in `cancer_group` would end up split into `Other low-grade gliomas` and `Other high-grade gliomas`.

I'll push what I have shortly so people can see for themselves.

jaclyn-taroni:
We're going to need a different approach for the oncoprint palettes as well. I'm pushing what I have with that commented out.

jashapiro:
> `Oligodendroglioma` might complicate our plans a bit, as there are cases where samples with `Oligodendroglioma` in `cancer_group` would end up split into `Other low-grade gliomas` and `Other high-grade gliomas`.

This seems to me like it might be the correct result, at least for the purposes of many of the display figures where we are distinguishing HGG and LGG. The limitation is that we will need to update any joining between the histology file and palette file to use both broad histology and cancer group. But we may need to do that anyway if we are going to make the "Other LGG" and "Other HGG" labels follow expectations in the case where we do not separately label all possible cancer groups within LGG/HGG.

jaclyn-taroni:
> Agree with above, this is expected and we have been joining on both in many of the module updates.

If that is the case, that assuages my concern that this is a bigger bite than I was hoping for. But I think it's still a bigger bite than I was hoping for because...

> what do you mean by updates for oncoprint? I was thinking we'd only be adding distinguishing colors for the LGAT tumor plot, but maybe you had something else in mind?

There will need to be a rewrite of how we handle cancer group colors for oncoprints. Now we're collapsing multiple cancer groups into, e.g., "Other low-grade gliomas." Previously, each cancer group under the "Other low-grade gliomas" umbrella would have gotten a random color from a greys palette in the oncoprint. Should we still be showing the individual cancer groups in the oncoprint? My assumption is yes = we need to rewrite how the oncoprint display palette is generated. And then what do we do about the "Low-grade glioma astrocytoma" --> "Other low-grade gliomas" in the oncoprint context? Maybe that's still fine to do.

jaclyn-taroni:
> I think that we might do something like: instead of the old HGG label, it'll now be "other HGG" and for any groups not colored in the LGG plot (non SEGA, pilocytic, pxa) and which were previously in the general LGG label, that would become "other LGGs". That seems easier than greys for every group, esp if we have a very small N in those groups?

Okay, to clarify – for the LGAT oncoprint, we'd expect the following groupings:

| `oncoprint_display` (same as `cancer_group_display` currently) | `cancer_group` |
|---|--|
| Other low-grade gliomas | Low-grade glioma astrocytoma <br> Gliomatosis cerebri <br> Diffuse fibrillary astrocytoma <br> Oligodendroglioma |
| Subependymal Giant Cell Astrocytoma | Subependymal Giant Cell Astrocytoma |
| Pilocytic astrocytoma | Pilocytic astrocytoma |
| Ganglioglioma | Ganglioglioma |
| Pleomorphic xanthoastrocytoma | Pleomorphic xanthoastrocytoma |  

Is that correct?

jharenza:
Looks right!

jaclyn-taroni:
Okay, functionally, I believe we no longer need an oncoprint-specific palette at this point. I will take it out in the interest of moving things along, but I imagine we may need to revisit when we revise the oncoprints to reflect the v22 release.

figures/mapping-histology-labels.Rmd
```diff
@@ -55,7 +55,7 @@ library(RColorBrewer)
 
 We should perform some minimal checking to make sure the provided release is _at least_ at version 21.
 ```{r}
-minimum_allowed_release <- 21
+minimum_allowed_release <- 22
 release_version <- as.numeric(stringr::str_match(params$release, "-v(\\d+)-")[[2]])
 if (release_version < minimum_allowed_release) {
   stop("This notebook should only be used with data release version 21 or above.")
```
sjspielman, line 61:
```suggestion
  stop("This notebook should only be used with data release version 22 or above.")
```

figures/mapping-histology-labels.Rmd
```diff
@@ -55,7 +55,7 @@ library(RColorBrewer)
 
 We should perform some minimal checking to make sure the provided release is _at least_ at version 21.
```
sjspielman, line 56:
```suggestion
We should perform some minimal checking to make sure the provided release is _at least_ at version 22.
```

figures/mapping-histology-labels.Rmd
```diff
@@ -103,7 +103,55 @@ included_labels_df <- histologies_df %>%
...
+)
+
+# To become "Other embryonal tumors"
+other_embryonal <- c(
+  "CNS Embryonal tumor",  # Has n >= 10, but we're trying to limit # of colors
+  other_cg_df %>%
+  filter(broad_histology == "Embryonal tumor") %>%
+  pull(cancer_group)
+)
+```
```
sjspielman, line 154: Can we print these three vectors at the end of this chunk? Helpful for following along in a notebook.

figures/mapping-histology-labels.Rmd
```diff
@@ -112,19 +160,22 @@ Check out the discussion on [#1174](https://github.com/AlexsLemonade/OpenPBTA-an
...
+                              "High-grade glioma",
+                              "Embryonal tumor",
+                              "Ependymal tumor",
+                              "Germ cell tumor",
+                              "Low-grade glioma",
+                              "Meningioma",
+                              "Mesenchymal non-meningothelial tumor",
+                              "Neuronal and mixed neuronal-glial tumor",
+                              "Tumor of cranial and paraspinal nerves",
+                              "Tumors of sellar region"),
```
sjspielman, line 178: Just pointing out this one is plural; make singular match?

figures/mapping-histology-labels.Rmd
```diff
@@ -301,25 +376,26 @@ broad_histology_order_df <- histologies_df %>%
...
   # And then add numeric column with the order
-  mutate(broad_histology = forcats::fct_reorder(broad_histology,
-                                                n,
-                                                .desc = TRUE) %>%
+  mutate(broad_histology_display = forcats::fct_reorder(broad_histology_display,
+                                                        n,
+                                                        .desc = TRUE) %>%
            forcats::fct_relevel("Benign tumor",
                                 "Other",
                                 after = Inf),
```
sjspielman, line 397: Something happened upstream where this isn't working and throws a warning, since the "Other" level isn't there - 
```
Warning: Problem while computing
`broad_histology_display = ... %>% ...`.
ℹ Unknown levels in `f`: Other
```

Though, I do see that "Other Tumor" _is_ a level here?

figures/mapping-histology-labels.Rmd
```diff
@@ -330,16 +406,16 @@ And now we're ready to add this to the palette data frame.
...
 ```
 
 ### Add `oncoprint_group` and `oncoprint_hex`
 
 For most plots that make use of the `cancer_group` palette, such as a box or violin plot, we will rely heavily on labels and therefore using the gray hex code for multiple groups will not be a problem.
 
-We will have four panels of individual oncoprints, where many `broad_histology` values will get grouped together into the `Other CNS` panel which you can see [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d31c927a27813ec0b8032fbe768002f31723636f/analyses/oncoprint-landscape/02-plot-oncoprint.R#L181).
-We can move this information into our palette data frame.
+We will have four panels of individual oncoprints, where many `broad_histology` values.
+We can move the information about which into our palette data frame.
```
sjspielman, line 418: Confused with the "about which". About which palette? About which oncoprint?

figures/mapping-histology-labels.Rmd
```diff
@@ -330,16 +406,16 @@ And now we're ready to add this to the palette data frame.
...
+            by = "broad_histology_display")
 ```
 
 ### Add `oncoprint_group` and `oncoprint_hex`
 
 For most plots that make use of the `cancer_group` palette, such as a box or violin plot, we will rely heavily on labels and therefore using the gray hex code for multiple groups will not be a problem.
 
-We will have four panels of individual oncoprints, where many `broad_histology` values will get grouped together into the `Other CNS` panel which you can see [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d31c927a27813ec0b8032fbe768002f31723636f/analyses/oncoprint-landscape/02-plot-oncoprint.R#L181).
-We can move this information into our palette data frame.
+We will have four panels of individual oncoprints, where many `broad_histology` values.
```
sjspielman, line 417: missing a verb?

figures/mapping-histology-labels.Rmd
jaclyn-taroni: This is what's in the histologies file, so if we were to do this it would mean changes later in the notebook as well, like "Low-grade astrocytic tumor" --> "Low-grade glioma"

figures/mapping-histology-labels.Rmd
jaclyn-taroni: I did something else that I prefer, but I believe still addresses your comment.

figures/mapping-histology-labels.Rmd
jaclyn-taroni: Tagging @jharenza  for thoughts

figures/mapping-histology-labels.Rmd
jaclyn-taroni: Should be fixed now

figures/mapping-histology-labels.Rmd
jharenza: This is probably just a miss - I think they maybe all used to be plural to go along with the WHO book, but then became singular as talking about one biospecimen. But, we had not been re-labeling plots to use the plural, though maybe we should (eg oncoprints have broad histology singular but the plot is of multiple tumors). I don't have a strong opinion, but if we want to be grammatically correct in plots, we probably should make `broad_histology_display` plural.

jharenza: @jaclyn-taroni and @sjspielman - just wanted to come back to this before running anymore modules/figures. I updated sample distribution plot [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/f0aa941c2e9b4f91ad33420eae3b7092e5a0fc88/figures/pdfs/fig1/panels/tumor_descriptor_proportion_panel.pdf) and have made some updates to interaction plots [here](url) (Note: legend still needs work to header the HGG and LGGs per @jashapiro's request), and so the palette as it stands looks good to me, but if we are to update `broad_histology_display` to plural, we should just do it here/now so it carries over to sample distribution and oncoprints (don't think it is anywhere else).

jaclyn-taroni: What about `cancer_group_display` then? Also make that plural? To me, it's not confusing either way so long as it is consistent and recoding fewer things is going to tend to be less error prone.

sjspielman: The main thing I was trying to point out with this comment was "Tumor of" vs "Tumors of." I think cancer groups with this naming scheme (tumor of XYZ, or "XYZ tumor") should all be singular or plural. Otherwise, anything is fine as long as consistent.

jaclyn-taroni: I went singular because there were fewer changes required.

jharenza (approved):
👍🏻

figures/mapping-histology-labels.Rmd
sjspielman: I'm not sure I see what/where you did this something else?

sjspielman: I agree with where you printed!

sjspielman (approved):
Looks ready!! I just have a couple small comments, but I don't need to see again unless you want me to. At your discretion - 

+ I might append `arrange(broad_histology_order)` to the pipeline ending on line 415 for easier viewing of the output.
+ I'd also suggest printing the _final_ `palette_df` at the end of the notebook, with all its expected columns.

Issue #1399, opened 2022-05-11
Create bash script for generating analysis files, separate out subtyping steps


jaclyn-taroni:
Related to a comment https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1286#issuecomment-1121072097

We're going to need to rerun subtyping ahead of release v22 (#1389). I'm proposing that we write a bash script that runs all the required molecular subtyping steps. It would probably live in the root of `analyses` and get documented in `analyses/README.md`.

I can start this this morning.

jaclyn-taroni:
New plan! This exists: [`scripts/run-for-subtyping.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e257cf5649a362137cbb7bbe82f172bb36b5d988/scripts/run-for-subtyping.sh)

But we have subtyping modules that are inconsistent on their reliance on `data/` vs. `analyses/` files _and_ that script currently does two things:

1. Generates analysis or derived files to be included in a release
2. Performs subtyping

I am going to split these two things up such that the procedure for generating a release is:

1. Generate analysis or derived files to be included in a release (new script!)
2. Add analysis or derived files to the release (i.e., `data/`)
3. Generate the subtypes, where to accomplish that I need to make the following updates:
    - Make sure all the subtyping modules consistently use `data/`
    - Make sure the `run-for-subtyping.sh` script runs steps with files that _don't get included in the data download_ (e.g., TP53 scores)

jaclyn-taroni:
Okay I have put something together on this branch https://github.com/jaclyn-taroni/OpenPBTA-analysis/tree/jaclyn-taroni/scripting-subtyping

#### Summary of changes

- The subtyping modules now all use `data/` files: https://github.com/jaclyn-taroni/OpenPBTA-analysis/commit/cfaed472c4ff95f8571562343f9418469bfcd24a, https://github.com/jaclyn-taroni/OpenPBTA-analysis/commit/efb9d6610cb00ff125fdef9249ba00baa0d21ced, https://github.com/jaclyn-taroni/OpenPBTA-analysis/commit/1bdc82327a971d89cefb76ebe2a5d78205e512bc, https://github.com/jaclyn-taroni/OpenPBTA-analysis/commit/273f2444127a5ba333e22bdb639954dd3fff2dbf
[...]

PR #1404, opened 2022-05-12
Remove short histology from TMB function grouping

jaclyn-taroni:
Related to https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1399#issuecomment-1124329628

⚠️  this is untested. I don't have a copy of the SNV database locally. 

On https://github.com/jaclyn-taroni/OpenPBTA-analysis/tree/jaclyn-taroni/scripting-subtyping, I created  [`scripts/generate-analysis-files-for-release.sh`](https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/f95853e50c92fa850a156f00cd914808e88cf098/scripts/generate-analysis-files-for-release.sh) to generate all the analysis files that are included in a release.

That includes the `snv-callers` module, which generates two TMB files (`pbta-snv-consensus-mutation-tmb-all.tsv` and `pbta-snv-consensus-mutation-tmb-coding.tsv`) included in the data download.  Those files contain `short_histology`. The `pbta-histologies.tsv` file, rather than the `pbta-histologies-base.tsv`, file is used in that module. (Using `pbta-histologies-base.tsv` would be consistent with how all other modules run via `scripts/generate-analysis-files-for-release.sh` are )

One approach is to remove `short_histology` from the TMB files like I'm doing here. We could instead incorporate logic that determines which histologies file to use. The right answer might be to do both, as I don't think it's necessary to include any metadata (outside of the experimental strategy) in the TMB files.


jaclyn-taroni:
This change is now included in #1420, so I'm going to close this.

PR #1415, opened 2022-05-19
Make sure non-ATRT/non-MB embryonal subtyping module uses data download

jaclyn-taroni:

Here, I'm changing the non-ATRT/non-MB embryonal subtyping module to make sure it uses analysis files `data/` (i.e., in the download), rather than files from results. In this case, there were two annotated files from individual CNV callers that _are not included in the data download_ that I've removed. This is because the results were inconclusive, so this seemed like a good way to keep the "always use `data/`" consistent.

I've uncovered some new issues with CI where we need to skip some chunks. I may not have found all of them quite yet, we'll see if this passes!

**Reviewers:** Please review [the `molecular-subtyping-embryonal` module in `master`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-embryonal) to ensure I didn't miss anything.


analyses/molecular-subtyping-embryonal/03-clean-c19mc-data.Rmd
```diff
@@ -106,7 +125,7 @@ histologies_df <- read_tsv(histologies_file,
 
 # copy number consensus file (analysis file)
 # TODO: get this from data download
-cn_consensus_seg <- file.path("..", "copy_number_consensus_call", "results",
+cn_consensus_seg <- file.path(data_dir,
```
sjspielman, line 128: The comment above this can be deleted, as the TODO is now done!

sjspielman (approved):
This looks all in order! Just one small comment about removing a now-deprecated comment.

PR #1418, opened 2022-05-19
Make sure MB subtyping uses data download

jaclyn-taroni:

Here, I'm changing the MB subtyping module to make sure it uses analysis files `data/` (i.e., in the download), rather than files from results.

**Reviewers:** Please review [the `molecular-subtyping-MB` module in `master`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-MB) to ensure I didn't miss anything.


sjspielman (approved):
Looks good, but make sure this line is as intended (or should it be the base file)?
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e72c11dbebed2377072df43150fd15f2ffa0262a/analyses/molecular-subtyping-MB/04-no-RNA-samples.R#L17

I don't need to see this one again.

jaclyn-taroni:
Good catch @sjspielman, thanks! The correct way to handle that (over and above https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1418/commits/14de872f559d38a57d54d8c4aa3816f39d2e4819) would be to control the histology file that's being used via a command line argument. But I think that'd be quite the rewrite because multiple things are hardcoded, so I'll leave it for now!

jharenza:
@jaclyn-taroni I am surprised  that last commit [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/2b49c0c0fa8e10565e6d35b1a0d73a2a1db8374d) worked because all short histology in the base file should be NA. 🤔

jaclyn-taroni:
That is not the case in v21.

jharenza:
Interesting, not sure how that snuck in. Pathology diagnosis should also work for MB, but good to note.

jaclyn-taroni:
I should say that `short_histology` is `NA` only _some of the time_ in v21. 

If it needs to get changed because the current release is incorrect re: `Medulloblastoma` values in `short_histology`, can you file an issue please?

PR #1414, opened 2022-05-19
Make sure HGG subtyping uses data download

jaclyn-taroni:
Splitting up changes described in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1399#issuecomment-1124329628. 

Here, I'm changing the HGG subtyping module to make sure it uses analysis files `data/` (i.e., in the download), rather than files from results. This also includes _skipping steps that were used to make earlier determinations re: subtyping_; these steps should not be rerun every time we want to make sure all subtyping is up-to-date.

**Reviewers:** Please review [the `molecular-subtyping-HGG` module in `master`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-HGG) to ensure I didn't miss anything.

PR #1413, opened 2022-05-19
Make sure EWS subtyping uses data download

jaclyn-taroni:
Splitting up changes described in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1399#issuecomment-1124329628. 

Here, I'm changing the EWS subtyping module to make sure it uses analysis files `data/` (i.e., in the download), rather than files from results.

**Reviewers:** Please review [the `molecular-subtyping-EWS` module in `master`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/molecular-subtyping-EWS) to ensure I didn't miss anything.

PR #1412, opened 2022-05-19
Rerun changes added

runjin326:
#### What scientific question is your analysis addressing?
Ran generate analysis for data release bash script and these are the changes after the re-run.

#### What was your approach?
Started an EC2 instance with 1T storage and 256G memory, pull the scripts, run the generate files script, and then push the generated files to S3 bucket.
*Note: I am not so sure about the memory requirement to run everything but the overall storage is ~60G so next time an instance with 100G should be enough.

#### What GitHub issue does your pull request address?
NA

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
No

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
Only gistic results were changed (expected) - and everything else remained the same.

#### What is your summary of the results?
NA

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

PR #1419, opened 2022-05-19
Update logic in modules where analysis files included in the data releases are generated

jaclyn-taroni:

There are a number of analysis files that are included in data downloads now. If you look at [the current version of `scripts/run-for-subtyping.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/d2f2c6e1786dabb0dcf5aa099a203c2baef5c8b9/scripts/run-for-subtyping.sh), you will see that some of these are currently run in that script. However, I will split up what is required for generating analysis files for release from what is required for subtyping in subsequent pull requests. 

All of the changes included here pertain to modules that will be run for generating analysis files. (Subtyping modules, where possible, will use `data/` [see: #1413, #1414, #1415, and #1418].) Because the analysis file generation will happen _prior to subtyping_, these steps still need to use the `pbta-histologies-base.tsv` file. The logic I am adding or modifying will allow us to do that in subsequent PR(s).

This might not be super clear at this point, so I'll go ahead and outline how data releases will work going forward after everything goes through:

1. We start a release that has all of the PBTA data files (i.e., upstream files) included
2. We run `scripts/generate-analysis-files-for-release.sh`, which should generate all the analysis files for a release and put them in `scratch/analysis_files_for_release`, and commit any changes to files that are included in the repository. (PR with this shell script coming very soon; #1412 is an example of the commit any changes part!)
3. We add all the analysis files to the release.
4. We run `scripts/run-for-subtyping.sh` and commit any changes to files that are included in the repository. (PR coming soon!)
5. We add `pbta-histologies.tsv` to the release


sjspielman:
Quick clarification here:

>2. We run scripts/generate-analysis-files-for-release.sh, which should generate all the analysis files for a release and put them in scratch/analysis_files_for_release, and commit any changes to files that are included in the repository. (PR with this shell script coming very soon;
https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1412 is an example of the commit any changes part!)
>3. We add all the analysis files to the release.

The files that are added to release are those in modules, or those in the `scratch/analysis_files_for_release`? My reading here is that those are the same files, but they are placed into `scratch/` for data release support. Am I reading this right?

jaclyn-taroni:
They are in both places but compiled into `scratch/analysis_files_for_release` for convenience (to support data release, as you say).

sjspielman (commented):
These are the modules I expect _may_ see changes in this PR, because they produce files towards the data release (based on [this file](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/doc/data-files-description.md)), and my notes about them...

+ `focal-cn-file-preparation`
	+ Uses metadata. Changes have been made.  ✅
+ `fusion-summary`
	+ Does not use metadata. No changes expected. ✅
+ `independent_samples`
	+ Uses metadata. Changes have been made.  ✅
+ `snv-callers`
	+ Uses metadata. The expected changes are made in #1420 ✅
+ `copy_number_consensus_call`
	+ Does not use metadata. No changes expected. ✅
+ `fusion_filtering`
	+ 	Uses metadata. Changes have been made.  ✅
+ `collapse-rnaseq`
	+ Uses metadata. **Changes may need to be made.** :warning:
	+ Actually something seems off here. In the current `scripts/run-for-subtyping.sh` , I see this is run as: `OPENPBTA_BASE_SUBTYPING=1 ../analyses/collapse-rnaseq/run-collapse-rnaseq.sh`, but I do not see a corresponding way to accept this arg in analysis script.

jaclyn-taroni:
> `collapse-rnaseq`
> 
>     * Uses metadata. **Changes may need to be made.** ⚠️
> 
>     * Actually something seems off here. In the current `scripts/run-for-subtyping.sh` , I see this is run as: `OPENPBTA_BASE_SUBTYPING=1 ../analyses/collapse-rnaseq/run-collapse-rnaseq.sh`, but I do not see a corresponding way to accept this arg in analysis script.

Yea, the current `scripts/run-for-subtyping.sh` is wrong. The only place the metadata gets used in this module is in this file: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e72c11dbebed2377072df43150fd15f2ffa0262a/analyses/collapse-rnaseq/00-create-rsem-files.R

Which not run via the shell script: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e72c11dbebed2377072df43150fd15f2ffa0262a/analyses/collapse-rnaseq/run-collapse-rnaseq.sh

Because we distribute one RSEM file for each selection strategy now. I think that Rscript _may be used upstream_, which is to say that I am afraid to touch it really.

sjspielman (approved):
👍

PR #1420, opened 2022-05-19
Update snv-callers to be able to be run from places besides the root of repo

### LINKED ISSUE #1406: Update the shell scripts in `snv-callers` to have a consistent working directory
@runjin326 is running the `generate-analysis-files-for-release.sh` script described here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1399#issuecomment-1124329628

The script fails at the SNV callers step with the following error:

```
Run SNV callers module for PBTA data
Setting up Database
python3: can't open file 'analyses/snv-callers/scripts/01-setup_db.py': [Errno 2] No such file or directory
```

Which stems from this line: https://github.com/jaclyn-taroni/OpenPBTA-analysis/blob/f95853e50c92fa850a156f00cd914808e88cf098/scripts/generate-analysis-files-for-release.sh#L63

And subsequently:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/c9b0f4456066f821f2c67983bdd822ed209d8b7a/analyses/snv-callers/run_caller_consensus_analysis-pbta.sh#L30

I _believe_ the source of the error is that there is nothing in the shell script `analyses/snv-callers/run_caller_consensus_analysis-pbta.sh` that causes it to have a consistent working directory. So, when it gets called from `scripts/generate-analysis-files-for-release.sh`, its working directory is `scripts/` which does not contain `analyses/snv-callers/scripts/01-setup_db.py`.

It would be consistent with other aspects of the project to set the shell script's working directory to where the file is located in the project, e.g., with:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/c9b0f4456066f821f2c67983bdd822ed209d8b7a/analyses/interaction-plots/01-create-interaction-plots.sh#L14

Then we'd change all of the paths to the scripts in `analyses/snv-callers/run_caller_consensus_analysis-pbta.sh` to be relative to `analyses/snv-callers`. If I recall correctly, all the scripts in the module already assume the paths to files (e.g., MAF files) supplied as arguments are relative to the root of the repository, so we won't have to do anything there.

Note: I haven't look at the TCGA shell script in `snv-callers`, but I'm willing to bet we'd have to do the same thing.

jaclyn-taroni:
⚠️  Review #1419 first! Unfortunately, I can't stack these in earnest because I'm using a fork!

Closes #1406. 

The shell scripts in `snv-callers` made an assumption that you'd run it from the root of the repository. That will be a problem when the script for generating analysis files (run from `scripts/`!!) is including in an upcoming PR. This PR introduces the following changes to the `snv-callers` module:

* Set the working directories of the PBTA and TCGA shell scripts to always be where the script is located, and update the paths in those scripts accordingly.
* Remove the assumption that input files will have paths relative to the root of the repository in the Rscripts `analyses/snv-callers/scripts/02-merge_callers.R` and `analyses/snv-callers/scripts/03-calculate_tmb.R`. We still need to make use of `rprojroot`, however, to make sure we're able to `source()` appropriately.


jaclyn-taroni:
Okay, two updates since filing:

* https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1420/commits/3b40a91b82ea75fdc24537dd545e41a30f944a3f removes `short_histology` from the TMB file. It's not necessary and we'd rather have folks joining `short_histology` from the appropriate histologies file, rather than using that one. I'm not 100% sure if all the downstream ramifications are going to get caught until we have v22 CI data with TMB files that do not contain `short_histology` to be frank!
* https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1420/commits/4e53522a097bf1971e9008569746b0366f50f688 adds functionality to use the base histologies file when running for a data release (related to #1419!)

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
```diff
@@ -8,60 +8,72 @@
 set -e
 set -o pipefail
 
+# Set the working directory to the directory of this file
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
 # The sqlite database made from the callers will be called:
-dbfile=scratch/snv_db.sqlite
+dbfile=../../scratch/snv_db.sqlite
```
sjspielman, line 15: When we made the S2 figures, this path got written in during review. To my knowledge figures are the _only_ thing that needs to use these databases after this module. So, we can either revert back the figures path, or save the databases there in the first place.
  
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e72c11dbebed2377072df43150fd15f2ffa0262a/figures/scripts/supp-snv-callers-panels.R#L26

analyses/snv-callers/run_caller_consensus_analysis-tcga.sh
```diff
@@ -8,14 +8,17 @@
 set -e
 set -o pipefail
 
+# Set the working directory to the directory of this file
+cd "$(dirname "${BASH_SOURCE[0]}")"
+
 # The sqlite database made from the callers will be called:
-dbfile=scratch/tcga_snv_db.sqlite
+dbfile=../../scratch/tcga_snv_db.sqlite
```
sjspielman, line 15: Same comment as before for the scratch path.

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
```diff
@@ -8,60 +8,72 @@
...
-cds_file=scratch/gencode.v27.primary_assembly.annotation.bed
-wgs_bed=scratch/intersect_strelka_mutect_WGS.bed
+cds_file=../../scratch/gencode.v27.primary_assembly.annotation.bed
+wgs_bed=../../scratch/intersect_strelka_mutect_WGS.bed
 
 # Set a default for the VAF filter if none is specified
 vaf_cutoff=${OPENPBTA_VAF_CUTOFF:-0}
 
+# If running during data generation, we want to use the BASE histologies file
+run_for_release=${OPENPBTA_BASE_RELEASE:-0}
```
sjspielman, line 28: For consistency with other modules, consider all caps `RUN_FOR_RELEASE`

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
```diff
@@ -8,60 +8,72 @@
...
+if [[ "$run_for_release" -eq "0" ]]
+then
+   histologies_file=../../data/pbta-histologies.tsv
+else
+   histologies_file=../../data/pbta-histologies-base.tsv
+fi
+
 # Unless told to run the plots, the default is to skip them
 # To run plots, set OPENPBTA_PLOTS to 1 or more
 run_plots_nb=${OPENPBTA_PLOTS:-0}
```
sjspielman, line 38: Same all caps suggestion, at your discretion.

analyses/snv-callers/util/tmb_functions.R
```diff
@@ -133,8 +133,7 @@ calculate_tmb <- function(tumor_sample_barcode = NULL,
     dplyr::group_by(
       #TODO: Make this column passing stuff more flexible with some tidyeval maybe
       Tumor_Sample_Barcode = tumor_sample_barcode,
-      experimental_strategy,
-      short_histology
+      experimental_strategy
```
sjspielman, line 136: The associated docs here still say:
```
 # Returns:
 # A calculated total region size, and TMB for the given Tumor_Sample_Barcode,
 # returned as a single row data.frame. `experimental_strategy`, `short_histology`
 # columns are carried along.
```

That said, I'm not sure why short_histology is being removed here?

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
jaclyn-taroni: This is for consistency _within this module_. Otherwise, I'd agree.

analyses/snv-callers/util/tmb_functions.R
jaclyn-taroni: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1420#issuecomment-1131933494

> It's not necessary and we'd rather have folks joining `short_histology` from the appropriate histologies file, rather than using that one. 

Or put another way, one could imagine it getting out of sync with the histologies file.

sjspielman: Woops my bad!!! Moving parts 🙃

jaclyn-taroni: Yep I'll still update the doc. I mostly quote because I'm lazy.

jaclyn-taroni: Updated the docs in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1420/commits/6e075b210df5b5b63bac84518dbde6713a13e544

analyses/snv-callers/run_caller_consensus_analysis-pbta.sh
jaclyn-taroni: I split the difference (kind of) in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1420/commits/054616bd42bde245e06335da4a68792633699186. It's a good call for modules to have their own directories in `scratch/`, rather than just hoping their are no file name collisions in `scratch/`! But I also think `snv-callers` is more appropriate than `snv-callers-panels`.

sjspielman (approved):
👍

PR #1410, opened 2022-05-18
Add a module for counting git contributions to the repository

jaclyn-taroni:
This can definitely wait until a release goes in, so filing as a draft for now.  For the moment, I'm going to quote from the `README.md` I'm adding...

> The purpose of this module is to count the git contributions to this repo in two ways:
> 
> 1. Count the number of analysis modules – directories in `analyses/` that are included in the manuscript, specifially – that individual contributors have authored commits in. 
>   You can find that information in `results/module_contribution_counts.tsv` and a _less summarized_ version that lists all contributors to a module in `results/module_contributors.tsv`.
> 2. Count the number of commits to the entire repository individual contributors have authored.

PR #1423, opened 2022-05-19
Docs changes related to #1399

jaclyn-taroni:
Here I'm adding the docs changes required as part of #1399. Specifically:

- Updating `analyses/README.md` to reflect the fact that subtyping modules are using `data/`.
- I've also added changes to `analyses/README.md` that reflect what's going on in #1410 (i.e., what's included in the manuscript). That's getting added here to avoid conflicts! 
- Adding `scripts/README.md`, with emphasis on instructions for data release preparation.
- Removing the subtyping instructions from the main README (because they're now in `scripts/README.md`) and updating the TOC accordingly.


sjspielman (commented):
Overall, this is pretty clear. My first impression of something that would be really helpful to include, I think, is an additional column in `analyses/README.md` that indicates if the module produces files that end up as part of data releases. Currently we document whether it produces result files that get _consumed by other analyses_, but not specifically whether the files go into the data release.

sjspielman:
> do you think we need to address your comment in this PR or can we file an issue for later?

I don't mind if we file an issue for later, but I do think it definitely needs to be done at some point. I can go ahead and open it.

sjspielman (approved):
I opened #1431 for additional documentation needs, so this is good to go!

PR #1425, opened 2022-05-19
Group R package installation to try to speed up build time

jaclyn-taroni:
We have some instances of failure in CI at the MB molecular subtyping due to

```
Error in library(medulloPackage) : 
  there is no package called ‘medulloPackage’
Calls: suppressPackageStartupMessages -> withCallingHandlers -> library
Execution halted

Exited with code exit status 1

CircleCI received exit code 1

```

Here's an example: https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/7000/workflows/42052230-8031-4020-bba4-1798181a9869/jobs/7864

I'm adding a step in the Dockerfile where we explicitly try to load `medulloPackage` to see if we can figure out what's going on.


sjspielman:
I remember @davidsmejia had some ideas (though I don't know what they were!) about how to speed up data downloads. These ideas might come in handy at this time.

sjspielman:
@jaclyn-taroni I'm a bit confused here - the title for this PR and the main comment for this PR seem to have different goals. The title suggests runtime (and comparing this CI runtime to other successful runs, I don't see any specific speedups during setup), but the comment suggests just debugging `medulloPackage`.

What do you want me to focus on for review?

jaclyn-taroni:
Fair! So I'd say that the dependencies I am grouping together should be considered stable at this point in the project. So by grouping things together now, we'd _potentially expect savings in the future_ as we add things to the end of the Dockerfile. I am also addressing the issue I initially posted by explicitly checking to see if we can load `medulloPackage`: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1425/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R226-R227


Because this passes CI, I think we can reasonably expect that I included everything I needed to include. So review for: is there anything obviously amiss here?

Dockerfile
```diff
@@ -37,14 +37,17 @@ RUN pip3 install \
   "wheel==0.34.2"
 
 # Install java
-RUN apt-get -y --no-install-recommends install \
+RUN apt-get update && apt-get -y --no-install-recommends install \
```
sjspielman, line 40: Line 9 already runs `RUN apt-get update`, so this may not be needed?

Dockerfile
```diff
@@ -76,142 +79,102 @@ RUN wget https://github.com/samtools/htslib/releases/download/1.9/htslib-1.9.tar
...
     lattice \
+    leafem \
+    leafpop \
+    lsa \
+    lwgeom \
+    mapview \
     MASS \
     Matrix \
     mgcv \
+    MM2S \
```
sjspielman, line 129: This might be one of the ones to _not_ group together. This is what's in master now - 
```
# Molecular subtyping MB
RUN R -e "remotes::install_github('d3b-center/medullo-classifier-package', ref = 'e3d12f64e2e4e00f5ea884f3353eb8c4b612abe8', dependencies = TRUE, upgrade = FALSE)" \
    && ./install_bioc.r MM2S
# More recent version of sva required for molecular subtyping MB
RUN R -e "remotes::install_github('jtleek/sva-devel@123be9b2b9fd7c7cd495fab7d7d901767964ce9e', dependencies = FALSE, upgrade = FALSE)"
```

Conceptually `MMS2` is coded there like `TCGAbiolinks` is coded here. 

Either way nothing is _missing_, so this is just an organizational aspect at your discretion.

sjspielman (approved):
A few places just to check, but this is looks good to me!

PR #1432, opened 2022-05-25
Generate script modified to include seg with status

runjin326:
#### What scientific question is your analysis addressing?
add `consensus_seg_with_status.tsv` from `scratch` to `scratch/analysis_files_for_release`

#### What was your approach?
Add one line:
`cp ${scratch_dir}/consensus_seg_with_status.tsv ${compiled_dir}`

#### What GitHub issue does your pull request address?
NA

#### Which areas should receive a particularly close look?
NA

#### Is there anything that you want to discuss further?
No

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
Yes

#### What types of results are included (e.g., table, figure)?
NA

#### What is your summary of the results?
Although I didn't run the entire module but I re-ran this particular line and it ran as expected.

#### Reproducibility Checklist

- [ ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

Issue #1426, opened 2022-05-19
Plan: Road to v22


jaclyn-taroni:
The purpose of this issue is to outline the order in which things need to be reviewed and run prior to the release of v22. 

### 1. Reviews before release

Pull requests that are marked as `blocking release` and `review before release` should be reviewed & ideally merged in the following order. cc: @sjspielman 

- [x] Refactoring the Docker image build and fixing the push to Dockerhub in CI: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1425

#### Analysis files generation

- [x] #1400
- [x] #1419
- [x] #1420 – depends on #1419 
- [x] #1421  ✅ Approved – depends on #1419 and #1420

- [x] #1412 includes the changes that come from running `scripts/generate-analysis-files-for-release.sh` 

#### Subtyping modules need to use `data/`

- [x] #1413
- [x] #1415
- [x] #1418

- [x] #1414

#### Subtype script & docs

- [x] #1422  ✅  Approved  – Running these changes is blocked by the above section, so let’s review and merge those first! 
- [ ] #1423 — this doesn’t strictly need to be reviewed _prior to release_ because this is just docs, but it would make sense conceptually.

### 2. Running subtyping

Once #1422 is in, we can use the PBTA data and analysis files from v22 to run `scripts/run-for-subtyping.sh`. We’ll need to create a branch to commit back any changes to files in the repository (akin to what’s happening with #1412).

Then we’ll upload `pbta-histologies.tsv` to S3 (related: #1424) .

### 3. Creating CI files

Once all the v22 files are on S3, we’re ready to generate the CI files (#1369).   Here’s what we need to do:

1. Create a branch to update the `create-subset-files` module as required to create the CI files. I’m going to call this `v22-ci` for illustrative purposes.

  Specifically, we need to update this line to `release-v22-20220505` reflect the new release: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e72c11dbebed2377072df43150fd15f2ffa0262a/analyses/create-subset-files/create_subset_files.sh#L10

  Then we should check to make sure it runs locally without issue with:

  ```
  RUN_LOCAL=1 ./analyses/create-subset-files/create_subset_files.sh
  ```

2. Start an AWS instance with at least 128 GB RAM
[...]

PR #1365, opened 2022-05-04
release V22

jharenza:
#### What scientific question is your analysis addressing?
This updates the histologies file with MB WGS samples as "To be classified", which were previously missed 

#### What was your approach?
1. Updated https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/molecular-subtyping-MB/04-no-RNA-samples.R for subtypes to say `MB, To be classified` instead of `To be classified`
2. Updated https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/46a9d5c0656742b79aa472eadbd78f8bdd720fe4/analyses/molecular-subtyping-pathology/pathology_free_text-subtyping-lgat.Rmd to recode `LGG, subtype` --> `SEGA, subtype`
3. Created `base-histologies.tsv` from v21 and reran `molecular-subtype-integrate` to get `pbta-histologies.tsv`.

#### What GitHub issue does your pull request address?
#1207 

#### Is there anything that you want to discuss further?
Notes:
- `pbta-histology-base.tsv` changes daily, and when rerunning using what I thought might be the base @kgaonkar6 used, I realized this module was recently rerun in [this PR](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1247) by @runjin326, perhaps using a more recent version of the file. Therefore, it looks like there are many more diffs than there really should be here.
- To make sure none of the data aside from the few samples needing subtype updates changed, I simply created a new base file using v21 `pbta-histologies.tsv` minus the columns for `harmonized_diagnosis` and `cancer_group`. I put this bit of code at the very top of the script, but I think we may want to comment it out before merge? Similarly, I also added some QC into this, but maybe it is fine because this is the _last_(?) version of the histologies file for OpenPBTA?
- After checking the diffs in the histologies file within this PR on GitHub, I realized some code which was implemented a while back for cancer groups never made it into the histologies file. This only pertains to the LGAT samples, but that means that it affects multiple figures. 
- Below are the diffs in cancer groups for LGAT `broad_histology` with the code we had in place (also in `01-integrate-subtyping.nb.html`:

```r
v21 %>%
  filter(short_histology == "LGAT") %>%
  select(cancer_group, experimental_strategy) %>%
  table()
                                     experimental_strategy
cancer_group                          RNA-Seq WGS
  Diffuse fibrillary astrocytoma            0   1
  Low-grade glioma astrocytoma            244 234
  Pilocytic astrocytoma                     1   2
[...]

PR #1433, opened 2022-05-26
V22 subtyping (1/N)

jharenza:
I am going to stagger these so the changes aren't so overwhelming. For this PR, I am running only GSVA (no changes) and then TP53:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e99c13550956c95fa83e9c811515e04a01a8c3d4/scripts/run-for-subtyping.sh#L12-L16

#### What was your approach?
Ran the lines above, but had some errors in the TP53 module which required some code changes:
1. Remove https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e99c13550956c95fa83e9c811515e04a01a8c3d4/analyses/tp53_nf1_score/run_classifier.sh#L90-L92 and add this to the data release. This script was using `pbta-histologies.tsv`, and we really don't need to rerun if we keep the file from the previous step of generate data analysis files. See new PR to maintain this file #1432. It is also now included in #1365 
2. Update `focal-cn-file-preparation/02-add-ploidy-consensus.Rmd` to pull the above seg status file from `/data`
3. Add code around https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/e99c13550956c95fa83e9c811515e04a01a8c3d4/analyses/tp53_nf1_score/run_classifier.sh#L114-L118 to skip these steps for subtyping, since the plots require fields created post-subtyping

#1426 - partial

code updates

No, everything now runs as expected. File changes are minor.

yes

TP53 module file changes

as expected

#### Documentation Checklist


analyses/molecular-subtyping-embryonal/00-embryonal-select-pathology-dx.R
```diff
@@ -18,7 +18,8 @@ output_file <- file.path(root_dir,
 # as we identified in 00-v17-embryonal-select-pathology-dx.Rmd are:
 path_dx_terms<- c(
   "Supratentorial or Spinal Cord PNET",
-  "Embryonal Tumor with Multilayered Rosettes"
+  "Embryonal Tumor with Multilayered Rosettes (ETMR)",
+  "Embryonal Tumor with Multilayered Rosettes (ETMR);High-grade glioma/astrocytoma (WHO grade III/IV)"
```
jaclyn-taroni, line 22: Is _this_ expected?

jharenza: Yes, I added due to a new dx

jharenza: I can revert this back until I get to embryonal. I did it right away so I didn't forget about it, but let me revert that.

jaclyn-taroni (approved):
👍🏻

Issue #1207, opened 2021-11-30
V22 data release


jharenza:
From the below conversation, we will push one more release to include the missed _to be classified_ MB subtypes into the histology file. This will require rerunning the molecular subtyping to create the new histology file.

I don't think we want to be in a situation where the release and the supplemental data are out of sync, but perhaps the new release can be the very last thing we do before we cut a release of the codebase.

__Originally posted by @jaclyn-taroni in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1206#issuecomment-982880365__

PR #1435, opened 2022-05-26
V22 run MB subtyping (3/N)

jharenza:
Run MB subtyping

#1207 #1426 

yes

No change in subtypes, only clinical data

#### Documentation Checklist


jaclyn-taroni:
Are all the changes in `analyses/molecular-subtyping-MB/input/subset-mb-clinical.tsv` expected?

jharenza:
> Are all the changes in `analyses/molecular-subtyping-MB/input/subset-mb-clinical.tsv` expected?

It looks like these are because of column order changes. I didn't think that would have a major effect, but let me go back and match the order of the new base to previous so it is easier to detect changes. Also, `parental_aliquot_id` will no longer exist in base since it was removed from our database.

jharenza:
@jaclyn-taroni looks a bit cleaner in [0e6f745](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1435/commits/0e6f745218411a186685fa5f0259085b3cc7a810). Expected changes due to base file.

jaclyn-taroni (approved):
👍🏻

PR #1444, opened 2022-05-26
V22 LGG subtyping (11/N)

jharenza:
run v22 LGG subtyping

#### Documentation Checklist


analyses/molecular-subtyping-LGAT/results/lgat_subtyping.tsv
```diff
@@ -52,11 +53,12 @@ sample_id	Kids_First_Participant_ID	Kids_First_Biospecimen_ID_RNA	Kids_First_Bio
 7316-1744	PT_3KK1F95W	BS_0X9EGHY2	BS_03SSN1N2	LGG, wildtype
 7316-1745	PT_2BJF83GQ	BS_02NZT8CE	BS_J5J0EA7Y	LGG, NF1-germline
 7316-175	PT_DEYCW197	BS_HV8PFRBR	BS_6XP91JM2	LGG, wildtype
-7316-176	PT_2PRV78FX	BS_E1CGF3KV	BS_17AXPP1Y	LGG, MYB/MYBL1
+7316-176	PT_2PRV78FX	BS_E1CGF3KV	NA	LGG, MYB/MYBL1
+7316-176	PT_2PRV78FX	NA	BS_17AXPP1Y	LGG, To be classified
```
jaclyn-taroni, line 57: These two BSIDs have different `pathology_free_text_diagnosis` values, which is why they are not being joined here despite having the same `sample_id` and `Kids_First_Participant_ID`. My questions are: 1. Is the difference in `pathology_free_text_diagnosis` values something we should be concerned about? 2. If not, should we be including `pathology_free_text_diagnosis` in what we join on? https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/705377ab2bd3727a7e852526aa0514d726bab334/analyses/molecular-subtyping-LGAT/04-LGAT-compile-subtypes.Rmd#L80

analyses/molecular-subtyping-LGAT/results/lgat_subtyping.tsv
```diff
@@ -40,7 +40,8 @@ sample_id	Kids_First_Participant_ID	Kids_First_Biospecimen_ID_RNA	Kids_First_Bio
 7316-1661	PT_F27SA15G	BS_YA8RBGD8	BS_2NN6ME5A	LGG, NF1-germline
 7316-1664	PT_ADTP7EHN	BS_K7MVS0Z7	BS_KG1YMTA4	LGG, KIAA1549-BRAF
 7316-1668	PT_Q25V9Y6M	BS_55P8HPYD	BS_9HDPJT21	LGG, KIAA1549-BRAF
-7316-167	PT_VZTTTP3Q	BS_E4QK839R	BS_16FT8V4B	LGG, IDH
+7316-167	PT_VZTTTP3Q	BS_E4QK839R	NA	LGG, To be classified
```
jaclyn-taroni, line 43: Same issue as with sample `7316-176`.

analyses/molecular-subtyping-LGAT/results/lgat_subtyping.tsv
```diff
@@ -72,7 +74,8 @@ sample_id	Kids_First_Participant_ID	Kids_First_Biospecimen_ID_RNA	Kids_First_Bio
 7316-1890	PT_2WVW55DA	BS_QAG1RDXN	BS_NYVVN7SD	GNT, RTK
 7316-1926	PT_5CNZ47XV	BS_JMHG7GCN	BS_QM1Z8R38	LGG, KIAA1549-BRAF
 7316-1928	PT_6MWPJ96F	BS_VR7FR1NE	BS_M8EBKZDG	LGG, KIAA1549-BRAF
-7316-193	PT_TXFYRDP4	BS_1G3WYEN3	BS_5Z4XQC9X	LGG, BRAF V600E
+7316-193	PT_TXFYRDP4	BS_1G3WYEN3	NA	LGG, To be classified
+7316-193	PT_TXFYRDP4	NA	BS_5Z4XQC9X	LGG, BRAF V600E
```
jaclyn-taroni, line 78: Same issue as sample `7316-176`.

jaclyn-taroni (commented):
There are some potential issues with `pathology_free_text_diagnosis` for 3 samples that we'll need to discuss before this goes in.

jharenza:
Yes. I noticed this yesterday and filed https://github.com/PediatricOpenTargets/ticket-tracker/issues/338

I see two options - I can manually update these at the base file level or wait until the data warehouse is updated and do a new pull of the file, but the latter will take longer, so I will probably do the former. I don't think this would affect any other modules based on the diagnoses with issues, but would you want me to rerun all modules?

jaclyn-taroni:
I don't think we need to rerun additional modules.

If we don't need to join on `pathology_free_text_diagnosis` (and maybe we make that decision based on if that impacts any results), removing that seems like it would have the benefit of being a quick change that is tracked in the code base.

jharenza:
I did update the code not to join on it, but unfortunately, we do need it for glialneuronal subtyping, which occurs within this module.

analyses/molecular-subtyping-LGAT/results/lgat_subtyping.tsv
jharenza: The differences look to be an upstream database error, in which one bs_id is getting the `pathology_diagnosis` in `pathology_free_text_diagnosis`. Odd.

jaclyn-taroni: Do you have an idea of how long it would take to fix upstream? We also have to factor in the time it would take to check and make sure there are no other differences in the base file.

jharenza: I was suggesting I manually update this in the base file I have so that we do not have to wait.

jharenza: See https://github.com/d3b-center/D3b-codes/pull/70. I updated the base file in S3 as well.

Issue #1403, opened 2022-05-12
Update HGG and LGG wording in figures


sjspielman:
From [this comment](https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/268#issuecomment-1125303759), the palette file should be updated and associated figures re-generated. This should be done alongside adding in the additional LGG groupings to relevant panels, which should naturally occur when the metadata and palette files are accordingly updated for v22.

Specifically, 

+ `broad_histology_display` should be updated:
    + "Low-grade astrocytic tumor" --> "Low-grade glioma"
    + "Diffuse astrocytic and oligodendroglial tumor" --> "High-grade glioma"
+ `cancer_group_display` should be updated:
    + "High-grade glioma astrocytoma" --> "Other high-grade gliomas" 
    + "Low-grade glioma astrocytoma" --> "Other low-grade gliomas"

The following panels need to be remade, and associated figures recompiled

- [x] Figure 1B
- [ ] Figure 2 A, C
- [ ] Figure 3 A, D, E
- [ ] Figure 4D
- [ ] Figure 5 A,B,C
- [ ] Figure S3B
- [ ] Figure S4 A,B

Related to https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1368.

sjspielman:
This is effectively closed by #1401 which updated `cancer_group_display` to use correct wording.

PR #1447, opened 2022-05-31
V22 sample distribution analysis rerun (14/N)

jharenza:
Rerun sample distribution analysis post v22 histology file generation

#### What was your approach?
Reran entire module. Added script 04 to the shell script, since it was missing.

I think these results look as expected - particularly, more LGG samples broken down by type of LGG. One question is, given https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1403, do we want to make this modification more upstream, as in #1446 for the cancer group? cc @sjspielman 

yes

#### Documentation Checklist


jaclyn-taroni (approved):
I admittedly do not fully understand this question:

> I think these results look as expected - particularly, more LGG samples broken down by type of LGG. One question is, given https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1403, do we want to make this modification more upstream, as in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1446 for the cancer group?

But these updates look fine, and I expect we could update this module to use the palette if needed.

jharenza:
> I admittedly do not fully understand this question:
> 
> > I think these results look as expected - particularly, more LGG samples broken down by type of LGG. One question is, given #1403, do we want to make this modification more upstream, as in #1446 for the cancer group?
> 
> But these updates look fine, and I expect we could update this module to use the palette if needed.

Yes, we discussed making this change to palette only

PR #1450, opened 2022-06-01
v22 CI fixes

jaclyn-taroni:
The problem was that we were skipping the header of TMB files, which meant we did not have any participant IDs associated with those files. Once we got to the `intersect` step here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L279-L280

We had no participant IDs, and similarly, we get back no participant IDs here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/addac5b6d1b1a0003592a03e268e8877b7ca4d40/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L283

So when we attempted to sample `polya_for_subset` here: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/addac5b6d1b1a0003592a03e268e8877b7ca4d40/analyses/create-subset-files/01-get_biospecimen_identifiers.R#L288

It threw the error noted here https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1426#issuecomment-1143943704

### DIFF
--- a/analyses/create-subset-files/create_subset_files.sh
+++ b/analyses/create-subset-files/create_subset_files.sh
@@ -5,9 +5,6 @@
 set -e
 set -o pipefail
 
-# Set the working directory to the directory of this file
-cd "$(dirname "${BASH_SOURCE[0]}")"
-
 # Set defaults for release and biospecimen file name
 BIOSPECIMEN_FILE=${BIOSPECIMEN_FILE:-biospecimen_ids_for_subset.RDS}
 RELEASE=${RELEASE:-release-v22-20220505}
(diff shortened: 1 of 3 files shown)


analyses/create-subset-files/create_subset_files.sh
```diff
@@ -31,6 +31,8 @@ cd "$script_directory" || exit
 # generated via these scripts
 FULL_DIRECTORY=../../data/$RELEASE
 SUBSET_DIRECTORY=../../data/testing/$RELEASE
+# If run subsetting only, we need to make this directory
+mkdir -p $SUBSET_DIRECTORY
```
jharenza, line 35: This actually needed brackets because it was still throwing an error, so I updated in my branch and added logic to make if does not exist.

${SUBSET_DIRECTORY}

PR #1457, opened 2022-06-21
V22 chromothripsis figure update to use cancer group display

jharenza:
❗ stacked on #1401 

Updates to use `cancer_group_display` for chromothripsis plot

#### What was your approach?
Updated code as above

#1368 

Just make sure I did not miss any

No

Yes

Updated figure 3 panel

#### Documentation Checklist


sjspielman (commented):
This looks good, and numbers make sense. I'll approve this once 1401 goes in.

In addition, one small change might be beneficial even though slightly beyond the scope here. The x-axis in the plot has angled labels but definitely doesn't need them - the labels should all fit fine at the default, so the angling just adds visual noise. Can we keep those at default? This just involves modifying line 102.

sjspielman (commented):
"Approved!" 🚀 

@jharenza which do you think will be easier for you - should I approve now but add a `don't merge` label, or wait to approve until #1401? Both fine with me!

jharenza:
> "Approved!" 🚀
> 
> @jharenza which do you think will be easier for you - should I approve now but add a `don't merge` label, or wait to approve until #1401? Both fine with me!

I may want to rerun after #1401 just in case we add samples to any groups, so either way works. definitely can use a dont merge regardless!

sjspielman:
Ok, if there's potential for a full re-review, I won't approve yet! I realize we should definitely make #1401 a blocker instead of (or on top of) a label since that will be a little more protected.

jharenza:
@sjspielman this one is also ready for re-review

figures/scripts/supp-S3-panels-BCD.R
```diff
@@ -7,6 +7,9 @@
 library(tidyverse)
 library(ComplexHeatmap)
 
+# set seed for jitter plots
+set.seed(1979)
```
sjspielman, line 11: :tada:

jharenza:
> Looks good! One requested change, something that I recently realized more generally and have been trying to update when I come across it. We hadn't previously set a seed with many of our jitter plots, like 3C-D, so these are not stable. Can you add a `set.seed(whatever you want!)` in that section of that `scripts/supp-S3-panels-BCD.R`, and re-run?

Done - I realized we do have one figure being copied over for Figure 3C, but I think we did not set seed in #1471 

```# Copy scatter plot from the chromothripsis module - Panel C
cp ${analyses_dir}/chromothripsis/plots/04-breakpoint-data/count_chromothripsis_cnv_and_sv_breaks_scatterplot.pdf pdfs/fig3/panels/count_chromothripsis_cnv_and_sv_breaks_scatterplot.pdf
```

Do you want me to a) copy that over or b) set seed + rerun the whole module/copy over? c) else?

sjspielman (dismissed):

sjspielman:
> Do you want me to a) copy that over or b) set seed + rerun the whole module/copy over? c) else?

Oh I totally missed that one of those was copied over! Yes, definitely feel free to do that here; I dismissed my review.

sjspielman (approved):
Ah, I see it was a _scatterplot_ that was being copied over, so we don't actually need the seed for manuscript figures. Certainly doesn't hurt for the non-manuscript jitter plots anyways. All good!

jharenza:
Ahhh right!

Issue #1455, opened 2022-06-21
Updated approach to running analyses and figures


sjspielman:
This approach emerged from discussion with @jaclyn-taroni and @jashapiro about a strategy to ensure analysis modules and figures are up-to-date and in sync, especially if/when there is a future data release.

Currently, we have *three scripts* that run aspects of PBTA:
+ `scripts/generate-analysis-files-for-release.sh`. This script runs analysis modules that need to be run to _enable_ subtyping.
+ `scripts/run-for-subtyping.sh`. This script runs molecular subtyping modules and produces the `pbta-histologies.tsv` file from the `base` version.
+ `figures/generate_figures.sh`. This script (currently under review in #1454) runs, for _each figure_ in the manuscript, the any relevant analysis modules needed to make the given figure, and then either runs associated figure scripts in `figures/scripts/`, or copies over figure panels created within analysis modules.

Ideally, `figures/generate_figures.sh` should _only_ be generating figures (running scripts in `figures/scripts/` or copying over figure panels created within analysis modules) without also running analysis modules. Therefore, we should do the following:

1. Create a fourth script, called something like `scripts/run-analysis-modules.sh`. This script will run all analysis modules that either i) were not run for subtyping, ii) were partially run for subtyping but now need to be fully run (e.g. GSVA), or iii) were run for sybtyping and need to be run again with the subtyped data.
2. Remove all analyses from `figures/generate_figures.sh` so that this script _only_ generates figures under the assumption analysis modules have been run and are ready to be used for figure generation.
3. Document changes accordingly, likely in `scripts/README.md`.

An additional item to consider is that we might want to name these scripts with prefixes `01_`, ..., `04_`. However, since `generate_figures.sh` lives in `figures/` and not in `scripts/`, we may only do `01` - `03` and leave `generate_figures.sh` as it is currently named.

Edit - one point I forgot to add is that, to build this up, we want to be running _each analysis module **individually**_ as PRs.

sjspielman:
> You need disease label information to find recurrent calls: 

I marked `focal-cn-file-preparation` as redundant because, even though the histology file is read in, the only columns used are `"Kids_First_Biospecimen_ID"`  and `"short_histology"`:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6071c5fe89afeb656ade29051a9ceeb17a0effb7/analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd#L87-L97


[...]

PR #1460, opened 2022-06-22
Re-run chromosomal instability with V22

sjspielman:
Partly addresses #1455 

This PR re-runs the chromosomal instability module with the V22 data release. The only diffs are plots which are not part of the manuscript, and one of these plots is entirely new. There is also a `sessionInfo()` diff, and I can confirm I ran this in the Docker. Note I also found/fixed a typo in the README.


jashapiro (approved):
This looks good overall. I had one main question looking at the plot changes, which is: What happened to the Metastases? I did some investigation, and it looks like it is a coding change, where metastatic tumors that have a pathology diagnosis that includes a primary tumor type are now assigned to that tumor type in `short_histology`, making the Metastases group now fall below the minimum count threshold.

So with that explanation satisfying my curiousity, I am happy to approve this!

PR #1461, opened 2022-06-22
Re-run tp53_nf1_score module with v22

sjspielman:
This PR re-runs the `tp53_nf1_score` analysis module with V22. The primary diffs are in the LGAT files, which makes sense for what we expect to see changing here.


sjspielman:
I've re-run this again now that #1401 has been merged. All `.tsv` diffs are just minor numerical tolerance differences, and overall results remain stable. I also updated the PDFs pointed out in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1461#pullrequestreview-1017131762 to use the correct palettes.

jashapiro (approved):
LGTM

PR #1448, opened 2022-05-31
EDIT: V22 rerun all tables

jharenza:
#### What scientific question is your analysis addressing?
<s>Rerun Tables 1, S1 as QC for the new v22 histologies file.</s>

Rerun all tables for manuscript with the following updates:
- Remove duplicates from Table 1 #1459 
- Add new hypermutator to Table 2 #1481 
- Add missing Key Resources #1407 

#### What was your approach?
<s>Reran only tables 1 and S1 from the Rmd of `output_tables.Rmd`. This probably can be kept as a draft until we get the remaining analysis updated for v22, then rerun this entire notebook.</s>

Made the above updates and reran all tables

#### What GitHub issue does your pull request address?
#1510 

#### Which areas should receive a particularly close look?
Code updates

#### Is there anything that you want to discuss further?
<s>Looks like this checks out to me.</s>

Note: Since we added `[BS_F0GNWEJJ](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1448/commits/e5739b97cd16c3f1c66b1b6c9ec5b6da0fd14fe8)` to the hypermutator table and this sample is a metastatic NBL, not a HGG and therefore not subtyped, I added a small notebook to assess the same (as HGG) _TP53_ status and additionally, MYCN status. NBLs are most broadly categorized as MYCN amp or non-amp. This sample was found to not have any _TP53_ alterations, although it does have a high _TP53_ score. It is not MYCN-amplified via consensus CNV analysis, and additional manual inspection shows no amplification at 2p. (attached below).
c67d0ab2-0dbf-4d77-93c1-54f3273b5ce0 controlfreec ratio log2

I further asked, and Sharon Diskin's germline lab team has shared their germline annotation and pathogenicity analysis for these six patients. Therefore, I added a column specifically for germline pathogenic variant (this was previously pulled from pathology report manual review by me). Here, we add one more pathogenic variant than was previously described via pathology reports. I will create an issue in the manuscript repository to add a short method for germline annotation/pathogenicity and add them as authors.

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?
<s>yes, for Tables 1, S1</s>

Yes

#### What types of results are included (e.g., table, figure)?
Updated tables

#### What is your summary of the results?
NA

#### Reproducibility Checklist

- [x] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [x] This analysis has been added to continuous integration.

#### Documentation Checklist

[...]

PR #1462, opened 2022-06-22
Update table 1 to remove duplicates & update ms tables to v22

### LINKED ISSUE #1459: Table 1 double counts histologies
The code linked below is used to derive Table 1 in the manuscript, but this code double counts samples where there are WGS and RNA-seq biospecimens. 

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/6071c5fe89afeb656ade29051a9ceeb17a0effb7/tables/output_tables.Rmd#L78-L86

I believe adding a `distinct(broad_histology, molecular_subtype, sample_id)` line to the above before counting should address the issue and return the proper count.

jashapiro:

The Table 1 generation script was double counting some samples that had both WGS and RNA-seq specimens

#### What was your approach?

Added a `distinct()` step to count each `sample_id` only once.

Also updated to v22

closes #1459 

The change to v22 and or rerunning the whole notebook resulted in a number of other changes. I have not checked these in detail to be sure that they are as expected, partly because they are all xls files, and partly because I am not sure I know all the expectations!

Yes

New Tables for the manuscript.

Most of the counts in table 1 went down, as expected!

#### Documentation Checklist


jharenza:
@jashapiro I had #1448 started for rerunning of all tables with v22 release so that we could make sure some of the changes would come through as expected. I can close that one if want to keep this one open. However, note, that the full script will need to be rerun again once some other modules are updated (mut sigs, survival), since those results are included in tables.

PR #1463, opened 2022-06-22
Re-run GSEA module with V22

sjspielman:
This PR re-runs the `gene_set_enrichment_analysis` module with V22. A couple changes had to be made along the way - one of the result tsv files got a bit large (61 MB), and while this won't break git, git doesn't like it and complained. So I updated the code so that all result files are now RDS, and we're back down to 49 MB. 

I additionally updated (but did not run!) the relevant figures script to be able to read in this new format.


jashapiro (commented):
I think this is fine, but I would have taken a slightly different approach to the size issue: I would have changed all output files to `*.tsv.gz` which maintains compatibility outside R and would invoke automatic compression by `write_tsv()`. I'll note also that `write_rds()` does not compress by default (in contrast to `saveRDS`), so if we are concerned about size, we should probably include compression.

sjspielman:
> I'll note also that write_rds() does not compress by default (in contrast to saveRDS), so if we are concerned about size, we should probably include compression.

Did not know this about its default! Good to know. I'll update this with `.tsv.gz`.

sjspielman:
@jashapiro This has been updated with `.gz` compression.

PR #1430, opened 2022-05-24
Updated survival notebooks with LGG broad histology

sjspielman:
Closes #1416 

This PR updates and re-runs the two survival notebooks (tp53 and immune) that consider an LGG grouping. The grouping should be the `broad_histology`, NOT the `cancer_group`. Results are broadly consistent.

PR #1464, opened 2022-06-22
Re-run survival with V22

sjspielman:
This PR re-runs the `survival-analysis` module with the V22 data release. Note that it was also recently updated with correct LGG groupings in #1430 also with V22, but that PR did not update the `survival-analysis_subtypes` notebook for V22, but this PR runs it all.


jharenza (approved):
When I saw this come into email for review, I was having deja vu and thought, I just did this.. but then saw we didn't run the one subtype notebook. I see the same results as before. Agree with @jashapiro and then 👍

sjspielman:
`sessionInfo()` added to the missing notebook!

PR #1465, opened 2022-06-22
Update Table 1 with distinct counts

### LINKED ISSUE #1459: Table 1 double counts histologies

jashapiro:
Per earlier comments, this PR (stacked on #1448, and targetted to  merge there) updates the code to generate manuscript Table 1 to only count each sample once,  rather than each biospecimen. Since this will require reruns as other modules finish, I did not include updates to the other tables, but I wanted to have the correct counts available for updating the manuscript per https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/275

Merging this will allow the completion of #1448 to close #1459


jharenza (approved):
Looks good to me

PR #1466, opened 2022-06-23
V22 rerun survival figures

jharenza:
❗ stacked on #1464 

This PR updates all associated survival figures for Figures 4 and 5.

#### What was your approach?
I reran the following scripts:
```bash
Rscript --vanilla scripts/fig4-hgg-kaplan-meier.R
Rscript --vanilla scripts/fig4-hgg-subtype-forest-plot.R
Rscript --vanilla scripts/fig5-forest-plot.R
```

and only ran the forest plot for `scripts/fig4-tp53-telomerase-panels.R`

#1368 

<s>I added the couple of scripts into the `generate-figures.sh` even though I know you have the script updated in PR, just in case those are missing currently (didn't check your PR!).</s disregard now that the PR is merged

I had to update the x-axis for the tp53 telomerase forest plot in order to get LGG group to show up, and then further updated it to be able to get the full confidence intervals to show up. The axis went back to scientific notation, but don't think this is a big deal. Let me know what you think.

See above

Yes

Updated KM plot and 3 forest plots

NA

#### Documentation Checklist


jharenza:
@sjspielman this is ready for review. ran with your latest updates in #1464

sjspielman:
We're going to need to take a step back here... Because this was "semi-stacked" on #1464, there are quite a few diffs in the survival module itself. But, since the module itself doesn't produce any manuscript figures, this PR should only be re-running, and updating where needed, these four scripts - `fig4-hgg-subtype-forest-plot.R`, `fig4-tp53-telomerase-panels.R`, `fig4-hgg-kaplan-meier.R`, and `fig5-forest-plot.R`. I'd therefore close this PR and open a new one which *only* runs (and updates where needed) the scripts of interest.

Also, we need to be cognizant that `fig4-tp53-telomerase-panels.R` is making other figures that are exported but will need to be updated separately, which I see is part of the list in #1368. But, I don't see that that script was actually _re-run_ here since I would have expected more byte-size PDF diffs (?). Even though we only want to capture the forest plot section for this PR, make sure the full scripts are run in case of bugs for the new PR.

jharenza:
>Also, we need to be cognizant that fig4-tp53-telomerase-panels.R is making other figures that are exported but will need to be updated separately, which I see is part of the list in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1368. But, I don't see that that script was actually re-run here since I would have expected more byte-size PDF diffs (?). Even though we only want to capture the forest plot section for this PR, make sure the full scripts are run in case of bugs for the new PR.

Yeah, noted above, I only ran the FP part of the code, but will open a new PR and run full scripts.

PR #1469, opened 2022-06-27
Add GitHub PAT to Docker builds in CircleCI to allow us to install R packages from GitHub without rate limiting

jaclyn-taroni:
We've been having issues with Docker images failing to build because we can't install R packages from GitHub. [Here's an example](https://app.circleci.com/pipelines/github/AlexsLemonade/OpenPBTA-analysis/7255/workflows/91fde3ca-d003-4bad-828f-249134a1a778/jobs/8148), and here's the relevant/informative bit of error message:

```
Error: Failed to install 'unknown package' from GitHub:
  HTTP error 403.
  API rate limit exceeded for 35.231.15.12. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

  Rate limit remaining: 0/60
  Rate limit reset at: 2022-06-27 17:26:38 UTC

  To increase your GitHub API rate limit
  - Use `usethis::browse_github_pat()` to create a Personal Access Token.
  - Use `usethis::edit_r_environ()` and add the token as `GITHUB_PAT`.
```

I've created a personal access token (PAT), which is scoped for public access only (i.e., [no scope](https://docs.github.com/en/developers/apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes)) for an account with very little public information, and added it as an environmental variable in CircleCI.

~This PR adds a step in the CircleCI workflow adding the PAT to `.Renviron`. It is my hope that this will circumvent us frequently getting rate limited.~


.circleci/config.yml
```diff
@@ -299,7 +299,8 @@ workflows:
   version: 2
   test-and-deploy:
     jobs:
-      - run_analyses
+      - run_analyses:
+          context: openpbta-docker
 
       - deploy:
```
jashapiro, line 305: I think we want this context only in the deploy step
```suggestion

      - deploy:
          context: openpbta-docker
```

.circleci/config.yml
```diff
@@ -299,9 +299,10 @@ workflows:
   version: 2
   test-and-deploy:
     jobs:
-      - run_analyses
+      - run_analyses:
```
jaclyn-taroni, line 302:
```suggestion
      - run_analyses
```

jaclyn-taroni:
Going to close in favor of #1477

PR #1471, opened 2022-06-28
Re-run chromothripsis module with V22

sjspielman:
This PR re-runs the `chromothripsis` analysis module with V22 data. 

Notable changes:
+ The figure jitter points have changed since no seed is set (but these aren;'t used for manuscript), but comparison of points shows ~identical results. 
+ The file [analyses/chromothripsis/results/shatterseek_results_per_chromosome.txt](https://github.com/AlexsLemonade/OpenPBTA-analysis/compare/master...sjspielman:chromothripsis-v22?expand=1#diff-052cca7aa93aafae832609bc46cb953a1b89c4f035f755094873b22ec13c9fec) has some changes which are almost entirely related to numerical precision. There is no seed set when generating this file, which _may_ be worth further discussion in and of itself.
+ There is one change in an HTML file in a `geom_point()` call, which is almost certainly because the HTML file was not previously up-to-date with the Rmd! This diff is therefore fine.


jharenza:
>The figure jitter points have changed since no seed is set (but these aren;'t used for manuscript), but comparison of points shows ~identical results.

Some of these are in supplemental (S3 C-D), and I didn't have them marked to re-generate - ha! But, since they are being generated here, can you also copy over to the supp figure folder?

jharenza:
> > Some of these are in supplemental (S3 C-D), and I didn't have them marked to re-generate - ha! But, since they are being generated here, can you also copy over to the supp figure folder?
> 
> The manuscript figures are not made in this module, but instead in `figures/scripts/supp-S3-panels-BCD.R`.

Ok, I see. Do you want me to tack that onto #1457 then? Wasn't planning since it was theoretically not to change due to lack of the cancer group labels, but we may as well run all figures at this point?

sjspielman:
> Ok, I see. Do you want me to tack that onto https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1457 then?

Sure, that should be fine! I do see this in #1368 as - 
> Figure S3B CNV genomic view via https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/supp-S3-panels-BCD.R

So we can check it off there too.

PR #1472, opened 2022-06-28
Deprecate packages to "avoid" rate limit

sjspielman:
This PR comments out a deprecated analysis module, and no longer installs some of that specific module's R package dependencies to speed up CI and/or possibly avoid a github API rate limit.


jashapiro (approved):
LGTM, but maybe change `DEPRECATE` to `Deprecated` for a bit less yelling.

jashapiro:
I think we can pause/close this PR in favor of #1477 which incorporates the same changes plus the PAT fix (and it does work when a cache is used and it can finish in under 3 hours)

sjspielman:
@jashapiro I'll add a don't merge tag here then.

jashapiro:
Closing now that #1477 is in.

PR #1473, opened 2022-06-29
Try using devtools for github installs

jashapiro:
This is a different approach to fixing the problem with rate limiting we have been running into as discussed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1469.

Based on [an issue in the `remotes` package](https://github.com/r-lib/remotes/issues/210) it appears we may be able to get around the API rate limit by going back to the `devtools` package for github installation, as that package does not use the github API, but rather pulls from raw github content. So let's see if it works!


jashapiro:
Nope

PR #1476, opened 2022-06-29
Re-run mutational signatures module with V22

sjspielman:
This PR re-runs the mutational signatures module, specifically by issuing `CNS_FIT_ONLY=1 bash run_mutational_signatures.sh`. Note that I had to clean up a bit of syntax in the run script due to weird spacing, and I fleshed out some of the comments in the run script for clarity.

There were several bugs that cropped up when re-running the notebooks that had to do with V22 changes, including how Oligodendrogliomas are handled and there _appears_ to now be an additional hypermutator (`BS_F0GNWEJJ` - this could use some more eyes to confirm it's really supposed to be there!). As a consequence, I had to update the figures in this PR as well since that was the only way to address the bugs and get the overall module to run. 

Overall, there are no strong changes in results except for the new (?) hypermutator sample. Note that I also added `set.seed()` to the sina and jitter plot code for reproducibility moving forward as this was missing.


jharenza:
Great, I will take a look at this now @sjspielman . Note that we did see `BS_F0GNWEJJ` before, but this was an NBL metastasis and we chose to focus our discussion only on HGGs, so did not include. I thought we added that to the plot code in some way, but will take a look!

jharenza:
> > Note that we did see BS_F0GNWEJJ before, but this was an NBL metastasis and we chose to focus our discussion only on HGGs, so did not include.
> 
> Ah, this is coming back to me! I guess we're seeing now that we need to explicitly filter it out in this module.

In thinking more about this, I think we ought to include. I am scanning for predisposition and germline variants, but we will need to assess TP53 status in this sample.

jharenza (commented):
I just have one question, and that is whether you want to update `exposures_presence_barplot.pdf` here since this PR reruns all figures and the text in the facet strips and the x-axis is a bit smushed.

sjspielman:
> I just have one question, and that is whether you want to update `exposures_presence_barplot.pdf` here since this PR reruns all figures and the text in the facet strips and the x-axis is a bit smushed.

Thanks for catching this, will do!

jharenza (approved):
beautiful!

sjspielman:
@jharenza I realize that, since this PR did all the mutational signatures figures too as a bonus treat, I should also copy them into `figures/pdfs/`! I'm re-requesting a review here to make sure those copies look ok to you!

jharenza (approved):
plots look good 👍

PR #1477, opened 2022-06-30
Apply GitHub PAT via Docker secrets

jashapiro:
This PR builds off the work in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1469 in an attempt to get past github rate limiting. 

The work here takes an environment variable that is provided by CircleCI project secrets containing a GitHub PAT, and applies it as a Docker secret. Docker secrets are then stored in a file in the`/run/secrets` directory, which it seems must be mounted by the steps that will use it. So here I am also adding a small script that reads from that file before running `remotes::install_github()`

Two potential followup ideas:
-  I could the script take multiple github repositories (tagged with commit refs as `user/repo@commit_id`) to install multiple files at once, but I thought leaving separate run steps probably still made sense overall, and certainly for now.
- To reduce the repetition of `--pat_file /run/secrets/gh_pat` in the Dockerfile, I could make that the default for the script. 

With the current project settings as I submit this,  the `GH_PAT` env variable is set to a `testing` value, so this should fail with an authentication error at the first call to `install_github.r`

In local testing, this worked as expected. We will se how things go in CI!


jaclyn-taroni (approved):
Looks good! Thanks for the docs updates (and for teaching me about `docopt`).

PR #1515, opened 2022-07-06
Re-generate S5 panels

### LINKED ISSUE #1492: Re-generate S5 panels
With the new cancer groups and palette in V22, this script needs to be re-run and updated as needed:
[`supp-S5-panels.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/supp-S5-panels.R). This script will generate all Figure S5 panels.

sjspielman:
Closes #1492 
This PR re-runs the script `figures/scripts/supp-S5-panels.R` to re-generate the three S5 panels with V22.

The scatterplots don't have any changes (same R and P-values), but the ROC plot shows a rather different AUC for the shuffled group of 0.32. This is lower than one expects the lower threshold of an AUC to be, but may not be unreasonable for the shuffled group.


sjspielman:
Having a look at the relevant script, it appears a random seed is set for `numpy` here:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/bfb7d798d37ab490392821f57c8bcf7b39a5a092/analyses/tp53_nf1_score/01-apply-classifier.py#L56

And a numpy method is used to perform the shuffling as well. I'm going to explore this a but further to see if this seed is indeed properly applied.

sjspielman:
@jharenza I've re-run this with the updated TP53. I have a question for you as well - what do you think of the current axis labels for the scatterplots? Specifically, I was thinking maybe they should include the +1, e.g. `TERC log(FPKM+1)`. What do you think?

jharenza:
> @jharenza I've re-run this with the updated TP53. I have a question for you as well - what do you think of the current axis labels for the scatterplots? Specifically, I was thinking maybe they should include the +1, e.g. `TERC log(FPKM+1)`. What do you think?

Sure, we can do that, since TERT has a lot of 0s.

sjspielman:
@jharenza to be clearer, this calculation was already performed, but the label didn't reflect it.

jharenza:
> @jharenza to be clearer, this calculation was already performed, but the label didn't reflect it.

Oh! yes, then we should update the label!

jharenza (approved):
Looks great!

sjspielman:
This code does not run through CI so we can merge before checks.

PR #1518, opened 2022-07-07
Update cancer display abbreviations

### LINKED ISSUE #1516: Update cancer group abbreviations for V22
In #1401, we updated cancer group display names and palette for V22. However, we did not sufficiently update cancer group abbreviations _which are only used in Figure 1B_! Currently, HGG and LGG abbreviations are used for "Other high-grade glioma" and "Other low-grade glioma," respectively, _but these abbreviations match broad histology groups!_ We therefore need a different abbreviation for these two cancer groups to avoid ambiguity between LGG representing a cancer group and a broad histology.

This will involve updating the palette generating file, but I do not think we'll strictly need to update any other figures other than 1B, since no other figures use this column.

Any thoughts on something like oLGG and oHGG? CC @jharenza @jaclyn-taroni @jashapiro

sjspielman:
Closes #1516 
This PR updates cancer group abbreviations for `Other low-grade glioma` and `Other high-grade glioma`, (edited) from LGG ->  Other LGG and HGG -> Other HGG, respectively.


jaclyn-taroni:
How bad an approach would "Other LGG" and "Other HGG" be? Would that work in Figure 1B at all?

jaclyn-taroni (approved):
👍🏻

PR #1519, opened 2022-07-08
Update oncoprint figures for V22

### LINKED ISSUE #1484: Update Figure 2: oncoprint plots
Figure 2 panels need to be updated by re-running, and updating as potentially needed for V22 cancer group and palette changes, this script: [`fig2-oncoprint-landscape.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/fig2-oncoprint-landscape.R).

sjspielman:
Closes #1484 

This PR updates and re-runs the script `figures/scripts/fig2-oncoprint-landscape.R` to make oncoplots with V22. The code was specifically updated to include V22 cancer groups such that `cancer_group_display` column is used for labeling, and `cancer_group_hex` is used for colors (since the `oncoprint_hex` column had been removed from palette). In addition, because the `oncoprint_include` column had also been removed from the palette file, I only filtered on `oncoprint_group` to identify which cancer groups to include. If this strategy captures too many cancer groups, we may need to incorporate some hard-coding. Finally, I added in a check in the script to error out if the expected `scratch/` directory files are present.


jharenza (approved):
I think this looks good! Not too many groups, imo.

jaclyn-taroni:
@sjspielman this needs another pass with the legend ordering: https://github.com/sjspielman/OpenPBTA-analysis/blob/bd63b4e60bdc85b2b9d74c0c5e39bb784150dfa6/figures/scripts/fig2-oncoprint-landscape.R#L216-L244

The legend order is hard-coded because it is intended to mirror whatever order the cancer groups are in in the figure itself, mostly because it can help a reader if they are having trouble distinguishing between two cancer groups with similar-ish colors. Granted, the comment here (https://github.com/sjspielman/OpenPBTA-analysis/blob/bd63b4e60bdc85b2b9d74c0c5e39bb784150dfa6/figures/scripts/fig2-oncoprint-landscape.R#L212-L214) could definitely be more explicit!

Taking the low-grade oncoprint as an example:

primary_only_lgat_oncoprint

The cancer group order, by my read, is: Pilocytic astrocytoma, Other low-grade glioma, Ganglioglioma, Pleomorphic xanthoastrocytoma, Subependymal Giant Cell Astrocytoma

Taking a look at the legend:

Screen Shot 2022-07-09 at 9 09 07 AM

If we were to put those next to each other as is, I think it'd potentially trip people up.

We should also drop DIPG from the HGAT legend. 

I believe the other two legends are all set, but I'd double (triple?) check just to be sure!

jaclyn-taroni (changes requested):
I'm going to request changes, since we have one approval and I don't think this should get merged until the legend issue is addressed!

sjspielman:
@jaclyn-taroni definitely missed that legend ordering! I was so excited to get the names right 😂  but you are absolutely correct we need to have the right order. On it!

jaclyn-taroni (approved):
👍🏻

PR #1521, opened 2022-07-08
Compiled Figure 3 for V22

### LINKED ISSUE #1488: Re-compile Figure 3
Figure 3 will need to be recompiled.

Note from #1393:
> There is some funky compilation-induced text in Figure 3D as well as overly thick lines in 3E. The compilation requires minor tweaks.

sjspielman:
Closes #1488 

This PR updates the compiled Figure 3 with V22 panels. In addition, I have re-made the `exposures_sina_IQR.pdf` plot at a different size to facilitate compilation. I copied this panel into `figures/pdfs/`. As such, that notebook was re-run, and there is a result file difference that emerged specifically because I changed cancer group label _wrapping_ to fit the new cancer group names.


jharenza (approved):
Looks good!

PR #1526, opened 2022-07-12
Re-run TP53 module with correct flag

sjspielman:
In #1461, we re-ran the TP53 module with V22. However, as detailed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1515#issuecomment-1181699404 that run did _not_ specify the flag `OPENPBTA_TP53_FIGURES=1` which enforces that data in the data release, instead of from the `collapse-rnaseq` module, be used. 

This PR fixes that oversight by running the module on V22 data _with_ this flag. The most notable change is that polya AUC goes from 0.32 --> 0.38.

I have also added some information to the module README and the run script to document the different flags that can be used with the module.


analyses/tp53_nf1_score/README.md
```diff
@@ -20,6 +20,18 @@ The analysis can be run with the following (assuming you are in the root reposit
 bash analyses/tp53_nf1_score/run_classifier.sh
 ```
 
+The script contains several additional flags that can be specified:
+- `OPENPBTA_BASE_SUBTYPING`: 
+  - if value is 1, then uses pbta-histologies-base.tsv for subtyping.
```
jharenza, line 25:
```suggestion
  - if value is 1, then uses `pbta-histologies-base.tsv` for subtyping. 
```

analyses/tp53_nf1_score/README.md
```diff
@@ -20,6 +20,18 @@ The analysis can be run with the following (assuming you are in the root reposit
 bash analyses/tp53_nf1_score/run_classifier.sh
 ```
 
+The script contains several additional flags that can be specified:
+- `OPENPBTA_BASE_SUBTYPING`: 
+  - if value is 1, then uses pbta-histologies-base.tsv for subtyping. 
+  - if value is 0 (DEFAULT), runs all modules with pbta-histologies.tsv
```
jharenza, line 26:
```suggestion
  - if value is 0 (DEFAULT), runs all modules with `pbta-histologies.tsv`
```

jharenza (approved):
Looks good - just minor suggestions to backtick the file names, but otherwise, 👍

sjspielman:
This CI timed out, but the relevant module `tp53_nf1_score` completed CI successfully before time out, so we can merge.

PR #1527, opened 2022-07-12
Re-generate snv-callers panels with V22

### LINKED ISSUE #1494: Re-generate  snv-callers panels for Figure S2
This script will need to be re-run and updated for V22: [`supp-snv-callers-panels.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/supp-snv-callers-panels.R)

Note that this needs to be done on AWS.

sjspielman:
Closes #1494 
This PR updates the snv-callers panels used in Figure S2 with V22 data by running the script [`supp-snv-callers-panels.R`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/scripts/supp-snv-callers-panels.R) on an EC2 instance (working with the SQL databases requires ~22 GB loaded into memory). In general there aren't any salient diffs (often no diffs!) with the new data release. The one plot with no changes at all was `pbta_vaf_cor_matrix.tiff` (recall that plot is `.tiff` because it is too large as a PDF), but all the PDFs are updated even if only just re-generated.


jharenza (approved):
👍

PR #1530, opened 2022-07-12
V22 Compiled Figure 2

### LINKED ISSUE #1485: Re-compile Figure 2
Once oncoprint panels are remade in #1484, we will need to re-compile panels to create the final Figure 2.

sjspielman:
Closes #1485 
This PR adds compiled versions of Figure 2 (oncoprint), both PDF and PNG. 
For this one, I ended up switching the order of the alteration and sex legends, so now the alterations are on top. I think this is more consistent with plots where sex is annotated _below_ the oncoplot.


jharenza:
Hi @sjspielman! Thanks for adding this.

Can you add a title to the legends for "Germline sex estimate" or "Sex" and "Alteration"? 

I also just noticed, and did not catch this in the old compiled figures, that we do not have the row annotation labels for "Sex" and "Cancer group". Can we add these? We only have two annotation rows and people can probably assume what's what there, but I think it should be clearly labeled on the oncoprint.

sjspielman:
Sure, I can add those back in. I actually agree with you here. I can't track it down now, but previous conversations at some point (can't remember exact details) had led to those being _removed._ I'll relabel!

jharenza:
I wonder if it was because they had underscores- I recall we couldn't remove them in the plot. Maybe the intention was to label during compilation? 🤷🏻‍♀️

sjspielman:
@jharenza I added back in legend titles.

sjspielman:
This PR does not modify code so does not need to run through CI.

PR #1534, opened 2022-07-13
Re-run survival with updated TP53 for V22

sjspielman:
Closes #1529 
This PR re-runs the survival module now that #1526 has been merged in. 
There do not appear to be any diffs (yay!), but I suggest we still merge this to track that the module _was_ re-run with the (albeit marginally) updated TP53 scores.

PR #1535, opened 2022-07-13
re-ran script

### LINKED ISSUE #1528: Update TP53 figures
Once #1526 is merged to update the TP53 module correctly, Figure 4 panels that use TP53 scores will need to be updated again. Strong changes are not expected, but we should still update. This just entails running the script `figures/scripts/fig4-tp53-telomerase-panels.R`.

sjspielman:
Closes #1528 
This PR re-runs `figures/scripts/fig4-tp53-telomerase-panels.R` to make use of the very slightly updated (aka the 12th decimal differs for a handful of scores) TP53 results merged in with #1526.
As in #1534 we have no diffs, but this PR can serve to track that the figures were indeed re-generated after #1526.


jharenza (approved):
These all look good to me!

sjspielman:
This PR does not modify code, and the code it ran to produce results does not run through CI either. Merging before checks.

PR #1538, opened 2022-07-13
Recompile Figure S4 for V22

### LINKED ISSUE #1499: Re-compile Figure S4
S4 panels will need to be re-compiled into an overall figure.

sjspielman:
Closes #1499 

This PR compiles S4 for V22. To support a final figure size <= 8.5x11, I changed some ggplot2 theme settings in the relevant mutational signatures notebook (axis text sizing, etc.) to reduce the figure panel size by code instead of manually. I re-ran this full notebook, and copied figure panels into `figures/` after doing the re-sizing. I also added a `useDingbats=FALSE` for the jitter plot which is needed to properly render the `stat_manual()` pointrange geom (without this argument, they are all literally Greek letter lambdas instead of points).


jharenza (approved):
The changes look good!

PR #1536, opened 2022-07-13
Rerun the count-contributions module with real stats

jaclyn-taroni:
I followed the instructions here for running this on the `AlexsLemonade/master` branch: https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/count-contributions#running-the-module-manually

And have committed the real stats!

PR #1540, opened 2022-07-14
Try to fix `workflow_dispatch` trigger in count contributions GHA workflow

jaclyn-taroni:
The changes in the GHA workflow in #1536 sure did not work. Here I'm trying to make the `workflow_dispatch` trigger work by putting it after the `cron` trigger and using a trailing `:`, which the internet told me is fine (but we'll find out).


jashapiro (approved):
Oh sure, believe the "internet"

PR #1543, opened 2022-07-14
Separate analyses from figures script

sjspielman:
This PR replaces a previous now-gone PR of the same goal, because I borked my fork and had to start over 😬 

This PR updates `generate-figures.sh` by removing code that runs analyses to prepare for figures. I place that code into a new script, `scripts/run-analyses-for-figures.sh`. This follows the new plan outlined in #1455. This script takes the place of the `run-analysis-modules.sh` script conceptualized below, but I named it as is here because the script doesn't fully run all modules, only those needed for figures.

> This approach emerged from discussion with @jaclyn-taroni and @jashapiro about a strategy to ensure analysis modules and figures are up-to-date and in sync, especially if/when there is a future data release.
> 
> Currently, we have three scripts that run aspects of PBTA:
> 
>    + scripts/generate-analysis-files-for-release.sh. This script runs analysis modules that need to be run to enable subtyping.
>    + scripts/run-for-subtyping.sh. This script runs molecular subtyping modules and produces the pbta-histologies.tsv file from the base version.
>    + figures/generate_figures.sh. This script (currently under review in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1454) runs, for each figure in the manuscript, the any relevant analysis modules needed to make the given figure, and then either runs associated figure scripts in figures/scripts/, or copies over figure panels created within analysis modules.
> 
> Ideally, figures/generate_figures.sh should only be generating figures (running scripts in figures/scripts/ or copying over figure panels created within analysis modules) without also running analysis modules. Therefore, we should do the following:
> 
>    + Create a fourth script, called something like scripts/run-analysis-modules.sh. This script will run all analysis modules that either i) were not run for subtyping, ii) were partially run for subtyping but now need to be fully run (e.g. GSVA), or iii) were run for sybtyping and need to be run again with the subtyped data.
>    + Remove all analyses from figures/generate_figures.sh so that this script only generates figures under the assumption analysis modules have been run and are ready to be used for figure generation.
>    + Document changes accordingly, likely in scripts/README.md.


jaclyn-taroni:
I am actually going to put this into the Epic that's tracking a version of the manuscript for distribution/comments. That's because having this in place will be really helpful if we get co-author comments that require us to rerun analyses (seems very plausible).

It can be wrapped up in parallel to the manuscript issues in that Epic in my opinion.

sjspielman:
This is now ready for review! I've requested a couple reviewers since I'd like a few eyes on this to ensure that all analyses that should be run are being run. You'll also notice I rearranged the order in which analyses are run so that `tp53_n1_score` gets run before modules that consume its results, which include `mutational-signatures` and `survival-analysis`.

jaclyn-taroni:
@sjspielman I would expect to see some docs changes in `figures/` and `scripts/` associated with these code changes.

sjspielman:
@jaclyn-taroni doc updates have been pushed!

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Run the chromothripsis module, which requires the chromosomal-instability module, for Figure 3
+bash ${analyses_dir}/chromosomal-instability/run_breakpoint_analysis.sh
+bash ${analyses_dir}/chromothripsis/run-chromothripsis.sh
+
+# Run the tp53 classifier, for Figures 4 and S5
+#  Note this module is run earlier in this script than the relevant 
+#  figures' placement because these modules use the TP53 scores:
+#  `mutational-signatures` and `survival-analysis`
+OPENPBTA_TP53_FIGURES=1 bash ${analyses_dir}/tp53_nf1_score/run_classifier.sh
```
jaclyn-taroni, line 50: I think the logic in this module is wrong? Or at least compared to the docs...

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/tp53_nf1_score/run_classifier.sh#L13-L15

And the logic:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/tp53_nf1_score/run_classifier.sh#L74-L85

Should it be `if [[ "$RUN_FOR_FIGURES" -eq "1" ]]`  or am I confused?

jaclyn-taroni: Coming back to this, I think the code is what we want, i.e., it is doing 

```
# If running for the purpose of figure generation (RUN_FOR_FIGURES will be 1), 
 # use the data in the analysis directory that has been freshly collapsed 
 ```

But the doc at the top of the script has things mixed up? 

At this point, if we move to an approach where everything _in this script_ is using files from the data release, we can go ahead and rip this out entirely?

File: scripts/run-analyses-for-figures.sh [hunk 2]
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Run the mutational-signatures module for Figures 3 and S4
+# We only run the part of the module used in the manuscript (i.e., not de novo)
+OPENPBTA_CNS_FIT_ONLY=1 bash ${analyses_dir}/mutational-signatures/run_mutational_signatures.sh
```
jaclyn-taroni, line 55: Uses release files:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/mutational-signatures/02-split_experimental_strategy.R#L35

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/mutational-signatures/02-split_experimental_strategy.R#L45

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/mutational-signatures/05-fit_cns_signatures.R#L36

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Run the telomerase activity prediction script, for Figures 4 and S5
+OPENPBTA_FOR_FIGURES=1 bash ${analyses_dir}/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh
```
jaclyn-taroni, line 59: When in run this manner, uses release file:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh#L22

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Run the survival module, for Figures 4 and 5
+bash ${analyses_dir}/survival-analysis/run_survival.sh
```
jaclyn-taroni, line 62: Uses release files: 

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/survival-analysis/survival-analysis_subtypes.Rmd#L63-L66

But I found an issue where we need to run another module first...

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Run the dimension reduction module, for Figures 5 and S6
+bash ${analyses_dir}/transcriptomic-dimension-reduction/dimension-reduction-plots.sh
+
+# Run the immune deconvolution module, for Figures 5 and S6
```
jaclyn-taroni, line 67: This needs to go before survival!

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/survival-analysis/survival-analysis_immune.Rmd#L30

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+
+# Run the telomerase activity prediction script, for Figures 4 and S5
+OPENPBTA_FOR_FIGURES=1 bash ${analyses_dir}/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh
+
+# Run the survival module, for Figures 4 and 5
+bash ${analyses_dir}/survival-analysis/run_survival.sh
+
+# Run the dimension reduction module, for Figures 5 and S6
+bash ${analyses_dir}/transcriptomic-dimension-reduction/dimension-reduction-plots.sh
```
jaclyn-taroni, line 65: Release files used throughout https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/analyses/transcriptomic-dimension-reduction/01-dimension-reduction.sh

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+# Run the immune deconvolution module, for Figures 5 and S6
+bash ${analyses_dir}/immune-deconv/run-immune-deconv.sh
+
+# Generate GSVA scores and test for cancer group differences, for Figure 5
+bash ${analyses_dir}/gene-set-enrichment-analysis/run-gsea.sh
```
jaclyn-taroni, line 71: Uses release files:

https://github.com/sjspielman/OpenPBTA-analysis/blob/6644d9df98ebb35adb7c332c0db8f950c019c5ac/analyses/gene-set-enrichment-analysis/run-gsea.sh#L31

jaclyn-taroni (commented):
Thanks for the doc updates! I think we should re-evaluate whether or not we include the steps that can not be run locally (SNV and CNV consensus) in `scripts/run-analyses-for-figures.sh` at all. As I detail in the comments we're returning, pretty much everything else that the script run points to `data/`. Even if we run those consensus steps, the subsequent steps won't make use of the newly generated files. 

I also found a couple issues: 

* Immune deconvolution needs to be run before survival analysis.
* There's an issue with docs in the shell script in `tp53_nf1_score` (I _think_ it's the docs and not the code).

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,73 @@
...
+
+#################################
+
+# Run modules that cannot be run locally due to memory requirements
+if [ "$RUN_LOCAL" -lt "1" ]; then
+  bash ${analyses_dir}/focal-cn-file-preparation/run-prepare-cn.sh       # Figures 2 and S3
+  bash ${analyses_dir}/snv-callers/run_caller_consensus_analysis-pbta.sh # Figure S2
+  bash ${analyses_dir}/snv-callers/run_caller_consensus_analysis-tcga.sh # Figure S2
+  bash ${analyses_dir}/copy_number_consensus_call/run_consensus_call.sh  # Figure S3 (heatmap)
+fi
```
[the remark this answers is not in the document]
sjspielman: The `snv-callers` plots in S2 need the SQL db's in `scratch/`, so this is annoyingly unavoidable.

This `copy_number_consensus_call` can be removed I think, and that figure script can be updated too since results from that module don't appear to ever be read in! I might want to add in running the module `cnv-chrom-plot`, but that can be run locally outside this `if`.
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/figures/scripts/figS3-panels-BCD.R#L35-L39

Finally, I had kept `focal-cn-file-preparation` because of this discussion - https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1455#issuecomment-1163114816

scripts/run-analyses-for-figures.sh
sjspielman: Thanks good catch!!

scripts/run-analyses-for-figures.sh
sjspielman: > Should it be if [[ "$RUN_FOR_FIGURES" -eq "1" ]] or am I confused?

You are not confused! Thanks for catching this. It should indeed be `eq "1"` here. Making that change now and updating that comment to appropriately match. 

> At this point, if we move to an approach where everything _in this script_ is using files from the data release, we can go ahead and rip this out entirely?

Yeah, I think we can do that. The `collapse-rnaseq` results don't change with subtyping, so I guess it shouldn't matter where we pull those files from?

jaclyn-taroni: Wait read the second comment!

analyses/tp53_nf1_score/run_classifier.sh
```diff
@@ -72,8 +72,8 @@ Rscript --vanilla ${analysis_dir}/00-tp53-nf1-alterations.R \
   --gencode ${cds_file}
 
 # If running for the purpose of figure generation (RUN_FOR_FIGURES will be 1),
-# use the data in the analysis directory that has been freshly collapsed
-if [[ "$RUN_FOR_FIGURES" -eq "0" ]]
+# use the data in the data release directory
+if [[ "$RUN_FOR_FIGURES" -eq "1" ]]
```
jaclyn-taroni, line 76: If you use the data release for figures, you don't need this at all. It was intended to let you use _results_ for figures, as that was an existing paradigm that we've since moved away from.

jaclyn-taroni (commented):
Oops, meant to leave a single comment

scripts/run-analyses-for-figures.sh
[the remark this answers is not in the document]
jaclyn-taroni: > The `snv-callers` plots in S2 need the SQL db's in `scratch/`, so this is annoyingly unavoidable.

That is annoying, yes.

> Finally, I had kept `focal-cn-file-preparation` because of this discussion [#1455 (comment)](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1455#issuecomment-1163114816), with the idea that this script serves as the "big run" after subtyping.

I think my comment there didn't take into account whether or not the output of `analyses/focal-cn-file-preparation/06-find-recurrent-calls.Rmd` _gets used_, and I don't think it does from a find all of `"consensus_seg_focal_cn_recurrent_genes.tsv"` and `"consensus_seg_recurrent_focal_cn_units.tsv"`...

scripts/run-analyses-for-figures.sh
sjspielman: Is this the second comment?
> If you use the data release for figures, you don't need this at all. It was intended to let you use results for figures, as that was an existing paradigm that we've since moved away from.

Meaning, we can rip out the whole flag.

jaclyn-taroni: This is the second comment: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1543#discussion_r935936382

The docs were wrong, not the behavior of the code. If running for figures in the past, we did not want to use the data in the release. But yes, we can rip out the flag and just always use the data in the release.

scripts/run-analyses-for-figures.sh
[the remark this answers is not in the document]
sjspielman: >  I don't think it does 

I agree that it doesn't. But this raises another question about how we should go about ensuring analysis modules that are in the paper, even if not directly used for figures, are up-to-date. That might be another script entirely.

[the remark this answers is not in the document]
jaclyn-taroni: > That might be another script entirely.

Yea, that's basically where I landed.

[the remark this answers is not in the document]
sjspielman: Oh actually we can probably scrap `snv-callers/run_caller_consensus_analysis-tcga.sh` from this script, since it's run in `generate-analysis-files-for-release.sh` and definitely doesn't need the PBTA histology file!

scripts/run-analyses-for-figures.sh
```diff
@@ -0,0 +1,77 @@
...
+analyses_dir="$BASEDIR/analyses"
+
+#################################
+
+# Run modules that cannot be run locally due to memory requirements
+if [ "$RUN_LOCAL" -lt "1" ]; then
+  # This module be run to SQL databases needed to make Figure S2 panels
+  # Note that the TCGA version would have been run in `generate-analysis-files-for-release.sh`
+  bash ${analyses_dir}/snv-callers/run_caller_consensus_analysis-pbta.sh # Figure S2
+fi
```
jaclyn-taroni, line 32: I'm confused why we would need one but not the other here?

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/2a4d691498c08cc985962447806ff5f2c2b92176/scripts/generate-analysis-files-for-release.sh#L61-L67

Is it because we're running it with the histology file in the release? (I think it's probably that, but the comment is maybe what I find confusing!)

I don't think it's a good idea to assume that someone will have the SQL database for TCGA around on their machine anyway, though.

jaclyn-taroni (approved):
This is most of the way there! We should run the SNV consensus steps for _both_ TCGA and PBTA for a reason I explain in a comment I am returning. Approving with the expectation that you will agree with me and implement that 😄

PR #1547, opened 2022-07-15
Compiled Figure S6 for V22

sjspielman:
This PR adds compiled Figure S6 for V22.

The UMAP panels should be in the correct order now.

Along the way, I realized some `immune-deconv` changes were needed so I made them in `immune-deconv` and copied the updated panels. There were two specific changes:
1) I fixed a "loss" -> "lost" in the subtypes
2) I updated the x-axis title for panel F to match that in panel E. Previously F said "molecular subtype of sample," but E said "molecular subtype of _tumor_ sample," so now both include the word tumor.

In addition, after a side-chat with @jaclyn-taroni (who agreed with my actions here), I ended up integrating only the _image_ for panel E into the compiled figure instead of the PDF precisely. In other words, if you open the PDF, that panel will not be made of vector objects. That is because this particular panel is, to be quite frank, a nightmare to compile - the angled x-axis text is never imported properly and requires a literal full work day to fix as well as a lot of sanity 😬 . Since this is "just" a supplemental figure, and a precise PDF version of this panel exists in the repo, I decided to just drop it in as a image for the overall compilation.


jharenza:
>That is because this particular panel is, to be quite frank, a nightmare to compile - the angled x-axis text is never imported properly and requires a literal full work day to fix as well as a lot of sanity 😬 .

Wow! Agreed. Our supp figures are already much prettier than most!!

It looks like the left vertical and top horizontal axes lines of A-D are much darker than the others - not sure if that is fixable. I am also now wondering if we can manually label A-D with the cancer groups?

sjspielman:
> It looks like the left vertical and top horizontal axes lines of A-D are much darker than the others - not sure if that is fixable.

Hmm that's really odd; it's not appearing like this in the Graphic app (it's one object around the whole plot, so one size!!), but I also see what you see in the export. I also see weird `alpha` point things happening with _some_ of the UMAP panels and _some_ of the points in panel F. Since this figure is misbehaving, I think I should just place all the panels (ala panel E), and it'll still look nicer than other supp figures out there! :)

>  I am also now wondering if we can manually label A-D with the cancer groups? 

Will do!!

jharenza (approved):
👍 looks good to me now!

PR #1550, opened 2022-07-18
Compiled Figure S2 for V22

### LINKED ISSUE #1497: Re-compile Figure S2
S2 panels will need to be re-compiled into an overall figure.

sjspielman:
Closes #1497 

This PR adds compiled S2. Panels A and D are too large to directly import, so these are placed images. In addition, there was a lot of text wonkiness from importing panels H/I, so these have also been directly placed. Panels, B-C and E-G are all proper PDFs in the compilation. While trying to import H/I though I noticed they were missing `useDingbats=FALSE` so I re-ran the figures and fixed a surprising but bad bug in that script! That said, those re-run figures aren't used directly because I ended up placing H/I since x-axis text was really messed up. Either way I fixed the bug in the script!

In addition, the upset plot files are modified because I removed a blank page that was in those files, left over from their generation. The plots themselves are unchanged.


jharenza:
>While trying to import H/I though I noticed they were missing useDingbats=FALSE so I re-ran the figures and fixed a surprising but bad bug in that script! That said, those re-run figures aren't used directly because I ended up placing H/I since x-axis text was really messed up. 

Do you mean you placed figures? I see in the png that the dots for H-I are OK, but look like they have shadows in the PDF.

>Either way I fixed the bug in the script!

👏

sjspielman:
> Do you mean you placed figures? I see in the png that the dots for H-I are OK, but look like they have shadows in the PDF.

I did place H-I because the axis labeling was a mess after importing a PDF directly. I don't see any shadows on my end - here's a screen shot of the PDF. Is yours different, or can you point me to where I'm not seeing?
Screen Shot 2022-07-18 at 2 13 45 PM
\

jharenza (approved):
👍

sjspielman:
Note to merge this _after_ #1547!

PR #1552, opened 2022-07-19
update Figure 3D to separate out "Other" cancer groups

jharenza:
This panel was previously updated to use `cancer_group_display` and when I did that, I overlooked the fact that this figure had been plotting the other cancer groups individually.

#### What was your approach?
- Update the code to use `cancer_group` when `cancer_group_display == "Other"`. 
- Update palette code accordingly

NA

code updates

No

Yes

#### Documentation Checklist


sjspielman (commented):
Thanks for looking into this! In comparing this panel with the previous one, I'm getting different _total_ amounts for the "Other" groups. Previously, we had 9/60 for "Other". In this PR, I see 8/38 when pooling the "Other" groupings. Do you also see this discrepancy?

jharenza:
> Thanks for looking into this! In comparing this panel with the previous one, I'm getting different _total_ amounts for the "Other" groups. Previously, we had 9/60 for "Other". In this PR, I see 8/38 when pooling the "Other" groupings. Do you also see this discrepancy?

totally missed this comment - will check!

jharenza:
> > Thanks for looking into this! In comparing this panel with the previous one, I'm getting different _total_ amounts for the "Other" groups. Previously, we had 9/60 for "Other". In this PR, I see 8/38 when pooling the "Other" groupings. Do you also see this discrepancy?
> 
> totally missed this comment - will check!

@sjspielman I think what is going on is that because we only plot groups with N>= 3 samples, we lose some of the 60 in the plot. I will create a table which shows the new groups and Ns so we can see the diffs.

jharenza:
@sjspielman - added some calculations in this and indeed, we had lost 22 samples due to N <3. Ready for re-review!

sjspielman (approved):
Great, this all makes sense!

sjspielman:
This code does not run through CI, so can merge before checks complete.

PR #1554, opened 2022-07-19
Update Table 1 to use broad_histology_display

### LINKED ISSUE #1553: Update Table 1 to use broad_histology_display
[Table 1](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/tables/results/Table1-molecular-subtypes.tsv) shows molecular subtypes across broad histologies, but this table needs to be further updated to use terminology in `broad_histology_display` and NOT in `broad_histology`, as concluded here - https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/268#issuecomment-1125303759

sjspielman:
Closes #1553 

This PR updates Table 1 to use `broad_histology_display` for the heading. I updated the table notebook to read in and incorporate the palette file and re-ran the notebook. The output column now says "broad_histology_display" which I think is appropriate because the values indeed differ from "broad_histology."

Something else I am realizing is that we may need to update Table S1 also (the histologies file), OR include docs for how to map "internal naming" back to the naming used in the manuscript, for when broad_histology != broad_histology_display. If we want to consider doing that I can add that into this PR.


tables/results/Table1-molecular-subtypes.tsv
```diff
@@ -64,4 +62,6 @@ Neuronal and mixed neuronal-glial tumor	GNT, KIAA1549-BRAF	2
 Neuronal and mixed neuronal-glial tumor	GNT, other MAPK	1
 Neuronal and mixed neuronal-glial tumor	GNT, other MAPK, FGFR	1
 Neuronal and mixed neuronal-glial tumor	GNT, RTK	2
-Tumors of sellar region	CRANIO, ADAM	27
+Other	CHDM, conventional	2
+Other	CHDM, poorly differentiated	4
```
jashapiro, line 66: While we use Other for display tables, this should probably remain as `Chordoma` for this table.

jashapiro:
> Something else I am realizing is that we may need to update Table S1 also (the histologies file), OR include docs for how to map "internal naming" back to the naming used in the manuscript, for when broad_histology != broad_histology_display. If we want to consider doing that I can add that into this PR.

I think we should probably leave S1 as is, but I agree about adding in documentation about the groups. If Table S1 is the only place that we use the `broad_histology` values, we might note it in the caption for that table.

tables/results/Table1-molecular-subtypes.tsv
sjspielman: Hm, maybe I should change this to a whole new column that uses `broad_histology_display` unless it's "Other", in which case `broad_histology`. Then, the column will not match either of our internal variables `broad_histology<_display>`, so I feel it needs a new name altogether to avoid confusion with those (or is this more confusing?). Maybe something like `pbta_broad_histology`?

jharenza:
> > Something else I am realizing is that we may need to update Table S1 also (the histologies file), OR include docs for how to map "internal naming" back to the naming used in the manuscript, for when broad_histology != broad_histology_display. If we want to consider doing that I can add that into this PR.
> 
> I think we should probably leave S1 as is, but I agree about adding in documentation about the groups. If Table S1 is the only place that we use the `broad_histology` values, we might note it in the caption for that table.

We can also possibly link to the palette file somewhere in the STAR methods?

PR #1556, opened 2022-07-19
Fix oncoprint bugs and S3A

### LINKED ISSUE #1548: Figure S3A includes the wrong groups and `n_per_cancer_group.tsv` has two "Other" rows
Per this comment:

Hi @sjspielman - in trying to go through these tables for the manuscript, I realized that we have two "Other CNS" oncoprints - one in Figure 2D and one in S3-A. Currently, both have the same oncoprint, but S3 should have one similar with groups similar to [this](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/b6ae33b3c10651efd3c417d1f3d9172ef894999e/figures/pdfs/supp/figs3/figureS3.pdf). I think some lines are getting crossed with two groupings called "Other". I also noticed (and my bad for not noticing during review), that we have two "Other" rows in the `*n_per_cancer_group.tsv` tables.

_Originally posted by @jharenza in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1498#issuecomment-1186651357_

We need to create a new Figure S3. I think this may be related to our updating of the palette when we removed some of the oncoprint labeling starting at line [377 here](https://github.com/AlexsLemonade/OpenPBTA-analysis/commit/f7852f0b267c5ddd39df71a37c9eb1ff2f3064c8). So I _think_, @jaclyn-taroni , this has to go back in, to color the "Other" (other) CNS cancer groups in the supplemental figure?

Below is where I think this should be updated in the figure script:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/045e5c6121c35377f7d7acbd2ce94ce25b9e2b90/figures/scripts/fig2-oncoprint-landscape.R#L363

sjspielman:
Closes #1548 

This PR fixes the oncoprint problems! Overall it turned out to be not _too_ many coding differences, so I've kept everything in one PR. I have done the following:

- Updated the palette file to include a column `oncoprint_main`, which is `TRUE` if the group is designated for *main text* and `FALSE` if it should head to the supplement. While I was coding, I found this variable name easier to follow, so I use this instead of `oncoprint_include`.
- Updated the oncoprint module itself so there is only one "Other" in the `n_per_cancer_group` files. This problem was caused by a filtering bug in the oncoprint script. I confirmed the cancer groups in the "primary-only" file match those in the landscape plots.
  -  I re-ran the oncoprint module, which resulted in these 2 file diffs and an updated PNG (but not diffs inside the PNG! just regenerated!).
- Updated the oncoprint figure script to filter on `oncoprint_main` for the supp figure. Since the supp panel uses `cancer_group` and not `cancer_group_display`, I had to slightly update a function in this script to handle the different cases. 
  - When I re-ran this script, other PDF panels were re-generated but with 0 bytes difference.


jharenza (commented):
yay! This looks good. My only comment is that the legend ordering in the supplemental figure is not the same as in the oncoprint display (I think this wasn't coordinated previously, but since it was mentioned on the previous main oncoprint PR, I wanted to mention this and ask if we want to update here as well).

sjspielman:
> My only comment is that the legend ordering in the supplemental figure is not the same as in the oncoprint display (I think this wasn't coordinated previously, but since it was mentioned on the previous main oncoprint PR, I wanted to mention this and ask if we want to update here as well).

Yeah, this wasn't previously coordinated. The way to do it would be hardcoding the legend order I think. I'll have a go.

sjspielman:
@jharenza I was able to get the legend matching the plot, but I'm not sure I like this...
The reason this script handles legends separately is because legends cannot be ordered on `maftools::oncoprint()`. The annotation legend will always be alphabetical. However, _samples_ can be ordered, so in this version the legend matches the annotation row, but the cancer group annotations are not in descending order in terms of the number of samples.

I can have the legend handled for this supp figure separately, similar to how main text legends are handled, but I'll have to rearrange the script quite a bit to get there.

jharenza:
>@jharenza I was able to get the legend matching the plot, but I'm not sure I like this...

Eh, being a supplement and not many samples, I think this could be ok. It is up to you if you want to put more work into a separate legend

sjspielman:
I'd rather not re-work the script too dramatically for just this one supplementary legend. I'm definitely ok with it as it is now!

jharenza (approved):

PR #1559, opened 2022-07-20
Compiled Fig2 with Illustrator

### LINKED ISSUE #1557: Re-compile Figure 2 with Illustrator
I'm just noticing the re-compiled Figure 2 has some funkiness to it - the boxes in the oncoprints have "extra lines" at the top and left sides: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pdfs/fig2/figure2.pdf

This is not visible in the Graphic app at all during compilation, and it is also not present in the PNG. It's only in the PDF export. Therefore this will need to be re-compiled in Illustrator.

_That said_, this issue should not necessarily block our ability to proceed towards a shareable Word document! Figure 2 as-is is ok for internal sharing, but we should clean it up for submission.

sjspielman:
Closes #1557 

This PR updates compiled Fig 2 using Illustrator instead of Graphic, due to some funkiness in the Graphic PDF export.

(Thanks again to @envest for super helpful Illustrator tips that made this process, for the first time?!, not at all painful! :tada:)


jharenza (approved):
Lines are gone!

PR #1562, opened 2022-07-20
Add notebook for setting manuscript author order

jaclyn-taroni:

Here I am adding a notebook (`manuscript-authorship/set-authorship-order.Rmd`) that sets the author order for the manuscript. I'm going to quote the top of the notebook here to explain the methodology.

> This notebook updates the current [manuscript metadata](https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/master/content/metadata.yaml) such that the authors are ordered in the following way:
> 
> * The first, last 4 authors, and consortia authors positions are pinned to reflect scholarship and contributor roles that are not captured in the Git history of this repository (e.g., substantive analytical code review, conceptualization, supervision, or project administration).
> * Manuscript authors that have contributed to the code base for the analysis repository -- specifically the number of analysis modules that were included in the paper that a manuscript author has contributed to -- are then ordered from the second position on in decreasing order.
> In the case of ties, the order is randomly selected.
> * Manuscript authors that did not directly contribute to the code base are then randomly ordered. 
> We set a seed directly before that shuffling step, using the year as the seed, to keep a consistent order in future runs if and when the Git contributions change.

To accomplish this, I also needed to update module that counts the modules people contributed to so that the names in the author metadata match the git contribution names (e.g., add middle initials).

https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/219

### Questions for reviewers

* Is the code implementation correct, i.e., does it match the methodology I've stated?
* Did I miss any authors where their names differ between GH display name and their name as listed in the manuscript metadata?
* For the section setting author order for authors that did not directly contribute to the code base, we don't want that order to always change or be impacted by changes in code contributions. I set a seed right before that step. Anything else I should be doing?
* Should this notebook go in some other directory? 
* Should this notebook get run in the workflow counting git contributions every Wednesday?


jashapiro (approved):
This all looks good to me. I think the location is okay, but I might put the notebooks in the `count-contributions` directory, as those seem to go together. I think adding running the notebook to the counting GHA also makes sense, but that could potentially be in a separate PR. I'm approving the PR, but if you add a GHA and want me to look at it, feel free to rerequest my review!

jaclyn-taroni:
@jashapiro  I moved the authorship ordering notebook to `analyses/count-contributions` where I also added it to the README and the shell script. The shell script is what is run via GHA, so this will get run, but I'm continuing to ignore the updated YAML file for now. What is "missing" is to create an action that files a PR with the new metadata file to the manuscript repo. I'm going to file an issue to track that and get this merged. Any debate about whether or not we should be ignoring the updated metadata YAML can take place on that new issue!

jaclyn-taroni:
This doesn't run in CircleCI. It _does_ get run via GHA every Wednesday (and I'll manually trigger it after merging to make sure it works as expected). I can also confirm that the shell script runs successfully in the project Docker container. That is to say, I'm going to merge this before CircleCI completes.

PR #1565, opened 2022-07-21
Reorder S2 panels H-I by median

### LINKED ISSUE #1563: Figure S2 H-I are no longer ordered by median TMB
TMB plots for S2 H-I became unordered when previously ordered by ascending median TMB, so I think these should be updated :(

Note: I don't think this blocks converting the manuscript to a document for distribution.

sjspielman:
Closes #1563 

This PR updates Figure S2 so that panels H-I are ordered by ascending median rather than cancer group frequency. For the PBTA plot, "Other" still remains on the end (not in order), but this can be quickly updated too! 
Since this is a very quick change, all of the following are updated here:
- Code to generate these 2 panels
- The panels themselves
- The compiled S2 (again as in #1550, panels H-I are placed)


jharenza:
Hi @sjspielman this looks good, sorry for the extra step. 

>"Other" still remains on the end (not in order), but this can be quickly updated too!

For this, I might suggest we remove this group since it isn't really informative here (or order it with the rest) - thoughts?

sjspielman:
> For this, I might suggest we remove this group since it isn't really informative here (or order it with the rest) - thoughts?

My thoughts were to keep "Other" at the end as we've done previously for other plots, but I don't have a strong preference on this. If you prefer one of those options I can go that direction!

jharenza:
>My thoughts were to keep "Other" at the end as we've done previously for other plots, but I don't have a strong preference on this. If you prefer one of those options I can go that direction!

oops, this never sent. I would suggest we remove for this specific figure since it would be hard to interpret.

sjspielman:
> I would suggest we remove for this specific figure since it would be hard to interpret.

On it!

sjspielman:
@jharenza "Other" has been removed from the PBTA panel and figure has been recompiled.

jharenza (approved):
🚀

sjspielman:
Code doesn't run through CI; merging now.

Issue #1567, opened 2022-07-21
Pull out `mutational-signature` figure code


sjspielman:
Plots from the `mutational-signatures` module continue to be made within the module rather than from scripts in `figures/scripts`. At this stage of the project, since we are revisiting these figures _again_ (#1566 and very possibly/likely more!!), it's going to be much more straightforward to establish stand-alone figure scripts.

We should therefore add scripts to `figures/scripts/` to create the following panels:
+ One script for panels currently made in [`07-plot_cns_fit.Rmd`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/mutational-signatures/07-plot_cns_fit.Rmd):
  + 3E, the sina plot of exposures 
  + S4A, the barplot of exposures across samples
  + S4B, the jitter plot of signature 1 weights vs tumor descriptors
+ One script for the 4E heatmap panel and its legend, currently made in [`08-explore_hypermutators.Rmd`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/mutational-signatures/08-explore_hypermutators.Rmd)

sjspielman:
Closed by #1569 and #1570

PR #1569, opened 2022-07-21
New mutational-signatures figures script (1/2)

### LINKED ISSUE #1566: Update panel 3E order to match 3A
Currently, in [Figure 3](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pdfs/fig3/figure3.pdf), we have different orderings of the _same_ cancer groups in panel A vs. E.
Based on conversations with @jashapiro, we should retain the more meaningful ordering in A and update panel E to match. This can be hard-coded!

sjspielman:
Partly addresses #1567 
Closes #1566 

This PR adds a new script, `figures/scripts/fig3_figS4-mutational-signatures-panels.R` to create panels 3E, S4A, and S4B. This way we do not have to re-run the mutational signatures notebooks and copy over PDFs when updating these panels.
Within the code, 3E facets are also re-ordered to match what we have in 3A. However, so are panels S4A and S4B. While we want this order for 3E, it's not yet clear that we want it for S4A/B, as discussed - https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/281/files#r926700818 and https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/281#pullrequestreview-1046502591


jharenza:
>However, so are panels S4A and S4B. While we want this order for 3E, it's not yet clear that we want it for S4A/B, as discussed -

I think it makes sense for these to match like @jashapiro suggested, especially since we added back the "other HGG/LGG"

jharenza (approved):
I think this looks good as it is for both Fig3 and S4

jashapiro:
> I think this looks good as it is for both Fig3 and S4

@jharenza My concern [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/281/files#r926700818) was that we are referring to some of the cancer groups that are not actually in the top set here, so it might make sense for figure S4 to include more groups. 

A related question I have is whether supplemental figures have to be one page? I've certainly seen many journals where they are not!

jashapiro (approved):
This looks good to me. I'm going to approve as is, but see my comment below. If we decide to  make figure S4 different,  you can use your judgement about whether to do that in a separate PR or dismiss my review here and update this one.

sjspielman:
> A related question I have is whether supplemental figures have to be one page? I've certainly seen many journals where they are not!

We have a limited amount of supplemental figures that we are allowed to include, but I'm not sure about size. This is the guidance for figure size, but I _think_ this might only be for main text? https://www.cell.com/figureguidelines

jharenza:
> @jharenza My concern [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/281/files#r926700818) was that we are referring to some of the cancer groups that are not actually in the top set here, so it might make sense for figure S4 to include more groups.

It looks like they were added in this PR (?) I see the "other LGG" and "other HGG" here now - or do you mean something else? Agree with keeping them the same as Figure 3.

>A related question I have is whether supplemental figures have to be one page? I've certainly seen many journals where they are not!

We have a max of supp figures, but do you mean splitting this into two pages as one figure? I have not seen where we can split one figure into two pages for Cell, but can check. I think they would then become two figures.

jashapiro:
> > @jharenza My concern [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/281/files#r926700818) was that we are referring to some of the cancer groups that are not actually in the top set here, so it might make sense for figure S4 to include more groups.
> 
> It looks like they were added in this PR (?) I see the "other LGG" and "other HGG" here now - or do you mean something else? Agree with keeping them the same as Figure 3.

Sorry, there are two points in that comment, one before the suggestion and one after. The one after is about DNETs and schwannomas, which do not appear in any of the figures. So I thought it might make sense to add them in figure S4

jharenza:
>Sorry, there are two points in that comment, one before the suggestion and one after. The one after is about DNETs and schwannomas, which do not appear in any of the figures. So I thought it might make sense to add them in figure S4

Oh, I see now. I hadn't read through the manuscript comments! I can also remove them from the manuscript text - I don't think that adds or takes away much and we have other groups we can reference, so that way S4 doesn't get huge.

sjspielman:
> Oh, I see now. I hadn't read through the manuscript comments! I can also remove them from the manuscript text - I don't think that adds or takes away much and we have other groups we can reference, so that way S4 doesn't get huge.

Sounds like the conclusion here is to keep cancer groups as-is in Fig S4 and update text instead! I'll get this merged.

sjspielman:
Code does not run through CI, so merging before checks.

PR #1544, opened 2022-07-14
Compiled Figure 4 for V22

sjspielman:
Closes #1489 
This PR adds compiled versions of Figure 4 in PDF and PNG.
Also, when I was compiling, I noticed that the heatmap patient legend order didn't match the figure patient order. I therefore updated the legend *manually.* Reviewers, please check that the updated labels are correct!

PR #1570, opened 2022-07-21
Figure script for mutational-signatures heatmap (2/2)

sjspielman:
Addresses rest of #1567 

This PR adds a script `figures/scripts/fig4-heatmap.R` to make the hypermutator heatmap and its legends so that the `mutational-signatures` module does not need to be rerun + panels copied over when updating these figures. I also took this time to update the code so that the patient order in the heatmap matches the order in the patient legend. Previously the legend was manually altered during compilation (see https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1544#issue-1305243019), but now it's all set to work programmatically! The colors may be in a slightly different order, but the patient colors are not in a meaningful order anyways. I confirmed that BS and PT ids match up here.


jharenza (approved):
✅ patient and bs_ids match and this looks good!

sjspielman:
Noting that `generate-figures.sh` is somewhat out of date because it is still copying panels but it should run this new script. I'm going to open a separate issue for that!

PR #1575, opened 2022-07-21
Move immune-deconv figure code

sjspielman:
Inspired by my PRs #1569 and #1570, I got a little excitable and also moved out the `immune-deconv` panel code for easier updating moving forward!

Along the way I made a couple figure updates including using `shape=16` to make Illustrator happier with transparent points (thanks again for the tip @envest!) and re-sizing some panels that are compiled with Graphic. I also updated the CD274 expression axis label to match the same format that we have in figure S5B-C (all `GENE log2(FPKM+1)`). I recompiled accordingly!

Again, the `generate-figures.sh` script will need to be updated and that's coming up next!


figures/scripts/fig5_figS6-immune-deconv-panels.R
```diff
@@ -0,0 +1,377 @@
...
+quantiseq_gather <- quantiseq_spread %>%  
+  gather(cell_type, score, -c(sample, library, method))
+
+# First, find the molecular subtypes in the of interest cancer groups AND excluding unclassified, with >=3 samples
+subtypes_of_interest <- palette_mapping_df %>%
+  filter(cancer_group_display %in% cancer_groups_of_interest, 
+         !(str_detect(molecular_subtype, "To be classified"))) %>%
+  count(molecular_subtype) %>%
+  filter(n >= 3) %>%
+  pull(molecular_subtype)
```
jharenza, line 113: The one thing I noticed over in the manuscript was that S6 immune deconv panels have the HGG subtype removed, I thought because other HGG was no longer in the plots. But, I see they did not return in the plots - I am pretty sure that HGG, H3 WT has N>3 - would you mind checking that here - we probably should recover them with updating these figures.

sjspielman: Ok, I see what's happening - I had defined `cancer_groups_of_interest` for *panel 5C* on line 36 as  - 
```
cancer_groups_of_interest <- c("Pilocytic astrocytoma", 
                               "Diffuse midline glioma",
                               "Craniopharyngioma", 
                               "Ganglioglioma",
                               "Ependymoma", 
                               "Medulloblastoma",
                               "Schwannoma",
                               "Neurofibroma Plexiform",
                               "Other embryonal tumor",
                               "Atypical Teratoid Rhabdoid Tumor",
                               "Meningioma",
                               "Dysembryoplastic neuroepithelial tumor")
```

This code you pointed out filters to those cancer groups, and "other HGG" is not in that list. So, for the S6 plots, we probably don't want to use this filtering. If I don't filter to that list and include all cancer groups, here are _all_ the molecular subtypes with N>= 3, arranged by subtype. We definitely can't be plotting all 33 of these though in one plot, so we will need to discuss a different approach to filtering down I think?

```
1 CHDM, poorly differentiated          4
 2 CNS Embryonal, NOS                  14
 3 CNS NB-FOXR2                         3
 4 CRANIO, ADAM                        24
 5 DMG, H3 K28                         20
 6 DMG, H3 K28, TP53 activated          9
 7 DMG, H3 K28, TP53 loss              39
 8 EPN, ST RELA                        28
 9 ETMR, C19MC-altered                  5
10 EWS                                 11
11 GNG, BRAF V600E                     12
12 GNG, KIAA1549-BRAF                   5
13 GNG, other MAPK                      4
14 GNG, RTK                             3
15 GNG, wildtype                       14
16 HGG, H3 G35                          4
17 HGG, H3 wildtype                    37
18 HGG, H3 wildtype, TP53 activated     5
19 HGG, H3 wildtype, TP53 loss         20
20 LGG, BRAF V600E                     26
21 LGG, BRAF V600E, CDKN2A/B            5
22 LGG, FGFR                            8
23 LGG, IDH                             3
24 LGG, KIAA1549-BRAF                 113
25 LGG, NF1-germline                    6
26 LGG, other MAPK                     12
27 LGG, RTK                            10
28 LGG, wildtype                       34
29 MB, Group3                          14
30 MB, Group4                          50
31 MB, SHH                             31
32 MB, WNT                             10
33 SEGA, wildtype                      11
```

jharenza: oh wow, can you do a quick cd8_cd4_ratio.pdf plot of those - not ideal, but we can select based on what might have higher ratios, if more than cranio + MB group 3/4? If nothing else, then we can assess? For the larger subtype plot, the major thing I highlight in the paper is CD8 for MB group 3/4 and monocytes, myeloid DCs for DMGs. 

Also, it looks like the number and order of subtypes in`quantiseq-cell_types-molecular_subtypes.pdf` and `cd8_cd4_ratio.pdf` are slightly different with the EPN, extra embryonals in the first and cranio being in a different spot. The struggle.

sjspielman: Yup, here's that plot with all N>=3 subtypes across all cancer groups. One HGG, H3 WT sample is a bit elevated, but nothing else really stands out strongly.
[cd8_cd4_ratio_ALL-SUBTYPES.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/9193615/cd8_cd4_ratio_ALL-SUBTYPES.pdf)


> Also, it looks like the number and order of subtypes inquantiseq-cell_types-molecular_subtypes.pdf and cd8_cd4_ratio.pdf are slightly different with the EPN, extra embryonals in the first and cranio being in a different spot. The struggle.

The struggle indeed 😂 Since these plots are next to each other, I think it would make sense to show the same molecular subtypes in each of them, and in the same order. Maybe let's pick for CD8/CD4 and then use that in `quantiseq-cell_types-molecular_subtypes.pdf`?

jharenza: ok, how about cranio, dmg, hgg, MB, and the embryonal types (since we have MB) here for both plots ... or if easier, just remove GNG/LGG/SEGA bc there is nothing really to note in those subtypes, and keep the rest (would then also include EPN)?

jharenza (commented):
Hi @sjspielman - thanks for the updates. I just noticed two things. For the cancer groups plot in Figure 5, I think we can have a slightly larger text in the facet strips? Also, for the CD8/4 ratio plot, it looks like the ETMR (part of embryonal) is out of order of the broad histology.

sjspielman:
> For the cancer groups plot in Figure 5, I think we can have a slightly larger text in the facet strips?

I entirely agree!! But whenever I bump this up from the current value, it gets too big right away! It has been mystifying me. Maybe I'll try using `rel()` instead of setting an exact size.

> Also, for the CD8/4 ratio plot, it looks like the ETMR (part of embryonal) is out of order of the broad histology.

Ok, will check this.

sjspielman:
@jharenza more updates - 
- I tried fighting more with the facet text size, and it just won't budge. Feel free to five it a go if you want! I might have to do it manually in Illustrator?
- I realized I was missing a few N >= 3 subtypes for those histologies, so I ended up removing my hardcoding to programmatically get the right subtypes and order. Are there any new subtypes that got added in from this code change we should _remove_ now though?
- I also changed the boxplot color in the MB expression to a darker gray, since it was a bit harder to see the current gray with the `shape=16` addition, which again helps avoid alpha weirdness during compiling.

jharenza:
>I tried fighting more with the facet text size, and it just won't budge. Feel free to five it a go if you want! I might have to do it manually in Illustrator?

AHhh, I tried a bunch of combinations of wrapping the text after fewer characters and then increasing the strip text size, but see what you mean. Some names are too long and if you wrap too soon, you get 3 rows for DNET and then you go in circles.

>I realized I was missing a few N >= 3 subtypes for those histologies, so I ended up removing my hardcoding to programmatically get the right subtypes and order. Are there any new subtypes that got added in from this code change we should remove now though?

This looks good to me!

>I also changed the boxplot color in the MB expression to a darker gray, since it was a bit harder to see the current gray with the shape=16 addition, which again helps avoid alpha weirdness during compiling.

👍

jharenza (approved):
will approve now!

sjspielman:
This code does not run through CI, so can merge before checks.

PR #1584, opened 2022-07-27
Update generate-figures.sh to run scripts instead of copy certain panels

### LINKED ISSUE #1576: Update generate-figures.sh
With PRs #1569, #1570, and #1575, there are some changes to how manuscript figure panels are generated. Plots previously made in `mutational-signatures` and `immune-deconv` modules and then copied over will now be made in their own scripts in `figures/scripts/` without needing to copy any panels over.

Therefore, the script `generate-figures.sh` should be updated to run those new scripts instead of copying files from modules.

sjspielman:
Closes #1576 

Recent PRs #1569, #1570, and #1575 have updated the generation of certain figure panels. Rather than creating these panels in modules, they are now created in scripts in `figures/scripts/`. This PR updates `generate-figures.sh` to no longer copy relevant panels over from modules, but instead run the given scripts.


figures/generate-figures.sh
```diff
@@ -148,16 +150,15 @@ cp ${analyses_dir}/mutational-signatures/plots/cns/hypermutator_sigs_heatmap_leg
 
 # Create directory
 mkdir -p pdfs/fig5/panels
-
+# # This directory is also created here because one of the scripts also generates S6 panels
+mkdir -p pdfs/supp/figs6/panels
```
jharenza, line 154: just a general question about ordering - maybe these directories can be all made at the top of the script so that it is easier to follow which are being made altogether?

jharenza (commented):
Maybe this would also be a good time to rename the following scripts to have the same convention as the others:
```
supp-S3-panels-BCD.R
supp-S5-panels.R
supp-snv-callers-panels.R
supp-subtype-umap.R
supp-tmb-compare-panels.R
```

figures/generate-figures.sh
sjspielman: Sure!

sjspielman:
> Maybe this would also be a good time to rename the following scripts to have the same convention as the others:

So, we can rename everything to start with the figure names it is making and remove `supp` (is this what you mean?). If I take that step, I'll also update this in `figures/README.md`:

https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/ce8d583b95750f35b9a600f83f16623b7eb7e2aa/figures/README.md?plain=1#L96-L97

I also realize this is the right time to update `figures/scripts/README.md` too and get all these scripts documented properly.

jharenza:
Yes!

sjspielman:
@jharenza I've made several changes here - 
- Updated both figure README files. The overall README now explains the convention, and the scripts README table should now be fully filled out!
- Renamed relevant scripts with new convention
- Updated `generate-figures.sh` to use those script names and create directories up front. (FYI after this PR is done I anticipate un-drafting and wrapping up #1543)
- Updated `generate-figures.sh` to call the palette Rmd with V22 (I missed this before! it's on line 33)
- I then _re-ran_ the figure scripts line-by-line in `generate-figure.sh` to ensure no bugs, _and_ that the re-generated figures are stable! I anticipated diffs of 0 bytes for all the regenerated PDFs. I observe this in the `files changed` tab here for all but TWO figures - the heatmap and panel S3B. Visual inspection of these files do not reveal any differences (no re-compilation necessary!), so I'm chalking up the small diffs up to PDF embedding stuff.

jharenza (approved):
Thanks for updating all of that - this looks great!

sjspielman:
Code does not run through CI, so merging now!

PR #1585, opened 2022-07-28
Update `count-contributions` to not count the bot

sjspielman:
Inspired by this excellent `git` contribution towards authorship https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1578#discussion_r931619986, I am proposing here to modify the `count-contributions` module so that the bot does not get counted as a contributor. The code is flexible enough to allow for future bots etc. down the line as needed.

Since I was poking around, I went ahead and add in a `sessionInfo()` as well. I re-ran the module to generate an updated HTML with sessionInfo output in it, but I am not including _result_ changes in this PR since this code has been run in my branch which incorrectly "inflates" my commit count.


jashapiro (commented):
This looks good. don't have any content comments, but some process ones...

Can you stack these changes on #1562, please? That has some changes to the notebook as well, so that might avoid some conflicts later (I don't anticipate any with the rmd, but the html file is likely to be a mess).

On that front, it is probably worth noting that this notebook is run in CI by the rocker/tidyverse:3.6.3 image. I don't think that really matters, as nothing here should be particularly sensitive to package versions, but in case you want to minimize future diffs.

sjspielman:
> On that front, it is probably worth noting that this notebook is run in CI by the rocker/tidyverse:3.6.3 image.

Ah, that's right. `sessionInfo` might get conflict-y then. Do you think I should remove that new chunk, maybe just for now?

jashapiro:
> > On that front, it is probably worth noting that this notebook is run in CI by the rocker/tidyverse:3.6.3 image.
> 
> Ah, that's right. `sessionInfo` might get conflict-y then. Do you think I should remove that new chunk, maybe just for now?

I think it is fine. From past experience, the more annoying conflicts will be from slightly different rmarkdown & pandoc versions that change the html in minor ways throughout the document.

sjspielman:
> Can you stack these changes on https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1562, please? That has some changes to the notebook as well, so that might avoid some conflicts later (I don't anticipate any with the rmd, but the html file is likely to be a mess).

Going to close this and re-open, filing _to_ the #1562 branch.

sjspielman:
This has been re-filed here: https://github.com/jaclyn-taroni/OpenPBTA-analysis/pull/4.

PR #1594, opened 2022-08-03
Script to run analyses after release

sjspielman:
Closes #1593 

This PR adds a new script `script/run-analyses-after-release.sh`, as well as associated documentation in `scripts/README.md`, that will run modules that:
- Are used in the manuscript
- Are not needed for figures
- Produce results that will change after subtyping

I have retained the `USE_LOCAL` flag because this script runs `focal-cn-file-preparation` which needs a lot of memory.

Issue #1593, opened 2022-08-03
Add script to run analyses not used in figures


sjspielman:
This issue emerged from this discussion: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1543#discussion_r936705688

We currently have the following scripts to run OpenPBTA in various ways:
- `scripts/generate-analysis-files-for-release.sh`: Runs analyses needed to perform subtyping 
- `scripts/run-for-subtyping.sh`: Runs subtyping
- `scripts/run-analyses-for-figures.sh`: [Pending in #1543]: Runs analyses needed for making figures
- `figures/generate-figures.sh`: Runs figure generation scripts and copies over a few manuscript panels from analyses

What is missing from this list is a script that can re-run analysis modules _after subtyping_ which aren't needed for figure generation. To this end, we should create a new script `scripts/run-analyses-after-release.sh` (feedback for name @jaclyn-taroni?) which will run additional analysis modules that are used in the manuscript but not specifically for figures, and need to be re-run with the subtyping information. From what I can tell, these modules include the following which will need to consume the subtyped histologies file -

- [focal-cn-file-preparation](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/focal-cn-file-preparation)
- [fusion_filtering](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/fusion_filtering)
- [run-gistic](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/run-gistic)

jaclyn-taroni:
Should this be everything that isn't deprecated? Which is more than the three you've listed.

sjspielman:
> Should this be everything that isn't deprecated?

Including or excluding those in `generate-analysis-files-for-release.sh`?

jaclyn-taroni:
That _is the question_. I'm trying to wrap my head around what might be the least confusing at this point.

sjspielman:
Closing in favor of #1595

PR #1596, opened 2022-08-05
Script to run manuscript analyses

### LINKED ISSUE #1595: Script for running analyses after release
Based on separate conversation with @jaclyn-taroni, we will be using this new (`final.final` 🤞) approach to running analyses/figures. _**This approach displaces #1455.**_

0. When it is time to cut a new data release, run [`scripts/generate-analysis-files-for-release.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/scripts/generate-analysis-files-for-release.sh). This script runs non-subtyping analysis modules that are needed to generate data for the release.
1. Perform molecular subyping to include in the release by running the script [`scripts/run-for-subtyping.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/scripts/run-for-subtyping.sh)
    - _At this time, a release is cut_ which includes the relevant results from modules as well as the metadata file `pbta-histologies.tsv` that is generated during subtyping.
    - Locally, one can run [`download-data.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/download-data.sh) to update to the new data release.
2. Run all analyses that are not deprecated  _and_ are used in the manuscript. A full table of analyses [here](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses#readme) indicates which analyses satisfies these conditions. This table also shows which modules consume results from other modules, which is helpful for ensuring analyses are run in the correct order. This script does not yet exist, but we will create it by modifying [`scripts/run-analyses-for-figures.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/scripts/run-analyses-for-figures.sh) to run all such analyses, rather than just analyses used for figures. It should also be renamed, something like `run-manuscript-analyses.sh`.
3. The manuscript figure panels can be generated with [`figures/generate-figures.sh`](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/generate-figures.sh). This script assumes that the previous script (`run-manuscript-analyses.sh`) has been locally run and all necessary input files (including some `SQL` databases in `scratch/`!) exist. Checking is not generally performed.

sjspielman:
Closes #1595 

This PR adds `scripts/run-manuscript-analyses.sh` as well as associated documentation. 

I made some minor clarifying updates to the `analyses/README.md` file as well and fixed a typo. Please let me know if the added sentence here makes sense.

When reviewing, also see this comment, since I have two lines in there for running the telomerase and we'll need to pick one: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1595#issuecomment-1206495621


jaclyn-taroni:
@jashapiro can I get your opinion on this as well? I don't remember why we deviated from the plan originally in #1455:

> Create a fourth script, called something like `scripts/run-analysis-modules.sh`. This script will run all analysis modules that either i) were not run for subtyping, ii) were partially run for subtyping but now need to be fully run (e.g. GSVA), or iii) were run for sybtyping and need to be run again with the subtyped data.

What was added in #1543 was only analyses that fit those criteria ☝🏻  _and_ were required for making figures. #1593/#1594 was adding a 5th script that was other analyses in the manuscript fit those criteria but were not required for making figures.

I can see the point of including everything like what we have here, but the script being added here doesn't actually include everything because subtyping is entirely absent. I don't think we want to include subtyping, so it makes more sense to me to go back to what was originally stated in #1455 and add the following criterion: iv) needs to be re-run so files are available for figure creation (e.g., SNV callers). 

I would like to hear other opinions, though.

jashapiro:
> What was added in #1543 was only analyses that fit those criteria ☝🏻 _and_ were required for making figures. #1593/#1594 was adding a 5th script that was other analyses in the manuscript fit those criteria but were not required for making figures.
> 
> I can see the point of including everything like what we have here, but the script being added here doesn't actually include everything because subtyping is entirely absent. I don't think we want to include subtyping, so it makes more sense to me to go back to what was originally stated in #1455 and add the following criterion: iv) needs to be re-run so files are available for figure creation (e.g., SNV callers).

To clarify, is the difference here that  #1593/#1594 would have created two scripts and this PR combines those into a single script? So @jaclyn-taroni your preference would be to reopen #1594?

jaclyn-taroni:
My preference, which is not 100% solidified, is to have this script run modules that fit the following criteria:

* Non-deprecated modules that are used in the manuscript in some manner
* Were not run for data generation _unless_ the underlying `scratch` files are required for figures
* Were not run for subtyping or need to be re-run now with the disease label information in the release histologies file

instead of "running everything" included in the manuscript, because it isn't running everything – subtyping is missing and it should stay that way in my opinion. My argument is, in part, that running everything but subtyping might lull someone into a sense that they are running all the relevant modules, and that keeping things limited to what is necessary for the manuscript post-release is more straightforward.

jashapiro:
> My preference, which is not 100% solidified, is to have this script run modules that fit the following criteria:
> 
> * Non-deprecated modules that are used in the manuscript in some manner
> * Were not run for data generation _unless_ the underlying `scratch` files are required for figures
> * Were not run for subtyping or need to be re-run now with the disease label information in the release histologies file
> 
> instead of "running everything" included in the manuscript, because it isn't running everything – subtyping is missing and it should stay that way in my opinion. My argument is, in part, that running everything but subtyping might lull someone into a sense that they are running all the relevant modules, and that keeping things limited to what is necessary for the manuscript post-release is more straightforward.

Looking through in a bit more detail, it seems that there are modules run here (`independent-samples` was the one I noted) that _were_ run for data release generation in `generate-analysis-files-for-release.sh`, and I would agree those should _not_ be rerun here unless required  to populate `scratch`.

But I am fine with combining all scripts that need to be run for figures _or other ms results_, as was my general interpretation of replacing #1593/#1594 with this.

Are we saying the same thing?

jaclyn-taroni:
I believe we are saying the same thing, yea.

sjspielman:
Just going to note that I (for now!) included `independent-samples` here because the code would _not_ have previously been run with the _subtyped_ metadata file - https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/independent-samples#generating-sample-lists
It's been my sense that anything run with `generate-analysis-files-for-release.sh` would _not_ have used the release's subtyping information, so relevant modules that could consume the subtyped metadata should be re-run. I do feel pretty strongly that whatever results are in the analyses modules at "release" (preprint/pub/send repo to Zenodo) time should be up-to-date, and it seems to me if we don't re-run analyses that consume the metadata file, then we have outdated analyses. Does this make sense/am I missing something?

jashapiro:
> Just going to note that I (for now!) included `independent-samples` here because the code would _not_ have previously been run with the _subtyped_ metadata file

True, but I don't think it uses or writes that data at all, so there should be no need to rerun it.

sjspielman:
I've updated the docs here to be more precise about which scripts are run, and I removed modules that I think should be removed:
- `collapse-rnaseq`
- `copy_number_consensus_call`
- `hotspot-detection`
- `independent-samples`

I _think_ `fusion_filtering` and or `fusion-summary` can also be removed, but I wasn't certain? Feedback on those modules as well as the ones I did remove would be great, thanks!

figures/README.md
```diff
@@ -78,6 +78,7 @@ RUN_LOCAL=1 bash ../scripts/run-analyses-for-figures.sh
 _⚠️ This requires 64GB of RAM to run successfully! You can set `RUN_LOCAL=1` for local testing that skips the RAM-intensive steps._
 
 The script [`generate-figures.sh`](./generate-figures.sh) runs all scripts in [`figures/scripts/`](./scripts/) to generate manuscript-ready figure panels, as well as copies certain panels from analyses that are meant for the manuscript.
+**Note that it is ASSUMED that all results, including certain files in `scratch/`, will have been generated with `../scripts/run-manuscript-analyses.sh` before running this script.**
```
jashapiro, line 81: Should we add a check to the `generate-figures.sh` script? Or is the concern here that there will be old versions of the files hanging around? If so, I would make sure to say this directly.

If we want to get fancy, we could check in the figure script that the relevant files in `scratch/`  are not too old and print a warning if they are older than, say, a week.

jashapiro (commented):
This looks generally fine to me, though I will admit I did go through with a fine comb to be sure all of the analyses are there and in order. I did have a comment about checking requirements for later scripts though...

jaclyn-taroni (approved):
I agree with what came out. I'm returning my thoughts on the fusion modules and approving with the expectation that those will be implemented.

figures/README.md
sjspielman: >  If so, I would make sure to say this directly.
👍 

> If we want to get fancy, we could check in the figure script that the relevant files in scratch/ are not too old and print a warning if they are older than, say, a week.

I think this is a good way to check!

sjspielman: > though I will admit I did go through
Do you mean "did not"? Seems like doing it is not an admission :) 

I am pretty certain that the order is good, based on what results feed into other modules.

sjspielman:
I removed `fusion-summary`, but kept `fusion_filtering`, and I added some `scratch/` checks into `figures/generate-figures.sh` along with some updated documentation.

figures/generate-figures.sh
```diff
@@ -43,6 +43,39 @@ mkdir -p pdfs/supp/figs6/panels
...
+
+# Check for snv-callers databases:
+snv_pbta_db_check=${scratch_dir}/snv-callers/snv_db.sqlite
+snv_tcga_db_check=${scratch_dir}/snv-callers/tcga_snv_db.sqlite
+snv_pbta_scratch_days=$((($(date +%s) - $(date +%s -r "$snv_pbta_db_check")) / 86400))  # https://unix.stackexchange.com/questions/102691/get-age-of-given-file
+snv_tcga_scratch_days=$((($(date +%s) - $(date +%s -r "$snv_tcga_db_check")) / 86400))  # https://unix.stackexchange.com/questions/102691/get-age-of-given-file
+if [ ! -f $snv_pbta_db_check ] || [ ! -f $snv_tcga_db_check ]; then
+  echo "The 'snv-callers' module scratch databases do not exist. Please re-run the 'snv-callers' module first, or run the script 'scripts/run-manuscript-analyses.sh'."
+  exit 0
+fi
```
jashapiro, line 72: This check should probably be before the date calculations, or I expect they will fail. 
Also, since this is an error, `exit 1`

sjspielman: done!

figures/generate-figures.sh
```diff
@@ -43,6 +43,40 @@ mkdir -p pdfs/supp/figs6/panels
...
+# Check for snv-callers databases:
+snv_pbta_db_check=${scratch_dir}/snv-callers/snv_db.sqlite
+snv_tcga_db_check=${scratch_dir}/snv-callers/tcga_snv_db.sqlite
+if [ ! -f $snv_pbta_db_check ] || [ ! -f $snv_tcga_db_check ]; then
+  echo "The 'snv-callers' module scratch databases do not exist. Please re-run the 'snv-callers' module first, or run the script 'scripts/run-manuscript-analyses.sh'."
+  exit 1
+fi
+
+snv_pbta_scratch_days=$((($(date +%s) - $(date +%s -r "$snv_pbta_db_check")) / 86400))  # https://unix.stackexchange.com/questions/102691/get-age-of-given-file
+snv_tcga_scratch_days=$((($(date +%s) - $(date +%s -r "$snv_tcga_db_check")) / 86400))  # https://unix.stackexchange.com/questions/102691/get-age-of-given-file
```
jashapiro, line 73: Wow that is sadly ugly.

jashapiro (approved):
LGTM

figures/generate-figures.sh
sjspielman: This was the _least_ ugly thing I found in stackoverflow/stackexchange land, which may not be very surprising.

PR #1597, opened 2022-08-08
Table 2: update TMB/make germline variant names consistent

jharenza:
Per comments from code review [here](https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/300), I am rounding the TMB to one decimal and adding a consistent germline variant annotation + P/LP to table 2, using the clinvar database version utilized in the analysis as input. This was downloaded from [here](https://ftp.ncbi.nlm.nih.gov/pub/clinvar/tab_delimited/archive/variant_summary_2020-05.txt.gz) and subset for the 3 specific variants of interest found in the germline analysis. 

NA

How does this look?

#### Documentation Checklist


tables/output_tables.Rmd
```diff
@@ -107,12 +109,25 @@ histology_df %>%
 
 ## Table 2: Patients with hypermutant or ultra-hypermutant tumors
 ```{r table 2}
+# read in clinvar variant annotations / add P/LP annotation
+clinvar <- read_tsv(clinvar_file) %>%
+  mutate(`Pathogenic germline variant` = case_when(ClinicalSignificance == "Pathogenic" ~ paste0(Name, " (P)"),
+                          ClinicalSignificance == "Likely pathogenic" ~ paste0(Name, " (LP)"))) %>%
```
jashapiro, line 115:
```suggestion
  mutate(`Pathogenic germline variant` = case_when(ClinicalSignificance == "Pathogenic" ~ paste0(Name, " (P)"),
                                                   ClinicalSignificance == "Likely pathogenic" ~ paste0(Name, " (LP)"))) %>%
```

tables/output_tables.Rmd
```diff
@@ -107,12 +109,25 @@ histology_df %>%
...
+  mutate(tmb = round(tmb,1)) %>%
   select(Kids_First_Biospecimen_ID, tmb) %>%
-  left_join(histology_df) 
+  left_join(histology_df) %>%
+  # add germline variants
+  left_join(clinvar) %>%     
+  # add "None detected" to pathogenic germline variant column
+  mutate(`Pathogenic germline variant` = case_when(is.na(`Pathogenic germline variant`) ~ "None detected",
+                                              TRUE ~ as.character(`Pathogenic germline variant`))
+  )
```
jashapiro, line 130:
```suggestion
  Spacing, and removing what should be an unnecessary `as.character`
```suggestion
  mutate(`Pathogenic germline variant` = case_when(is.na(`Pathogenic germline variant`) ~ "None detected",
                                                   TRUE ~ `Pathogenic germline variant`))
```
```

File: tables/output_tables.Rmd [hunk 3]
```diff
@@ -149,15 +164,6 @@ soi_df <- tmb_hist %>%
       sample_id == "7316-2189" ~ "Lynch Syndrome",
       Kids_First_Biospecimen_ID == "BS_85Q5P8GF" ~ "Lynch syndrome",
```
jashapiro, line 165: I just noticed that the capitalization of "Lynch syndrome" here is inconsistent.  Also, can we be consistent about the id we use here? Presumably `sample_id` is fine for all of these.

Finally, elsewhere, there now appear to be no samples with only one predisposition, so we can probably remove the `(s)` from the header.

jashapiro (commented):
This looks good, I had only a few minor comments for consistency.

tables/output_tables.Rmd
jharenza: Sure, yeah I think I may not have used sample_id for all since there are some diffs between T/N sample ids and it was hard for me to keep track, but I think better to just use patient ID here, since it is a germline predisposition and it is patient-level, so I can update.

jharenza: Ah, I think I had `7316-515` as Other, NOS since that path report did not have a specific syndrome listed, but then another sample from the same pt `7316-2085` did have Lynch annotated. Since this is patient, not tumor-level, I updated this to match for the same patient.

jashapiro (approved):
LGTM

jharenza:
@jashapiro I just went back to path reports for 7316-515 and 7316-2085 to double confirm, and that patient actually has CMMRD, constitutional mismatch repair disease, which is a bit different than Lynch mechanistically, so I updated that one. Just wanted to make sure you saw since it was post-approval.

jashapiro:
> @jashapiro I just went back to path reports for 7316-515 and 7316-2085 to double confirm, and that patient actually has CMMRD, constitutional mismatch repair disease, which is a bit different than Lynch mechanistically, so I updated that one. Just wanted to make sure you saw since it was post-approval.

Thanks for letting me know. My only concern here is whether to use the abbreviation or full name.

jharenza:
I think abbreviation is ok for this since well-known and it can be written out in the caption.

PR #1610, opened 2022-08-29
update table 2 per KAC suggestions

jharenza:
This PR updates Table 2 per the following comments from Kristina Cole:

>For the table with the hypermutant samples, I would list radiation first, followed by the chemo since this is the order it's given (small detail). Also, do you think it will be clear to readers that for the progressive samples that the prior therapy (after the initial sample) would also have included radiation therapy?

I moved Radiation to earlier in the treatment list and for any pt with primary + sequential sample, I included previous Tx since Tx is noted as "post-biopsy"


tables/output_tables.Rmd
```diff
@@ -146,13 +146,13 @@ soi_df <- tmb_hist %>%
   filter(experimental_strategy == "WGS") %>%
   # manually add therapy field and update cancer predispositions not included in histologies file per [ticket](https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1318)
   mutate(`Therapy post-biopsy` = case_when(
-      Kids_First_Biospecimen_ID == "BS_VW4XN9Y7" ~ "Temozolomide, CCNU, Radiation",
-      Kids_First_Biospecimen_ID == "BS_20TBZG09" ~ "Temozolomide, Irinotecan, Bevacizumab, Radiation",
+      Kids_First_Biospecimen_ID == "BS_VW4XN9Y7" ~ "Radiation, Temozolomide, CCNU",
+      Kids_First_Participant_ID == "PT_3CHB9PK5" ~ "Radiation, Temozolomide, Irinotecan, Bevacizumab",
```
sjspielman, line 150: I'm not sure I understand why this ID was changed from biospecimen to patient?

jaclyn-taroni: There's no impact on the table outside of the intended impact, which is to include previous treatment for the progressive BS ID (`BS_8AY2GM4G`) in addition to the Initial CNS Tumor (`BS_20TBZG09`).

jaclyn-taroni (approved):
👍🏻  this code appears to do what you stated in your comment. I don't think we need to address anything in a caption.

jaclyn-taroni:
This does get run in CI as the last step, so it might be a bit of a pain to get through right now (with the frequency I expect we will hit a cache). Perhaps we do not need to hold up merging https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/338 so long as everything agrees.

jharenza:
@sjspielman since this is rerunning the whole tables script, I can update SnpEff --> SnpSift here, if OK with you?

sjspielman:
Noting this PR also closes #1611

sjspielman (approved):

PR #1617, opened 2022-10-25
UMAPs highlighting sequencing center

### LINKED ISSUE #1616: Revision: UMAP colored by sequencing center
Related to https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/366

To explore the potential for batch effects, we can make UMAPs colored by histology and _shaped_ by sequencing center. We can do this twice: once for all histologies, and once only for histologies that are shared among centers. These may be helpful supplementary figure panels (there's lots of room in S5!) to clarify that we do not anticipate major batch effects across centers.

sjspielman:
Closes #1616 
Related to https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/366 

This PR begins the process of exploring batch effects that may arise from samples being processed at different sequencing center. To this end, I created a script (that for now makes panels for figure S6, but this not necessarily set in stone!) to make three UMAPs, and we can discuss whether all/some/none of these are helpful. 
1. Colored by sequencing center and shaped by broad histology, and considers ALL histologies
2. Colored by sequencing center and shaped by broad histology, but considers ONLY histologies processed at >1 center
3. Colored by sequencing center with no indications for histology, and considers ALL histologies

The resulting UMAPs do not show very strong clustering by sequencing center, but it's a bit tricky to see some of the triangle and square points. If we want to use these plots, it might be helpful to code them out rather than using the `plot_dimension_reduction()` wrapper function, or we can add some arrows pointing to triangles/squares in Illustrator (or `ggplot2` if feeling extra fancy).

The main questions for reviewers are as follows:
- Are any of these useful inclusions for sequencing center batch concerns, and if so, do they fully address the concerns?
- Would additional non-UMAP visualizations (something like barplots of sequencing center/histology counts?) or tables be helpful?
- If something here is a useful inclusion, should it actually be panels in S6 or incorporated elsewhere?


figures/scripts/figS6_UMAPs_seq-center.R
```diff
@@ -0,0 +1,119 @@
...
+#  set order for `seq_center` as group size
+set.seed(2022)
+umap_df <- umap_df %>%
+  dplyr::sample_n(nrow(umap_df)) %>%
+  dplyr::mutate(seq_center = forcats::fct_infreq(seq_center))
+
+# UMAP of all histologies
+umap_all_bh <- plot_dimension_reduction(umap_df,
+                                        point_color = "broad_histology_display",
+                                        point_shape = "seq_center",
```
jashapiro, line 75: For this plot, I think it might be more clear to have the color by sequencing center and the broad histology shapes. Or you could just incorporate the shapes into the later plot where you color by sequencing center? I just found it hard to pick out the sequencing centers on this plot.

figures/scripts/figS6_UMAPs_seq-center.R
```diff
@@ -0,0 +1,119 @@
...
+# first, which histologies are shared, aka processed at >1 center?
+shared_histologies <- umap_df %>% 
+  dplyr::count(seq_center, broad_histology_display) %>%
+  dplyr::count(broad_histology_display) %>%
+  dplyr::filter(n>1) %>%
+  dplyr::pull(broad_histology_display)
+
+umap_shared_bh <- plot_dimension_reduction(dplyr::filter(umap_df, broad_histology_display %in% shared_histologies),
+                                           point_color = "broad_histology_display",
+                                           point_shape = "seq_center",
```
jashapiro, line 94: I liked this reduction in the number of points: it was definitely easier to see what was going on. I wonder if we can plot to a finer detail in this reduced plot though? Can we show this with cancer group, and if so, does it make a difference?

figures/scripts/figS6_UMAPs_seq-center.R
```diff
@@ -0,0 +1,119 @@
...
+pal_file <- file.path("figures", "palettes", "broad_histology_cancer_group_palette.tsv")
+rsem_umap_file <- file.path(
+  dim_red_dir,
+  "results",
+  "rsem_stranded_log_umap_scores_aligned.tsv"
+)
+
+
+# Output:
+output_dir <- file.path("figures", "pdfs", "supp", "figS5", "panels")
```
jashapiro, line 23: Assuming we are making a figure S6, as implied by the title.
```suggestion
output_dir <- file.path("figures", "pdfs", "supp", "figS6", "panels")
```

figures/scripts/figS6_UMAPs_seq-center.R
sjspielman: I agree that it's hard to pick out the centers. Something else I think will help is using outlined points w/ color and fill for low N sequencing centers, which means not using the plot wrapper (which is very fine).

figures/scripts/figS6_UMAPs_seq-center.R
sjspielman: S5 is definitely what it should _all_ say 😬 😝. This will go "away" in a notebook, either way. We can save potential figure exports for later.

sjspielman:
> One initial thought I had is that this might make more sense to start off as an analysis notebook, rather than as a "final" figure script, until we decide what the results we want to present really are.

:100: good call. I think an exploratory notebook in `transcriptomic-dimension-reduction` is appropriate/good enough for all of this.

sjspielman:
@jashapiro Thanks for the feedback, this has been updated as a notebook! I also modified the plotting approach overall to one I think is easier to see. HTML preview is here: https://htmlpreview.github.io/?https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/034a40a4e90af380e6e950897154948abadf9d48/analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.nb.html 

I ended up also including polyA data in the tables, but I'm not sure that I needed to. I figured can't hurt for an exploratory notebook anyways.

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -0,0 +1,311 @@
...
+  dplyr::distinct(cancer_group_display) %>%
+  dplyr::pull(cancer_group_display)
+
+
+stranded_df %>%
+  dplyr::filter(cancer_group_display %in% shared_cg) %>%
+  ggplot() + 
+  aes(x = UMAP1, y = UMAP2, 
+      shape = seq_center, 
+      fill = cancer_group_display,
```
jashapiro, line 294: Here too I think it would be worth included lower level diagnoses. We end up here with a lot of "other", which is not particularly helpful when we might be trying to dive deeper and understand the data. I know that we don't have a defined palette for these diagnoses, and probably not enough visual discrimination, but I don't think that is necessarily a problem at this point, when we are just exploring the data a bit.

jashapiro (commented):
Thanks for the updates! I definitely like the notebook better for exploring the results at this stage. 

To the end of exploration, I think my main comment here (repeated 2x in different places) is that I am curious to see what happens when we go beyond the "display" terms that we had chosen and back down to the detailed data. If some of the differences are due to finer scale diagnoses and not batch effects, we might expect to see that when we break things up more, even if it does not show up in the cancer grouping decisions that we made.

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -0,0 +1,311 @@
...
+
+### All diagnoses
+
+```{r}
+# How many sample for each diagnosis/center?
+umap_df %>%
+  dplyr::count(RNA_library, seq_center) %>%
+  tidyr::spread(seq_center, n)
+
+# Lets hone in on stranded BGI and BGI@CHOP - what are the diagnoses?
```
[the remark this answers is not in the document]
sjspielman: Hm, this is a good question! I think I did it _because_ these two were the least common, so I suspected they may have a much narrower set of diagnoses, vs NantOmics.
This does not need to stay, though..

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
sjspielman: I'm looking into this now with the original cancer groups, before we collapsed them into the display groups, and it seems there are N=59 raw cancer groups in stranded data. But, there are only N=9 such raw cancer groups that are sequenced in >1 center. Moreover, these cancer groups are ones that made it uncollapsed into `cancer_group_display`, so I'm not sure there's much more exploration here other than color explosions?
So maybe rather than going all the way back to original cancer groups, just "uncollapse" the "Other" grouping. I'll look into this next.

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -0,0 +1,433 @@
...
+      fill = broad_histology_display, 
+      color = seq_center, 
+      alpha = seq_center) + 
+  geom_point() + 
+  scale_shape_manual(values = c(21, 22, 24)) + 
+  scale_fill_manual(values  = pal_bh) +
+  scale_color_manual(values = c("grey70", "black", "black")) +
+  scale_alpha_manual(values = c(0.4, 0.6, 0.6)) +
+  ggpubr::theme_pubr() + 
+  guides(fill = FALSE, alpha = FALSE)
```
jashapiro, line 277: Can we have the `fill` guide here? I don't know my cancer types that well. Or maybe only on the plots where we reduce to histologies shared across sites?

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -0,0 +1,433 @@
...
+```
+
+Since there are `r length(total_cg)` cancer groups, we'll focus on visualizing only the `r length(shared_cg_full)` cancer groups that are processed at >1 center.
+
+
+```{r}
+stranded_df_subset <- stranded_df %>%
+  dplyr::filter(cancer_group %in% shared_cg_full)
+
+# Legend does not render for cancer group colors properly, so create a hack legend first -
```
jashapiro, line 397: Something still looks off on the legend here. There is a purple color that doesn't seem to be present in the legend? Or maybe it is just the pink and a trick of the light? But there should be 9 cancer groups, right? and there are only 7 in the legend.

jashapiro (commented):
All of this looks good to me, and I do appreciate the addition of the more detailed cancer types. 

I think the main thing missing at this point is a little summary of the results. Which might be just to say that there is not obvious pervasive clustering by sequencing center. Certainly for Medulloblastoma, the disease effects seem stronger than any sequencing center effect. It's maybe a bit harder to say for some of the other cancer types, but it is clear that there is not a huge amount of data by which we could make a clear determination (and even less from which we could really correct it)

jaclyn-taroni:
@jashapiro giving this a bump

sjspielman:
I've updated this with a conclusion and a UMAP with arrows pointing to potential "areas of caveat."
It was also re-run with V23 and there were no changes as expected!

[04-explore-sequencing-center-effects.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10384537/04-explore-sequencing-center-effects.nb.html.zip)

jashapiro (approved):
Thank you for adding the final section; I do think it helps with somebody coming into this fresh.

I added a note about the `ggplot2::annotate()` function, which you may already know about, but was quite a revelation to me when I found out about it.

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -0,0 +1,494 @@
...
+Overall, UMAP visualization shows that sequencing center is likely not a severe batch effect, and these visualizations emphasize that sequencing center is highly confounded with diagnosis making formal batch correction challenging.
+
+That said, there does appear to be a group of samples with differing from BGI@CHOP which cluster together, which may be a batch-induced effect.
+In addition, there are a couple DMGs that cluster with MBs, though there are multiple sequencing centers in this "cluster."
+Arrows are shown in the plot below pointing to these two caveated regions.
+
+
+```{r}
+cg_plot + 
+  geom_segment(
```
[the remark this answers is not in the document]
sjspielman: In this case I used `geom_segment()` since they're more or less equivalent but can easily switch it all to `annotate()`.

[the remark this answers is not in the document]
jashapiro: In end result yes, as long as the output isn't a pdf, especially if you want to edit in illustrator.

Issue #1621, opened 2022-11-03
Revision: Tumor purity and transcriptomics exploration


sjspielman:
From https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/377

Add an exploratory notebook that looks at relationship between tumor purity and some results from transcriptomics analyses. For example, we can look at:
- EXTEND scores vs tumor purity
- tp53 scores vs tumor purity

It's not immediately clear to me what we could do with GSVA scores, so tagging in some folks for discussion @jaclyn-taroni @jashapiro @jharenza 

Since this notebook will rely on the output from a couple other modules (`telomerase-activity-prediction` and `tp53_nf1_score` for example), it might be a good idea to go ahead and add a `tumor_purity_exploration` module to hold this notebook.

jaclyn-taroni:
My knee jerk reaction upon reading the relevant reviewer comment was "regress out" tumor purity. To quote [this blog post from Rob Alderman](https://www.rga78.com/blog/2015/9/23/intro-to-regression-part-8-multiple-regression-regressing-on-two-numeric-variables):

> By "regress out", we mean to "remove any correlative influence" of one variable from another, so that we can examine the variable in "isolation", so to speak. We do this by building a simple regression model for the two variables, treating one as the response and the other as the regressor (the one being "regressed out"), then taking the residuals and making some use of them (e.g. correlating them with another variable).
> 
> The residuals represent the "leftover" variance in the response variable after removing (i.e. "accounting for") any correlative relationship with the regressor variable. In other words, the residuals represent the portion of the response variable's variance that is NOT explained by the regressor variable. 

Knee jerk reactions, famously, are not always good.

sjspielman:
It sounds to me like you are proposing (with knee jerk caveat!) something conceptually along these lines - 
`lm(EXTEND ~ histology + tumor purity)`, and then regress out `tumor purity`. What to then do with the resulting residuals?

sjspielman:
Another idea (h/t @envest!) could involve running transcriptomics analyses again but only focusing on samples with >=X % tumor purity. 
If _vibes_ match between all samples and, say, only samples that are >=95% tumor, then we're good.

sjspielman:
[...]

PR #1622, opened 2022-11-03
Initiate tumor purity module

### LINKED ISSUE #1620: Revision: Explore tumor purity metadata
We would like to explore tumor purity a bit further as part of revisions. From https://github.com/AlexsLemonade/OpenPBTA-manuscript/issues/377, 

> We can also consider reporting more summary stats about tumor purity, e.g. mean/median and SD/range, and assess whether tumor purity is correlated to histology.

This can be an exploratory notebook in the `sample-distribution-analysis` module.

sjspielman:
Closes #1620 

This PR initiates a `tumor-purity-exploration` module for looking into tumor purity (called `tumor_fraction` in our metadata). I started a notebook for overall exploration as a jumping off point. As part of this review, please feel free to suggest other initial explorations we can do in this notebook. If we like this overall structure, next we can address transcriptomics aspects in other notebooks (or in more sections of this notebook?) in this module to address #1621.

The module has been added to CI and the `analyses/README.md` file with `TBD` in the column for whether this is used in the MS.


sjspielman:
> I was looking through and wondering if we should really be using the DNA tumor purity for RNA, though it does [appear](https://www.nature.com/articles/ncomms3612) they might correlate.

The initial exploration (_very initial_) I have done here is already using DNA tumor purity since we don't have the RNA equivalent in the metadata. I think it's overkill to derive for RNA.

jharenza:
>I think it's overkill to derive for RNA.

Understood! :)

sjspielman:
In f52472d  I added some chunks to explore sample distributions with some filtering thresholds for tumor purity as discussed here: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1621#issuecomment-1332201013

sjspielman:
These thresholds will need to be re-explored once #1629 is done - how do thresholds change when only co-extraction samples are considered?

sjspielman:
This is now actually ready for real review! The notebook now includes sections for exploring thresholds after filtering down to only `Same Extraction` samples, including both global thresholds and group-specific thresholds. A couple metadata file versions are exported that contain only relevant samples that make it past different filtering steps (same extraction, and then thresholding), which can potentially be used to facilitate sample filtering in other analyses.

[01_explore-tumor-purity.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10393015/01_explore-tumor-purity.nb.html.zip)

jharenza:
Hi @sjspielman thanks for working on this. I was able to go through and I had a few comments/suggestions.
1. I know you have this plotted, but can you also print (in the notebook), the values of the `cancer_group_tumor_threshold`? I generally like this idea, but I also wonder if we still need some minimum threshold across the board (for example, PXA have a low median so many of these tumors do not have high purity at all, so should we drop them and how do we do this statistically?)
2. I saw [this](https://www.frontiersin.org/articles/10.3389/fgene.2021.642759/full) paper, which assesses tumor purity in a few TCGA cancers and then uses purity as a covariate for different expression analyses. I don't think we need to create new analysis models, but it was nice to see that our overall median (~0.7) of tumor purity was exactly what they saw in their 3 cancers assessed. They then used this as their cutoff of "high purity" samples. 

So I think approach 1 is sound, does it answer the question of whether the overall cohort results recapitulate those of highly pure samples? Not to make this more complicated, but what we could do is use the approach in 2 to run a few modules with that data subset (specifically GSVA and UMAP) and see if we reproduce our initial results. I think GSVA will be reproduced if we have a large enough N because without purity correction, we have great signal in both of these modules. I just don't know if they would improve because of the sample size reduction. If we fail to, then it may be due to samples with still low purity in the mix, although using approach 1 may reduce our sample N enough to get a great UMAP.

At this time, I don't think we need to worry about TP53/EXTEND.

jharenza:
Can you also order the cancer group tumor_violins by ascending median? I think this would make a good supplemental figure.

sjspielman:
@jharenza For EXTEND and TP53, I'm thinking exactly what you say - replot distributions (or re-run ROC) and see if the conceptual interpretations hold up. We'd like to see if certain trends across cancer groups (like we're showing in Fig 4) are recapitulated with the high tumor purity data. Nothing _too_ fancy besides some `if` statements for filtering and re-running existing code (🤞 ).

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+The quantity `tumor_fraction` provides information about what percent of the sample was likely tumor. 
+We'll explore this variable:
+
+```{r}
+# Note that this quantity is only recorded, if ever, with WGS libraries: 
+presence_table <- table(metadata$experimental_strategy,
+                        is.na(metadata$tumor_fraction))
+colnames(presence_table) <- c("Tumor fraction known", "Tumor fraction unknown")
+presence_table
+```
```
jaclyn-taroni, line 61: Not a strongly held opinion, but this doesn't seem that useful to me.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+         
+# plotting function:
+tumor_violins <- function(df, display_name, display_hex) {
+  ggplot(df) + 
+    # order by ascending median
+    aes(x = forcats::fct_reorder({{display_name}}, tumor_fraction),
+        y = tumor_fraction, 
+        fill = {{display_hex}}) + 
+    geom_violin(alpha = 0.5) +
+    stat_summary() +
```
jaclyn-taroni, line 113: Is the default point for `stat_summary()` the median? If so, the two next plots don't look as I would I expect if ordering by median.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+tumor_violins(tumor_palette_df, broad_histology_display, broad_histology_hex)
+```
+
+
+
+
+Distribution across sequencing centers, highlighting cancer groups:
+
+```{r}
+# use jitter to highlight diagnoses
```
jaclyn-taroni, line 131: `{ggfunfetti}`

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+Estimating tumor fraction from RNA directly is challenging since methods focus on estimating proportion of immune cells and assuming the rest is tumor, which is not a fair assumption for many of these diagnoses.
+
+Therefore, for RNA data tumor purity estimates, we should instead directly use tumor fraction values but _only for those samples where DNA and RNA sequencing were performed on the same extraction_.
+We cannot really know the RNA tumor purity for samples that have separate DNA/RNA extractions.
+
+This extraction information is present in data release `>=v23`.
+Only RNA are annotated in this column, hence the `NA`s below:
+
+```{r}
+table(metadata$extraction_type, useNA = "ifany")
```
jaclyn-taroni, line 166: I think this is similar to my comment about the presence table above – if the `NAs` aren't useful, maybe we should drop them before you make the table.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+```
+
+Let's now consider only "Same extraction" samples (which is happily the majority of samples) whose RNA library was stranded.
+We're going to map this information using the `sample_id` information, as was done for the oncoprint module:
+https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/oncoprint-landscape/01-map-to-sample_id.R
```
jaclyn-taroni, line 171: Permalink here as well please.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+cat("\nunweighed mean", unweighted_mean)
+
+calc_weighted_mean <- function(df, group_var) {
+  df %>%
+    drop_na(tumor_fraction) %>%
+    add_count({{group_var}}) %>%
+    mutate(contribution = tumor_fraction * (n/sum(n))) %>%
+    summarize(weighted = sum(contribution)) %>% 
+    pull(weighted)
+}
```
jaclyn-taroni, line 290: My strong preference is that all these functions interspersed throughout should get their own section at the top of the notebook and a little bit more documentation (as a treat).

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+  count(broad_histology_display, name = "count_remaining") %>%
+  # join again to get n_tumors
+  inner_join(total_bh_df) %>%
+  mutate(percent_remaining = count_remaining/n_tumors) 
+remaining_bh
+```
+
+
+```{r}
+plot_remaining_frac2 <- function(df, var, title, fill) {
```
jaclyn-taroni, line 421: More descriptive function name?

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+    labs(
+      y = "Percent of samples that remain after thresholding",
+      title = title, 
+      subtitle = "text labels are remaining counts"
+    ) +
+    theme(axis.text.x = element_text(angle = 30, size = 8, hjust = 1))
+}
+
+plot_remaining_frac2(remaining_cg, cancer_group_display, "Remaining cancer groups", "darkolivegreen1")
+plot_remaining_frac2(remaining_bh, broad_histology_display, "Remaining broad histology groups", "darkolivegreen4")
```
jaclyn-taroni, line 438: This doesn't _really matter_ but my person preference would probably be to have the y-axis go to 1 and not use a guideline (makes for some difficult to read text as it currently stands)

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,503 @@
...
+```
+
+
+Let's also export this thresholded version of the data frame.
+
+```{r}
+readr::write_tsv(rna_stranded_df_thresholded, 
+                 rna_stranded_thresholded_df_file)
```
jaclyn-taroni, line 446: Only the IDs please! Or IDs and tumor fraction. If the other fields changed because of changes to the palette or histologies file, someone could be working off of outdated information in this file alone. Hopefully this doesn't matter in practice because those things don't change, but as a matter of practice...

jaclyn-taroni (commented):
Thanks for putting this together @sjspielman. I didn't note anything to be incorrect in this initial review. I did have a few organizational comments, mostly about where the functions should live. However, this notebook is also very long. I wonder if generating the results files – after the decisions are recorded in the notebook – is a logical place to split this up. 

I don't think you should spend that much more time on this pull request, though, so some of my comments are up for discussion.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
sjspielman: This table was helpful for me while coding this up see literally do we know _anything_ about tumor purity in the first place, so I prefer to keep it.

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
sjspielman: The default is `mean_se()` but I can set it to use median now that the plot got reordered 👍

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
sjspielman: ha woops, I can't believe I committed this 😂

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,259 @@
...
+         tumor_fraction, 
+         cancer_group_display, 
+         cancer_group_hex, 
+         broad_histology_display,
+         broad_histology_order,
+         broad_histology_hex) %>%
+  drop_na() %>%
+  # reorder by median fraction and ensure "Other" is always last
+  mutate(cancer_group_display = forcats::fct_reorder(cancer_group_display, tumor_fraction), 
+         cancer_group_display = forcats::fct_relevel(cancer_group_display, "Other", after = Inf),
```
jaclyn-taroni, line 122: Making sure Other is last is about to get "undone" by the function `plot_tumor_violins()`. I'd either remove this and the analogous broad histology step or move putting Other last into the function. I actually think what we might want is the option to drop Other that's `TRUE` by default in `plot_tumor_violins()` – the category is broad enough as to be confusing in this plot – and order by median no matter what. Alternatively, drop Other before you `plot_tumor_violins()` (this might be what I favor given the ease of implementation).

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
```diff
@@ -0,0 +1,259 @@
...
+
+As part of this, we want to find the ambiguous `sample_id` samples so that we don't include them, as described:
+https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/5fe10b88aaa5176fa6577ffd6c8388be05098e83/analyses/oncoprint-landscape/01-map-to-sample_id.R#L153
+
+> An ambiguous sample_id will have more than 2 rows associated with it in the histologies file when looking at tumor samples -- that means we won't be able to determine when an WGS/WXS assay maps to an RNA-seq assay
+
+
+```{r}
+# find the ambiguous samples to flag for removal
+# here the composition does not matter, unlike in the linked oncoprint filtering
```
jaclyn-taroni, line 188: Hmm... I'm not sure I agree with this. Not necessarily here... but when we look at tumor purity above, it strikes me as a little bit strange to include cell lines? I think this comment may be more important for the next notebook with thresholds or when we think about making plots using the thresholds.

However, it seems that it won't matter in practice because only one cell line is going to pass the same extraction filtering step.

analyses/tumor-purity-exploration/02_explore-tumor-purity-thresholds.Rmd
```diff
@@ -0,0 +1,357 @@
...
+              color = "red") +
+    labs(
+      y = "Percent of samples that remain after thresholding",
+      title = title, 
+      subtitle = "text labels are remaining counts"
+    ) +
+    theme(axis.text.x = element_text(angle = 30, size = 8, hjust = 1)) + 
+    ylim(c(0,1))
+}
+```
```
jaclyn-taroni, line 103: Not something you have to change, but there's no real reason not to put all the functions in one chunk. You're not adding Markdown formatted text around them to add explanation, and you're not showing any output.

analyses/tumor-purity-exploration/02_explore-tumor-purity-thresholds.Rmd
```diff
@@ -0,0 +1,357 @@
+---
+title: "Tumor purity exploration"
```
jaclyn-taroni, line 2: This could use a title that's distinct from the other notebook, she says as she has both HTML files open in different tabs...

analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd
jaclyn-taroni: Update: briefly looked at the _TP53_ classifier and EXTEND modules (analysis modules, didn't look at figures), I think we lump in cell lines to those analyses so 🤷🏻 

Maybe the composition included in plots should be made more clear like the fact that we always use the standed data.

analyses/tumor-purity-exploration/README.md
```diff
@@ -0,0 +1,15 @@
+## Tumor purity exploration module
+
+Author: SJ Spielman `@sjspielman`
+
+This analysis modules explores the tumor purity values and potential correlates.
+
+The full module can be run with `bash run_tumor-purity`.
+
```
jaclyn-taroni, line 8: Can't actually get this to work as a suggestion for code block reasons, but this is my suggestion

The full module can be run with (assuming your current directory is the root of this module):

```sh
bash run_tumor-purity.sh
```

Code block makes this easier to scan. At the very least, need to add `.sh`.

jaclyn-taroni (approved):
I've left some thoughts about composition that I think are mostly future thoughts for either figure legends or if we want to think about what exactly we're plotting when we revise `figures/`.  The only comment that's _must do_ here is the one about the README. I am going to go ahead and approve.

PR #1619, opened 2022-11-01
update EPN fusion gene `MAML2` in fusion summary module

jharenza:
#### What scientific question is your analysis addressing?
Fusion summary was mistakenly searching for the wrong EPN fusion for YAP1 subtype

#### What was your approach?
Update fusion summary to recode `YAP1-MAMLD2` --> `YAP1--MAML2`

#### What GitHub issue does your pull request address?
Requirement for #1618 

#### Which areas should receive a particularly close look?
Note: the header shifted

#### Is there anything that you want to discuss further?
No

#### Is the analysis in a mature enough form that the resulting figure(s) and/or table(s) are ready for review?

Yes

#### What is your summary of the results?
4 bs_ids with `YAP1--MAML2` fusion:
```r
> epn %>%
+   filter(`YAP1--MAML2` == 1) %>%
+   pull(Kids_First_Biospecimen_ID)
[1] "BS_QP6MSKRE" "BS_WYERXS4G" "BS_1J2WQ08M" "BS_3DCBDJAT"
```

#### Reproducibility Checklist

- [ ] The dependencies required to run the code in this pull request have been added to the project Dockerfile.
- [ ] This analysis has been added to continuous integration.

#### Documentation Checklist

- [ ] This analysis module has a `README` and it is up to date.
- [ ] This analysis is recorded in the table in `analyses/README.md` and the entry is up to date.
- [ ] The analytical code is documented and contains comments.

PR #1638, opened 2022-12-20
v23 epn add YAP1--MAML2 to fusions list and YAP1 calling, rerun (4/n)

jharenza:
While the fusion summary file was updated to add `YAP1--MAML2` (instead of `YAP1--MAMLD2` in [this](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1619) previous PR, the module was not updated to handle this extra fusion. Here, I am adding that fusion to the list (curiously, `YAP1--MAMLD2` was not previously in the list), and then adding it to the requirements list for YAP1 subtyping. 
<s>However, when I did this, the output file now has shifted columns and I am not well-versed in python to update this. </s> just kidding, I think my rstudio rendering was smushed, this seems fine.

Note: this will not pass CI until we add the new files, since this fusion is missing from the summary file.

#### Documentation Checklist


jaclyn-taroni (approved):
Seems reasonable!

PR #1640, opened 2022-12-21
V23 integrate (9/n)

jharenza:
#### What was your approach?

- Add a notebook for RNA extraction type `00-prepare-extraction-table.Rmd`
- Update `01-integrate-subtyping.Rmd` to add the `extraction_type` to `base_histology` and then to `pbta-histologies.tsv`
- Rerun module to get finalized histology file.

Double check the extraction status looks OK and if this is sufficient for downstream work/ any modifications are suggested.

Yes

`pbta_extraction_status`
`pbta-histologies.tsv` also has the newly subtyped YAP1 EPN samples now

```r
Same extraction	910			
Different tissue aliquot	49			
Same tissue aliquot	44			
NA	11	
```

#### Documentation Checklist


jharenza:
Although a private repo, I am tracking here that the upstream data wrangling for the `*aliquots.tsv` files has been submitted in PR [here](https://github.com/d3b-center/D3b-codes/pull/79) and will be reviewed by @ewafula. @ewafula please comment here after you have a chance to review! Thanks.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
```diff
@@ -0,0 +1,87 @@
...
+  mutate_all(na_if,"") %>%
+  # rename to be compatible with CBTN data
+  dplyr::rename(parental_aliquot_id = parent_aliquot_id)
+```
+
+# Combine CBTN and PNOC RNA sample and aliquot ids with extraction status
+```{r}
+combined <- pnoc_aliquots_status %>%
+  # only keep those with RNA-Seq (had extraction annotated)
+  filter(!is.na(extraction_type)) %>%
```
jaclyn-taroni, line 65: I think this is going to miss one case of RNA-seq (sample id `7316-4998`) that has `NA` for `extraction_type`. I assume that means it won't be from the same extraction?

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
```diff
@@ -0,0 +1,87 @@
...
+  dplyr::rename(sample_id = sample_id_rna,
+                aliquot_id = aliquot_id_rna) %>%
+  select(sample_id, aliquot_id, parental_aliquot_id, extraction_type) %>%
+  # combine with CBTN
+  bind_rows(cbtn_aliquots) %>%
+  # clean up parental_aliquot id
+  mutate(parental_aliquot_id = gsub("^.*\\[|\\]", "", parental_aliquot_id)) %>%
```
jaclyn-taroni, line 72: Unclear to me why this is necessary since we're dropping it elsewhere?

jaclyn-taroni (commented):
Thanks for getting this started @jharenza. I am returning one question and one instance where I think there might be a bug (or at the very least something I don't understand). 

Can you also commit the HTML output of the new R Markdown file to the repository please? That's the only way we'll record the output of `sessionInfo()`.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
jharenza: Oh, yes at this point, there is one RNA-Seq which had no matched DNA, so it doesn't have an extraction type

jaclyn-taroni: Should we include that though? Either as an `NA` or something else descriptive like `No matched DNA sample`

jharenza: It becomes NA in the final histology file, but yeah, maybe we should include here, since there are other NAs. There are also 20 which it seems Noel couldn't find and so they were not propagated into my list - Let me add those here as NA also so we can have all 1035 RNA ids in that extraction file.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
jharenza: Not _really_ necessary except that I aimed for a cleaner file 🤷‍♀️

ewafula:
@jharenza, I'll work to review. Hopefully will donentomorrow AM.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
jharenza: @jaclyn-taroni I update this to not only include this one, but add the remaining 20. Turns out, there are two samples with the same sample and aliquot ids, so the total is now 1033 in the final file.

jharenza:
@jaclyn-taroni I also found an instance in which the aliquot id was broken up for 7316-135 into 3 different IDs when doing the search for parental aliquot upstream, so I updated that sample in the CBTN file as well. Ready for re-review.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
```diff
@@ -0,0 +1,104 @@
...
+cbtn_aliquots <- read_tsv(file.path(input_dir, "cbtn_aliquots.tsv")) %>%
+  # necessary for combining with PNOC later
+  mutate(aliquot_id = as.character(aliquot_id),
+         # update aliquot id for id which was split into 3
+         aliquot_id = case_when(sample_id == "7316-135" ~ "116304.116311.116318",
+         TRUE ~ aliquot_id)) %>%
+  distinct()
+
+# Read in histologies file to get all RNA sample and aliquot ids
+rna_aliquots <- read_tsv(file.path(data_dir, "pbta-histologies.tsv"), guess_max = 3000) %>%
```
jaclyn-taroni, line 43: I believe you'd want this to be:

```suggestion
rna_aliquots <- read_tsv(file.path(data_dir, "pbta-histologies-base.tsv"), guess_max = 3000) %>%
```

analyses/molecular-subtyping-integrate/README.md
```diff
@@ -11,6 +11,11 @@ bash run-subtyping-integrate.sh
 
 ### Module contents
 
+`00-prepare-extraction-table.Rmd`creates a table `pbta_extraction_status.tsv` of RNA samples with `extraction_type` as one of the following:
```
jaclyn-taroni, line 14: If you're going to use the histologies file, I'd recommend adding this to `run-subtyping-integrate.sh` as well

```
Rscript -e "rmarkdown::render('00-prepare-extraction-table.Rmd', clean = TRUE)"
```

Do I expect we will need to run this again? And do I expect it would change even if we did? I hope not on both points, but if we need to rerun this for whatever reason, I'd expect the bash script to run both steps _unless_ you explicitly state that it won't in the docs. So that's the other option – write a note under the usage section that this not included in the bash script.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
```diff
@@ -0,0 +1,104 @@
...
+# How many samples in each group
+```{r}
+combined %>%
+count(extraction_type) %>%
+  arrange(desc(n))
+```
+
+
+```{r}
+sessionInfo()
```
jaclyn-taroni, line 102: The output from this chunk not in the HTML file I'm looking at locally. Can you hit run all and commit the output? If you take my bash script suggestion and rerun it, that would be feeding two birds with one scone.

jaclyn-taroni (approved):
A few more edits that I trust you will make. I don't need to see this again ✅

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
jharenza: yes, good catch!

analyses/molecular-subtyping-integrate/README.md
jharenza: yes, will add.

analyses/molecular-subtyping-integrate/00-prepare-extraction-table.Rmd
jharenza: yes, realize this didn't output because I reran the bash script without having this script in it. All good now.

ewafula (approved):
@jharenza, I have reviewed the code logic is look ok. I have reproduced the results using the `pbta-histologies.tsv` file of  OpenPBTA `release-v22-20220505` and the CBTN and PNOC input files included in the module. 

The result reproduced when reviewing are consistent with those on the OpenPBTA [PR #1640](https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1640) in checkout branch `v23-integrate` - same md5 checksum

DB3 v23 integrate
```
(base) wafulae@88665a17de0f:~/Projects/PediatricOpenTargets/TEMP/D3b-codes/OpenPBTA_v23_release_QC/results$ ~/Software/homebrew/bin/md5sum cbtn_aliquots.tsv pnoc_aliquots.tsv 
7ce0e6b3fa346f330c25a2a3614f0084  cbtn_aliquots.tsv
c9d10b426ae34063360a6b966750c740  pnoc_aliquots.tsv
``` 
PBTA v23integrate
```
(base) wafulae@88665a17de0f:~/Projects/PediatricOpenTargets/TEMP/OpenPBTA-analysis/analyses/molecular-subtyping-integrate/input$ ~/Software/homebrew/bin/md5sum cbtn_aliquots.tsv pnoc_aliquots.tsv 
7ce0e6b3fa346f330c25a2a3614f0084  cbtn_aliquots.tsv
c9d10b426ae34063360a6b966750c740  pnoc_aliquots.tsv
```

PR #1643, opened 2023-01-04
Revision: Update Table 1 to include Patient Ns

jharenza:
#### What was your approach?
Response to Reviewer 1, point 6 - add Patient Ns to Table 1

NA

logic ok?

I also updated the table headings so not to have to manually update in the paper later

Note: this may need a rerun after release due to added EPN subtypes.

Yes

Table 1 changes

Added Patient N column

#### Documentation Checklist


tables/output_tables.Rmd
```diff
@@ -99,16 +99,33 @@ histology_df %>%
...
+# Count broad histology and subtype combinations
+tumor_level <- subtype_n %>%
+  select(-Kids_First_Participant_ID) %>%
+  count(`Broad histology group`, `OpenPBTA molecular subtype`) %>%
+  dplyr::rename(Tumors = n)
+
+# Patient level
+pt_level <- subtype_n %>%
+  select(-sample_id) %>%
+  distinct() %>%
```
jaclyn-taroni, line 115: So this works because the only remaining identifier in this data frame will be `Kids_First_Participant_ID`, correct?

jaclyn-taroni (approved):
One question to check my understanding, but if the answer to that question is yes 👍🏻  no need for me to look again

tables/output_tables.Rmd
jharenza: yes, sorry for the delay!

PR #1653, opened 2023-01-20
TERTp mutations and EXTEND scores

### LINKED ISSUE #1651: Revision: update telomerase analysis to assess whether samples with TERTp have high scores
https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/master/analyses/telomerase-activity-prediction

Reviewer commented that we did not assess whether samples with TERT promoter (TERTp) mutations have high telomerase scores

Plot telomerase scores and highlight tumors with  _TERT_ promoter mutations C228T and C250T? 

SNV consensus MAF + hotspot MAF (to ensure we capture all TERTp)
telomerase scores

this week?

?

sjspielman:
Closes #1651 
This PR adds a short script that explores whether samples with specific TERTp mutations `C228T` and `C250T` also have higher EXTEND scores, which is part of our reviewer response.
There are 9 samples in our data that have these two mutations, and only 6 of those samples have corresponding stranded RNA libraries that we can use for EXTEND score comparison.
A wilcox test for EXTEND scores across groups is P=0.1196, so we have no evidence of increased scores for these samples. The sample size is of course _tiny_ though, so this is a major caveat.

My sense is we can pop the figure I made here along with the test results into the response to reviewers, and call it a day? I'm also curious if we prefer for me to make this a notebook since it's more exploratory?

#### Documentation Checklist


jharenza (commented):
Thanks for doing this - I think this will be fine, and with the combination of low sample N and the fact that the telomerase score is derived from not just one gene, but 13(I think?, will look up), makes this not unexpected. There is also no cutoff for "high" vs "low" scores. Now thinking about this more, I think we can also point to the TERT and TERC expression correlations to scores. Could be some samples have mutations, but have some yet to be known compensatory mechanism keeping telomerase in check. Now I am curious where these samples lie on the TERT and TERC expression plots, Figure S5, B-C? This would perhaps partially answer the question of whether these components of telomerase are over-expressed and likely or not likely to contribute to activation. Could you pop those into this script as well? 😬

sjspielman:
@jharenza I've added in those scatterplots. We see some trend with TERT (all points are above the line except one which is on the line) but less so with TERC. But, this does not really suggest that these samples have higher expression of either gene.
[TERTp_mutations_TERC_TERT_expression.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10481412/TERTp_mutations_TERC_TERT_expression.pdf)

jharenza:
> @jharenza I've added in those scatterplots. We see some trend with TERT (all points are above the line except one which is on the line) but less so with TERC. But, this does not really suggest that these samples have higher expression of either gene. [TERTp_mutations_TERC_TERT_expression.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10481412/TERTp_mutations_TERC_TERT_expression.pdf)

Thanks! OK, I think we can use this and explain there is heterogeneity - we see some have high scores and high TERT expression, some low/low, etc, but numbers are low.

jharenza (commented):
This looks good to me - do you want to change that TERT/TERC ggplot background and we can call it a day - these can go into supplemental? I think that plot is more informative than the boxplot (shows what the box plot shows + more info)

sjspielman:
> do you want to change that TERT/TERC ggplot background and we can call it a day - these can go into supplemental

That sounds good to me! But, I'd prefer do that in a separate PR that focuses just on modifying [the FigS5 script](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/5fe10b88aaa5176fa6577ffd6c8388be05098e83/figures/scripts/figS5-all-panels.R) to include this information, which means this PR would be more exploration towards that end.

jharenza (approved):
👍

PR #1657, opened 2023-01-25
EXTEND: re-analyze trends at tumor purity threshold

### LINKED ISSUE #1625: Revision: Recalculate EXTEND scores as re-assess at threshold
To assess the affects of tumor purity on our results, we should re-run the EXTEND score calculations, considering only samples that have `tumor_fraction>= median tumor fraction` (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1622).

We should then assess how well the following results hold up:

- TERT and TERC correlations
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L208
- Aggressive vs. low-grade tumor trends
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L209

sjspielman:
Closes #1625 
This PR re-assess EXTEND scores if generated from only tumors that pass a tumor purity threshold. A couple steps are taken:

- The EXTEND script itself now has an additional flag to calculate scores only from threshold-passing samples. The EXTEND algorithm does appear to use all genes for baseline calculations, so a full re-calculation was necessary here (as opposed to just using existing EXTEND scores but filtering down to only those samples that pass the threshold). Those scores are outputted in the `results/` directory for consumption by the new notebook....
- The module run script will now also run a notebook that compares new results to those we report in the manuscript
- There are a few other tiny changes:
  - Removed an extra back tic from tumor purity README that made it through! woops!
  - At some juncture one of the figure scripts got opened in VS Code, and it ate all the EOL spaces. No other changes are in that file.

Everything is broadly consistent with our existing results, but I do find some areas we could rephrase during revisions anyways. 
Here is the notebook:
[07_EXTEND-at-threshold.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10501543/07_EXTEND-at-threshold.nb.html.zip)


analyses/telomerase-activity-prediction/01-run-EXTEND.R
```diff
@@ -32,6 +37,35 @@ output_file <- opt$output
...
+    dplyr::pull(Kids_First_Biospecimen_ID) %>%
+    unique()
+
+  # Subset the `PTBA_GE_Standard` variable to those IDs
+  PTBA_GE_Standard <- PTBA_GE_Standard[,bs_ids]
+
+  # Quick check on filtering:
+  if (!length(bs_ids) == ncol(PTBA_GE_Standard)) {
+    stop("Error filtering IDs to use in EXTEND calculation with tumor purity thresholding on.")
+  }
```
jaclyn-taroni, line 65: I'm not sure what this check on length is accomplishing for us.

If `PTBA_GE_Standard` is a data frame (hard to tell given the context in the PR alone), would we expect an informative-ish error if an identifier in `bs_ids` was not present in the column names? 

Would it be better to check if the column names are identical to the vector of identifiers?

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
```diff
@@ -0,0 +1,206 @@
...
+
+# input files
+metadata_file <- file.path(data_dir, "pbta-histologies.tsv")
+palette_file <- file.path(root_dir, "figures", "palettes", "broad_histology_cancer_group_palette.tsv")
+extend_file_thresholded <- file.path(extend_dir, 
+                         "results",
+                         "TelomeraseScores_PTBAStranded_FPKM_thresholded.txt")
+extend_file_original <- file.path(extend_dir, 
+                         "results",
+                         "TelomeraseScores_PTBAStranded_FPKM.txt")
```
jaclyn-taroni, line 42: Possibly some editor weirdness here with the spacing? Not sure.

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
```diff
@@ -0,0 +1,206 @@
...
+The plot above shows distributions before (top) and after (bottom) filtering, ordered by original EXTEND score median values.
+Previously, we observed from https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L209:
+
+> We found aggressive tumors such as HGGs (DMGs and other high-grade gliomas) and MB had high `EXTEND` scores (**Figure {@fig:Fig4}D**), while low-grade lesions such as schwannomas, GNGs, DNETs, and other low-grade gliomas had among the lowest scores (**Table S3**).
+
+We see the order is broadly the same, so this means overall EXTEND score trends are largely the same.
+There are some minor differences and items to note:
+
+- First, there are no longer DIPGs for this plot (but these were excluded from the main text figure anyways since it's such a small group)
+- It's notable that pediatric meningiomas are _not_ highly aggressive, but have high scores (both without and with thresholding), so we may wish to clarify this further in the manuscript. also had high scores when analyzed previously so this result is not necessarily different but we might clarify it further.
```
jaclyn-taroni, line 154: Bullet could use another pass. See if I've retained your meeting.

```suggestion
- It's notable that pediatric meningiomas are _not_ highly aggressive, but have high scores (both without and with thresholding). 
Meningiomas had high scores when analyzed previously, so this result is not necessarily different, but we might clarify it further in the manuscript.
```

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
```diff
@@ -0,0 +1,206 @@
...
+TERT correlation:
+```{r}
+cor.test(tert_df$extend_score, tert_df$log2_fpkm)
+
+```
+
+
+Previously, we observed from https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L208
+
+> As expected, `EXTEND` scores significantly correlated with _TERC_ (R = 0.619, p < 0.01) and _TERT_ (R = 0.491, p < 0.01) expression (**Figure {@fig:S5}B-C**).
```
jaclyn-taroni, line 196: Any idea why these results are different? Maybe the lack of transformation before?

jaclyn-taroni (approved):
Returning some comments/questions, the most substantive of which is why we see a different numerical result than what we report in the manuscript.

I don't think what's in the notebook is incorrect, and I don't really need to take another look at the code. Thus, I am going to approve.

analyses/telomerase-activity-prediction/01-run-EXTEND.R
sjspielman: Yep, this was a CI-fixing adventure I had yesterday in modules where this was a df (and not a matrix like here)! This check should just look for names then, but checks where it's a df in other PRs' thresholding notebooks aren't needed.

As I write this, I'm now wondering if we could do with some more modularity. I'd like your thoughts @jaclyn-taroni for this -  I have some form of this code (identical except for whether the subsetting is on a matrix or df, which is easy to handle!) floating around all the thresholding notebooks I've been writing. I'm now thinking this should be a `util` function in the tumor purity module instead that each of those notebooks can call if filtering is turned on. I would make that change here, and then we'd probably just wait to merge this until tackling the other ones.

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
sjspielman: The data going into the test is different (here, we're just testing a subset of the full stranded cohort), so I wouldn't expect the same values anyways? 
The data were previously transformed as well; this `mutate()` happens before modeling - https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/60791aac1eaa252137eaa1ddbe57bb4a7a1c51b1/figures/scripts/figS5-all-panels.R#L135

jaclyn-taroni: Oh I see, I got confused - I thought the two lines of code were old results vs. new results and not two different genes.

analyses/telomerase-activity-prediction/01-run-EXTEND.R
jaclyn-taroni: A `util` function in the tumor purity module module makes sense to me. But I'd recommend breaking that out into it's own PR intended to go in first, then propagate the use of that function in the currently open PRs. It gives us more flexibility in getting things in, which seems good.

jaclyn-taroni: I'm going to slap a don't merge label on the open PRs...

sjspielman:
@jaclyn-taroni The tumor filtering function has now been incorporated here. I re-ran the module and there were no diffs in any results, so seems good as long as CI agrees 😄

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
sjspielman: Cleaned this up in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657/commits/1b5164e14ea05429b6cdb2d1bf868f1c9f15ef2b.

analyses/telomerase-activity-prediction/01-run-EXTEND.R
```diff
@@ -8,6 +8,7 @@ suppressPackageStartupMessages({
   library(optparse)
 })
 stringsAsFactors=FALSE
+`%>%` <- dplyr::`%>%`
```
jaclyn-taroni, line 11: I assume this got added because of `filter_expression_by_tumor_purity()`. From a design perspective, tucking this into the function itself might be preferable.

jaclyn-taroni:
@sjspielman reviewing now - I think something funky is happening with the HTML output that's committed/pushed at the moment. Specifically in the re-analysis 2 section and beyond.

analyses/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh
```diff
@@ -21,33 +21,38 @@ if [ "${RUN_FOR_FIGURES}" == "1" ]; then
...
 
-else 
+  # Generate again, but this time with tumor purity thresholding
+  Rscript --vanilla 01-run-EXTEND.R --apply_tumor_purity_threshold --input ../../data/pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds --output results/TelomeraseScores_PTBAStranded_FPKM_thresholded.txt
 
-  #generate collapsed data for count files 
+  # Explore thresholded results
+  Rscript --vanilla -e "rmarkdown::render('07_EXTEND-at-threshold.Rmd')"
+
+else
```
jaclyn-taroni, line 30: Off the top of my head, I don't really know why we'd expect this to be run when we're running things for figures. Can you tell me about your thought process here? Related (at least in my mind): There's no corresponding change in `analyses/telomerase-activity-prediction/README.md`.

analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd
```diff
@@ -0,0 +1,207 @@
...
+There are some minor differences and items to note:
+
+- First, there are no longer DIPGs for this plot (but these were excluded from the main text figure anyways since it's such a small group)
+- It's notable that pediatric meningiomas are _not_ highly aggressive, but have high scores (both without and with thresholding). 
+Meningiomas had high scores when analyzed previously, so this result is not necessarily different, but we might clarify it further in the manuscript.
+- Similarly, schwannomas are not actually terribly low in either set of distributions, so we may want to rethink this emphasis in the manuscript as well.
+
+
+
+## Re-analysis 2: Relationship to TERT and TERC expression
```
jaclyn-taroni, line 160: I wouldn't make this change, but I think I better understand my own previous confusion in the comment below now – it's a bit odd to plot both the original and high purity scores in re-analysis 1 but not here.

jaclyn-taroni (commented):
Returning a few comments @sjspielman. I don't want to approve in the absence of documentation updates + the funkiness with the HTML output. 

I am also a little confused about why the Fig 4 panels are being update here, which was probably something included earlier that I did not notice!

analyses/telomerase-activity-prediction/RUN-telomerase-activity-prediction.sh
sjspielman: Running for figures, in my mind, is equivalent to "manuscript-related tasks" and re-running with the filter like one of those situations to me!

jaclyn-taroni: I was going to say wouldn't we want to run it to make sure all analyses in the manuscript are up to date _or_ when we run for figures but then I noticed that we use this flag when we re-run all analyses included in the manuscript: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/059047a19025f4afd9143fceff174e6da3392747/scripts/run-manuscript-analyses.sh#L87

That strikes me as a little bit weird, but okay, not a hill I want to die on.

sjspielman: > That strikes me as a little bit weird, but okay, not a hill I want to die on.

We agree!

analyses/telomerase-activity-prediction/01-run-EXTEND.R
sjspielman: Done in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657/commits/15b058b61eaacebfa1ffa6b1ddc9f2655c1bc267

sjspielman:
@jaclyn-taroni I've fleshed out the docs with more information for these additions, and I added in some details about running the module in general. I'm glad I did this because some docs were missed in #1653 _and_ that script (06) didn't get added to CI 😬 . Should be set now!! 
That said, I'll note more generally that this module's `README` could be expanded quite a bit in other ways that are beyond the scope of this PR, and in the interest of staying on task I avoided too many recreational changes there.

I also went back and forth a bit thinking about this comment https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657#discussion_r1098928446, and I think that it really does make sense to always make sure that analysis is run. I moved the thresholding script/notebook out of the logic entirely so it'll always get called.

jaclyn-taroni:
I don't believe you've resolved the issue with the HTML noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657#issuecomment-1421054716 at this point, @sjspielman.

sjspielman:
> I don't believe you've resolved the issue with the HTML noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657#issuecomment-1421054716 at this point, @sjspielman.

Woops, somehow missed your comment up there about the HTML when I went ahead to re-request review! Fixing it now. Sorry about that confusion! At some point I was testing locally with CI data, and it borked the notebook.


> I am also a little confused about why the Fig 4 panels are being update here, which was probably something included earlier that I did not notice!

Hm, great question... Somehow they got opened along the way and VS Code ate all the EOL spaces are the diffs there, and then further somehow I missed it in `git status`. A series of unfortunate events. I'll go ahead and revert those files!

sjspielman:
Ok @jaclyn-taroni, feeling confident 🤞 that this is ready. Plots are reverted and here's the properly-rendered HTML:
[07_EXTEND-at-threshold.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10686662/07_EXTEND-at-threshold.nb.html.zip)

jaclyn-taroni (approved):
👍🏻

Issue #1601, opened 2022-08-12
Differences in mitochondrially-encoded gene expression between RNA-seq samples sequenced at BGI@CHOP vs. NantOmics


jaclyn-taroni:
This issue is reported by @ginamawla. I am including Gina’s writing and visualizations here with some modifications for the issue.

#### What data file(s) does this issue pertain to?

Transcriptomic data, specifically TPM: 

```
pbta-gene-expression-rsem-tpm.stranded.rds
pbta-gene-expression-rsem-tpm.polya.rds
```

But I believe this is most relevant to the stranded data.

#### What release are you using?

Unknown

#### Put a link to the relevant section of the [OpenPBTA-manuscript](https://github.com/AlexsLemonade/OpenPBTA-manuscript/) here.

https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/ee2e934a9be7cdc77d883c031cf763e10ed49035/content/06.methods.md#data-generation

#### Put your question or report your issue here.

_Summary_

There are differences in mitochondrially-encoded gene expression – very likely to be artifactual in nature – between samples sequenced at BGI@CHOP and samples sequenced at NantOmics, despite no reported differences in processing in the manuscript. These differences have precluded classifying pediatric high-grade gliomas based on the RNA-seq data.

_From @ginamawla_

> 1. All of the pHGG tumor samples sequenced at BGI@CHOP that I am working with in my studies (n=12) contain ~200-300 fold levels of mitochondrial 12S and 16S rRNAs (encoded by the genes _MT-RNR1_ and _MT-RNR2_).  This result initially led us to believe that there was a novel subgroup of pHGG tumors, which we called the "mito" class. 
> 2. We were surprised to find that two samples, `BS_6VPKXXMR` and `BS_M85CXHDV`, originated from the same sample, but were sequenced at different centers (BGI@CHOP and NantOmics, respectively).  Interestingly, `BS_6VPKXXMR` fell into the "mito" subgroup, as defined by high expression of _MT-RNR1_ and _MT-RNR2_, whereas `BS_M85CXHDV` did not. 
> 3. Upon closer inspection, this time taking particular notice of the sequencing center, we found that  all 12/12 of the "mito" samples were sequenced by BGI@CHOP, and 12/12 BGI@CHOP-sequenced samples were of the "mito" group.
[...]

PR #1658, opened 2023-01-26
UMAPs without mitochondria

sjspielman:
This PR explores, as part of manuscript revisions, how dimension reduction might change if we do not include mitochondrial genes, as inspired by this issue: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1601
Specifically, it seems that one sequencing center has some problematic mitochondrial gene expression distributions that _may have_ resulted from using a different kit whose identity may also be lost to the ages.

There are two main changes in this module:

1. I updated the `transcriptomic-dimension-reduction` module to include an option to remove mitochondrial genes. I added this run for stranded RSEM with log2 (skipping tnse) in order to generate data for this exploration. But, I did not add this run to the 02 plot script , since we don't really need those.

2. I added a notebook to `transcriptomic-dimension-reduction` to explore the mitochondrial expression, first via just looking at FPKMs of mito genes for relevant diagnoses and then via UMAPs made with and without mitochondrial genes. I don't see anything _too_ different here, but welcome thoughts on interpretation. Here's that notebook:
[05-seq-center-mitochondrial-genes.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10510197/05-seq-center-mitochondrial-genes.nb.html.zip)

Note that I had some merge conflicts bringing `master` in, so I had to open and clean up some the tumor purity scripts (including re-rendering a notebook), and along the way there VS Code ate lots of EOL spaces, so that's why those diffs are here!


analyses/transcriptomic-dimension-reduction/01-dimension-reduction.sh
```diff
@@ -73,7 +73,24 @@ Rscript --vanilla scripts/run-dimension-reduction.R \
...
   --log2_transform
+  
+  
 
+#### RSEM log2 stranded, excluding mitochondrial genes and skipping t-sne ------------------------------------
+# Note that these results are only used for exploration in `05-seq_center-mitochondrial-genes.Rmd`
+Rscript --vanilla scripts/run-dimension-reduction.R \
+  --expression ../../data/pbta-gene-expression-rsem-fpkm.stranded.rds \
+  --metadata ${METADATA} \
+  --filename_lead rsem_stranded_no-mito_log \
```
jashapiro, line 84: I see here that the script is saving as `no-mito`, but I don't see any option where it is actually excluding mitochondrial genes. Am I missing something?

Maybe?
```suggestion
  --remove_mito_genes \
  --filename_lead rsem_stranded_no-mito_log \
```

analyses/transcriptomic-dimension-reduction/scripts/run-dimension-reduction.R
```diff
@@ -167,6 +173,32 @@ if (opt$log2_transform) {
...
+    dplyr::filter(!(stringr::str_starts(gene_symbol, "MT-"))) %>%
+    # remove wrangle columns
+    dplyr::select(-ensembl_id, -gene_symbol)
+  
+  # Convert back to data frame
+  expression_data <- as.data.frame(expression_data_tibble)
+ 
+  # Remove extra column and add row names back in
+  expression_data$ensembl_gene <- NULL
+  rownames(expression_data) <- expression_data_tibble$ensembl_gene
```
jashapiro, line 199: Do this all in one step? (I'm assuming this is in the tidyverse version we have, but I'm pretty sure it is)
```suggestion
  expression_data <- expression_data_tibble %>%
    tibble::column_to_rownames("ensembl_gene")
```

analyses/transcriptomic-dimension-reduction/05-seq-center-mitochondrial-genes.Rmd
```diff
@@ -0,0 +1,225 @@
...
+  dplyr::pull(broad_histology) %>%
+  unique()
+
+# Data frame of log2(FPKM) of mitochondrial genes for relevant diagnoses
+mito_seq_center_df <- expression_df %>%
+  dplyr::filter(stringr::str_starts(gene_symbol, "MT-")) %>%
+  tidyr::gather("Kids_First_Biospecimen_ID", 
+                "fpkm", 
+                tidyselect::starts_with("BS_")) %>%
+  dplyr::mutate(log2_fpkm = log2(fpkm+1)) %>%
```
jashapiro, line 110: Just want to note my minor discomfort with `+1` as a zero-correction for FPKM, but it does not really matter in this case.

jashapiro (commented):
This looks good to me code-wise, but I am a bit concerned about whether the UMAPs are what they should be, given https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1658/files#r1089064606

I'll wait for confirmation/rerunning before approval...

analyses/transcriptomic-dimension-reduction/05-seq-center-mitochondrial-genes.Rmd
sjspielman: For better or worse, this is all up in OpenPBTA :/

analyses/transcriptomic-dimension-reduction/scripts/run-dimension-reduction.R
sjspielman: yup, right version!

analyses/transcriptomic-dimension-reduction/01-dimension-reduction.sh
sjspielman: the return of the flag!

sjspielman:
@jashapiro this now filters as expected! For the record, here's how we bounce around scripts in that module:

- User calls `dimension-reduction-plots.sh` (or CI version) to run the module
- That shell script runs  `01-dimension-reduction.sh`, which itself calls `scripts/run-dimension-reduction.R` with various options
	- One of those options **now properly has** the `remove_mito_genes` flag
- `scripts/run-dimension-reduction.R` has a `remove_mito_genes` option and an associated if-statement for filtering out `^MT` genes

I regenerated the results, and they are about the same either way. I added a small "Conclusions" section to the notebook to be more clear about that.

analyses/transcriptomic-dimension-reduction/05-seq-center-mitochondrial-genes.Rmd
```diff
@@ -0,0 +1,234 @@
...
+```
+
+## Conclusions
+
+- Samples from `BIG@CHOP` sequencing center have unique mitochondrial gene distributions.
+- Removing mitochondtial genes from the UMAP does not have a strong qualitiative effect on whether broad histologies or cancer groups tend to cluster together.
+Visually, it seems the "mixed histology" groupings identified in `04-explore-sequencing-center-effects.Rmd`
+notebook still are grouped together in these UMAPs.
+
+Overall, this suggests that mitochondrial genes likely do not have a strong influence on UMAP visualizations.
```
jashapiro, line 228:
```suggestion
Overall, this suggests that mitochondrial genes alone do not have a strong influence on UMAP visualizations.
However, there may still be protocol differences associated with BIG@CHOP samples that influence expression values more generally, and can not be corrected by removing only mitochondrial genes.
```

jashapiro (commented):
I was about to approve this and say that this looks good to me, but then I realized that this may not actually be sufficient for "removing" the mitochondrial genes. The FPKM calculation will still be using the mitochondrial genes in the denominator, which will affect the expression values for all of the other genes _before_ we get to this point. So we may want to remove the mitochondrial genes from the count files, then recalculate FPKM. 

I think we can do this using the count matrices that we have, though I believe the original the FPKM calculation occurs upstream, so we might not get precisely the same results if we reimplement the FPKM calculations. (So we should recalculate with and without MT)  Before attempting this, we should probably discuss whether it is the right thing to spend time on at this moment.

analyses/transcriptomic-dimension-reduction/05-seq-center-mitochondrial-genes.Rmd
sjspielman: >  Before attempting this, we should probably discuss whether it is the right thing to spend time on at this moment.

👍

jashapiro:
> I was about to approve this and say that this looks good to me, but then I realized that this may not actually be sufficient for "removing" the mitochondrial genes. The FPKM calculation will still be using the mitochondrial genes in the denominator, which will affect the expression values for all of the other genes _before_ we get to this point. So we may want to remove the mitochondrial genes from the count files, then recalculate FPKM.
> 
> I think we can do this using the count matrices that we have, though I believe the original the FPKM calculation occurs upstream, so we might not get precisely the same results if we reimplement the FPKM calculations. (So we should recalculate with and without MT) Before attempting this, we should probably discuss whether it is the right thing to spend time on at this moment.

My main followup thought here is that we can probably do a rough version of this using the existing FPKM or TPM matrices: With TPM, I think we can re-sum the TPM values for each sample, excluding mitochondrial genes, divide by the sum/10^6 to rescale. That won't be quite right for FPKM, but it will probably be close enough. Still, I'd probably start with TPM for this notebook (calculating UMAP from the original TPM and then from the mito-excluded & rescaled TPM) as we are just looking to see if there is an effect of removing the mitochondrial genes. If we feel we need to go back to FPKM calculations later for better consistency with previous analyses, we can still do that.

sjspielman:
@jashapiro We have some TPM UMAPs 😄 
Here's the re-rendered notebook exploring the results. The plots don't look wildly different from FPKM which is a nice little sanity check to see. Overall trends look roughly the same with(out) mitochondrial genes.
[05-seq-center-mitochondrial-genes.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10679713/05-seq-center-mitochondrial-genes.nb.html.zip)

For this review, please me know in particular any feedback on how I did the TPM prep (scripts being called from right places, and `scratch/` export choice). I also kept the `COUNT_THRESHOLD` bash variable for filtering out low counts at 100, which seems more or less fine in the end.

analyses/transcriptomic-dimension-reduction/scripts/prepare-tpm-for-umap.R
```diff
@@ -0,0 +1,73 @@
...
+      key = "bs_ids",
+      value = "tpm",
+      tidyselect::starts_with("BS_")
+      ) 
+
+# Re-normalize now that mito genes are removed
+tpm_df <- tpm_df %>%
+  # calculate sample-level sum
+  dplyr::group_by(bs_ids) %>%
+  dplyr::mutate(total_tpm = (sum(tpm))/1e6) %>%
```
jashapiro, line 58: Minor style thing, but I find it is easier for me to follow what is done when it is all in a single step/equation. 
```suggestion
  # renormalize tpm
  dplyr::mutate(tpm = 1e6 * tpm/sum(tpm)) %>%
```

jashapiro (commented):
This looks fine. Glad to see that renomalizing doesn't really have much of an effect, and I like the additional comment you added about how this doesn't correct for everything that we might see from protocol differences.

The html file in the repo is not up to date, however, so that needs to be added before approving. I don't know if other data files are up to date, but I image a full rerun will address both.

Minor comments that do _not_ be to be addressed here:
I probably wouldn't have created a separate script for the TPM matrix filtering, as you already had most of the filtering code in the main script, but I don't see any major problem for this check.

If this were more than a one-off, I might also suggest not doing the gather/spread, but instead converting to a matrix and using matrix ops for efficiency, but for this it really doesn't matter.

It also seems odd to me that the umap tsv files have all of the metadata in them, but not worth the effort to change here.

analyses/transcriptomic-dimension-reduction/scripts/prepare-tpm-for-umap.R
jashapiro: Ugh, this comment isn't the full change...

analyses/transcriptomic-dimension-reduction/scripts/prepare-tpm-for-umap.R
```diff
@@ -0,0 +1,73 @@
...
+  # divde back out for new tpm
+  dplyr::mutate(tpm = tpm/(total_tpm)) %>%
+  # remove total
+  dplyr::select(-total_tpm) %>%
```
jashapiro, line 62: This is the change I meant to suggest:
```suggestion
  # renormalize tpm
  dplyr::mutate(tpm = 1e6 * tpm/sum(tpm)) %>%
```

analyses/transcriptomic-dimension-reduction/01-dimension-reduction.sh
```diff
@@ -74,6 +74,45 @@ Rscript --vanilla scripts/run-dimension-reduction.R \
...
+  --neighbors ${NEIGHBORS} \
+  --low_count_threshold ${COUNT_THRESHOLD} \
+  --skip_tsne \
+  --log2_transform
+  
+# Perform dimension reduction on nomito TPM
+Rscript --vanilla scripts/run-dimension-reduction.R \
+  --expression ${NOMITO_TPM_FILE} \
+  --metadata ${METADATA} \
+  --remove_mito_genes \
```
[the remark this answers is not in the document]
sjspielman: Ah yes, and the flag in general is not needed _at all_ anymore if we're not doing this with FPKM. I'll sort that all out.

sjspielman:
@jashapiro As of now, CI is passing through this module, so it's ready for another look! I updated the module to wholesale remove the `remove_mito_genes` flag since it's not needed at all anymore. I also re-ran the module to properly generate the HTMLs with the full data (not the CI data as was there before, woops!).  Here’s the notebook:
[05-seq-center-mitochondrial-genes.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10700632/05-seq-center-mitochondrial-genes.nb.html.zip)

Since the tumor purity had a diff (see my last bit here https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1658#issue-1558238784), I also decided just to be safe to re-run that module as well. All the notebook diffs are HTML miscellany.

PR #1661, opened 2023-01-27
Add Table 1 "Total" row

sjspielman:
As part of checking the updated Table 1 in https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/394, I wanted to ensure the totals were correct, and I do not trust my own arithmetic 😃 

I added a `Total` row to the bottom of Table 1 so that it exactly matches what's in the MS, and indeed it matches!


jharenza (approved):
Cool! I didn't know how to do that, so I was doing math - your way is better :)

PR #1663, opened 2023-01-30
Add tumor purity util filtering function

### LINKED ISSUE #1662: Add tumor purity util function
From: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1657#discussion_r1089040018

We should add a `util` function to the tumor purity module which can be used by other modules to filter the collapsed FPKM matrix (or data frame, if it was transformed) to only threshold-passing samples.
This function should then be incorporated into relevant bits of EXTEND, GSEA, and immune deconvolution modules, with potential to go into more modules as needed later.

sjspielman:
Closes #1662 

This PR adds a function, as well as associated docs, to filter expression data (either a df or matrix) to only samples that pass our tumor purity threshold. It can be used by other modules that need to be re-run at the threshold.


analyses/tumor-purity-exploration/util/function_filter-by-threshold.R
```diff
@@ -0,0 +1,39 @@
...
+  # Define vector of IDs to retain
+  bs_ids <- readr::read_tsv(tumor_purity_file) %>%
+    dplyr::pull(Kids_First_Biospecimen_ID) %>%
+    unique()
+  
+  # Subset the provided variable to these IDs
+  # To handle that the input can either be a data frame or matrix, we first
+  #  subset `bs_ids` to only the intersection with the input's names.
+  #  This allows us to use the same code to filter on either input object structure.
+  bs_ids <- intersect(bs_ids, colnames(expression_data))
```
jaclyn-taroni, line 32: Do we want to warn when samples are missing from the expression data frame or matrix? Potentially annoying in the CI case, but useful elsewhere (untested):

```suggestion
  bs_ids <- intersect(bs_ids, colnames(expression_data))
  
  missing_bs_ids <- setdiff(bs_ids, colnames(expression_data)
  if ( length(missing_bs_ids) > 0 ) {
     warn(glue::glue("{missing_bs_ids} is missing from expression data"))
  } 
```

analyses/tumor-purity-exploration/README.md
```diff
@@ -16,4 +16,6 @@ Contents of this module include the following:
...
 
 - `02_explore-tumor-purity-threshold.Rmd`: Notebook to explore resulting sample sizes after filtering samples based on different thresholds for tumor purity values.
-This notebook outputs the file `results/thresholded_rna_stranded_same-extraction.tsv`, which is a subsetted version of `results/rna_stranded_same-extraction.tsv` that contains only those tumors whose purity is >= the respective cancer group's median tumor purity value.
\ No newline at end of file
+This notebook outputs the file `results/thresholded_rna_stranded_same-extraction.tsv`, which is a subsetted version of `results/rna_stranded_same-extraction.tsv` that contains only those tumors whose purity is >= the respective cancer group's median tumor purity value.
+
+- The `function_filter-by-threshold.R` script define a functions `filter_expression_by_tumor_purity()` that can be used by _other modules_ to perform filtering on expression data based on the median per-cancer-group tumor purity threshold.
```
jaclyn-taroni, line 21:
```suggestion
- The `util/function_filter-by-threshold.R` script define a functions `filter_expression_by_tumor_purity()` that can be used by _other modules_ to perform filtering on expression data based on the median per-cancer-group tumor purity threshold.
```

analyses/tumor-purity-exploration/util/function_filter-by-threshold.R
```diff
@@ -0,0 +1,39 @@
...
+filter_expression_by_tumor_purity <- function(expression_data) {
+  
+  # tumor purity metadata file which has been filtered to only biospecimens that
+  #  survive the cancer-group-level threshold
+  tumor_purity_file <- file.path(
+    rprojroot::find_root(rprojroot::has_dir(".git")),
+    "analyses",
+    "tumor-purity-exploration",
+    "results",
+    "thresholded_rna_stranded_same-extraction.tsv"
```
jaclyn-taroni, line 20: I think this is _fine_ for this particular function, but registering my discomfort with the practice of hardcoding this file in here instead of using an argument 🙃

jaclyn-taroni (approved):
I think this looks good overall. One part of the path is missing in the docs, and I had a suggestion about maybe adding a warning when there are biospecimens missing from the expression data.

analyses/tumor-purity-exploration/util/function_filter-by-threshold.R
sjspielman: Don't disagree with you! This could also be an argument with that path as default.

sjspielman: Noting that I did that ^

sjspielman: ^ Noting I disagreed with myself since I decided I don't like having such a prescribed path, even if relative, as an argument. I did more things in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1663/commits/c96f96d3747cf886becb1b49f20ace75d4b3d691 and named argument to be named `bs_ids_file` and not have default. Instead, the file w/ ids to filter to can be specified in scripts that use it, which need to spell out a lot of the relevant to source this function anyways!

sjspielman:
@jaclyn-taroni I'll ask you to have one more peek here since I adventured a bit with specifying the file. Now it'll just be an argument provided to the function wholesale, no default.

jaclyn-taroni (approved):
👍🏻

Issue #1624, opened 2022-12-12
Revision: Run TP53 classifier and assess scores at tumor fraction threshold


sjspielman:
To assess the affects of tumor purity on our results, we should re-run the TP53 score calculation and classifier, considering only samples that have `tumor_fraction>= median tumor fraction` (https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1622).

We should then assess how well the following results hold up:

- ROC scores
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L194-L195
- activated/lost score difference is NS
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L197
- activated/lost expression difference is signif
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L199 
- overall trends in TP53 score distributions
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L200
- TP53 scores of LFS samples -
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L203
  - Note that we plainly state that 2 of these samples have low tumor purity of 16% and 37%. These samples would be excluded from any re-analysis, but this caveat sentence is the MS is fine/good to keep assuming this section is also fine to keep.
- TP53 scores for hypermutants samples (if those are included after thresholding, which is something to check!)
  - https://github.com/AlexsLemonade/OpenPBTA-manuscript/blob/b0fdf6f24fe9c021f22e8b7da348708f68877ee5/content/03.results.md?plain=1#L227

PR #1664, opened 2023-02-08
Update TP53 module to accommodate tumor purity filtering

sjspielman:
Part of #1624 

This PR begins the process of running TP53 with tumor purity filtered data. Since a lot of module changes had to be made to generate new results, this PR focuses only on that step. There will be a second PR that adds a final notebook to this module in order to compare these results to the original ones reported in the manuscript (those next steps are started in this branch: https://github.com/sjspielman/OpenPBTA-analysis/tree/tumor_purity-tp53-notebook).

There are a variety of changes that I made here, attempting to keep code as-is as much as possible. I created a new script `run_classifier-tumor-purity-threshold.sh` which runs the relevant scripts in this module to re-generate results with this filtered data. This script specifically calls (and does not call) the following for stranded data only:
  - `01-apply-classifier.py`
    - This script now takes two additional flags - one to indicate if we're turning tumor purity filtering on, and another with the path to the TSV file with ids to filter to. In addition, it ensures output files are identifiable as being tumor purity filtered.
  - ❌ `02-qc-rna_expression_score.Rmd` is not run as it does not produce any output that is consumed later.
  - `03-tp53-cnv-loss-domain.Rmd` and `04-tp53-sv-loss.Rmd`
    - These files produce inputs that are needed for `05-tp53-altered-annotation.Rmd`.
    - They now include an Rmd param and file name scheme to handle whether the data is tumor purity filtered.
  - `05-tp53-altered-annotation.Rmd`
    - Also now includes an Rmd param and file name scheme for export
  - `06-evaluate-classifier.py` 
    - generates files that are used to plot ROC curves, consuming output from  05-tp53-altered-annotation.Rmd`
  -  ❌ `07-plot-roc.R` is not run. ROC plots will be separately made in the forthcoming notebook 
  -  ❌ `08-compare-molecularsubtypes-tp53scores.R` and `09-compare-histologies.R` are not run since they are not really relevant here.

This script is documented in the README and also is in CI. After running this through and generating result files that can be analyzed, I also ran the normal pipeline again to ensure notebooks are rendered from the full dataset. 

I'll request review once checks pass!


sjspielman:
Since I'll be out for a couple days, just wanted to note where this is heading next. This notebook is still draft-y and not part of this PR, but sharing in case anyone is curious.
Link to current notebook Rmd:
https://github.com/sjspielman/OpenPBTA-analysis/blob/0a3202f71a73dad7dedec6a82612e0f74edb493d/analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd
HTML for download: [10-tp53-tumor-purity-threshold.nb.html.zip](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10701009/10-tp53-tumor-purity-threshold.nb.html.zip)

Importantly, I uncovered a couple areas where we reported outdated P-values, and 1-2 other small inconsistencies in the MS we should have our eyes on.

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
```diff
@@ -0,0 +1,82 @@
...
+python3 ${analysis_dir}/01-apply-classifier.py -f ${collapsed_stranded} -t
+
+# Run notebook to produce the `loss_overlap_domains_tp53` file consumed by 05-tp53-altered-annotation.Rmd
+Rscript -e "rmarkdown::render('${analysis_dir}/03-tp53-cnv-loss-domain.Rmd',params=list(tumor_purity_threshold = 1))"
+
+# Run notebook to produce the `fusion_bk_tp53_loss` and `sv_overlap_tp53` files consumed by 05-tp53-altered-annotation.Rmd
+Rscript -e "rmarkdown::render('${analysis_dir}/04-tp53-sv-loss.Rmd',params=list(tumor_purity_threshold = 1))"
+
+# Run notebook to produce `tp53_altered_status` file consumed by 06-evaluate-classifier.py
+Rscript -e "rmarkdown::render('${analysis_dir}/05-tp53-altered-annotation.Rmd',params=list(tumor_purity_threshold = 1))"
```
jaclyn-taroni, line 74: For all of these, I think you should utilize the `output_file` argument to `rmarkdown::render()` to save separate HTML outputs to the repo itself. We use `params` for controlling behavior in CI elsewhere without specifying a separate CI file, but those aren't actual results; this feels a bit different.

You could even put the HTML output in the tumor purity cases in a separate directory if you wanted.

I know you said: 

> After running this through and generating result files that can be analyzed, I also ran the normal pipeline again to ensure notebooks are rendered from the full dataset.

But I have seen the future and I've forgotten to do this step 🙃

analyses/tp53_nf1_score/util/plot_roc.R
```diff
@@ -8,13 +8,13 @@ suppressPackageStartupMessages({
...
 plot_roc <- function(roc_df, plots_dir, fname){
-
+  
   # add legend 
   roc_df <- roc_df %>%
     mutate(auroc = round(auroc, 2),
            shuffled = ifelse(shuffled, 'TP53 Shuffle', 'TP53'),
            Classifier = paste0(shuffled, ' (AUROC = ', auroc,')'))
-
+
```
jaclyn-taroni, line 17: These seem like the kind whitespace changes we don't want? Careful with `git add` !

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
```diff
@@ -0,0 +1,82 @@
...
+    > $cds_file
+
+  # Prep the SNV consensus data for evaluation downstream
+  Rscript --vanilla ${analysis_dir}/00-tp53-nf1-alterations.R \
+    --snvConsensus ${snvconsensus_file} \
+    --cnvConsensus ${cnvconsensus_file} \
+    --histologyFile ${histology_file} \
+    --outputFolder ${analysis_dir}/results \
+    --gencode ${cds_file}
+fi
```
jaclyn-taroni, line 59: Makes sense.

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# expression files for prediction from data release
+collapsed_stranded="${data_dir}/pbta-gene-expression-rsem-fpkm-collapsed.stranded.rds"
+
+# Run classifier and ROC plotting for stranded data, with tumor purity threshold turned on (`-t` flag)
+python3 ${analysis_dir}/01-apply-classifier.py -f ${collapsed_stranded} -t
```
jaclyn-taroni, line 65: I know there's a default value, but including the `--ids_file` here seems like it will make any potential future debugging or following the analysis easier.

jaclyn-taroni: Also just expressing my disdain for short flags in analytical code for the same reason. (You're following convention in `run_classifier.sh`, I realize. Don't change anything.)

analyses/tp53_nf1_score/03-tp53-cnv-loss-domain.Rmd
```diff
@@ -7,6 +7,10 @@ params:
     label: "1/0 to run with base histology"
     value: 0
     input: integer
+  tumor_purity_threshold:
+    label: Indicates whether this notebook should consume/output tumor purity thresholded or non-thresholded results for stranded data specifically. 1 indicates thresholded, and 0 indicates regular.
```
jaclyn-taroni, line 11:
```suggestion
    label: Indicates whether this notebook should consume/output tumor purity thresholded or full (non-thresholded) results for stranded data specifically. 1 indicates thresholded, and 0 indicates full results.
```

analyses/tp53_nf1_score/04-tp53-sv-loss.Rmd
```diff
@@ -7,6 +7,10 @@ params:
     label: "1/0 to run with base histology"
     value: 0
     input: integer
+  tumor_purity_threshold:
+    label: Indicates whether this notebook should consume/output tumor purity thresholded or non-thresholded results for stranded data specifically. 1 indicates thresholded, and 0 indicates regular.
```
jaclyn-taroni, line 11:
```suggestion
    label: Indicates whether this notebook should consume/output tumor purity thresholded or full (non-thresholded) results for stranded data specifically. 1 indicates thresholded, and 0 indicates full results.
```

analyses/tp53_nf1_score/04-tp53-sv-loss.Rmd
```diff
@@ -57,8 +61,44 @@ if ( params$base_run ==0 ){
...
+
+  # Load the function to filter IDs
+  source(
+    file.path(tumor_purity_dir, "util", "function_filter-by-threshold.R")
+  )
+
+  # Filter the expression data
+  stranded_expr <- filter_expression_by_tumor_purity(stranded_expr,
+                                                     tumor_purity_file)
+}else {
```
jaclyn-taroni, line 94:
```suggestion
} else {
```

analyses/tp53_nf1_score/05-tp53-altered-annotation.Rmd
```diff
@@ -7,6 +7,10 @@ params:
     label: 1/0 to run with base histology
     value: 0
     input: integer
+  tumor_purity_threshold:
+    label: Indicates whether this notebook should consume/output tumor purity thresholded or non-thresholded results for stranded data specifically. 1 indicates thresholded, and 0 indicates regular.
```
jaclyn-taroni, line 11:
```suggestion
    label: Indicates whether this notebook should consume/output tumor purity thresholded or full (non-thresholded) results for stranded data specifically. 1 indicates thresholded, and 0 indicates full results.
```

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
jaclyn-taroni: As I'm going through the changes to the Rmd themselves, I think doing this and skipping saving plots in `analyses/tp53_nf1_score/05-tp53-altered-annotation.Rmd` still makes sense.

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
```diff
@@ -0,0 +1,82 @@
...
+
+# Run notebook to produce `tp53_altered_status` file consumed by 06-evaluate-classifier.py
+Rscript -e "rmarkdown::render('${analysis_dir}/05-tp53-altered-annotation.Rmd',params=list(tumor_purity_threshold = 1))"
+
+# Produce files needed for ROC
+python3 ${analysis_dir}/06-evaluate-classifier.py \
+  -s ${analysis_dir}/results/tp53_altered_status_tumor-purity-threshold.tsv \
+  -f ${analysis_dir}/results/pbta-gene-expression-rsem-fpkm-collapsed.stranded_classifier_scores_tumor-purity-threshold.tsv \
+  -c ${histology_file} \
+  -o stranded_tumor-purity-threshold
```
jaclyn-taroni, line 81: Is this missing `--apply_threshold`?

analyses/tp53_nf1_score/06-evaluate-classifier.py
```diff
@@ -56,13 +56,25 @@
...
 scores_file = options.filename
 clinical = options.clinical
 outputfilename = options.outputfile
 
+# Set filename suffix for specifying output TSV filenames
+if options.apply_threshold is True:
+    filename_suffix = "_tumor-purity-threshold"
+else:
+    filename_suffix = ""
+
```
jaclyn-taroni, line 77: I don't think this gets used in `analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh`, and when you call this script in the shell script [you encode the fact that you're using tumor purity filtered data in the output base name](https://github.com/sjspielman/OpenPBTA-analysis/blob/09100897d2890d0a664ab8637e207bb303671096/analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh#L81): 

```
  -o stranded_tumor-purity-threshold
```

So is this (and the associated changes below) required at all?

jaclyn-taroni (commented):
Thanks for getting the changes to this module started @sjspielman. I agree with most of the choices you've made here. To summarize the most substantive comments:

* We should utilize the `output_file` parameter of `rmarkdown::render()` to specify separate HTML outputs. They could go in a different directory. You could even `.gitignore` them, I guess (I probably wouldn't recommend that). I am concerned about someone running this and forgetting to rerun and commit the HTML output of the full results.
* I don't think the alterations to `06-evaluate-classifier.py` are necessary, but I might have missed something.
* Once you make the HTML output change, you'll need to document it. (Note to myself: This is the reason that I haven't looked carefully at the `README.md` changes yet; I will in my next pass.)

There are also a few docs comments and preferences about being more verbose for future us throughout.

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
sjspielman: > But I have seen the future and I've forgotten to do this step 🙃

Evergreen 🌲 "Future You" problems indeed

analyses/tp53_nf1_score/util/plot_roc.R
```diff
@@ -9,13 +9,13 @@ suppressPackageStartupMessages({
 
 plot_roc <- function(roc_df, plots_dir, fname){
 
-  # add legend 
+  # add legend
```
sjspielman, line 12: Noting my `{styler}` options cleaned up EOL spaces here.

analyses/tp53_nf1_score/06-evaluate-classifier.py
sjspielman: Ah, yes I see I've duplicated this information in output file names! The `--apply_threshold` flag and all that goes with it are indeed not needed - the `filename` input options provides scores from the thresholded analysis as well, so everything is properly specified without needing this.

Therefore, while the `--apply_threshold` flag was indeed missing as noted in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1664/files#r1106490269, it's not going to get added in since it's being removed.

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
sjspielman: Yes, but now that flag is going away: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1664/files#r1107354897

analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh
```diff
@@ -0,0 +1,91 @@
...
+cd "$analysis_dir" || exit
+
+data_dir="../../data"
+scratch_dir="../../scratch/tp53-classifier"
+# Make sure scratch directory exists
+mkdir -p $scratch_dir
+
+# Output directory for tumor purity threshold results and HTML notebooks
+output_dir="${analysis_dir}/results/tumor-purity-threshold"
+mkdir -p $output_dir
```
sjspielman, line 36: Note this addition since last review.

sjspielman:
This is now ready for another look! 

- The biggest change is that all associated results from this new pipeline will live in `results/tumor-purity-threshold/` along with the notebooks. At first I did this just for notebooks as suggested, but then I decided a bit more organization would be nice since there are many result files. This involved code updates such as:
	- As suggested, I now use the `output_file` argument in `rmarkdown::render()` to specify separate HTML outputs. 
	- Within the threshold logic in associated scripts/Rmds, I made sure `results_dir` points to the right directory, while still making sure polyA results are always read in from the primary `results/` directory (since polyA wasn’t invited to the tumor purity party)
	- In the `06` python script, I did have to add some opt parse and function arguments to handle this. I set their defaults to what the main pipeline uses, which was hardcoded in here before these changes.
- The diffs in `results/tp53_scores_vs_molecular_subtype_Ependymal_tumor.tsv` (and its associated plot `plots/tp53_scores_vs_molecular_subtype_Ependymal_tumor.png`) are present because there is actually one fewer molecular subtype at `v23` compared to when this module was last run, hence the one fewer subtype in these results.
- I didn’t catch before - there’s a chunk in the `05` notebook that exports several plots of tp53 altered status by cancer predisposition. I “turned this chunk off” for the tumor threshold pipeline as we do not need to create those plots.

analyses/tp53_nf1_score/03-tp53-cnv-loss-domain.Rmd
```diff
@@ -76,7 +101,8 @@ score_stranded_df <- read_tsv(file.path(results_dir, "pbta-gene-expression-rsem-
   as.data.frame()
 
 # 2) Read in scores from tp53-nf1-classifier for polya
-score_polya_df <- read_tsv(file.path(results_dir, "pbta-gene-expression-rsem-fpkm-collapsed.polya_classifier_scores.tsv")) %>%
+# Note that this path must "results" even if the `results_dir` was updated, so we use `polya_results_dir`
```
jaclyn-taroni, line 104:
```suggestion
# Note that this path must be "results" even if the `results_dir` was updated, so we use `polya_results_dir`
```

jaclyn-taroni (approved):
👍🏻  thanks for these changes

Issue #1652, opened 2023-01-17
Revision: Replace "Ependymal tumor(s)" --> "Ependymoma" in figures


jharenza:
#### What analysis module should be updated and why?
Per reviewer comment, harmonize "ependymal tumor" and "ependymoma". We only have a few instances of "Ependymal tumor", so we can replace with "Ependymoma". 

#### What changes need to be made? Please provide enough detail for another participant to make the update.
1. Updates to broad histology display in the palette.
2. Update figures with "Ependymal tumor".
  - Figure 1B
  - Figure 5A
  - Figure S6B
  - Figure S6E

sjspielman:
Noting that this will hopefully "just happen" as part of the final Big Run™️, given changes being made in #1666 (which addresses item 1 in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1652#issue-1535779766).

sjspielman:
All set as part of the Big Figure Run™️: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1677#issuecomment-1438770171

PR #1666, opened 2023-02-15
Use Ependymoma instead of Ependymal tumor in broad histology display palette

jaclyn-taroni:
This pull request addresses the first part of #1652 that doesn't involve rerunning figures. (We might as well do that all together, I think!) Specifically, I'm updating the palette notebook such that the `broad_histology_display` value for tumors with `broad_histology == "Ependymal tumor"` will be `Ependymoma`. It is my hope that everything downstream will just work.

I also updated the palette notebook to use the most recent data release while I was at it because it seemed prudent!


figures/mapping-histology-labels.Rmd
```diff
@@ -55,10 +55,10 @@ library(RColorBrewer)
 
 We should perform some minimal checking to make sure the provided release is _at least_ at version 22.
 ```{r}
-minimum_allowed_release <- 22
+minimum_allowed_release <- 23
```
sjspielman, line 58: I feel both ways about this change...
- The code will run on v22, so that could be interpreted as minimal version
- We'd _prefer_ for the code to run on v23, which is another interpretation.

I leave this call to you! But, just make sure the text on line 56 matches how the code ends up.

jaclyn-taroni: Updated the comment and also took the following note out:

> **Note: This notebook was originally written with `release-v21-20210820`. Future releases may necessitate revisiting some of this code.**

Seemed a little unnecessary (future releases may necessitate revisiting _any code_ in the repo) and potentially confusing.

PR #1669, opened 2023-02-16
Recompiled S5 with TERTp regressions

### LINKED ISSUE #1656: Revision: Update compiled FigS5
From #1655, we'll need to update the compiled version of figure S5.

sjspielman:
Closes #1656 

This PR updates Figure S5 such that panels B and C now highlight tumors where known TERTp mutations are present. 
Note that these panels are directly "placed in" as images instead of being true PDF/vector objects (which is something we've done a lot of for SI figures but _not_ for main text figures).


jaclyn-taroni (approved):
LGTM

sjspielman:
There is no code committed in this PR (only `pdf` and `png` files), so this does not need to wait for checks to complete before merging.

PR #1671, opened 2023-02-16
Panels for a tumor purity figure, part 1

sjspielman:
Towards #1665 

This PR updates three modules, including code and docs, to export plots that can be used towards the tumor purity SI figure. As proposed in #1665, these are (with nested bullets indicating relevant changed files) - 

- **Panel A** distribution of tumor purity by cancer group: `analyses/tumor-purity-exploration/plots/cancer_group_tumor_fraction.pdf`
  - `analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd` and html
  - `analyses/tumor-purity-exploration/README.md`

- **Panel D** remake UMAP with high purity tumors only: `analyses/transcriptomic-dimension-reduction/plots/umap_tumor-purity-threshold.pdf`
  - `analyses/transcriptomic-dimension-reduction/analyses/transcriptomic-dimension-reduction/06-umap-tumor-purity-threshold.Rmd` and html
  - `analyses/transcriptomic-dimension-reduction/README.md`

- **Panel E** quanTIseq estimations of immune fractions across cancer groups: `analyses/immune-deconv/plots/tumor-purity-threshold_quantiseq-cancer-groups.pdf`
  - `analyses/immune-deconv/03-quantiseq-at-threshold.Rmd` and html
  - `analyses/immune-deconv/README.md`

Issue #1665, opened 2023-02-12
Revision: Assemble a supplemental figure comprised of tumor purity threshold panels


jaclyn-taroni:
We have a number of notebooks exploring transcriptomic analyses with high purity tumors. These are often in whatever module the analysis is performed in. Organizing them in this manner in the code base makes sense, but it does mean it's not that easy to point someone to a single place that captures all of the re-analysis. So we should perform the additional step of assembling these panels into a supplemental figure _regardless of whether or not it ultimately gets included in the manuscript_, which depends on the journal's guidelines.

Here's what I'm thinking (plan devised by looking at main display items in the manuscript):

- [x] **Panel A** distribution of tumor purity by cancer group https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/78e4969b99b4ebbc22915b9d8b58b5b3f6a9cad5/analyses/tumor-purity-exploration/01_explore-tumor-purity.Rmd#L128-L129
- [x] **Panel B** _TP53_ classifier ROC for stranded data (related to: https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1624)
- [x] **Panel C** _TP53_ Score (#1624) and EXTEND box plots https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/78e4969b99b4ebbc22915b9d8b58b5b3f6a9cad5/analyses/telomerase-activity-prediction/07_EXTEND-at-threshold.Rmd#L126
- [x] **Panel D** remake UMAP with high purity tumors only (I don't think this exists yet? So up for discussion.)
- [x] **Panel E** quanTIseq estimations of immune fractions across cancer groups https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/78e4969b99b4ebbc22915b9d8b58b5b3f6a9cad5/analyses/immune-deconv/03-quantiseq-at-threshold.Rmd#L93

This will probably require alterations to the notebooks that perform the new calculations/re-analysis to either save numerical results or the plots to get copied over to `figures/`.

sjspielman:
> remake UMAP with high purity tumors only (I don't think this exists yet? So up for discussion.)

Indeed, this does not exist yet, but I'm looking into it now. 

[...]

PR #1678, opened 2023-02-20
Tumor purity supp panels

sjspielman:
Towards #1665 

This PR takes additional steps to establish a tumor purity supp figure, which we're going to call S7.

- I wrote a new script `figures/scripts/figS7-tp53-telomerase-tumor-purity-threshold.R` to make a boxplot+legend that matches main text Figure 4D. I realize in #1671 I did export a [PDF of colorful violins](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/tp53_nf1_score/results/tumor-purity-threshold/tp53-cancer-groups_tumor-purity-threshold.pdf), but since an EXTEND plot was still needed I just went for it here. I guess we won't need the colorful violins, but they don't hurt anyone just sitting there! 
  - Note that this new script is heavily inspired by this section of an existing `figures/` script:  https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/f924b025bfa02be5dec3c777c0641f6035553b01/figures/scripts/fig4-tp53-telomerase-panels.R#L248-L456
  - I also added this script to `figures/scripts/README.md`, but indicated TBD which panel this is for now 
  - Note that I found a wrong link in the README which I fixed here (`supp-subtype-umap-panels.R` -> (`figs6-subtype-umap-panels.R`)
- I updated `figures/generate-figures.sh` to copy over previously-made panels and run `figures/scripts/figS7-tp53-telomerase-tumor-purity-threshold.R`.

After this PR, the final step towards completing #1665 is to compile panels. I'm thinking something like (and I'll find a place for the UMAP panel when I dive in?)
Note Feb 20, 2023


figures/generate-figures.sh
```diff
@@ -182,6 +182,32 @@ Rscript --vanilla scripts/figS6-subtype-umap-panels.R
...
+# Copy panels for this plot (specific panel order is TBD):
+
+# TP53 scores across TP53 status
+cp ${analyses_dir}/tp53_nf1_score/results/tumor-purity-threshold/tp53-score-status_tumor-purity-threshold.pdf pdfs/supp/figs7/panels/
+
+# TP53 expression across TP53 status
+cp ${analyses_dir}/tp53_nf1_score/results/tumor-purity-threshold/tp53-fpkm-status_tumor-purity-threshold.pdf pdfs/supp/figs7/panels/
+
+# TP53 classifier ROC curve
+cp ${analyses_dir}/tp53_nf1_score/results/tumor-purity-threshold/tp53-roc_tumor-purity-threshold.pdf pdfs/supp/figs7/panels/
```
jaclyn-taroni, line 201: TP53 shuffle is cutoff in this panel currently. Probably an upstream issue but noting it here.

figures/generate-figures.sh
```diff
@@ -182,6 +182,32 @@ Rscript --vanilla scripts/figS6-subtype-umap-panels.R
...
+
+# quanTIseq fractions across cancer groups
+cp ${analyses_dir}/immune-deconv/plots/tumor-purity-threshold_quantiseq-cancer-groups.pdf pdfs/supp/figs7/panels/
+
+# UMAP highlighting broad histologies
+cp ${analyses_dir}/transcriptomic-dimension-reduction/plots/umap_tumor-purity-threshold.pdf pdfs/supp/figs7/panels/
+
```
jaclyn-taroni, line 208: Have all of these been adequately accounted for in `scripts/run-manuscript-analyses.sh`? That script hasn't changed since August, and I know you added a separate shell script at least for the TP53 classifier (`analyses/tp53_nf1_score/run_classifier-tumor_purity_threshold.sh`) since then. Should probably add `tumor-purity-exploration` as well. So, I've answer my own question and now suppose the question is what else needs to make it back in there?  New 🎫 required.

jaclyn-taroni (approved):
I am approving because the new script looks good and so do the additions to `generate-figures.sh`. I did, however, uncover some issues:

* The TP53 ROC curve is cutoff on the right margin. You can fix that here or elsewhere (hence the approval.
* More of a lift/issue: the tumor purity threshold analyses have not yet (all) made it to `run-manuscript-analyses.sh`, which is a problem if we're including it in `generate-figures.sh` since the model is to run  `run-manuscript-analyses.sh` before `generate-figures.sh`. New ticket required for this.

figures/generate-figures.sh
sjspielman: Fixed here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1678/commits/34b9f40d39885607c3af6616f23997555e9ad010
Copied over to `figures/pdfs/supp/figs7/panels/` here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1678/commits/937adfbc6ff391acbf87fc9e3bde69831c331165

sjspielman:
@jaclyn-taroni I'm going to dismiss your review here and request a new one soon, since the code is wrong and we both missed it the first time!

In the new script I added here, the code currently reads in the "original" module result files:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/4c5a70e42de85d68d6e5897a2f0667ef5d1b2255/figures/scripts/figS7-tp53-telomerase-tumor-purity-threshold.R#L49-L53

And then filters down to relevant IDs from there:
https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/4c5a70e42de85d68d6e5897a2f0667ef5d1b2255/figures/scripts/figS7-tp53-telomerase-tumor-purity-threshold.R#L111-L112

That was wrong! I should be reading in the threshold result files in the first place, without needing to filter in this script. The files I should be reading in are as follows, and then `inner_join`ing with hypermutators for point colors.
- `analyses/tp53_nf1_score/results/tumor-purity-threshold/tp53_altered_status_tumor-purity-threshold.tsv`
- `analyses/telomerase-activity-prediction/results/TelomeraseScores_PTBAStranded_FPKM_thresholded.txt`

jaclyn-taroni (approved):

PR #1684, opened 2023-02-22
Compiled figure S7

### LINKED ISSUE #1665: Revision: Assemble a supplemental figure comprised of tumor purity threshold panels

sjspielman:
Closes #1665 

This PR adds a compiled figure S7, which shows results generated at a tumor purity threshold. This figure may or may not be used in the MS (given supplementary figure journal restrictions), but we have it in case as well as to use during response to reviews.

In addition to the PNG and PDF files, there are some minor code changes:
- I had to add `useDingbats=FALSE` to `ggsave()` statements in `analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd` for the two tp53 violin plot panels (`tp53-score-status_tumor-purity-threshold.pdf` and `tp53-fpkm-status_tumor-purity-threshold.pdf`)
  - I re-ran this notebook to re-generate those plots (heads up, that rendered HTML lives elsewhere: `analyses/tp53_nf1_score/results/tumor-purity-threshold/10-tp53-tumor-purity-threshold.nb.html`)
  - I copied over the updated figure panels to `figures/pdfs/supp/figs7/panels/`
- I updated comments in `figures/generate-figures.sh` and the table in `figures/scripts/README.md` to indicate which panels the plots ended up as


sjspielman:
Note to self, this does need to run through CI before merge since `tp53_nf1_score` code changed.

analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd
```diff
@@ -336,17 +336,19 @@ score_plot <- plot_tp53_status(tp53_thresholded_df, tp53_score, "TP53 Score")
...
 ggsave(
-  tp53_score_status_pdf, 
-  score_plot, 
+  tp53_score_status_pdf,
+  score_plot,
   width = 4,
-  height = 4
+  height = 4,
+  # needed for compilation
+  useDingbats = FALSE
```
jashapiro, line 351: I'm putting this comment here, because I can't really comment on the line I want to comment on or the figure, but I find the varying alpha in these figures to be a bit disconcerting. It made me wonder if there was meaning to the alpha or if there were stacked points. Is there a reason we can't just set all to alpha = 0.25?

jashapiro (commented):
This looks good overall, but I spent too long staring at the plots to try to determine if there was meaning to the alpha values. I think there is not? Or maybe it is to match varying alpha in Panel E (or that is an optical illusion)? If it is the latter, then I would suggest we should use shape instead.

analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd
sjspielman: The varying `alpha` in these figures is a strategy I used to make the points visible. Originally `alpha` was set to 0.25, but because the first two distributions have VERY low N (and hence no overlapping points), it was barely possible to actually see the points, just way too light. But, if you increase alpha, the third distribution starts to look wild with its much higher N and becomes too opaque. This was my middle ground..if you have other suggestions here I'll take em!

sjspielman: Here are the panels. The "activated" points are just hard to see/find, but looking at these plots fresh today, maybe it's not as _too_ problematic as I was worried about since the violin and median +/ IQR guide are there to help.

[tp53-fpkm-status_tumor-purity-threshold.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10814825/tp53-fpkm-status_tumor-purity-threshold.pdf)
[tp53-score-status_tumor-purity-threshold.pdf](https://github.com/AlexsLemonade/OpenPBTA-analysis/files/10814826/tp53-score-status_tumor-purity-threshold.pdf)

jashapiro: I don't find those bad at all. Most people will be looking at the summaries from the violin & IQR, as you note.

jashapiro (approved):
LGTM!

PR #1685, opened 2023-02-23
Code changes from running V23 on AWS

### LINKED ISSUE #1680: Revision: Big Run on AWS
We want to do Big Run on AWS to obtain a time/cost estimate.

```
RUN_LOCAL=0 bash scripts/run-manuscript-analyses.sh
RUN_LOCAL=0 bash figures/generate-figures.sh
```

This requires ^ scripts be all set for resubmission.

sjspielman:
Closes #1680 
This PR adds some small code changes that were needed to run on AWS, which indeed successfully ran! Two main code changes:

- `analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd` and associated notebook: Change in file system case sensitivity (Linux is case sensitive but Mac is not!)
- For the `analyses/snv-callers/run*` files, I had to add some flags to ensure files indeed get overwritten - specifically, the SQL dbs and a tarball. The code fails if these files already exist, so we need overwrite flags. For this, I ended up adding some variables for this such that these overwrite flags are only used when we run for the manuscript from here: `scripts/run-manuscript-analyses.sh`, rather than _always_ overwriting the sql db and tarball.


jashapiro (approved):
LGTM

PR #1687, opened 2023-02-24
Update ms N's notebook

sjspielman:
This PR does a quick update of a notebook in `sample-distribution-analysis` - it had the wrong data release param! Upon re-running, happy to report there are no differences in any outputted numbers (but please check me here!)

Before: https://htmlpreview.github.io/?https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/analyses/sample-distribution-analysis/04-ns-for-manuscript.nb.html

After: https://htmlpreview.github.io/?https://github.com/sjspielman/OpenPBTA-analysis/blob/v23-ms-numbers/analyses/sample-distribution-analysis/04-ns-for-manuscript.nb.html


jashapiro (approved):
LGTM

PR #1689, opened 2023-02-24
Updated table S2 with oncoprint alterations

sjspielman:
As discussed in https://github.com/AlexsLemonade/OpenPBTA-manuscript/pull/414#issuecomment-1444404958, we'd like to add another sheet to Table S2 with the primary only alterations, by cancer group.

I added that new sheet here, and also incidentally removed references to V22 in the tables Rmd file since we're well past V22!
I re-ran the tables module and did add all excel here, but you can see there are no diffs (very few bytes!) except for table S2, which has diff in the KBs: https://github.com/AlexsLemonade/OpenPBTA-analysis/compare/master...sjspielman:oncoprint-supp-table?expand=1#diff-9e47e4437e08269ed9d3066f8e81e9fd3a7168c5f78df07cb8e514d87be06275


jharenza (approved):
Looks great! Thanks, can also use this in reviewer comments.

sjspielman:
Noting I made one additional change - I realized there were some all-NA columns, and those are now removed!

PR #1703, opened 2023-03-20
Fix polya/stranded barplot

### LINKED ISSUE #1701: Figure S7 panel fix
From this comment:

https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1698#discussion_r1142559984

> I think there may be a bug here, which may be due to using participants instead of samples here - can you check out PT_9S6WMQ92 - this patient has two different events sequenced, one is stranded and one is polyA (medullo). Seems like we capture stranded in the plot but not the polyA. Caught my eye.

For this issue, have a look at what's up!

sjspielman:
Closes #1701 

This PR fixes a bug caught by @jharenza (🎉 🙏 ) where we had some wrong samples in a plot, because `sample_id` needed to be used for filtering out duplicates.
Code and plot fixed - you can see MB is non-0 now!


jharenza (approved):
👍🏻👏🏻

Issue #1692, opened 2023-03-15
Assemble files for new Zenodo submission


jaclyn-taroni:
This issue will almost certainly need to be broken up into more narrowly scoped tasks, but I am filing this to communicate an overview of what we need to accomplish. ⚠️  Please set any more narrowly scoped issues as blocking this one!

### Contents

We are going to create an upload to Zenodo that contains:

* A CSV file for each set of tabular data that underlies individual figure panels. In some cases, multiple panels use a lot of the same data – for example, Figure 4B and 4D both use TP53 classifier scores but have different factors on their x axes – so this is perhaps not as simple as one CSV per panel!
  * _Proposed rule of thumb_: If `figures/generate-figures.sh` copies a figure panel from module in `analyses`, output the underlying data in the analysis module and copy it to the right place in `figures/generate-figures.sh`; otherwise, include the step that outputs the CSV file in the figure script ([`figures/scripts/`](https://github.com/AlexsLemonade/OpenPBTA-analysis/tree/627ec427ad0a8d9d913e614c9db50546c56d8283/figures/scripts)) and write it directly where we need it to go.
  * All of these should be **ordered by `sample_id`**.
* A CSV file that contains a table of molecular alterations included in Figure 2 for all tumor and cell line samples in the PBTA dataset (implementation note: that can have their RNA and DNA specimens mapped to one another). This is where I recommend starting: https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/627ec427ad0a8d9d913e614c9db50546c56d8283/analyses/oncoprint-landscape/01-map-to-sample_id.R
* A `README.md` that describes all of these files.

### Proposed Structure and Conventions

I welcome feedback on this proposed structure, in particular. One alternative idea I had was be to put `zenodo-upload` in the root of the repository.

* Create `tables/zenodo-upload/` which will contain everything to be included in the upload.
* The subdirectory `tables/zenodo-upload/figure-data` will contain the CSV files with underlying figure data. 
  * Filenames can follow the convention: `<figure number and panel>-<hyphen separated description of data>.csv` For example: `fig2A-lgat-oncoplot-matrix.csv`
* `tables/zenodo-upload/summarized-molecular-alterations.csv` will contain the table of molecular alterations included in Figure 2 for all tumor and cell line samples in the PBTA dataset.
  * We might consider including a script in `tables` that generates this, and then running it via `tables/run-manuscript-tables.sh`.
[...]

PR #1704, opened 2023-03-21
Prepare for Zenodo tables: Update tables organization

sjspielman:
Towards #1692 

This PR begins the process of establishing CSVs for Zenodo upload by updating the `tables/` directory to accommodate forthcoming changes, as follows:

- `run-tables.sh` is now the name of the main script that runs the tables module. 
- `output_tables.Rmd` is renamed to `write-manuscript-tables.Rmd`, and tables are outputted to slightly different paths. Rather than sending all output to `results/`, output tables for main and supp tables respectively, go to `manuscript-tables/` and `manuscript-tables-supp/`. `results/` is now gone.
- Set up a directory `zenodo-upload/` to hold the forthcoming CSV files. It has a `.gitkeep` for now which will get removed once it is populated.
- I created `util/zenodo-utils.R` to facilitate writing CSV files. We will always need to order by a given sample column and then write out, so this function can be used throughout OpenPBTA where needed. 
- `README.md` updates accordingly
- Circle CI updates accordingly, including running the bash script `run-tables.sh` in CI, not only the notebook.

#### What to expect when you're expecting more PRs

Subsequent PRs will export figure data in one of two ways:

- For figures created in `figures/scripts/`, CSVs will be written directly to `tables/zenodo-upload/`. Individual scripts in `figures/scripts/` will get modified accordingly.
- For figures creates in `analysis/` modules, the CSVs will first be written out _within_ those analysis modules (as described in #1692). Then, we will need to copy those CSVs into `zenodo-upload/`. A script `copy-zenodo-tables.sh`, which is already planned for in `README.md`, will be added here to copy those specific files (a lot like copying figure panels in `figures/generate-figures.sh`).
  - Note, we will not want to run `copy-zenodo-tables.sh` in CI, so it will eventually get added to `run-tables.sh` inside some appropriate logic
  
#### Goals for reviewing this PR

Now is the time for feedback on this overall _organizational framework_ before I dive into writing CSVs!


tables/util/zenodo-utils.R
jashapiro, line 1: I kind of question the need for this file/function. I think it will be more clear to simply add in the two `arrange%>%write_csv` lines as needed. To use this, we still have to add an `source` line and then somebody has to go back and find the function definition.

More philosophically, I'd hold off on the introduction of a function until a PR where it is actually being used, for better context.

tables/write-manuscript-tables.nb.html
jashapiro, line 1: More changes in this file than I would expect. Maybe due to adding `clean=TRUE` when calling?

jashapiro (commented):
The organization part of this looks fine. I left a comment about dropping the `write_zenodo_table` function, at least from this PR, but probably overall. If it really is going to be used through a bunch of packages, I think that is a level of indirection/dependency that we would be better off avoiding.

tables/manuscript-tables/supp/TableS1-histologies.xlsx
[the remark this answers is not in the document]
sjspielman: In my experience, excel likes to do this. I'll make sure to run everything once more through for this and notebook diffs!

tables/util/zenodo-utils.R
sjspielman: Totally fine, I see this both ways for sure! I can purge the function, and just `arrange %>% write()` directly

tables/write-manuscript-tables.nb.html
```diff
@@ -3205,7 +3205,7 @@ <h2>Table 2: Patients with hypermutant or ultra-hypermutant tumors</h2>
...
+{"columns":[{"label":["Kids_First_Biospecimen_ID"],"name":[1],"type":["chr"],"align":["left"]},{"label":["cancer_group"],"name":[2],"type":["chr"],"align":["left"]},{"label":["short_histology"],"name":[3],"type":["chr"],"align":["left"]}],"data":[{"1":"BS_02YBZSBY","2":"High-grade glioma astrocytoma","3":"HGAT"},{"1":"BS_20TBZG09","2":"High-grade glioma astrocytoma","3":"HGAT"},{"1":"BS_8AY2GM4G","2":"High-grade glioma astrocytoma","3":"HGAT"},{"1":"BS_F0GNWEJJ","2":"Neuroblastoma","3":"Neuroblastoma"},{"1":"BS_HM5GFJN8","2":"Diffuse midline glioma","3":"HGAT"},{"1":"BS_P3PF53V8","2":"High-grade glioma astrocytoma","3":"HGAT"},{"1":"BS_VW4XN9Y7","2":"High-grade glioma astrocytoma","3":"HGAT"}],"options":{"columns":{"min":{},"max":[10]},"rows":{"min":[10],"max":[10]},"pages":{}}}
```
jashapiro, line 3208: Why did this change? This is a real output change, so it has my curiousity piqued.

jashapiro: never mind... this was a change in an intermediate diff... maybe you were testing with a limited dataset.

sjspielman: Yes, there had been some testing locally, which is all cleaned up now! The output rendered HTML is the same as what's currently in `master`.
Just for posterity:

## Currently in `master`
> https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/83545e5b38f9a71db376c3c1d34f5cb01e3e2bfe/tables/output_tables.nb.html
Screenshot 2023-03-21 at 3 33 48 PM


## This PR
Screenshot 2023-03-21 at 3 34 03 PM

sjspielman:
@jashapiro I'm pretty sure I fixed CI here. Happily there was a data cache so these fresh checks should be done soon-ish!
Function has been purged as well.

jashapiro (approved):
Assuming CI passes, this looks good to go.

PR #1705, opened 2023-03-22
First set of zenodo CSVs

sjspielman:
This PR gets the ball rolling with CSV files towards #1692 by doing a first round of each "paradigm" of table making: a) copying from an analysis module (interaction plots), b) directly made in `figures/scripts/`. (fig1b barplot).

- I exported data files for panels, named as `data-figure-<num figure><panel letter>.csv`. Note I'm keeping these panels in lowercase due to mac/linux filesystem differences...lowercase is safer code-wise imo.
- I established `tables/copy-zenodo-tables.sh`, including add it to the `run-tables.sh` script.
- I added some docs to `tables/README.md` about how CSV files are named and other background info for running the tables module.

my plan is to stack PRs from here, after we're definitely on board with the approaches I've taken here! 🎢


tables/README.md
```diff
@@ -68,4 +72,5 @@ Table S5: Key Resources
 
 ## Zenodo Upload
 
-> This section is forthcoming.
+Files in this directory are named as: `data-figure-<figure number><figure panel>.csv` (or `.csv.gz` if compressed.)
```
jashapiro, line 75: Naming suggestion, because of how my brain works... I don't know if this is really worth doing, but here we are:
`figure3a-data.csv` For supplementals, I would allow for capital letters, because the lower case looks very strange: `figure-S2b-data.csv`.

Leaving this as a standalone comment, but I'll review more as I go. But if you want to change these, I figure earlier better than later.

jashapiro (commented):
Looks good overall, but I made a couple of suggestions for this particular export...

tables/README.md
jashapiro: I actually meant to go back and add that... I had it for the supplemental because that one really does need it, so everywhere makes sense...

analyses/interaction-plots/scripts/03-plot_interactions.R
```diff
@@ -396,8 +405,22 @@ combined_plot <- disease_plot2 + cooccur_plot2 +
...
+
+# First the `cooccur_df` for 3A.
+cooccur_df %>%
+  dplyr::arrange(gene1) %>%
+  readr::write_csv(fig3a_csv)
+
+# Second, the `disease_df_fct` for 3B.
+disease_df_fct %>%
+  dplyr::arrange(gene) %>%
+  readr::write_csv(fig3b_csv)
```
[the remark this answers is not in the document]
sjspielman: good catch!

tables/copy-zenodo-tables.sh
```diff
@@ -0,0 +1,35 @@
...
+set -eo pipefail
+
+# Find current directory based on this script
+WORKDIR=$(dirname "${BASH_SOURCE[0]}")
+cd "$WORKDIR"
+
+# Get base directory of project
+cd ..
+BASEDIR="$(pwd)"
+cd -
```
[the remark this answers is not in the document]
sjspielman: I accepted this, but it doesn't actually work for me running from any directory... Probably something about directory mounting in docker. So, I'm reverting this, unless you have another idea here.

tables/copy-zenodo-tables.sh
```diff
@@ -0,0 +1,35 @@
...
+set -eo pipefail
+
+# Find current directory based on this script
+WORKDIR=$(dirname "${BASH_SOURCE[0]}")
+cd "$WORKDIR"
+
+# Get base directory of project
+cd ..
+BASEDIR="$pwd"
+cd -
```
jashapiro, line 18: It failed because of the earlier `cd`... 
```suggestion
# Find current directory based on this script
WORKDIR="$(dirname "${BASH_SOURCE[0]}")"

# Get base directory of project
BASEDIR="$(dirname "$WORKDIR")"
```

jashapiro (approved):
LGTM. Paths.

sjspielman:
@jashapiro one more update. I realized the problems (which I still have some of..) with paths are b/c of differences when running from `run-tables.sh` calling it directly.
I've found a maybe middle ground here by just assuming it's being run from `tables/`. I'm re-requesting review since this was a Choice.

jashapiro (approved):
LGTM

PR #1710, opened 2023-03-23
Zenodo CSVs: S2H-I and S6A-D

sjspielman:
The march to #1692 continues

These two panels are the only S2 panels where I do not need to be on AWS (those are forthcoming probably next week and are in adifferent script). Tried to prune down here to relevant variables without getting too aggressive.

This wraps up S6 exports! I did some `purrr` and felt cute.
I did a double take at first since all cancer groups are present in all UMAPs, which felt like it should be wrong, but upon re-looking at the figure it is actually correct.

Also If we're concerned with the wild commit/merging history from the stacking extravaganza, I can plop this into a new branch as well. Due to squashing I think it's ok though.


figures/scripts/figS2-tmb-compare-panels.R
```diff
@@ -178,7 +183,32 @@ ggsave(tcga_tmb_cdf_pdf, tmb_tcga_plot, width = 5, height = 6.15, # because labe
...
+                -cancer_group, -cancer_group_abbreviation) %>%
+  # remove \n from wrapped display column
+  dplyr::mutate(cancer_group_display = stringr::str_replace(cancer_group_display, "\n", " ")) %>%
+  # arrange
+  dplyr::arrange(Kids_First_Biospecimen_ID) %>%
+  # export
+  readr::write_csv(fig2h_csv)
+
+
+# S2H
```
jashapiro, line 201:
```suggestion
# S2I
```

figures/scripts/figS2-tmb-compare-panels.R
```diff
@@ -178,7 +183,32 @@ ggsave(tcga_tmb_cdf_pdf, tmb_tcga_plot, width = 5, height = 6.15, # because labe
...
-
+# Export Zenodo CSV files --------------------
+
+# S2H
+tmb_pbta_plot_df %>%
+  # reorder columns and remove unnecessary columns, but keep hex since identity was plotted
+  dplyr::select(Kids_First_Biospecimen_ID, everything(),
+                -cancer_group, -cancer_group_abbreviation) %>%
+  # remove \n from wrapped display column
+  dplyr::mutate(cancer_group_display = stringr::str_replace(cancer_group_display, "\n", " ")) %>%
```
jashapiro, line 194: excess display output
```suggestion
  # remove \n from wrapped display column
  dplyr::mutate(Primary_diagnosis = stringr::str_replace(Primary_diagnosis, "\n", " ")) %>%
  # remove "n = " from sample size column
  dplyr::mutate(sample_size = stringr::str_replace(sample_size, "n = ", "")) %>%
```

jashapiro (approved):
Looks good, with a couple little suggestions. I don't think I need to see it again.

sjspielman:
Code does not run through CI; merging before checks.

PR #1711, opened 2023-03-24
Figure S7 Zenodo CSVs

sjspielman:
Towards #1692.

This PR is a bit large and contains all S7 panels, since I couldn't help myself what with the branch name...  But, all exports are very straightforward!  
This figure has not yet been re-compiled (3 new panels are being added during this round of revisions), but here is [the previous compilation](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pngs/figureS7.png). The current panels you see are getting scooted down to become D-I, and new panels are A-C.

Here's the inventory. ⚠️ indicates that the code is in `analyses/` modules, and therefore copying was also added to `table/zenodo-upload/`. Nothing here is particularly tricky, so I have very few notes!
I also learned that it seems our tidyverse version does not give you `everything()` and similar helpers just because you're calling them inside of `dplyr::select()` 😞 

- Panel A: `figures/scripts/figS7-seqcenter-barplot.R`
  - This is a barplot of counts, so I went ahead and just counted for them (since plot was made with `geom_bar()`). - Panel B: `figures/scripts/figS7-UMAP-libraries.R`
- Panel C: ⚠️ `analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd` 
  - While I was here I also removed a ggplot argument `show.legend = FALSE` which apparently is not in our ggplot version, and it throws a warning during runtime and then does absolutely nothing in the plot. So, it's gone!
- Panels D-F: ⚠️ `analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd` 
  - Note that this notebook output gets sent to `tp53_nf1_score/results/tumor-purity-threshold/`
  - Panels E and F have the same data frame, so I `select`ed relevant columns for each export.
- Panel G: `figures/scripts/figS7-tp53-telomerase-tumor-purity-threshold.R` 
- Panel H: ⚠️ `analyses/transcriptomic-dimension-reduction/06-umap-tumor-purity-threshold.Rmd` 
- Panel I: ⚠️ `analyses/immune-deconv/03-quantiseq-at-threshold.Rmd` 

#### What to expect when you're expecting more PRs

Let's take a second to check in overall here: What's left?

- Panels for Figure S2 that need to be dealt with on AWS. This could get weird since data is _HUGE_, so we'll see if we can even put some of those files in github at all. 
- Panel S3A, which will be exported from from the `tumor-purity-exploration` module.
- Oncoprint CSV files
- Sample-level molecular alteration files associated with oncopints
- Finish the README


analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd
```diff
@@ -432,6 +459,18 @@ ggsave(
...
+We'll also export this plot's data:
+```{r}
+tp53_thresholded_df %>%
+  # clean up columns
+  dplyr::select(Kids_First_Biospecimen_ID, dplyr::everything(),
+                -cancer_group_hex, -tp53_score) %>%
+  # arrange on biospecimen
+  dplyr::arrange(Kids_First_Biospecimen_ID) %>%
+  readr::write_csv(figS7f_csv)
+```
```
jashapiro, line 471: With 7e and 7f including most of the same data (dropping a different column), maybe we can just include both in one file?

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
```diff
@@ -505,14 +512,12 @@ cg_plot +
     "segment",
     x = -2, y = 0.25, xend = -1, yend = 0.25,
     color = "red", lwd=1.5,
-    arrow = arrow(), 
-    show.legend = FALSE) + 
+    arrow = arrow()) +
```
jashapiro, line 515: Just wondering about this change. Is `show.legend=FALSE` not needed?

jashapiro (approved):
This all looks fine. I would probably combine the outputs for S7e&f, but I don't know that it is worth the effort here.

analyses/transcriptomic-dimension-reduction/04-explore-sequencing-center-effects.Rmd
sjspielman: Yep, turns out it's an argument that doesn't do anything but toss a warning at you that it's not a real argument in the `ggplot2` version we have - 

> While I was here I also removed a ggplot argument `show.legend = FALSE` which apparently is not in our ggplot version, and it throws a warning during runtime and then does absolutely nothing in the plot. So, it's gone!

analyses/tp53_nf1_score/10-tp53-tumor-purity-threshold.Rmd
sjspielman: I came very close to doing this! But I didn't because I guess I wanted my life to be harder ¯\_(ツ)_/¯

PR #1713, opened 2023-03-24
Zenodo CSVs for oncoprint plots

sjspielman:
Ever towards #1692 

This PR adds some CSVs for oncoprint plots, in [Figure 2](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/868c3bfd6fbcb6c806935398a3e929f7f181c12a/figures/pngs/figure2.png) and [S3b](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/868c3bfd6fbcb6c806935398a3e929f7f181c12a/figures/pngs/figureS3.png). 
There are 5 new CSVs total for each of the oncoprints.

I exported data that roughly matches what was described in https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1692#issuecomment-1479951510, except I made it in a "long" format for more interoperability.

> |         | FGFR1             | H3F3A  | BRAF |
> |---------|-------------------|--------|-------|
> | Sample1 |  | | KIAA1549--BRAF |       
> | Sample2 |                   | p.K28M |       |
> | Sample3 |       Amplification            | |       |
>

To accomplish this, I modified the function that creates the maf objects to also create an analogous data frame that can be exported. Note that I also at one point made sure to keep `broad_histology_display` inside of `histologies_df` so that column could make it into export. The figure 2 CSV files are also compressed because 1 of them was 20 MB; this is well below github's limit, but I wanted to keep overall size low here. If we want, I can just `gz` that particular file, or I can not `gz` anything.


figures/scripts/fig2_figS3-oncoprint-landscape.R
```diff
@@ -116,16 +125,49 @@ prep_histology_maf <- function(included_cancer_groups,
...
+  
+  # Create a data frame that can be used to export this information
+  join_cols <- c("Hugo_Symbol", "Tumor_Sample_Barcode", "Variant_Classification")
+  
+  maf_export_df <- histology_maf_df %>%
+    dplyr::left_join(
+      histology_fusion_df, 
+      by = c(join_cols, "Variant_Type")) %>%
+    dplyr::left_join(histology_cnv_df,
+                     by = join_cols) %>%
```
jharenza, line 145: I guess we want full join here in case samples have no SNVs but only fusions or CNVs etc?

jharenza:
Hi @sjspielman! Thanks for starting this. I think that these tables are huge for a few reasons: 
1) We want them to show only what is in the oncoprint, that is, the top 20 altered genes (from the goi list for the histology) and samples with those alterations.  We could add all samples, including ones not shown, and just have NA or blanks in those genes' spaces for those samples. You might have to apply the `get_histology_goi` function, or subset by the goi  for that oncoprint and [these](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/868c3bfd6fbcb6c806935398a3e929f7f181c12a/analyses/oncoprint-landscape/util/oncoplot-functions.R#L60-L78) variant classifications.
2) Looks like _all_ variant classifications are included here, but I am actually not seeing fusions or CNVs - we would want to subset to only the variant classifications which get displayed - eg there are a ton of genes with unknown symbol and which are IGR classification not being displayed. IGR shouldn't be in the output per 1 above.
3) There are duplicate rows. I think this is happening because there are multiple alterations (eg two missense) for one sample in the same gene. 

I don't think the clinical info is 100% necessary for this table, so the wide format can still be accomplished, but I suppose I can be convinced of long as well - long is back to something like a MAF format without all of the columns. I suggested a wide format oncomatrix since it is a format which can be input to pedcbio or complexheatmap for oncoprint generation. There are two ways to make the oncomatrix - one is to just use general terms (more suitable for oncoprint generation) and one is to write the exact alterations per gene (would need wrangling to be useful for own oncoprint generation). Based on the review, I think we should do the latter, so long or wide probably doesn't matter (ie exact protein change for SNV, Amp/Del for CNV, fusion partners for fusion, as in the comment above). I think reviewer wants to be able to query which samples have KIAA1549--BRAF fusions or which samples have BRAF V600E? 

One other thought is we can theoretically export all alterations in any histology-specific goi rather than top 20. I can go either way on this since we plan to export the entire alteration matrix (of the combined goi, or full MAF??, not sure we decided or discussed) for all samples as a separate table.

Sorry for the long explanation, but hopefully that makes sense.

jharenza:
Thanks @sjspielman! Totally fine with the path of least resistance- updating captions!

figures/scripts/fig2_figS3-oncoprint-landscape.R
```diff
@@ -369,6 +426,14 @@ for (type_iter in seq_along(data_input_list)) {
     create_legend(legend_df$cancer_group_display,
                   legend_df$cancer_group_hex,
                   legend_output_pdf)
+    
+    # Export CSV for Zenodo upload, after filtering to histology_goi genes
+    maf_export <- histology_maf_object_df$maf_df %>%
```
jashapiro, line 431: suggested naming above
```suggestion
    maf_export <- histology_maf_pair$maf_df %>%
```

figures/scripts/fig2_figS3-oncoprint-landscape.R
```diff
@@ -383,9 +448,10 @@ for (type_iter in seq_along(data_input_list)) {
         dplyr::pull(cancer_group)
 
       # Prep MAF object for plot
-      histology_maf_object <- prep_histology_maf(included_cancer_groups,
+      histology_maf_object_df <- prep_histology_maf(included_cancer_groups,
```
jashapiro, line 451:
```suggestion
      histology_maf_pair <- prep_histology_maf(included_cancer_groups,
```

jashapiro (dismissed):
I had only a small naming suggestion, as I got confused. But as I finished it I realized what you had meant by your variable name (object AND df!), so it is probably fine as it was

jashapiro (approved):
LGTM

sjspielman:
Code does not run through CI, so can merge before checks are complete.
Also noting that from separate discussion, we are going to keep the tables in long form for more interoperability with any downstream platform.

PR #1717, opened 2023-03-27
Zenodo CSVs: `snv-callers` panels

sjspielman:
Part of #1692 
This PR adds (ish!) CSVs for `snv-callers` panels in [Figure S2, A-G](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/868c3bfd6fbcb6c806935398a3e929f7f181c12a/figures/pngs/figureS2.png). The PBTA panels in particular are made from data in a big SQL db, so we have some things to discuss about how to include these files (see https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1692#issuecomment-1483186513).
Here's what I've done for this PR:

- Finally removed the `.gitkeep`!
- Modified the one script `figures/scripts/figS2-snv-callers-panels.R` to export **three** CSVs only:
  - 1 CSV for panels A,B,C which have the same PBTA originating data. This file is ~1GB so it is currently in a `.gitignore` file here. To maintain the file, I've put it for now into the `OpenPBTA` google drive folder.
  - 1 CSV for panels D,E,F which have the same TCGA originating data.
  - 1 CSV for panel G, just a violin plot.
 - While I was working on this script, I found another rube goldberg machine I wrote on lines 188-198 so I cleaned that up quite a bit, and confirmed plots are the same! This involved some minor changes to `detect_mat` defining so that I could use that variable again in place of the rube goldberg.

Part of review here should therefore be some discussion about handling the 1 GB file, since it's not getting added to github! Whatever we decide, I'll then update `README.md` here to explain the whole situation.


figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -109,9 +126,13 @@ for (dataset in c("tcga", "pbta")) {
...
+    all_caller_df %>%
+      dplyr::select(Tumor_Sample_Barcode, everything()) %>%
+      dplyr::arrange(Tumor_Sample_Barcode) %>%
+      readr::write_csv(figS2def_csv)
   }
 
-
   ## Upset plots -------------------------------------------------------
   print("Preparing data for upset plot")
   detect_mat <- all_caller_df %>%
```
jashapiro, line 138: This is no longer being used as (only prep for) a detection matrix, so rename to `vaf_mat`?

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -185,17 +203,19 @@ for (dataset in c("tcga", "pbta")) {
...
-    # 1:n() column for spread/gather bookkeeping which we do to remove `VAF_` from strips in this plot
-    mutate(n = 1:n()) %>%
-    # we want to remove
-    gather(caller, vaf, -n) %>%
-    mutate(caller = str_replace_all(caller, "VAF_", "")) %>%
-    spread(caller, vaf) %>%
-    select(-n) %>%
-    GGally::ggpairs(aes(alpha = 0.05)) +
+  # Correlate VAFs across callers
+  detect_mat_df <- as.data.frame(detect_mat) %>%
```
jashapiro, line 207: I'm thinking maybe rename this for clarity?
```suggestion
  vaf_mat_df <- as.data.frame(vaf_mat) %>%
```

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -123,22 +144,21 @@ for (dataset in c("tcga", "pbta")) {
   dimnames(detect_mat)[[1]] <- all_caller_df$index
 
   # Turn into detected or not
-  detect_mat <- !is.na(detect_mat)
-
+  detect_mat_logical <- !is.na(detect_mat)
```
jashapiro, line 147: If you follow my renaming suggestion
```suggestion
  detect_mat_logical <- !is.na(vaf_mat)
```

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -98,6 +109,12 @@ for (dataset in c("tcga", "pbta")) {
       as.data.frame() %>%
       rowid_to_column("index")
 
+    # Export
+    all_caller_df %>%
+      dplyr::select(Kids_First_Biospecimen_ID = Tumor_Sample_Barcode, everything()) %>%
+      dplyr::arrange(Kids_First_Biospecimen_ID) %>%
+      readr::write_csv(figS2abc_csv)
```
jashapiro, line 116: Do we need `everything()` here? Never mind... keeping the position & variant info seems worthwhile to me. I would drop `index` though, as noted.

jashapiro (commented):
This looks mostly fine.

I got confused by some of your refactoring for a bit, as a variable name that was before used temporarily to hold something other than its name (`detect_mat`) got kept around holding that other thing. So I suggested a renaming.

As for where to keep the file: I think the google drive is fine. If we want to be fancy we could keep a hash for the file, but I don't think we need to. I think we could also move `figure-S2d-S2e-S2f-data.csv.gz` out of the repo, but it doesn't matter too much.

I also suggested dropping the `index` column that comes out of the database. Pretty sure we don't need that for anything; certainly not for the plots.

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -135,30 +131,27 @@ for (dataset in c("tcga", "pbta")) {
...
   print("Preparing data for upset plot")
-  detect_mat <- all_caller_df %>%
-    # Bring over VAF columns and the index
+  vaf_mat <- all_caller_df %>%
+    # Bring over VAF columns
     select(starts_with("VAF_")) %>%
     as.matrix()
 
-  # Store the indices as dimnames
-  dimnames(detect_mat)[[1]] <- all_caller_df$index
```
jashapiro, line 123: I think I might have left this, though I can't quite remember why we wanted it.

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -135,30 +131,27 @@ for (dataset in c("tcga", "pbta")) {
...
+  # Turn into logical matrix
```
jashapiro, line 139: A logical matrix of what? I actually liked the previous name, because the VAF data is gone at this point.
```suggestion
  # Turn into logical matrix of detected or not (NA means the mutation was not detected)
```

jashapiro (commented):
This looks fine, but I think I want to revert some variable name & amend some comment changes.

figures/scripts/figS2-snv-callers-panels.R
sjspielman: Well, the `index` column is gone now, so this line is a bug unless I go back and revert removing the `index` column..

jashapiro: Right, but I figured you would catch that!

figures/scripts/figS2-snv-callers-panels.R
```diff
@@ -185,17 +204,19 @@ for (dataset in c("tcga", "pbta")) {
...
-    # we want to remove
-    gather(caller, vaf, -n) %>%
-    mutate(caller = str_replace_all(caller, "VAF_", "")) %>%
-    spread(caller, vaf) %>%
-    select(-n) %>%
-    GGally::ggpairs(aes(alpha = 0.05)) +
+  # Correlate VAFs across callers
+  vaf_mat_df <- as.data.frame(detect_mat) %>%
+    # reorder columns; everything() will cover vardict which is only in PBTA
+    select(VAF_lancet, VAF_mutect, VAF_strelka, everything())
```
sjspielman, line 210: Just want to note that this was the unneeded pivoting that I purged, and it had nothing to do with `index`.

sjspielman:
@jashapiro thanks for your patience here, and sorry for misunderstandings about `index`! I believe I've gotten it back to a good place, and it's running as expected.

figures/scripts/figS2-snv-callers-panels.R
sjspielman: ^ this says is outdated now, but it's not in terms of the spirit of my "fyi comment"

jashapiro (approved):
LGTM (Did the `figure-S2d-S2e-S2f-data.csv.gz` file come back? I thought it was gone for a while, but maybe not)

sjspielman:
>  I thought it was gone for a while, but maybe not

I don't think it was, but maybe so! In any case it's here now and small enough to be ok, and fewer moving parts for this csv situation is better as there's enough going on already..

sjspielman:
This code does not run through CI, so can merge before checks complete.

Issue #1718, opened 2023-03-27
Zenodo CSVs: Update to include sample information


sjspielman:
Part of #1692 

Some of the exported files towards #1692 probably need to be revisited since they do not have sample ids, but _could_. This issue has some inventory for CSVs without this information and for which ones it can reasonably be added in: first, for CSVs that cannot reasonably accommodate this data without fundamental changes, and then for CSVs that can accommodate this data. Feedback much appreciated here!

#### Cannot accommodate
- [x] ROC curves in 4A, S5A, S7D. Just not applicable
- [x] Co-occurrence plot in 3A. The data seems way too derived to nail back down to sample level, but tagging @jashapiro for thoughts.

#### Could accommodate with some changes
- [x] Forest plot results from survival models in 4F, 4G, 5D. The models themselves as CSVs cannot take sample ids, but we could change these files to contain samples that went into the model along with their value for variables in the model. Either way some information gets sacrificed, so we should pick which.

#### Needs updating
- [x] S7A barplot: This counts the stranded/polyA across diagnoses. Since the _purpose_ of this plot is to show counts, I felt like counts CSV made more sense, but "uncounting" is matter of removing the `dplyr::count()` that I added! Highly achievable.
- [x] 3B: mutant samples. Rather than counting the number of samples, make this longer form with sample ids.
- [x] 3D: this plot summarizes fractions of cancer groups with chromothripsis events. The data can be changed to have `TRUE`/`FALSE` for whether each sample has an event
- [x] 4H: this is a KM curve. Data can be joined in here I'm sure.

sjspielman:
🚀

PR #1719, opened 2023-03-27
Fig S7A CSV redux

sjspielman:
Part of #1718 - we need sample IDs everywhere we can, apologies for this oversight!!
This PR fixes the stranded/polyA barplot in S7A and arranges on the retained sample ids.


jashapiro (approved):
LGTM!

PR #1720, opened 2023-03-27
Fig 3D csv redux

sjspielman:
Part of #1718 
Again, this CSV needs to include sample ids! I've redone the export with the non-summarized data.


jashapiro (approved):
LGTM

PR #1721, opened 2023-03-27
Fig 3B redux

sjspielman:
Part of #1718 

This PR updates the 3B Zenodo export so that it can also include sample ids, and be overall in a less summarized form.
This means the export has to happen from an earlier script in the interaction plots module.

While I was here, I thought about this module's zenodo exports overall. I wondered if b/c this module is written so flexibly, we might just lock this export down a bit with a flag. I added that to the R scripts and shell script that calls them. If this was overkill and reviewers prefer that I revert, no problem!


analyses/interaction-plots/scripts/02-process_mutations.R
```diff
@@ -158,6 +160,12 @@ option_list <- list(
     default = 0,
     help = "Minimum sequencing depth for called mutations",
     metavar = "numeric"
+  ),
+  make_option(
+    opt_str = "--write_zenodo_csv",
+    action = "store_true",
+    default = FALSE,
+    help = "When this flag is used, output tabular data associated with manuscript figures targeted for Zenodo upload to file."
```
jashapiro, line 168: if I were a stickler, I would want this to include a check that we were really using all data here, but I think it is fine.

jashapiro (approved):
LGTM. I was at first confused by the removal of disease types, but since that doesn't show up in figure 3b, I think that is fine.

analyses/interaction-plots/scripts/02-process_mutations.R
sjspielman: 🎼

PR #1722, opened 2023-03-28
Zenodo CSVs: Fix forest plots

sjspielman:
Part of #1718 

This PR remakes 3 csv exports for forest plots so that the model's underlying data, not the model itself, is what is exported. The figures are - 

- [4F](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pdfs/fig4/panels/forest_survival_tp53_telomerase_panel.pdf) in `figures/scripts/fig4-tp53-telomerase-panels.R`. This panel was derived from the `tp53_telomerase` survival module notebook. 
- [4G](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pdfs/fig4/panels/forest_hgg_subtypes.pdf) in `figures/scripts/fig4-hgg-subtype-forest-plot.R`. This panel was derived from the `subtypes` survival module notebook. 
- [5D](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/pdfs/fig5/panels/forest_MB_quantiseq_CD274.pdf) in `figures/scripts/fig5-forest-plot.R`. This panel was derived from the `immune` survival module notebook.  

The exports themselves are fairly simple, _but_ I had to update the survival module to be able to get the data written out for use in these figure scripts. That update was small, just a quick change to the function in `analyses/survival-analysis/util/survival_models.R`. Rather than only exporting the model object (which does not appear to still hold the input data) in RDS files, I updated the code to export a list of the model _and_ the data that went into the model. This does not affect any survival notebooks beyond the need to re-run them, and the only scripts that consume survival model output are figure scripts (**edit** - and MS tables, so I also updated reading in models accordingly in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1722/commits/0c8d64ae91dbf7f59ce4e6829911b000c65940e1). So, now we can read in data for export along with the model for plotting.

I therefore had to re-run the survival module, so all the `RDS` result files have changed _as expected_ - they are slightly larger now since they contain the data too. Hence, the diffs.

I also ensured that the exported CSVs have `LIVING` and `DECEASED` statuses instead of 1/0 (this was already the case for one of the data frames, but had to be updated for the other two).


analyses/survival-analysis/util/survival_models.R
```diff
@@ -264,10 +264,13 @@ fit_save_model <- function(df,
...
   }
   
-  # Save model RDS
-  readr::write_rds(fit$model, 
-                   # remove spaces in case (common for cancer groups)
-                   str_replace_all(output_file, " ", "-"))
+  # Save list of model and originating data to RDS
+  readr::write_rds(
+    list(model = fit$model, 
+         data  = df),
```
jashapiro, line 270: There's a funny bit here where `fit$original_data` already includes the data frame if the `survival_analysis()` function was used, but not for the multivariate case. But this should work just fine.

jashapiro (approved):
LGTM. I noted a place where the original design might have already accommodated most of what we need, but there is no reason to go deep in it at this point when what you have works fine.

analyses/survival-analysis/util/survival_models.R
sjspielman: Yeah, this survival code has been around the block and then some...............

sjspielman:
Noting this PR needs to run through CI.

PR #1723, opened 2023-03-29
Zenodo CSV: Fix 4H export

sjspielman:
Part of #1718 
This PR fixes data export for the KM curve in panel 4H so that it includes the data itself. Very excitingly, this table was already in the model object (note this is a _different object structure_ from what you see in cox regressions in #1722) in `$original_data`, so this is very straightforward!


jashapiro (approved):
LGTM!

sjspielman:
Code does not run through CI, so merging now!

PR #1725, opened 2023-03-29
Molecular alteration CSV for Zenodo upload

sjspielman:
Part of #1692 

This PR _begins_ (code needs help/feedback during review!) the process of getting a molecular alterations csv. At this point, I decided to file for feedback since my brain keeps getting twisted thinking of the right way to join back with DNA and RNA ids for the ambiguous samples.

The new code involves:
- A script `tables/tabulate-molecular-alterations.R` to generate this CSV file
- A script `tables/util/functions-molecular-alterations.R` with some extremely context-specific functions, as indicated in comments in that file.

The `tables/tabulate-molecular-alterations.R` script takes the following steps until we get to:
1. Process maf, cnv, fusion data
    - mirroring the approach in https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/627ec427ad0a8d9d913e614c9db50546c56d8283/analyses/oncoprint-landscape/01-map-to-sample_id.R.
2. Combine those processed df's into a big alteration df
3. Filter to genes present in the manuscript oncoplots, and create a presence/absence (with "Y"/"N") data frame for whether each gene/alteration is present in each of the 1074 tumors.
4. Map back to biospecimen ids for export 🥴  

Step 4 is either fairly straightforward, or very not! 
- I have created `export_identifiable_ids` which is an export-ready wide data frame for _non-ambiugous_ ids that have 1:1 DNA:RNA. 
- But, I still need a solid strategy for dealing with the non 1:1 mappings. There's two types of problematic mappings:
  - Where there are >1 RNA or DNA experiments, so we can't directly line up a DNA with RNA. This is a 1:many situation, like we discussed.
  - Where there are >1 of a given modality - eg 2 RNAs, or even a case with 6 DNAs!. This is, well, just a plain "many."

One potential approach that I was thinking of is to just add these `sample_id`s and alterations into the main export df (called `export_identifiable_ids` at the moment), _but_ keep their biospecimen id values as `NA` . Then, we could have a separate file to indicate what those ids might be, with columns: 
- `sample_id`
- `Kids_First_Biospecimen_ID`
- `experimental_strategy`
I don't think this approach will make a downstream user's life easier, but I also think they're (we're?) just in for a hard time anyways with these samples...


sjspielman:
Requested @jaclyn-taroni for review here mostly for purposes of strategizing ambiguous ids as needed!

sjspielman:
This closes #1726.

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,264 @@
...
+)))
+
+
+
+#### Prepare maf, cnv, fusion data ------------------
+
+maf_df <- prepare_maf(maf_df)
+fusion_df <- prepare_fusion(fusion_df)
+cnv_df <- prepare_cnv(cnv_df)
+
```
jashapiro, line 167: Each of these functions is used only once, so I'm not sure the value of having them separated as functions, especially in a separate file. At the least, I would like to see here something about what this preparation is: what are the changes that are being made by these functions?

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,264 @@
...
+}
+
+# Convert to presence/absence of _all combinations_ of genes and variants for each sample_id
+alterations_presence_df <- alterations_df %>%
+  # If the row is here already, then the alteration is present
+  dplyr::mutate(present = "Y") %>% 
+  # Fill in all combinatins of gene/variant for each sample, 
+  #  filling in "N" (no) for present for any new rows 
+  tidyr::complete(sample_id, Hugo_Symbol, Variant_Classification, 
+                  fill = list(present = "N")) %>%
```
jashapiro, line 217: Any reason not to use T/F here?
```suggestion
  dplyr::mutate(present = TRUE) %>% 
  # Fill in all combinatins of gene/variant for each sample, 
  #  filling in FALSE for present for any new rows 
  tidyr::complete(sample_id, Hugo_Symbol, Variant_Classification, 
                  fill = list(present = FALSE)) %>%
```

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,264 @@
...
+# - Second, the ambiguous samples: where DNA and RNA ids cannot be reliably linked, OR 
+#   where there are multiple same-modality experiments (since we are working with all samples, not `primary-only`)
+
+## Combine alterations_presence_df with biospecimen sample ids ----------------------
+  
+## First the fully non-ambiguous samples:
+export_identifiable_ids <- tumor_sample_ids_df %>%
+  # only ids of interest
+  dplyr::filter(!(ambiguous), 
+                !(multiple_exp_strategy)) %>%
```
jashapiro, line 233: If we don't filter here, we will end up with multiple rows for each `sample_id` where there may be more than one DNA or RNA sample after joining. If we then pass this through a `group_by()`, we should be able to collapse DNA & RNA columns with a `paste` or similar, keeping everything else the same. 

Does that make sense as a strategy?

That at least should work for multiple modalities... I'm not sure what the ambiguous samples really look like.

jashapiro (commented):
Some initial comments here, the last of which is a potential strategy comment.

I'm confused by the difference between ambiguous and multi-mapping. From your description they both sound like multimapping, just of degree? Either way, I think making the DNA/RNA columns be semicolon-separated lists of ids is the most straightforward, and accomplishing that with an appropriate group_by/summarize seems the simplest approach. But I could be missing something major here!

tables/tabulate-molecular-alterations.R
sjspielman: The value of having them in separate functions is mostly for my brain, less clutter during coding. All of that code was all originally in this script, but as the script got bigger and bigger....

Can definitely de-function them in the end, and I'll add some more comments here.

sjspielman:
Hm I'd just like to note, for the record, that this branch name, which I personally named, is....not how my name is spelled.

sjspielman:
>I'm confused by the difference between ambiguous and multi-mapping. From your description they both sound like multimapping, just of degree? 

Yes, you've got it! Both are ambiguous situations, but I should have been clear to say that one flavor is ambiguity for mapping RNA-DNA (1:many situation), and the other flavor is, for example, which DNA id (if there are several) is being referred to. The latter is not necessary 1:many.

> Either way, I think making the DNA/RNA columns be semicolon-separated lists of ids is the most straightforward, and accomplishing that with an appropriate group_by/summarize seems the simplest approach. But I could be missing something major here!

I don't disagree at all, but I have in my notes from yesterday's chat about this that we decided to avoid that strategy and instead have a separate column to indicate if the given _single pair_ of RNA/DNA ids is a 1:1 or 1:many situation. 
Maybe I'll go for a `why-not-both.gif` situation - have semi-colon separated ids _and_ a T/F column to indicate "ambiguous mapping".

sjspielman:
> Maybe I'll go for a `why-not-both.gif` situation - have semi-colon separated ids and a T/F column to indicate "ambiguous mapping".

As a move towards next steps, I went ahead and did this in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/commits/42d7ee84feb110fdced906871471be3d57a50f7a. The output CSV is manageable at 8MB without compression, but can compress if we want a tiny file!

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,433 @@
...
+# First, let's "combine" the ambiguous ids: 
+# For multiple DNA or RNA for a given sample_id, convert to a semi-colon separated single value
+# For example, two DNA ids on separate rows named BS_1 and BS_2 would be collapsed into `BS_1; BS2`
+merged_ambiguous_ids <- tumor_sample_ids_df %>%
+  # get only samples of interest
+  dplyr::filter(ambiguous | multiple_exp_strategy) %>%
+  # for each sample_id, combine biospecimen ids by modality to have 
+  # at most one "value" (semi-colon separated list) for each modality 
+  dplyr::group_by(sample_id, modality) %>%
+  dplyr::summarize(Kids_First_Biospecimen_ID = paste(Kids_First_Biospecimen_ID, collapse="; ")) %>%
```
jashapiro, line 369: I'd leave out the space: easier to unparse later.

Also, just in case, adding unique and sort here: 
```suggestion
  dplyr::summarize(
    Kids_First_Biospecimen_ID = paste(
      sort(unique(Kids_First_Biospecimen_ID)),
      collapse=";")
  ) %>%
```

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,433 @@
...
+  dplyr::ungroup() %>%
+  # bring back the other variables from `tumr_sample_ids_df`
```
jashapiro, line 371:
```suggestion
  # bring back the other variables from `tumor_sample_ids_df`
```

tables/tabulate-molecular-alterations.R
jashapiro: Further thought: if you don't filter to just ambiguous, you can put that indicator here as well, and save having to have two flows, I think. 

```suggestion
    dplyr::summarize(
    Kids_First_Biospecimen_ID = paste(
      sort(unique(Kids_First_Biospecimen_ID)),
      collapse=";"),
    ambiguous_biospecimen_mapping = dplyr::n()>1
  ) %>%
```

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,433 @@
...
+  dplyr::select(-ambiguous, -multiple_exp_strategy) %>%
+  # add indicator that these are NOT ambiguous mappings
+  dplyr::mutate(ambiguous_biospecimen_mapping = FALSE) %>%
+  # combine with prepared ambiguous samples
+  dplyr::bind_rows(merged_ambiguous_ids) %>%
+  dplyr::distinct()
+
+# Again, check that we've got everything: there should be `openpbta_n_tumors` 
+#  unique sample_id values
+if (!(length(unique(tumor_sample_ids_df$sample_id))) == openpbta_n_tumors) {
```
jashapiro, line 400:
```suggestion
if (length(unique(tumor_sample_ids_df$sample_id)) != openpbta_n_tumors) {
```

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,433 @@
...
+  dplyr::inner_join(
+    alterations_presence_df,
+    by = "sample_id"
+  ) %>%
+  # update columns and their order
+  dplyr::select(
+    sample_id, 
+    Kids_First_Biospecimen_ID_DNA = DNA,
+    Kids_First_Biospecimen_ID_RNA = RNA,
+    ambiguous_biospecimen_mapping,
```
jashapiro, line 417: There may be two values here, right? One each for DNA & RNA? do we need to account for that in the spread?

jashapiro (commented):
I think we are getting there. I had a suggestion to simplify a bit more, but I also had a question about the final output for the ambiguous indicator... I feel like it may be easiest to just leave that to the very end and add it if there is a semicolon in DNA or RNA columns, rather than trying to maintain it by other means through the transformations.

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,438 @@
...
+# We'll need to identify those samples, including their sample_id and biospecimen.
+
+# In addition, we'll note if a sample_id is ambiguous. Ambiguous sample_id's will 
+# have more than 2 rows associated with it in the histologies file when looking at 
+# tumor samples -- that means we won't necessarily be able to determine **when an 
+# WGS/WXS assay maps to an RNA-seq assay** for the purpose of tabulating molecular alterations
+
+tumor_sample_ids_df <- histologies_df %>%
+  dplyr::filter(sample_type == "Tumor",
+                composition == "Solid Tissue") %>%
```
jaclyn-taroni, line 237: We want to include cell lines, too, per the response to reviewers and https://github.com/AlexsLemonade/OpenPBTA-analysis/issues/1692#issue-1626351515 (adding emphasis here):

> A CSV file that contains a table of molecular alterations included in Figure 2 for all tumor **and cell line samples** in the PBTA dataset

jaclyn-taroni: I expect this also impacts your use of `openpbta_n_tumors` throughout. You could add `openpbta_n_cell_line` and then have the check be for the length of `openpbta_n_tumors + openpbta_n_cell_line`.

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,438 @@
...
+
+# Filter dfs to those ids:
+histologies_df <-  dplyr::filter(histologies_df, Kids_First_Biospecimen_ID %in% bs_ids) 
+maf_df         <-  dplyr::filter(maf_df, Tumor_Sample_Barcode %in% bs_ids) 
+fusion_df      <-  dplyr::filter(fusion_df, Sample %in% bs_ids) 
+cnv_df         <-  dplyr::filter(cnv_df, biospecimen_id %in% bs_ids) 
+
+# What are all the bs ids that end up being considered? 
+# This variable can be used when joining things back up
+final_bs_ids <- length(unique(
```
jaclyn-taroni, line 284: I would probably name the variable such that it's clear this is a _count_.

sjspielman:
@jashapiro while working to continue cleaning this process up, I came across this neat situation which I did not anticipate, but glad I caught it:
Screenshot 2023-03-29 at 3 37 22 PM

The two RNA samples have "different" sexes - `Male` and `NA`. So, if we have these ids on the same row, then we have a conflict with this metadata variable. On one hand, we could defer to the non-`NA` value here and proceed with `Male`, but also I am wary about changing metadata here. But back to the first hand, we could document here that we use the non-`NA` sex if this conflict occurs.

And on a third hand, it seems that there are only 2 sample id's like this! For such a limited scenario, maybe just using the non-`NA` and explaining in `README` is indeed the way to go.
```
tumor_sample_ids_df %>% 
  dplyr::select(sample_id, germline_sex_estimate) %>% 
  dplyr::distinct() %>%
  dplyr::count(sample_id) %>%
  dplyr::filter(n>1)
# A tibble: 2 x 2
  sample_id     n
  <chr>     <int>
1 7316-161      2
2 7316-85       2
```

jaclyn-taroni (commented):
Returning some initial comments. More importantly, **I was wrong yesterday!** I’ve re-reviewed the reviewer’s comments, and I don’t think this should be the presence or absence of alteration types (e.g., “Intron”) in the oncoprint. Instead, we’re being asked to include the _specific alterations_ in a table like so (this is a made-up example):

|  ids ... `histologies` data columns | TP53 |  BRAF |
|---|--|--|
| … |  p.A159P | None  |
| … | None |  KIAA1549--BRAF  |

The genes as columns will be the genes you have in `gene_list`. This might require a lot less wrangling for the fusions, for example (I very well might be wrong about that). 

In the case of _multiple alterations_, we can semi-colon separate them.

sjspielman:
> For 1:1 mappings between assays, I would expect one row per `sample_id`.

Going back to update code accordingly to have 1 row where possible, one item occurred to me as a _potential_ reason to maybe not do that that I'd like to hear thoughts on! The review itself asks for much more metadata to be part of this export, including whether certain experimental modalities are available (column for `has_RNAseq`, `has_WGS`, etc.). These kinds of columns will end up doubling down on some ambiguity - e.g. sure WGS might be available, but which biospecimen is it? Having 1 row for each biospecimen locks down this particular metadata item, and I think it gives more clarity as to which biospecimen is each type of experiment. 

Either way, I'm working now towards 1-row-per either way, and this is something to think about!

jaclyn-taroni:
@jashapiro I'm making edits to this and will tag you when it is ready for a look on

jaclyn-taroni:
Okay, we now have a CSV file with the following columns:

| `sample_id` | `Kids_First_Biospecimen_ID_DNA` | `Kids_First_Biospecimen_ID_RNA` | `multiple_assays_within_type` | `sample_type` | `composition` | `germline_sex_estimate` | `broad_histology` | `cancer_group` | `ACVR1` | ... | `ZIC1` |
|-----------|---|---|---|---|---|---|---|---|---|---|---|

Where: 

* (Almost) each row is a unique `sample_id`-`composition` pair – more on the "almost" in a second.
* `Kids_First_Biospecimen_ID_DNA` are all DNA assay biospecimen IDs associated with the `sample_id`-`composition` pair in that row, separated by `; `.
* * `Kids_First_Biospecimen_ID_RNA` are all RNA assay biospecimen IDs associated with the `sample_id`-`composition` pair in that row, separated by `; `.
* `multiple_assays_within_type` indicates whether there is more than one DNA or more than one RNA assay for that `sample_id`-`composition` pair.
* A gene column will have all alterations found in at least one biospecimen for each assay type (RNA and DNA). For multiple alterations, they are separated by `; `. If no alterations were detected given the available assays, it says "None."

You can see how I dealt with the issue mentioned in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725#issuecomment-1489204244 here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/files#diff-7eed238487c7c49c705f8ab849980cf0183d136fd84054fc833cdbe656cc14e7R365-R375

One `sample_id`-`composition` pair has multiple cancer groups / molecular subtype, resulting in 1075 rows:

```r
   > histologies_df %>% filter(sample_id == "7316-3230") %>% select(sample_id, Kids_First_Biospecimen_ID, composition, broad_histology,cancer_group, molecular_subtype)
# A tibble: 4 × 6
  sample_id Kids_First_Biospecimen_ID composition  broad_histology                               cancer_group                  molecular_subtype          
  <chr>     <chr>                     <chr>        <chr>                                         <chr>                         <chr>                      
1 7316-3230 BS_5968GBGT               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss     
2 7316-3230 BS_AF5D41PD               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss     
3 7316-3230 BS_EE73VE7V               Solid Tissue Diffuse astrocytic and oligodendroglial tumor High-grade glioma astrocytoma HGG, H3 wildtype, TP53 loss
4 7316-3230 BS_HYKV2TH9               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss   
```


The instance with `HGG, H3 wildtype, TP53 loss` reports a different primary site. Not really sure what to do about this issue here. cc: @jharenza

jaclyn-taroni:
@jashapiro this is ready for a look. There an issue with one sample I noted here: https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725#issuecomment-1492444139

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,392 @@
...
+
+## Alterations from DNA and RNA ------------------------------------------------
+
+alterations_df <- dna_alterations_df %>%
+  # Full join of DNA and RNA alterations by sample ID and gene symbol
+  dplyr::full_join(rna_alterations_df,
+                   by = c("sample_id",
+                          "Hugo_Symbol",
+                          "composition"),
+                   suffix = c("_DNA", "_RNA")) %>%
```
sjspielman, line 245: 👀 this is cute, new to me

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,392 @@
...
+  # Combine DNA and RNA alterations into one column where applicable
+  dplyr::mutate(final_alteration = dplyr::case_when(
+    !is.na(final_alteration_DNA) & 
+      !is.na(final_alteration_RNA) ~ paste(final_alteration_DNA,
+                                           final_alteration_RNA,
+                                           sep = "; "),
+    is.na(final_alteration_DNA) ~ final_alteration_RNA,
+    is.na(final_alteration_RNA) ~ final_alteration_DNA
+  )) %>%
```
sjspielman, line 254: I confirmed there are no `NA`s at this point in `final_alteration` 👍

sjspielman (commented):
Thanks for tackling a lot of this while I was out!! I'm noticing some ids that may not have made it in the end, but still working out what's up. Pasting in here some preliminary investigating - 

```
# All the non-Normal ids
expected_sample_ids <- histologies_df %>% 
  dplyr::filter(sample_type != "Normal") %>% 
  dplyr::pull(sample_id) %>% 
  unique()
length(expected_sample_ids)
# [1] 1080

# What actually made it in the end?
final_sample_ids <- unique(final_alterations_df$sample_id)
length(final_sample_ids)
# [1] 1064

# Hm, those lengths are different. What's missing?
expected_sample_ids[!(expected_sample_ids %in% final_sample_ids)]
#  [1] "A14709" "A19649" "A16538" "A15170" "A17997" "A16405" "A19015"
#  [8] "A18777" "A15726" "A20091" "A16915" "A14565" "A14567" "A18552"
# [15] "A17096" "A12911"

histologies_df %>%
  dplyr::filter(sample_id %in% expected_sample_ids) %>%
  dplyr::select(sample_id, experimental_strategy, sample_type, composition) %>%
  head()
# #A tibble: 6 x 4
# sample_id experimental_strategy sample_type composition           
# <chr>     <chr>                 <chr>       <chr>                 
#  1 7316-2558 WGS                   Tumor       Solid Tissue          
#  2 7316-2753 WGS                   Normal      Peripheral Whole Blood
#  3 7316-2660 WGS                   Tumor       Solid Tissue          
#  4 7316-897  WGS                   Normal      Peripheral Whole Blood
#  5 7316-362  RNA-Seq               Tumor       Solid Tissue          
#  6 7316-3299 WGS                   Tumor       Solid Tissue 

# Hm, why/how is there a Normal coming up there?!
# ...because it's both...!
histologies_df %>%
  dplyr::filter(sample_id == "7316-2753") %>%
  dplyr::select(sample_id, experimental_strategy, sample_type, composition)
# A tibble: 3 x 4
#  sample_id experimental_strategy sample_type composition           
#   <chr>     <chr>                 <chr>       <chr>                 
# 1 7316-2753 WGS                   Normal      Peripheral Whole Blood
# 2 7316-2753 WGS                   Tumor       Solid Tissue          
# 3 7316-2753 RNA-Seq               Tumor       Solid Tissue    
```

jaclyn-taroni:
Shooting from the hip – I suspect the missing sample IDs have something to do with the copy number alteration data because I believe all of those samples are assayed with WXS. We only have CNV data for WGS samples, if I recall correctly.

jaclyn-taroni:
Now having looked, I'm reasonably sure it's the handling of `NA` germline sex estimate for WXS samples.

jaclyn-taroni:
I made some changes in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/commits/d13c5086ff6a8e85e0e6b8c82fa0f69e75f79466. Please take another look @sjspielman!

jharenza:
>One sample_id-composition pair has multiple cancer groups / molecular subtype, resulting in 1075 rows:

   > histologies_df %>% filter(sample_id == "7316-3230") %>% select(sample_id, Kids_First_Biospecimen_ID, composition, broad_histology,cancer_group, molecular_subtype)
# A tibble: 4 × 6
  sample_id Kids_First_Biospecimen_ID composition  broad_histology                               cancer_group                  molecular_subtype          
  <chr>     <chr>                     <chr>        <chr>                                         <chr>                         <chr>                      
1 7316-3230 BS_5968GBGT               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss     
2 7316-3230 BS_AF5D41PD               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss     
3 7316-3230 BS_EE73VE7V               Solid Tissue Diffuse astrocytic and oligodendroglial tumor High-grade glioma astrocytoma HGG, H3 wildtype, TP53 loss
4 7316-3230 BS_HYKV2TH9               Solid Tissue Diffuse astrocytic and oligodendroglial tumor Diffuse midline glioma        DMG, H3 K28, TP53 loss   
The instance with HGG, H3 wildtype, TP53 loss reports a different primary site. Not really sure what to do about this issue here. cc: @jharenza

This was the one sample in which VarDict was the only algorithm which called H3K28M - it was definitely real, and prompted us to use clinical reports of H3 status (below), but this sample did not have a post-mortem clinical report and thus, it was decided we would not change this one to match the other samples taken at the same time (_I think_). I think it would be fine to merge all of the DNA for this sample into one field and use the combination of molecular alterations (eg if one sample has ABC and one has BCD, then the resulting alterations would be ABCD).



https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/cf6328ebd7c3bdffa18b4bd6b8d856713a8414aa/analyses/molecular-subtyping-pathology/03-incorporate-clinical-feedback.Rmd#L35-L36

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,396 @@
...
+    readr::read_tsv(cnv_xy_file)
+  )
+
+## Prepare the maf data --------------------------------------------------------
+
+# Only keep maf columns relevant for annotating specific alterations
+maf_keep_cols <-c("Tumor_Sample_Barcode",
+                  "Hugo_Symbol",
+                  "Variant_Type",
+                  "HGVSp",
```
jharenza, line 82: Can we use the `HGVSp_Short` here?

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,396 @@
...
+  dplyr::bind_rows(hotspot_df) %>%
+  # filter to only relevant genes and ids
+  dplyr::filter(Hugo_Symbol %in% gene_list) %>%
+  dplyr::distinct() %>%
+  dplyr::mutate(
+    # assign the final alteration based on which piece
+    # of information is known, in this order of priority
+    final_alteration = dplyr::case_when(
+      HGVSp != "." ~ HGVSp,
+      HGVSc != "." ~ HGVSc,
```
jashapiro, line 103: Agree that short form makes sense here, but we probably also want to use the nucleotide change if there is no change in the protein, which I am doing here by grabbing the last character, which will be `=` if there is no change. 
```suggestion
      HGVSp_Short != "."  & 
        stringr::str_sub(HGVSp_Short, -1) != "=" ~ HGVSp_Short,
      HGVSc != "." ~ HGVSc,
```

jashapiro (commented):
This looks good to me. I initially had the one comment about whether we want `; ` or `;` as the separator (I favor the latter), but then I did a little MAF diving and found something else that may be worth changing.

tables/tabulate-molecular-alterations.R
```diff
@@ -0,0 +1,392 @@
...
+               "KMT2C", "KMT2D", "KRAS", "KSR2", "MAML2", "MAMLD1", 
+               "MET", "MIR512-2", "MYB", "MYCN", "NF1", "NF2", 
+               "NTRK2", "NTRK3", "PDGFRA", "PIK3CA", "PIK3R1", 
+               "PPM1D", "PRKAR1A", "PTCH1", "PTEN", "PTPN11", 
+               "QKI", "RAF1", "RELA", "ROS1", "SETD2", "SMARCA4", 
+               "TACC1", "TCF4", "TERT", "TP53", "YAP1", "ZIC1")
+
+# Later we'll need to use a list to fill in NAs with tidyr::replace_na()
+gene_fill_list <- list()
+for (gene in gene_list) gene_fill_list[[gene]] <-  "None"
```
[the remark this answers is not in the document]
jaclyn-taroni: I find the `for` loop easier to read in this case (not always the case!)

sjspielman:
Here is something we need to solidify - Pictured are the rows whose given `sample_id` appears in more than one row in the `final_alteration_df`. A single sample can be in multiple rows if there is a different composition associated, which seems fair to me. But, have a look at the `multiple_assays_within_type` column, and let's consider the two cases:
- `7316-1746`: Has `TRUE` in this column for _both_ associated rows, even though one of the compositions does not have more than 1 assay
- `7316-1769`: Has `FALSE` in this column for both associated rows. 

So, how do we want to consider if there is >1 assay? Is this at the level of `sample_id`, or _joint_ `sample_id` and `composition`. 


Screenshot 2023-04-03 at 3 22 21 PM

sjspielman:
Also we need to have a specific look at one of those samples `"7316-3230"` which has, _excitingly_, different cancer groups, and therefore ends up as an "incorrectly duplicated" row.
Screenshot 2023-04-03 at 3 35 56 PM

From `histologies`:
Screenshot 2023-04-03 at 3 37 03 PM


Not really sure what to make of this one in general...

jaclyn-taroni:
> Also we need to have a specific look at one of those samples `"7316-3230"` which has, _excitingly_, different cancer groups, and therefore ends up as an "incorrectly duplicated" row. Screenshot 2023-04-03 at 3 35 56 PM
> 
> From `histologies`: Screenshot 2023-04-03 at 3 37 03 PM
> 
> Not really sure what to make of this one in general...

Are you seeing this after the changes in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/commits/b0f9d2881ef41954bf87644179d7ded4dc068a04?

jaclyn-taroni:
> 

Thanks for catching this! Should be fixed in https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/commits/ed6f28fbb11b089db138aabe51a6c3a605108dfd

sjspielman:
Ah, I had missed https://github.com/AlexsLemonade/OpenPBTA-analysis/pull/1725/commits/b0f9d2881ef41954bf87644179d7ded4dc068a04! :tada: it's not there, success :)

sjspielman:
I can't approve this as it is "my PR", but I think we're there, at least as far as I can see!

jashapiro (approved):
> I can't approve this as it is "my PR", but I think we're there, at least as far as I can see!

I can!

PR #1730, opened 2023-04-05
Complete the Zenodo CSV readme

sjspielman:
Closes #1716 
This PR finishes the README for Zenodo files. I filled in remaining table items, and did one smol tweak on the last line of the file (pluralify `sample_id` -> `sample_id`s - does this look weird though in the rendered view?). 

A key part to review here is the "Notes on Tabular Data" - should anything here be rephrased or reorganized, and/or is there anything we think I've missed? I wondered, for example, if we might want to link the `tumor-purity-exploration` module in the high tumor purity notes - I'm really on the fence for that, so can figure out in review!

I also expanded the `.gitignore` to ignore the uncompressed versions of compressed files, which felt like a safe thing to do..!

PR #1731, opened 2023-04-05
Update `tables/README` to include Zenodo upload info

sjspielman:
Right after merging #1730 (of course...), I realized we need one more thing for Zenodo, which happily allows me to use this branch name that I only dreamed I would get to use! I added a bit of content to the overall `tables/README.md` that was still missing.


jaclyn-taroni (approved):
👍🏻

jaclyn-taroni:
Docs only change. Merging!

Issue #1736, opened 2023-04-24
Figure 1 Revisions for Publication


sjspielman:
- Panel B
  -  bump strip padding and legend
  - x and y axes could get bigger
  - 2 rows of legend maybe
- Panel C
  - bump up font size

PR #1743, opened 2023-04-24
Fix 1B Zenodo csv export

sjspielman:
While working to update panel 1B as part of #1736, I saw that the Zenodo table export was rather wrong. A column with `\n` snuck in, so the previous CSV export was in fact not a properly formatted CSV. I removed that column, as well as a couple other columns that can be purged.


sjspielman:
Noting that CI is failing due to `apt-get` fails while building docker, likely transient errors. Either way code changes in this PR do not get run through CI!

jashapiro (approved):
LGTM

sjspielman:
Merging in spite of unrelated CI fail, since this code isn't part of CI anyways.

PR #1745, opened 2023-04-24
Update debian sources list 

jashapiro:
Looks like Debian finally pulled the plug on Debian 9 (Stretch) so now we have to deal with that. This is a first attempt, but it is failing locally when trying to build java, so we'll see.


jaclyn-taroni (approved):
The relevant step passes CI 👍

Issue #1738, opened 2023-04-24
Figure 3 Revisions for Publication


sjspielman:
- A/B overall should be larger so they can be wider in the overall figure. 
  -  Also bump panel A legend text
- C/D should be narrower for compilation, but make sure to not shrink font!
- Increase size of panel E x-axis

PR #1748, opened 2023-04-25
Update Figure 3AB

jashapiro:
Here, as part of #1738, I updated sizing and placement for the legends and labels for the interaction plots (Figure 3 A&B, as well as the disease-specific results files).

I ended up leaving the images the same size, but increased font sizes a bit for everything but the gene names (I tried going up a point, but it looked really cramped).

I pulled the legends in a bit to give some flexibility when merging and a more compact appearance.

Let me know what you think of the changes!


jharenza (approved):
👍 looks good!

PR #1749, opened 2023-04-25
Figure 4 with larger text

### LINKED ISSUE #1739: Figure 4 Revisions for Publication
- All panels need to be somewhat larger; make smaller to make bigger
- Increase size of legends for panels A, D, H
- Increase text size of HR/P-values for panels F, G

sjspielman:
Closes #1739 

This PR updates Figure 4, both with code updates to get larger text, as well as direct compilation updates to get larger text where code couldn't accomplish it. I made a couple small rearrangements to increase text size as well, mostly legend placements.

Let me know what you think! More work might be needed here (are panel text sizes too inconsistent?), but I think this is most of the way done.


jashapiro:
My first impression is that the text in B&C could probably get a bit bigger without disrupting anything. 

We also might make the panel labels a bit smaller? Right now they kind of dwarf the actual legend text, which seems odd.

jharenza:
> My first impression is that the text in B&C could probably get a bit bigger without disrupting anything.
> 
> We also might make the panel labels a bit smaller? Right now they kind of dwarf the actual legend text, which seems odd.

Agree with all of this and for H, legend can be slightly smaller since it looks bigger than panel F and G text, but I can go either way on it!

sjspielman:
Thanks for the feedback @jashapiro and @jharenza, I implemented all suggested changes. Let me know what you think now!

jashapiro (approved):
LGTM!

PR #1752, opened 2023-04-26
Figure 1 and 2 in RGB colors

sjspielman:
This PR updates the compiled Figs 1-2 with RGB settings (not rbg as the branch name says, but I stand by the branch name nonetheless!).

For Figure 1, I replaced panel B. For Figure 2, it's all replaced with itself, but better! 
I did not do this for Figure 4 since really on the heatmap would be strongly affected, and there aren't really "shared" palettes in this figure that matter as much as the others besides the binary one, which is broadly used for anything binary! I can definitely update that panel if we really want though!


jharenza (approved):
looks good!

